PR #1219, opened 2022-07-05
config/v1/types_cluster_version: Add architecture field

jottofar:
As described in [enhancement], add architecture fields to ClusterVersionSpec and to ClusterVersionStatus. The former indicates the desired architecture and the later indicates the current architecture. A cluster's architecture can be either homogeneous or heterogeneous. A single valid architecture value, e.g. "amd64", indicates an homogeneous cluster and "multi" indicates a heterogeneous cluster.

[enhancement]: https://github.com/openshift/enhancements/blob/836516786ce47bf74a8a3dfe8a2fa7f1143e94c9/enhancements/multi-arch/heterogeneous-architecture-clusters.md

### DIFF
--- a/config/v1/types_cluster_version.go
+++ b/config/v1/types_cluster_version.go
@@ -56,6 +56,13 @@ type ClusterVersionSpec struct {
 	// +optional
 	DesiredUpdate *Update `json:"desiredUpdate,omitempty"`
 
+	// desiredArchitecture is an optional field that indicates the desired
+	// value of the cluster architecture. Setting this value will trigger
+	// an upgrade if the value does not match the current cluster
+	// architecture.
+	// +optional
+	DesiredArchitecture ClusterVersionArchitecture `json:"desiredArchitecture,omitempty"`
+
 	// upstream may be used to specify the preferred update server. By default
 	// it will use the appropriate update server for the cluster and region.
 	//
@@ -81,11 +88,11 @@ type ClusterVersionSpec struct {
 	Overrides []ComponentOverride `json:"overrides,omitempty"`
 }
 
-// ClusterVersionStatus reports the status of the cluster versioning,
-// including any upgrades that are in progress. The current field will
-// be set to whichever version the cluster is reconciling to, and the
-// conditions array will report whether the update succeeded, is in
-// progress, or is failing.
+// ClusterVersionStatus reports the cluster architecture and the status
+// of the cluster versioning, including any upgrades that are in progress.
+// The current field will be set to whichever version the cluster is
+// reconciling to, and the conditions array will report whether the update
+// succeeded, is in progress, or is failing.
 // +k8s:deepcopy-gen=true
 type ClusterVersionStatus struct {
 	// desired is the version that the cluster is reconciling towards.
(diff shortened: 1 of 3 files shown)


jottofar:
/retitle WIP: config/v1/types_cluster_version: Add archiecture field

jottofar:
/retitle config/v1/types_cluster_version: Add architecture field

config/v1/types_cluster_version.go
```diff
@@ -95,6 +95,13 @@ type ClusterVersionStatus struct {
 	// +required
 	Desired Release `json:"desired"`
 
+	// architecture indicates whether the cluster is a homogeneous or
```
deads2k, line 98: What transitions are allowed for this value?

deads2k, line 98: If it's an enum, it should have enumerated values.

deads2k, line 98: If this is determined live from inspection of nodes, why just single and multi instead of a list of architectures with numbers of nodes?  

If this is not determined live from inspection of nodes, why is this a status field instead of a spec field where a user states their intent?

deads2k, line 98: Once we support multi, why would a user ever want to choose single?

jottofar: It's determined by inspection of the payload at load time with this change https://issues.redhat.com/browse/ART-4129. And per https://coreos.slack.com/archives/CB95J6R4N/p1657141996473319 ART will only include the new metadata setting `multi` if it's a heterogeneous arch, i.e. ART does not set an arch for a homogeneous payload in which case CVO pulls it via runtime.GOARCH.

jottofar: As noted in comment below, the payload will have new metadata set to `multi` only for heterogeneous payloads. All other payloads are unchanged. For the homogeneous payloads CVO will determine arch using runtime.GOARCH. So I would think that the enum values would be `multi` + the valid OpenShift architectures.

jottofar: This will also be set upon initial install/CVO startup which could be occurring with an homogeneous payload.

wking: all transitions are theoretically allowed.  multi -> amd64 and amd64 -> multi should be fairly straightforward.  To do amd64 -> s390x, you'd need to do some fairly exciting dances, likely passing through multi in between.

deads2k: Since this value can change (see the Trevor's response above), this is looking like a spec field where the desired state is set and a status field where the actual current state of the cluster is reflected.  That way it becomes possible to manage a transition.

config/v1/types_cluster_version.go
```diff
@@ -56,6 +56,13 @@ type ClusterVersionSpec struct {
 	// +optional
 	DesiredUpdate *Update `json:"desiredUpdate,omitempty"`
 
+	// desiredArchitecture is an optional field that indicates the desired
+	// value of the cluster architecture. Setting this value will trigger
+	// an upgrade if the value does not match the current cluster
+	// architecture.
+	// +optional
```
deads2k, line 63: What does empty mean?

config/v1/types_cluster_version.go
```diff
@@ -95,6 +102,12 @@ type ClusterVersionStatus struct {
 	// +required
 	Desired Release `json:"desired"`
 
+	// architecture identifies the architecture of the cluster nodes. If
+	// the cluster architecture is heterogeneous, whereby the cluster
+	// has nodes running on different architectures, "multi" is used.
+	// +optional
```
deads2k, line 108: explain whether empty is a temporary condition or a long term condition to help consumers know what to expect.

config/v1/types_cluster_version.go
Prashanth684: are we saying that if there is a transition from adm64->s390x, it will involve a middle stage of first moving to multi and then to the other single arch? 

i'm wondering if single -> multi should be the only allowed transition.

jottofar:
/hold
Still a lot decisions to be worked out. Being tracked by https://issues.redhat.com/browse/OTA-658 and https://issues.redhat.com/browse/OTA-659

PR #1220, opened 2022-07-06
Bug 2103680: Change the type of DisableNetworkDiagnostics to a pointer

kyrtapz:
Change the type of DisableNetworkDiagnostics to a pointer to allow for the variable to be unset.

Signed-off-by: Patryk Diak <pdiak@redhat.com>


deads2k:
/hold

You cannot change this type. By making this a pointer, it means that older clients (say 4.11, when upgrading to 4.12), may end up in a state where they read a network.operator.openshift.io that contains nil for this field and cause a failure to deserialize the API content.  That will cause an inability to read the desired state and an inability to write the actual state.

You can replace this field an enum field and after 1 year or four releases, you can remove this field.

cc @JoelSpeed @Miciah @bparees for an example of "cannot change the type because old clients will be broken".

PR #1221, opened 2022-07-07
Bug 2104481: operator/ingress: Allow PROXY protocol for Private endpoint publishing strategy

Miciah:
Allow configuring an IngressController with the "Private" endpoint publishing strategy type to use PROXY protocol.

* `operator/v1/types_ingress.go` (`PrivateStrategy`): Add `Protocol` field with type `IngressControllerProtocol`.
* `operator/v1/0000_50_ingress-operator_00-ingresscontroller.crd.yaml`:
* `operator/v1/zz_generated.swagger_doc_generated.go`: Regenerate.

### DIFF
--- a/operator/v1/types_ingress.go
+++ b/operator/v1/types_ingress.go
@@ -616,6 +616,34 @@ type HostNetworkStrategy struct {
 // PrivateStrategy holds parameters for the Private endpoint publishing
 // strategy.
 type PrivateStrategy struct {
+	// protocol specifies whether the IngressController expects incoming
+	// connections to use plain TCP or whether the IngressController expects
+	// PROXY protocol.
+	//
+	// PROXY protocol can be used with load balancers that support it to
+	// communicate the source addresses of client connections when
+	// forwarding those connections to the IngressController.  Using PROXY
+	// protocol enables the IngressController to report those source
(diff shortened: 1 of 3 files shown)


operator/v1/types_ingress.go
```diff
@@ -616,6 +616,34 @@ type HostNetworkStrategy struct {
...
+	// PROXY protocol.
+	//
+	// PROXY protocol can be used with load balancers that support it to
+	// communicate the source addresses of client connections when
+	// forwarding those connections to the IngressController.  Using PROXY
+	// protocol enables the IngressController to report those source
+	// addresses instead of reporting the load balancer's address in HTTP
+	// headers and logs.  Note that enabling PROXY protocol on the
+	// IngressController will cause connections to fail if you are not using
+	// a load balancer that uses PROXY protocol to forward connections to
```
deads2k, line 630: I bet I asked last time too.  Can we detect this and have a metric counting failures?

deads2k:
/approve

/assign @candita 
for final review and lgtm.

operator/v1/types_ingress.go
Miciah: Good idea—it might be observable using the `haproxy_frontend_request_errors_total` metric.

PR #1223, opened 2022-07-11
Bug 2106086: operator/ingress: Fix healthCheckInterval pattern

Miciah:
Reject values for `healthCheckInterval` such as "0abc".

* `operator/v1/types_ingress.go` (`IngressControllerTuningOptions`): Fix validation for `healthCheckInterval`.
* `operator/v1/0000_50_ingress-operator_00-ingresscontroller.crd.yaml`: Regenerate.


deads2k:
/lgtm
/hold

working out whether we have time to fix for 4.11.0.  If we do, then we can release this hold.  If we do not, then we cannot merge this.

Miciah:
```
Error: unable to parse option "output:dir=/tmp/tmp.0BvA8aFtqE": [binary literal has no digits (at <input>:1:10)]
```
/test verify

PR #1225, opened 2022-07-11
NE-586: ingress: add ReloadInterval API field

Ethany-RH:
add ReloadInterval API field to IngressControllerTuningOptions struct


operator/v1/types_ingress.go
```diff
@@ -1453,6 +1453,30 @@ type IngressControllerTuningOptions struct {
...
+	// reloadInterval defines the minimum interval at which the router is allowed to reload
+	// to accept new changes. Increasing this value can prevent the accumulation of
+	// HAProxy processes, depending on the scenario. Increasing this interval can
+	// also lessen load imbalance on a backend's servers when using the roundrobin
+	// balancing algorithm. Alternatively, decreasing this value may decrease latency
+	// since updates to HAProxy's configuration can take effect more quickly.
+	//
+	// The value must be a time duration value; see <https://pkg.go.dev/time#ParseDuration>.
+	// Currently, the minimum value allowed is 1s, and the maximum allowed value is
+	// 720h (= 30 days). Minimum and maximum allowed values may change in future versions
```
deads2k, line 1466: Why would we ever want a reload interval larger than the graceful termination time of the router pod?  If the reload interval is longer than the graceful termination, we're better off simply having the pod restart, right?  And the pod restart time is already chosen for minimal disruption.

operator/v1/types_ingress.go
```diff
@@ -1453,6 +1453,30 @@ type IngressControllerTuningOptions struct {
...
+	// of OpenShift.
+	//
+	// An empty reloadInterval tells the IngressController to choose the default, which
```
deads2k, line 1469: What is the difference between zero and nil?

operator/v1/types_ingress.go
Miciah: The grace termination period and reload interval seem like orthogonal concepts to me.  To be clear (and maybe this needs to be clarified in the godoc), `reloadInterval` controls the reload interval for the haproxy process within the container.  The router's graceful shutdown period, which is hard-coded to 45 seconds, controls the period of time that the pod will hang around and continue handling traffic when it gets a SIGTERM (presumably because the pod is being terminated).  A user might want to increase the reload interval to very high values to ensure stability during periods of time when the user anticipates high traffic loads.  During such periods of time, the user may prefer that the haproxy process not reload, and probably also would prefer that the router pod not terminate.  However, if this does happen, the user probably still wants the router have a shutdown period sufficient to drain traffic but no longer than that.

operator/v1/types_ingress.go
```diff
@@ -1453,6 +1453,30 @@ type IngressControllerTuningOptions struct {
 	// +kubebuilder:validation:Optional
 	// +optional
 	MaxConnections int32 `json:"maxConnections,omitempty"`
+
+	// reloadInterval defines the minimum interval at which the router is allowed to reload
+	// to accept new changes. Increasing this value can prevent the accumulation of
```
deads2k, line 1458: do these changes include new routes?  If this value were set to 30 days and a new route was created, would it be served?  If not, is there a way in the API to know that this has happened?

operator/v1/types_ingress.go
deads2k: if new routes are not served until this interval passes, then I think anything measured in more than seconds needs to have a way for the route author to know that his traffic is not being served.

operator/v1/types_ingress.go
Ethany-RH: I'm not sure if I understand your question. Are you asking what should be the difference b/w giving the field 0 and nil? I think we want to have the reload interval be the default of 5s if no input is given and 1s if the value given is lower than the floor (i.e. 0s).

operator/v1/types_ingress.go
Ethany-RH: From Miciah: "`RELOAD_INTERVAL` does affect new routes.  There is no direct feedback to the user on the route status or otherwise whether the HAProxy configuration has been updated and reloaded."

operator/v1/types_ingress.go
Miciah: > I thought the pod termination grace period for ingress pods was set to a number of hours.

Sorry, the pod termination grace period is set to 1 hour (3600 seconds; [source](https://github.com/openshift/cluster-ingress-operator/blob/b7dbcb92d672c7ac71c47721808ce6ee3695d81a/pkg/operator/controller/ingress/deployment.go#L240-L242)).  The router's own shutdown grace period is 45 seconds [(source](https://github.com/openshift/router/blob/601ba575b7fadc1e05e7b4e2f35a660859db9788/pkg/cmd/infra/router/template.go#L701-L705)).  If a router pod ever takes much more than 45 seconds to terminate, then something is probably broken.  

> A 30 day reload interval seems excessive to me and from a practical perspective means the pod would need to be restarted. Contrasting with say 4-12h as a maximum, which still allows for change.

It does seem excessive.  However, I could see someone wanting to turn off reloads entirely during an anticipated period of peak load.  In this scenario, the user doesn't want HAProxy reloads or pod restarts.  

> if new routes are not served until this interval passes, then I think anything measured in more than seconds needs to have a way for the route author to know that his traffic is not being served.

Ideally the route status would tell the user not only whether the controller had admitted the route but also whether the HAProxy config had been updated.  That's a long-standing gap.  Right now, we tell people that they need to try connecting to the route to determine whether it's ready.  I'd like for us to cover that gap, but that would be another feature requiring some design and consideration as to how it would impact API load.

deads2k: > if new routes are not served until this interval passes, then I think anything measured in more than seconds needs to have a way for the route author to know that his traffic is not being served.

> Ideally the route status would tell the user not only whether the controller had admitted the route but also whether the HAProxy config had been updated. That's a long-standing gap. Right now, we tell people that they need to try connecting to the route to determine whether it's ready. I'd like for us to cover that gap, but that would be another feature requiring some design and consideration as to how it would impact API load.

If I'm interpreting this correctly, the PR would change the maximum amount of time for the new route to be observed from 5s to 30 days.  This changes the gap from a minor annoyance to something that could leave a user's route broken for weeks after they expect it to work.  If I have correctly understood the impact, then I think the API gap must be closed before we can add a feature that makes that gap go from seconds to weeks.

Miciah: > the PR would change the maximum amount of time for the new route to be observed from 5s to 30 days.

Only if the cluster admin chooses to increase it to 30 days.  

What is the maximum amount of time we should allow the cluster admin to increase it to?

Ethany-RH: Hi @deads2k,
Miciah looked into various customer cases and Bugzilla reports and found that the longest value for ReloadInterval currently in use is 90s. There was 1 Bugzilla report that mentioned a ReloadInterval of 1 hour, but that came from an engineer trying to turn off reloads while testing a change in the OpenShift router. If we think that turning off reloads for testing should not be the responsibility of the ReloadInterval API field, then I think it would be fine to make the maximum-allowed interval much shorter. Do you have any suggestions for what the maximum should be? (maybe 5 to 10 minutes?)

Ethany-RH: (I also added a warning in the godoc about setting the interval too high.)

operator/v1/types_ingress.go
```diff
@@ -1453,6 +1453,33 @@ type IngressControllerTuningOptions struct {
...
+	// Currently, the minimum value allowed is 1s, and the maximum allowed value is
+	// 120s. Minimum and maximum allowed values may change in future versions of OpenShift.
+	//
+	// An empty reloadInterval tells the IngressController to choose the default, which
+	// is currently 5s.
+	//
+	// This field expects an unsigned duration string of decimal numbers, each with optional
+	// fraction and a unit suffix, e.g. "100s", "1m30s". Valid time units are "s" and "m".
+	//
+	// Note: Setting a large value (e.g. > 90s) can cause significant latency in observing
```
deads2k, line 1474: Also indicate that the ingress operator will reject values greater than 90s and use a default determined by the IngressController, that is subject to change in the future and is currently 5s.

operator/v1/types_ingress.go
```diff
@@ -1453,6 +1453,33 @@ type IngressControllerTuningOptions struct {
...
+	// updates to routes and their endpoints. HAProxy's configuration will be reloaded less
+	// frequently, and newly created routes will not be served until the subsequent reload.
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Pattern=^(0|([0-9]+(\.[0-9]+)?(s|m))+)$
+	// +kubebuilder:validation:Type:=string
+	// +optional
+	ReloadInterval *metav1.Duration `json:"reloadInterval,omitempty"`
```
deads2k, line 1482: if this is a metav1.Duration, we could have 0 mean "use a default", avoid nil checks, and make this option obvious

operator/v1/types_ingress.go
Ethany-RH: Miciah asked me to set the max as 120s. Should I change this to 90s then? The current code also caps values instead of rejects them (i.e. if a user sets a value over 120s, it gets capped to 120s and not replaced by the default). I can state this more clearly in the comments.

Ethany-RH: Done

operator/v1/types_ingress.go
Ethany-RH: Following up on this, is the current solution acceptable?

operator/v1/types_ingress.go
Ethany-RH: There is no difference between zero and nil. Both will get the default of 5s.

operator/v1/types_ingress.go
Ethany-RH: Following up here, the max is now 120s.

operator/v1/types_ingress.go
deads2k: > Following up on this, is the current solution acceptable?

Given a max of 120s, yes.  If it ever gets extended, the route API needs to grow a way to represent that.

config/v1/types_infrastructure.go
```diff
@@ -574,6 +574,10 @@ type IBMCloudPlatformStatus struct {
 	// CISInstanceCRN is the CRN of the Cloud Internet Services instance managing
 	// the DNS zone for the cluster's base domain
 	CISInstanceCRN string `json:"cisInstanceCRN,omitempty"`
+
+	// DNSInstanceCRN is the CRN of the DNS Services instance managing the DNS zone
+	// for the cluster's base domain
+	DNSInstanceCRN string `json:"dnsInstanceCRN,omitempty"`
```
deads2k, line 580: How is this related to the ReloadInterval?  Bad push?

machine/v1/types_controlplanemachineset.go
```diff
@@ -0,0 +1,313 @@
...
+// +kubebuilder:subresource:scale:specpath=.spec.replicas,statuspath=.status.replicas
+// +kubebuilder:printcolumn:name="Desired",type="integer",JSONPath=".spec.replicas",description="Desired Replicas"
+// +kubebuilder:printcolumn:name="Current",type="integer",JSONPath=".status.replicas",description="Current Replicas"
+// +kubebuilder:printcolumn:name="Ready",type="integer",JSONPath=".status.readyReplicas",description="Ready Replicas"
+// +kubebuilder:printcolumn:name="Updated",type="integer",JSONPath=".status.updatedReplicas",description="Updated Replicas"
+// +kubebuilder:printcolumn:name="Unavailable",type="integer",JSONPath=".status.unavailableReplicas",description="Observed number of unavailable replicas"
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp",description="ControlPlaneMachineSet age"
+// Compatibility level 1: Stable within a major release for a minimum of 12 months or 3 minor releases (whichever is longer).
+// +openshift:compatibility-gen:level=1
+type ControlPlaneMachineSet struct {
```
deads2k, line 25: this looks like an accident....

operator/v1/types_ingress.go
```diff
@@ -1453,6 +1453,38 @@ type IngressControllerTuningOptions struct {
...
+	//
+	// The value must be a time duration value; see <https://pkg.go.dev/time#ParseDuration>.
+	// Currently, the minimum value allowed is 1s, and the maximum allowed value is
+	// 120s. Minimum and maximum allowed values may change in future versions of OpenShift.
+	// Note that if a duration outside of these bounds is provided, the value of reloadInterval
+	// will be capped/floored and not rejected (e.g. a duration of over 120s will be capped to
+	// 120s; the IngressController will not reject and replace this disallowed value with
+	// the default).
+	//
+	// A nil or zero value for reloadInterval tells the IngressController to choose the default,
```
deads2k, line 1472: nil isn't allowed

operator/v1/types_ingress.go
```diff
@@ -1453,6 +1453,38 @@ type IngressControllerTuningOptions struct {
...
+	// which is currently 5s and subject to change.
```
deads2k, line 1473: subject to change without notice.

Ethany-RH: Done

deads2k:
/lgtm
/approve

/hold

holding in case you'd like another viewer.  /hold cancel at your discretion.

Ethany-RH:
cc @ShudiLi for QE approval
cc @ahardin-rh for docs approval

Ethany-RH:
cc @cfields651 for px approval

operator/v1/types_ingress.go
```diff
@@ -1453,6 +1453,38 @@ type IngressControllerTuningOptions struct {
...
+	// This field expects an unsigned duration string of decimal numbers, each with optional
+	// fraction and a unit suffix, e.g. "100s", "1m30s". Valid time units are "s" and "m".
+	//
+	// Note: Setting a value significantly larger than the default of 5s can cause latency
+	// in observing updates to routes and their endpoints. HAProxy's configuration will
+	// be reloaded less frequently, and newly created routes will not be served until the
+	// subsequent reload.
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Pattern=^(0|([0-9]+(\.[0-9]+)?(s|m))+)$
```
candita, line 1484: The `+` shouldn't be at the end of the regex, as it then allow something like 1s2m.

candita: But you still have to figure out how allow 1m30s

ShudiLi:
/label qe-approved
@Ethany-RH thanks.

operator/v1/types_ingress.go
```diff
@@ -1481,6 +1481,38 @@ type IngressControllerTuningOptions struct {
...
+	// +kubebuilder:validation:Type:=string
+	// +optional
+	ReloadInterval metav1.Duration `json:"reloadInterval,omitempty"`
```
candita, line 1515: We discussed the fact that this is the inaugural use of a non-pointer duration in the `tuningOptions`.  We are trying to phase out pointers so that empty fields are discoverable (an empty pointer duration hides the field).  For now, this is the only one.

PR #1228, opened 2022-07-15
route/v1: Add CRD

Miciah:
* `Makefile`: Add an "update-codegen-crd-route" target.
* `route/v1/route.crd.yaml`: New file.  Define a CRD for the Route API.
* `route/v1/route.crd.yaml-patch`: New file.  Patch `route-crd.yaml` with validation that cannot be expressed using kubebuilder tags.
* `route/v1/test-route-validation.sh`: New file.  Verify that the API admits valid routes and rejects invalid routes.
* `route/v1/types.go`: Add kubebuilder validation tags.

---

The motivation for this PR is to supersede <https://github.com/openshift/router/blob/master/deploy/route_crd.yaml>, which does not perform any validation, for use in environments where OpenShift router is used without openshift-apiserver.  

This implements most of the validation that openshift-apiserver does; see <https://github.com/openshift/openshift-apiserver/blob/master/pkg/route/apis/route/validation/validation.go>, and see <https://gist.github.com/Miciah/cfa92327b66bf697a1d66cccb7633cc7> for a shell-script that tests the validation.  One remaining gap is that the CRD doesn't default `spec.host`.  The CRD also doesn't do the ratcheting logic for `spec.host` validation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,7 @@ $(call add-crd-gen,operatorcontrolplane,./operatorcontrolplane/v1alpha1,./operat
 $(call add-crd-gen,machine-beta,./machine/v1beta1,./machine/v1beta1,./machine/v1beta1)
 $(call add-crd-gen,machine,./machine/v1,./machine/v1,./machine/v1)
 $(call add-crd-gen,monitoring-alpha,./monitoring/v1alpha1,./monitoring/v1alpha1,./monitoring/v1alpha1)
+$(call add-crd-gen,route,./route/v1,./route/v1,./route/v1)
 
 RUNTIME ?= podman
 RUNTIME_IMAGE_NAME ?= openshift-api-generator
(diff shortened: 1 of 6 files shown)


oglok:
This PR is of great interests for the MicroShift project. Thanks @Miciah for implementing the route CRD.

deads2k:
that's quite red.

deads2k:
Verify fails on `Error: missing 'api-approved.openshift.io' annotation pointing to openshift/api pull request in ./route/v1/route.crd.yaml`

please add.

/approve

deads2k:
Link to a gist so I don't forget: https://gist.github.com/Miciah/cfa92327b66bf697a1d66cccb7633cc7

/lgtm

deads2k:
This is part of microshift delivery, labelling so it doesn't apply for feature freeze.

deads2k:
now that gist is included.

/lgtm

PR #1231, opened 2022-07-19
Add storage capability

qJkee:
make storage operator optional

### DIFF
--- a/config/v1/0000_00_cluster-version-operator_01_clusterversion.crd.yaml
+++ b/config/v1/0000_00_cluster-version-operator_01_clusterversion.crd.yaml
@@ -68,6 +68,7 @@ spec:
                           - marketplace
                           - Console
                           - Insights
+                          - Storage
                       x-kubernetes-list-type: atomic
                     baselineCapabilitySet:
                       description: baselineCapabilitySet selects an initial set of optional capabilities to enable, which can be extended via additionalEnabledCapabilities.  If unset, the cluster will choose a default, and the default may change over time. The current default is vCurrent.
@@ -174,6 +175,7 @@ spec:
                           - marketplace
                           - Console
                           - Insights
+                          - Storage
                       x-kubernetes-list-type: atomic
                     knownCapabilities:
                       description: knownCapabilities lists all the capabilities known to the current cluster.
@@ -187,6 +189,7 @@ spec:
                           - marketplace
                           - Console
                           - Insights
+                          - Storage
                       x-kubernetes-list-type: atomic
                 conditionalUpdates:
                   description: conditionalUpdates contains the list of updates that may be recommended for this cluster if it meets specific required conditions. Consumers interested in the set of updates that are actually recommended for this cluster should use availableUpdates. This list may be empty if no updates are recommended, if the update service is unavailable, or if an empty or invalid channel has been specified.
--- a/config/v1/types_cluster_version.go
+++ b/config/v1/types_cluster_version.go
@@ -225,7 +225,7 @@ type UpdateHistory struct {
 type ClusterID string
 
 // ClusterVersionCapability enumerates optional, core cluster components.
-// +kubebuilder:validation:Enum=openshift-samples;baremetal;marketplace;Console;Insights
+// +kubebuilder:validation:Enum=openshift-samples;baremetal;marketplace;Console;Insights;Storage
 type ClusterVersionCapability string
 
 const (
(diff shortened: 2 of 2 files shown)


config/v1/0000_00_cluster-version-operator_01_clusterversion.crd.yaml
```diff
@@ -68,6 +68,7 @@ spec:
                           - marketplace
                           - Console
                           - Insights
+                          - storage
```
deads2k, line 71: PascalCase, so Storage, please.

deads2k:
/hold

Can I get a reference for making Storage optional?

/cc @jsafrane

gnufied:
Did we want to disable snapshot controller as well?

qJkee:
@gnufied 
Yeah, i have a task to make a `csi-snapshot-controller` optional

jsafrane:
@qjkee do you have an enhacement with some explanation how is OCP without storage useful?

jerpeter1:
> @qJkee do you have an enhacement with some explanation how is OCP without storage useful?

The closest thing I've been able to find is this doc: [What is “Telco 5G” and Why Should I Care?](https://mojo.redhat.com/people/markmc/blog/2020/01/21/what-is-telco-5g-and-why-should-i-care) which has some background on RAN and DU workloads. In summary, the customers which have requested this capability are running fixed workloads which don't utilize storage at all. For these clusters, reducing the overall footprint of OCP is paramount.

Thus far, we've resolved that documenting the consequences of disabling storage and the other capabilities that have been requested is sufficient preparation for any potential fallout. I'm uncertain how broadly telco's use-case is applicable, but if we can't think of anyone else who might possibly want to operate a cluster with 100% stateless workloads, I guess maybe we should consider adding some sort of limits which prohibit enabling this capability on non-SNO/DU clusters somehow.

jsafrane:
(rephrasing our discussion on slack)

Disabling storage and snapshots in SNOs makes perfect sense, still, we don't have a good API to limit storage removal *only* to SNO. If the capability is exposed on OpenStack and RHEV and users disable storage there, then the installation fails because of registry. We either need to teach registry operator not to create PVC if Storage capability is disabled, or somehow disable registry too. This needs to be approved by registry folks and recorded for the future, preferably as an enhancement.  Slack, this PR nor telco requirement doc is not a good place for such record.

cc @dmage

jsafrane:
Please update the commit message to mention that this optional capability aims at trimming down resource requirements of bare metal clusters + that OpenStack and RHEV require `Storage` capability to be installable.

Basically, write the note everywhere you can. With that notes everywhere I am fine approving this.

ggiguash:
CNF-5645

qJkee:
Hi @jsafrane 
I've added notes to the commit message and over the type comments.
I will provide proper documentation as well later before we merge operator annotations

config/v1/types_cluster_version.go
```diff
@@ -254,6 +254,14 @@ const (
 	// collects anonymized information about the cluster to generate
 	// recommendations for possible cluster issues.
 	ClusterVersionCapabilityInsights ClusterVersionCapability = "Insights"
+
+	// ClusterVersionCapabilityStorage manages the storage operator which
+	// is responsible for providing cluster-wide storage defaults
+	// WARNING: Do not disable this capability when deployed to
+	// bare metal clusters, RHEV and Openstack without reading the docs.
```
jsafrane, line 261: Please remove note about bare metal clusters - cluster-storage-operator actually does not do anything there. I would even recommend to disable the capability on bare metal.

OpenStack and RHEV is correct.

config/v1/types_cluster_version.go
```diff
@@ -254,6 +254,14 @@ const (
 	// collects anonymized information about the cluster to generate
 	// recommendations for possible cluster issues.
 	ClusterVersionCapabilityInsights ClusterVersionCapability = "Insights"
+
+	// ClusterVersionCapabilityStorage manages the storage operator which
+	// is responsible for providing cluster-wide storage defaults
+	// WARNING: Do not disable this capability when deployed to
+	// RHEV and OpenStack without reading the docs.
+	// Thiese clusters heavily rely on that capability and may cause
```
jsafrane, line 262: typo: thiese

jsafrane:
/lgtm
/approve
from storage POV

qJkee:
@deads2k could you please review as well?

deads2k:
> /lgtm
> /approve
> from storage POV

storage acknowledged
/approve
/hold cancel

jerpeter1:
@jinyunma - it doesn't make sense to do QE until we put up the PR that actually uses these changes in the storage operator

PR #1233, opened 2022-07-21
Add capabilities tests

raspbeep:
This test might be useful in future realeases. Taking into account the movement of separating core components and making the operators installation optional via OLM, this test verifies that set of capabilities available in preceding versions are a subset of the ones in the following.

e.g. None ⊂ v4.11 ⊂ v4.12 ⊂ Current

It verifies, that no error or mismatch occurs during adding new capabilities or releasing new OCP versions.

Also, minor change to the variable names due to builtin function name collision(`cap`).


config/v1/types_cluster_version_test.go
```diff
@@ -10,14 +11,49 @@ import (
...
+		0: ClusterVersionCapabilitySetNone,
+		1: ClusterVersionCapabilitySet4_11,
+		2: ClusterVersionCapabilitySetCurrent,
+	}
+
+	var keys []int
+	for k := range capabilitySetOrder {
+		keys = append(keys, k)
+	}
+	sort.Ints(keys)
```
stlaz, line 41: /hold

if I were you, I would use a completely different homogeneous data structure.

stlaz (changes requested):

config/v1/types_cluster_version_test.go
raspbeep: Replaced map for array in [here](https://github.com/openshift/api/pull/1242).

(Moved to new PR due to my mistake in git work)

PR #1235, opened 2022-07-25
ensure omitempty set on default placement to ensure ingress serialization comptaible for production hypershift deployments

relyt0925:
There is currently a problem when defaultPlacement does not have omitempty where the ingress config serialization of hypershift is incompatible when hypershift is managing 4.10 clusters. In order for that engine to appropriately work (the hypershift operator manages multiple independent clusters that can be at independent minor versions within a skew) this status field needs to be omitempty to ensure an invalid field is not serialized that then breaks the config being serialized in the control-plane operator (scoped to an individual cluster) associated with the 4.10 API.

### DIFF
--- a/config/v1/types_ingress.go
+++ b/config/v1/types_ingress.go
@@ -128,7 +128,7 @@ type IngressStatus struct {
 	//
 	// +kubebuilder:validation:Enum:="ControlPlane";"Workers";""
 	// +optional
-	DefaultPlacement DefaultPlacement `json:"defaultPlacement"`
+	DefaultPlacement DefaultPlacement `json:"defaultPlacement,omitempty"`
 }
 
 // ComponentRouteSpec allows for configuration of a route's hostname and serving certificate.


relyt0925:
cc @csrwng

csrwng:
@omertuc @deads2k The change introduced here: https://github.com/openshift/api/pull/1172 added "" as a valid enum value, but an empty value is not accepted if you don't have the `omitempty` tag on the field.

omertuc:
Not including omitempty was a [conscious](https://coreos.slack.com/archives/CE4L0F143/p1649710005788959?thread_ts=1649706201.918029&cid=CE4L0F143) decision. Can't say I entirely understand the reasoning, but you may want to review that thread before proceeding with this change

csrwng:
@omertuc thx I see. The blank as a option does work for current clients, it just doesn't work for previous version clients. @relyt0925 let me think of a way that we can fix this.

relyt0925:
closing as it was decided this wasn't best path

PR #1242, opened 2022-08-09
Add capabilities test

raspbeep:
This test might be useful in future releases. Taking into account the movement of separating core components and making the operators installation optional via OLM, this test verifies that set of capabilities available in preceding versions are a subset of the ones in the following.

e.g. None ⊂ v4.11 ⊂ v4.12 ⊂ Current

It verifies, that no error or mismatch occurs during adding new capabilities or releasing new OCP versions.

Also, minor change to the variable names due to builtin function name collision(cap).

Signed-off-by: Pavel Kratochvil <pakratoc@redhat.com>

### DIFF
--- a/config/v1/types_cluster_version_test.go
+++ b/config/v1/types_cluster_version_test.go
@@ -10,14 +10,44 @@ import (
 func TestKnownClusterVersionCapabilities(t *testing.T) {
 	exists := struct{}{}
 	known := make(map[ClusterVersionCapability]struct{}, len(KnownClusterVersionCapabilities))
-	for _, cap := range KnownClusterVersionCapabilities {
-		known[cap] = exists
+	for _, capability := range KnownClusterVersionCapabilities {
+		known[capability] = exists
 	}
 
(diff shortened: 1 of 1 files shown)


config/v1/types_cluster_version_test.go
```diff
@@ -10,14 +10,44 @@ import (
...
+
+// TestSubsetsVersionCapabilities verifies that all preceding sets
+// (order given by capabilitySetOrder) are subsets of following sets
+func TestSubsetsVersionCapabilities(t *testing.T) {
+	// defines the order by which subsets are tested
+	var capabilitySetOrder = []ClusterVersionCapabilitySet{
+		ClusterVersionCapabilitySetNone,
+		ClusterVersionCapabilitySet4_11,
+		ClusterVersionCapabilitySet4_12,
+		ClusterVersionCapabilitySetCurrent,
```
bparees, line 34: i don't think there's any guarantee that the sets only grow in size from version to version.

We could decide that some cap is no longer enabled by default, at some point (perhaps the cap gets deprecated/removed entirely).

PR #1250, opened 2022-08-16
OCPSE-25: introduce EvaluationConditionsDetected ClusterStatusConditionType

aravindhp:
This will be used to indicate the result of the detection logic that was added to a component to evaluate the introduction of an invasive change that could potentially result in highly visible alerts, breakages or upgrade failures on user clusters.

Ref: https://github.com/openshift/enhancements/pull/1216

### DIFF
--- a/config/v1/types_cluster_operator.go
+++ b/config/v1/types_cluster_operator.go
@@ -188,6 +188,13 @@ const (
 	// allow updates when this condition is not False, including when it is
 	// missing, True, or Unknown.
 	OperatorUpgradeable ClusterStatusConditionType = "Upgradeable"
+
+	// EvaluationConditionsDetected is used to indicate the result of the detection
+	// logic that was added to a component to evaluate the introduction of an
+	// invasive change that could potentially result in highly visible alerts,
+	// breakages or upgrade failures. You can concatenate multiple Reason using
+	// the "::" delimiter if you need to evaluate the introduction of multiple changes.
+	EvaluationConditionsDetected ClusterStatusConditionType = "EvaluationConditionsDetected"
 )
 
 // ClusterOperatorList is a list of OperatorStatus resources.


config/v1/types_cluster_operator.go
```diff
@@ -188,6 +188,14 @@ const (
...
 	// missing, True, or Unknown.
 	OperatorUpgradeable ClusterStatusConditionType = "Upgradeable"
+
+	// ClusterFleetEvaluation is used to indicate the result of the detection
+	// logic that was added to a component to evaluate the introduction of an
+	// invasive change that could potentially result in highly visible alerts,
+	// breakages or upgrade failures on user  clusters. You can concatenate
+	// multiple ConditionStatus/Reason if you need to evaluation the
+	// introduction of multiple changes.
+	ClusterFleetEvaluation ClusterStatusConditionType = "ClusterFleetEvaluation"
```
bparees, line 198: so is the idea that a component would set this to "true" when they detect one or more conditions they are investigating?

The condition name doesn't naturally lend itself to taking on a true/false value like our other conditions do.

I don't know how wordy we want to get, but "ClusterFleetEvaluationConditionPresent" or Detected would work better with a true/false value.

aravindhp: I see your point. `EvaluationConditionDetected` might be more appropriate though I wonder if it should be singular or plural given there could be multiple conditions.

aravindhp:
/retitle OCPSE-25: introduce ClusterFleetEvaluation ClusterStatusConditionType

aravindhp:
/retitle OCPSE-25: introduce EvaluationConditionDetected ClusterStatusConditionType

config/v1/types_cluster_operator.go
deads2k: `EvaluationConditionsDetected` works for me.

aravindhp: @deads2k what is the delimiter for concatenating multiple Reasons? I think adding that to the doc will help developers.

aravindhp: I have renamed the condition to `EvaluationConditionsDetected`

deads2k: > @deads2k what is the delimiter for concatenating multiple Reasons? I think adding that to the doc will help developers.

The delimiter is `::` from https://github.com/openshift/library-go/blob/master/pkg/operator/status/condition.go#L154

aravindhp: Addressed @deads2k. Please review.

aravindhp:
@bparees can I get a lgtm on this?

aravindhp:
/retitle OCPSE-25: introduce EvaluationConditionsDetected ClusterStatusConditionType

config/v1/types_cluster_operator.go
```diff
@@ -188,6 +188,14 @@ const (
 	// allow updates when this condition is not False, including when it is
 	// missing, True, or Unknown.
 	OperatorUpgradeable ClusterStatusConditionType = "Upgradeable"
+
+	// EvaluationConditionsDetected is used to indicate the result of the detection
+	// logic that was added to a component to evaluate the introduction of an
+	// invasive change that could potentially result in highly visible alerts,
+	// breakages or upgrade failures on user  clusters. You can concatenate
```
wking, line 195: nit: two spaces in `user  clusters`.

wking: also "on user clusters" is probably not necessary, because all of these ClusterOperator types only ever exist in OpenShift clusters.

aravindhp: Removed. Please lgtm @wking.

wking (approved):
/lgtm

PR #1253, opened 2022-08-18
Add feature flags for SDN migration

pliurh:
Allow users to decide whether to migration the configuration of
the egress IP, egress firewall and multicast automatically when
conducting SDN migration.

Signed-off-by: Peng Liu <pliu@redhat.com>


operator/v1/types_network.go
```diff
@@ -123,6 +123,35 @@ type NetworkMigration struct {
 	// changing the MTU for the default network will be rejected.
 	// +optional
 	MTU *MTUMigration `json:"mtu,omitempty"`
+
+	// features contains the features migration configuration. Set this to migrate
+	// feature configuration when changing the cluster default network provider.
+	// +optional
+	Features *FeaturesMigration `json:"features,omitempty"`
```
msherif1234, line 130: is it better to use `featuresMigration` instead of `features` since features is very generic ?

operator/v1/0000_70_cluster-network-operator_01.crd.yaml
```diff
@@ -373,6 +373,22 @@ spec:
                   description: migration enables and configures the cluster network migration. The migration procedure allows to change the network type and the MTU.
                   type: object
                   properties:
+                    features:
+                      description: features contains the features migration configuration. Set this to migrate feature configuration when changing the cluster default network provider.
+                      type: object
+                      properties:
+                        egressFirewall:
+                          description: egressFirewall specifies whether or not the EgressFirewall configuration is migrated automatically when changing the cluster default network provider. If unset, this property defaults to 'true' and EgressFirewall configure is not migrated.
```
kyrtapz, line 381:
```suggestion
                          description: egressFirewall specifies whether or not the EgressFirewall configuration is migrated automatically when changing the cluster default network provider. If unset, this property defaults to 'true' and EgressFirewall configure is migrated.
```

operator/v1/0000_70_cluster-network-operator_01.crd.yaml
```diff
@@ -373,6 +373,22 @@ spec:
...
+                          type: boolean
+                          default: true
+                        egressIP:
+                          description: egressIP specifies whether or not the EgressIP configuration is migrated automatically when changing the cluster default network provider. If unset, this property defaults to 'true' and EgressIP configure is not migrated.
```
kyrtapz, line 385:
```suggestion
                          description: egressIP specifies whether or not the EgressIP configuration is migrated automatically when changing the cluster default network provider. If unset, this property defaults to 'true' and EgressIP configure is migrated.
```

operator/v1/0000_70_cluster-network-operator_01.crd.yaml
```diff
@@ -373,6 +373,22 @@ spec:
...
+                          type: boolean
+                          default: true
+                        multicast:
+                          description: multicast specifies whether or not the multicast configuration is migrated automatically when changing the cluster default network provider. If unset, this property defaults to 'true' and multicast configure is not migrated.
```
kyrtapz, line 389:
```suggestion
                          description: multicast specifies whether or not the multicast configuration is migrated automatically when changing the cluster default network provider. If unset, this property defaults to 'true' and multicast configure is migrated.
```

operator/v1/types_network.go
kyrtapz, line 130: What happens to feature migration when `Features` is not set?

pliurh: This field is under `migration`. The full-path would be in `spec.migration.features`. Would it be clear enough?

pliurh: I added the description for this scenario.

operator/v1/types_network.go
```diff
@@ -123,6 +123,34 @@ type NetworkMigration struct {
 	// changing the MTU for the default network will be rejected.
 	// +optional
 	MTU *MTUMigration `json:"mtu,omitempty"`
+
+	// features contains the features migration configuration. Set this to migrate
+	// feature configuration when changing the cluster default network provider.
```
danwinship, line 128: Do we support feature migration in both directions? Or is this one-way? Should clarify if the latter.

operator/v1/types_network.go
```diff
@@ -123,6 +123,34 @@ type NetworkMigration struct {
...
+	// +kubebuilder:default:=true
+	EgressIP bool `json:"egressIP,omitempty"`
+	// egressFirewall specifies whether or not the EgressFirewall configuration is migrated
+	// automatically when changing the cluster default network provider.
+	// If unset, this property defaults to 'true' and EgressFirewall configure is migrated.
+	// +optional
+	// +kubebuilder:default:=true
+	EgressFirewall bool `json:"egressFirewall,omitempty"`
+	// multicast specifies whether or not the multicast configuration is migrated
+	// automatically when changing the cluster default network provider.
```
danwinship, line 149: Does multicast work _during_ the migration? If not, it would be nice to mention that here.

operator/v1/types_network.go
pliurh: We support in both directions.

operator/v1/types_network.go
pliurh: No, it doesn't. Neither does EgressIP or EgressFirewall since the current migration solution is a disruptive one. The target of feature migration in 4.12 is only to make sure the configuration of these features can be migrated automatically.

operator/v1/types_network.go
```diff
@@ -123,6 +123,34 @@ type NetworkMigration struct {
...
+type FeaturesMigration struct {
+	// egressIP specifies whether or not the EgressIP configuration is migrated
+	// automatically when changing the cluster default network provider.
+	// If unset, this property defaults to 'true' and EgressIP configure is migrated.
+	// +optional
+	// +kubebuilder:default:=true
+	EgressIP bool `json:"egressIP,omitempty"`
+	// egressFirewall specifies whether or not the EgressFirewall configuration is migrated
+	// automatically when changing the cluster default network provider.
+	// If unset, this property defaults to 'true' and EgressFirewall configure is migrated.
```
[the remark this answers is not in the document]
pliurh: fixed.

PR #1256, opened 2022-08-22
ODC-6778: Add console developer catalog customization

debsmita1:
**Story:**
https://issues.redhat.com/browse/ODC-6778

Add dev catalog types customization API based on https://github.com/openshift/enhancements/pull/1206


operator/v1/types_console.go
```diff
@@ -114,7 +114,7 @@ type ConsoleCustomization struct {
 	// SVG format preferred
 	// +optional
 	CustomLogoFile configv1.ConfigMapFileReference `json:"customLogoFile,omitempty"`
-	// developerCatalog allows to configure the shown developer catalog categories.
+	// developerCatalog allows to configure the shown developer catalog categories and it's types.
```
jerolimov, line 117: nit: its types or just types, or?

```suggestion
	// developerCatalog allows to configure the shown developer catalog categories (filters) and types (providers).
```

operator/v1/0000_70_console-operator.crd.yaml
```diff
@@ -122,6 +122,11 @@ spec:
                                 type: array
                                 items:
                                   type: string
+                        disableTypes:
+                          description: disable types which are shown in the developer catalog.
+                          type: array
+                          items:
+                            type: string
```
jerolimov, line 129: There are two spaces behind the type: string, so that the verify step doesn't pass:

```suggestion
                            type: string
```

operator/v1/types_console.go
```diff
@@ -149,6 +149,10 @@ type DeveloperConsoleCatalogCustomization struct {
 	// +kubebuilder:validation:Optional
 	// +optional
 	Categories []DeveloperConsoleCatalogCategory `json:"categories,omitempty"`
+	// disable types which are shown in the developer catalog.
+	// +kubebuilder:validation:Optional
+	// +optional
+	DisableTypes []string `json:"disableTypes,omitempty"`
```
jerolimov, line 155: The first word references the YAML/JSON type anywhere else. Recommend to align the doc to Quickstarts and add actions. See https://github.com/openshift/api/blob/874dd9fe45490eb7347db7422582e43ddffa2e3d/operator/v1/types_console.go#L192-L196 and https://github.com/openshift/api/blob/874dd9fe45490eb7347db7422582e43ddffa2e3d/operator/v1/types_console.go#L200-L206

```suggestion
	// disableTypes is a list of developer catalog IDs that are not shown to users.
	// +kubebuilder:validation:Optional
	// +kubebuilder:validation:MinItems=1
	// +optional
	DisableTypes []string `json:"disableTypes,omitempty"`
```

jerolimov (changes requested):

jerolimov:
/retitle ODC-6778: Add console developer catalog customization

operator/v1/types_console.go
```diff
@@ -149,6 +149,9 @@ type DeveloperConsoleCatalogCustomization struct {
 	// +kubebuilder:validation:Optional
 	// +optional
 	Categories []DeveloperConsoleCatalogCategory `json:"categories,omitempty"`
+	// disabledTypes is a list of developer catalog IDs that are not shown to users.
+	// +optional
+	DisabledTypes []string `json:"disabledTypes,omitempty"`
```
jerolimov, line 154: If I'm correct we agreed on @JoelSpeed idea to add an `enabledTypes` as well, or?

We can maybe also add a hint where to find the types = sub-catalog IDs. Wdyt?

```suggestion
	// enabledTypes is a list of developer catalog types (sub-catalogs IDs) that will be shown to users.
	// Sub-catalogs are added via console plugins, the available types (sub-catalog IDs) are available
	// in the console on the cluster configuration page, or when editing the YAML in the console.
	// +optional
	EnabledTypes []string `json:"enabledTypes,omitempty"`
	// disabledTypes is a list of developer catalog types (sub-catalogs IDs) that are not shown to users.
	// Sub-catalogs are added via console plugins, the available types (sub-catalog IDs) are available
	// in the console on the cluster configuration page, or when editing the YAML in the console.
	// +optional
	DisabledTypes []string `json:"disabledTypes,omitempty"`
```

jerolimov (changes requested):
Just one small comment, but in general this is fine:

operator/v1/types_console.go
jerolimov: > obvious to an end user?

Hopefully, at least the term is used in our UI and in our frontend code. Fyi:

image

jerolimov (approved):

operator/v1/types_console.go
```diff
@@ -143,12 +143,32 @@ type ProjectAccess struct {
...
+	// Types (sub-catalogs) are added via console plugins, the available types (sub-catalog IDs) are available
+	// in the console on the cluster configuration page, or when editing the YAML in the console.
+	// A new type will not be shown to the user until it is added in the enabled list.
+	// If `enabled` list is empty, complete developer catalog should be hidden.
+	// +optional
+	Enabled []string `json:"enabled"`
+	// disabled is a list of developer catalog types (sub-catalogs IDs) that are not shown to users.
+	// Types (sub-catalogs) are added via console plugins, the available types (sub-catalog IDs) are available
+	// in the console on the cluster configuration page, or when editing the YAML in the console.
+	// If all the available sub-catalog types are added in the `disabled` list, complete developer catalog should be hidden.
```
JoelSpeed, line 158: With enabled, there's a way to enable all (this is the default when neither list is configured). We then have a way to enable only certain items, and to disable only certain items.

There's currently no way to permanently disable all items (new catalogs could be added but not in the list).

I wonder if we need a wildcard option, let me see how others tackle this

operator/v1/types_console.go
```diff
@@ -143,12 +143,32 @@ type ProjectAccess struct {
...
+	Disabled []string `json:"disabled,omitempty"`
+}
+
 // DeveloperConsoleCatalogCustomization allow cluster admin to configure developer catalog.
 type DeveloperConsoleCatalogCustomization struct {
 	// categories which are shown in the developer catalog.
 	// +kubebuilder:validation:Optional
 	// +optional
 	Categories []DeveloperConsoleCatalogCategory `json:"categories,omitempty"`
+	// types allows enabling/disabling of sub-catalog types that user can see in the Developer catalog.
```
JoelSpeed, line 169: You need to add a comment that explains what happens when this config is omitted, IMO that should be all catalogs are shown (which would be backwards compatible)

operator/v1/types_console.go
```diff
@@ -143,12 +143,32 @@ type ProjectAccess struct {
 	AvailableClusterRoles []string `json:"availableClusterRoles,omitempty"`
 }
 
+// If sub-catalog type(s) is(are) mentioned in both `enabled` and `disabled` list, the sub-catalog type(s) should be hidden.
```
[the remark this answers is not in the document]
deads2k: > Having reviewed other enabled/disabled pairs, it's an error to have the value in both. I think we should maintain that behaviour here. Do you have a webhook that validates this type already?

We promoted CEL to beta for CRD validation in 1.25.  If there isn't a webhook, I'm open to considering an IOU for CEL validation before we ship 4.12.

operator/v1/types_console.go
deads2k: > I wonder if we need a wildcard option, let me see how others tackle this

Interesting observation.  Kube admission configuration is often used as an example that allows a default list with default on and default off as an option.  Then configuration values that forceEnable and forceDisable.  Very much like (exactly like?) this API.  But we didn't consider the case of, "disable everything".  I'm open to a wildcard value like "*".

JoelSpeed: There is an alternative suggestion to this API as well, which is to make a discriminated union which would look something like below:
```
state: All|Enabled|Disabled|None
enabled: []
disabled: []
```

Though I'm concerned about the 4 values of the enum, I'm not sure it's intuitive that All and Enabled are different from one another, likewise with Disabled and None.

Perhaps if the list was All, Explicit and None and then the structure would be
```
state: All|Explicit|None
explicit:
  enabled: []
  disabled: []
```
I think I prefer the current approach to the DU approach personally

jerolimov: We're fine with both, please let us know which direction we should follow...

JoelSpeed (commented):
Two observations of things to work on here:
- I think we should go with the enabled and disabled lists as they are right now, but I think we should add a special `*` value to allow users to disable all catalogs. Which will then need a comment in the disabled option to explain that
- Is there any way that a user can use a command to look up the valid values for these lists? If not, its my understanding they must look at the console, can we include in the docs a couple of examples of valid values that would be added to these lists

operator/v1/types_console.go
```diff
@@ -143,12 +143,32 @@ type ProjectAccess struct {
...
+	// If the list is non-empty, a new type will not be shown to the user until it is added to list.
+	// If the list is empty or omitted, the complete developer catalog will be shown apart from items explicitly listed in the disabled types list.
+	// +optional
+	Enabled []string `json:"enabled"`
+	// disabled is a list of developer catalog types (sub-catalogs IDs) that are not shown to users.
+	// Types (sub-catalogs) are added via console plugins, the available types (sub-catalog IDs) are available
+	// in the console on the cluster configuration page, or when editing the YAML in the console.
+	// If all the available sub-catalog types are added in the `disabled` list, complete developer catalog should be hidden.
+	// +optional
+	Disabled []string `json:"disabled"`
```
JoelSpeed, line 160: Can we try
```suggestion
	// +kubebuilder:validation:UniqueItems:=true
	Disabled []string `json:"disabled"`
```

operator/v1/types_console.go
```diff
@@ -143,12 +143,32 @@ type ProjectAccess struct {
...
 
+// A sub-catalog type cannot be in both `enabled` and `disabled` list.
+type DeveloperConsoleCatalogTypes struct {
+	// enabled is a list of developer catalog types (sub-catalogs IDs) that will be shown to users.
+	// Types (sub-catalogs) are added via console plugins, the available types (sub-catalog IDs) are available
+	// in the console on the cluster configuration page, or when editing the YAML in the console.
+	// If the list is non-empty, a new type will not be shown to the user until it is added to list.
+	// If the list is empty or omitted, the complete developer catalog will be shown apart from items explicitly listed in the disabled types list.
+	// +optional
+	Enabled []string `json:"enabled"`
```
JoelSpeed, line 154: Can we try
```suggestion
	// +kubebuilder:validation:UniqueItems:=true
	Disabled []string `json:"disabled"`
```

operator/v1/types_console.go
```diff
@@ -143,12 +143,32 @@ type ProjectAccess struct {
 	AvailableClusterRoles []string `json:"availableClusterRoles,omitempty"`
 }
 
+// A sub-catalog type cannot be in both `enabled` and `disabled` list.
+type DeveloperConsoleCatalogTypes struct {
```
JoelSpeed, line 147: We either need a webhook or CEL (https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules) for validating that an item is in either enabled or disabled but not both

The CEL validation would be for checking two listSets are disjoint
```
self.set1.all(e, !(e in self.set2))
```
I'll work out how this is applied as this is new for K8s 1.25

operator/v1/types_console.go
JoelSpeed: In this discriminated union:
* Default state to Enabled for backwards compatibility 
* Enabled + empty enabled[]: show everything
* Enabled + non-empty enabled[]: show only the items in the enabled list
* Disabled + empty disabled[]: disable everything
* Disabled + non-empty disabled[]: disable only the items in the disabled list
```
state: Enabled|Disabled
enabled: []
disabled: []
```
This means that you cannot use enabled and disabled at the same time, which, as we don't have specific lists of enabled and disabled by default (default is all on), we don't need to use them at the same time

operator/v1/types_console.go
debsmita1: we don't need this if we add state as a unionDiscriminator

operator/v1/types_console.go
debsmita1: used `+listType=set`

operator/v1/types_console.go
```diff
@@ -143,12 +143,51 @@ type ProjectAccess struct {
...
+type CatalogTypesState string
+
+const (
+	Enabled  CatalogTypesState = "Enabled"
+	Disabled CatalogTypesState = "Disabled"
+)
+
+// DeveloperConsoleCatalogTypesState defines the state of the sub-catalog types.
+// +union
+type DeveloperConsoleCatalogTypesState struct {
```
JoelSpeed, line 156: Based on the testing I'm doing here you may need a couple of changes to make sure this works. You have a required field within an optional struct, we need to make sure it's omitted correctly.

https://github.com/openshift/api/blob/b4c38862d26f227af4221d01dd41331e2b78201a/example/v1/types_stable.go#L70-L87

I think you need `omitempty` for state, and then we can add the CEL as in that example for the enabled and disabled fields here

JoelSpeed:
Have you done any manual testing of the API to check that when you don't set the developer console new config, that it doesn't complain about a required field? Just want to check the validations here

debsmita1:
@JoelSpeed This is what I have tested
https://user-images.githubusercontent.com/22490998/191085477-7acc8daa-f3ba-4d7b-ad31-edd66cb67b80.mov

JoelSpeed:
Thanks for the video, are you happy to wait for the rebase to 1.25 to land before you merge this? I think it should happen this week. Once 1.25 is merged we can test this properly

The alternative is to build off of #1298 and add some tests for this project based on what we are doing there (eg #1297 for CPMS), that could help prove the validations

jerolimov:
Tested this API via https://github.com/openshift/console-operator/pull/676 on a cluster bot.

Could verify that all the API change works fine.

/label qe-approved
/lgtm

jerolimov:
Propagate from the ODC Epic:
/label px-approved
/label docs-approved

JoelSpeed:
I'm happy with this API and that testing will follow up once the 1.25 rebase has landed, deferring to @deads2k for final approval

authorization/v1/0000_03_authorization-openshift_01_rolebindingrestriction.crd.yaml
```diff
@@ -211,5 +211,3 @@ spec:
                 type: object
             type: object
         type: object
-    served: true
```
deads2k, line 214: why was this removed?

deads2k:
verify passed so I'll assume the change in rolebindingrestrictions was correct.

/approve

JoelSpeed:
/hold

I disagree, I don't think the served and storage versions can be removed, pretty sure if we try to install that CRD into a cluster it will not allow the CRD to be created

JoelSpeed:
Yep, just tried to add an integration test for this and the API returned errors trying to install the CRD
```
                              {
                                  Type: "FieldValueInvalid",
                                  Message: "Invalid value: []apiextensions.CustomResourceDefinitionVersion{apiextensions.CustomResourceDefinitionVersion{Name:\"v1\", Served:false, Storage:false, Deprecated:false, DeprecationWarning:(*string)(nil), Schema:(*apiextensions.CustomResourceValidation)(nil), Subresources:(*apiextensions.CustomResourceSubresources)(nil), AdditionalPrinterColumns:[]apiextensions.CustomResourceColumnDefinition(nil)}}: must have exactly one version marked as storage version",
                                  Field: "spec.versions",
                              },
                              {
                                  Type: "FieldValueInvalid",
                                  Message: "Invalid value: []string(nil): must have at least one stored version",
                                  Field: ...
```

JoelSpeed:
This [commit](https://github.com/JoelSpeed/api/commit/fc37c0532e03bdb6a5c46b641a2ed313860645fc) can help demonstrate that the CRD change here is invalid, if you pull this branch (based on this PR) and from the root of the API repo run `make -C authorization/v1 test` it will error right now

operator/v1/types_console.go
```diff
@@ -213,16 +254,16 @@ type AddPage struct {
...
-	Enabled      PerspectiveState = "Enabled"
-	Disabled     PerspectiveState = "Disabled"
-	AccessReview PerspectiveState = "AccessReview"
+	PerspectiveEnabled      PerspectiveState = "Enabled"
+	PerspectiveDisabled     PerspectiveState = "Disabled"
+	PerspectiveAccessReview PerspectiveState = "AccessReview"
 )
 
-// PerspectiveAccessReview defines the visibility of the perspective depending on the access review checks.
+// ResourceAttributesAccessReview defines the visibility of the perspective depending on the access review checks.
```
jerolimov, line 262: Nit: A generic type name `ResourceAttributesAccessReview` with a really specific job "defines the visibility of the perspective", `PerspectiveVisibilityAccessReview` might be an option - or updating the doc.

jerolimov:
> I disagree, I don't think the served and storage versions can be removed, pretty sure if we try to install that CRD into a cluster it will not allow the CRD to be created

Thanks for finding and blocking this. It's fixed now. As this was the last blocker we can continue with this. We still can add some tests when your testing framework is ready.

/unhold
/lgtm

PR #1260, opened 2022-08-23
Drop GA-ed features (CSIMigrationAWS, CSIMigrationGCE, PodSecurity)

soltysh:
This is needed for landing k8s 1.25, since these 3 features are GA-ed upstream and locked to always on.

/assign @jsafrane @stlaz


jsafrane:
/lgtm
CSI migration changes

PR #1263, opened 2022-08-23
[CNF-5646] Add CSISnapshot capability

qJkee:
Add CSISnapshot capability which allows disabling CSI snapshot
controller operator which is responsible for watching the
VolumeSnapshot CRD objects and manages the creation and
deletion lifecycle of volume snapshots.


deads2k:
/approve
/assign @jsafrane 

Jan for lgtm.

PR #1271, opened 2022-08-26
STOR 437- Create API for vsphere CSI topology configuration

gnufied:
xref https://issues.redhat.com/browse/STOR-437


operator/v1/types_csi_cluster_driver.go
```diff
@@ -96,6 +96,44 @@ type ClusterCSIDriverSpec struct {
...
+	// will be applied.
+	// +kubebuilder:validation:Optional
+	// +optional
+	DriverConfig *CSIDriverConfigSpec `json:"driverConfig,omitempty"`
+}
+
+// CSIDriverConfigSpec defines configuration spec that can be
+// used to optionally configure a specific CSI Driver.
+// +union
+type CSIDriverConfigSpec struct {
```
bertinatto, line 112: I would vote for this API to be platform-agnostic. I think we should consider using a `runtime.RawExtension` here. 

For example, in order to backport a CSI driver configuration to a previous OCP version we would have to backport the API as well. I definitely can see this happening in the future.

gnufied: Not sure I understand. Are we going to backport topology support to 4.11 and older versions of OCP? It could be tricky. Do we have a concrete case where this may happen?

bertinatto: Actually I'm talking about other configurations (vSphere or not) that need to be backported to previous releases in order to fix bugs.

For example, in OCP 4.14 we fix a bug in GCP Filestore that require the user to provide a `clustercsidriver.Spec.Config.NetworkName` field. Then someone asks us to backport that fix to OCP 4.13. If `Config` field is hardcoded type in the API, we will need to backport the API as well. If it's a `runtime.RawExtension`, we don't need to backport the API changes (just the code in the operator).

Do you think this is something that can happen?

BTW, this has happened to me with proxy support. Because `ObservedConfig` was a [raw field](https://github.com/bertinatto/library-go/blob/fef421ec24f46a3abfb64fab52a52585059329f2/vendor/github.com/openshift/api/operator/v1/types.go#L88), we didn't have to backport the API changes.

jsafrane: From user perspective it does not really matter if an API is provided by `RawExtension` or a real structs / fields. It's still an API and they need to know what yaml / json / protobuf files they can use. And we don't backport new APIs, regardless how they are implemented internally.

Therefore I am for proper structs + fields, as they are more explicit.

`ObservedConfig` was technically backport of a new API that nobody noticed.

operator/v1/types_csi_cluster_driver.go
```diff
@@ -96,6 +96,44 @@ type ClusterCSIDriverSpec struct {
...
+	// DriverName indicates which of the driver we are
+	// installing
+	// +kubebuilder:validation:Required
+	// +unionDiscriminator
+	// +required
+	DriverName CSIDriverName `json:"driverName"`
```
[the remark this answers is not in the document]
dobsonj: ^ this is a good point IMO, I don't really understand the purpose of the new `DriverName` field.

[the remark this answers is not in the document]
gnufied: yeah I am going to rename it. But we do need a field here as a union discriminator which will basically make it obvious that if `driverName: vsphere.vmware.com` then - only `vsphere` will be set.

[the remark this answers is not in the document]
gnufied: https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#discriminated-unions

I am torn about renaming `driverName` to `Name`. Elsewhere in OCP docs, union discriminators are called `Type` - if we do that, then we probably have to define a new enum.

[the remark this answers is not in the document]
dobsonj: Ah, I don't have a strong preference on renaming the field, I think `DriverName` is fine. I just didn't understand the purpose, but that doc on discriminated unions explains it.

operator/v1/types_csi_cluster_driver.go
```diff
@@ -96,6 +96,44 @@ type ClusterCSIDriverSpec struct {
...
+	// +kubebuilder:validation:Optional
+	// +optional
+	VSphere *VSphereCSIDriverConfigSpec `json:"vsphere,omitempty"`
+}
+
+// VSphereCSIDriverConfigSpec defines properties that
+// can be configured for vsphere CSI driver.
+type VSphereCSIDriverConfigSpec struct {
+	// TopologyCategories indicates categories with which
+	// vcenter resources as hostcluster or datacenter were tagged with.
```
JoelSpeed, line 131: Can you clarify what this means? It doesn't make sense to me, do you mean `resources such as`?

operator/v1/types_csi_cluster_driver.go
```diff
@@ -96,6 +96,44 @@ type ClusterCSIDriverSpec struct {
...
+	// If unspecified CSI driver configuration will default to configuration
+	// specified in cluster configuration.
```
JoelSpeed, line 133: Where does this configuration live? On the Infrastructure object?

operator/v1/types_csi_cluster_driver.go
```diff
@@ -96,6 +96,44 @@ type ClusterCSIDriverSpec struct {
...
+	// +kubebuilder:validation:Optional
```
JoelSpeed, line 134: Don't need this
```suggestion
```

operator/v1/types_csi_cluster_driver.go
gnufied: Yes - basically what @jcpowermac is proposing in his KEP.

operator/v1/types_csi_cluster_driver.go
```diff
@@ -96,6 +96,44 @@ type ClusterCSIDriverSpec struct {
...
 	// +optional
 	StorageClassState StorageClassStateName `json:"storageClassState,omitempty"`
+
+	// DriverConfig field can be used to optionally specify
+	// driver specific configuration.
+	// If this field is nil, platform chosen default configuration
+	// will be applied.
+	// +kubebuilder:validation:Optional
+	// +optional
+	DriverConfig *CSIDriverConfigSpec `json:"driverConfig,omitempty"`
```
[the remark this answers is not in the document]
gnufied: I was reading about API conventions and although I have accepted this suggestion, one of the things which I find sometimes problematic is - `CSIDriverConfigSpec` says `DriverName` is a *required* field and hence by that definition - an empty `CSIDriverConfigSpec` isn't a valid object because `DriverName` is not set.  I know on the wire, if field is empty - it is entirely omitted but I think that after deserializing we basically have a GO object that is essentially invalid and does not conform to scheme it defines itself, which is weird. 

So I think for union types which have a required union discriminator I think that a nil pointer is only sensible default if users don't want to specify one.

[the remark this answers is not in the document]
JoelSpeed: It's a trade off we have decide to make for the discoverability benefits. But yes, you need to teach your controller to handle the empty `DriverConfig`. It shouldn't be too difficult and isn't dissimilar to the nil check you'd have to otherwise do
```
emptyConfig := operatorv1.CSIDriverConfigSpec{} 
if spec.DriverConfig == emptyConfig {
  // No driver config was specified
  return nil
}
```

operator/v1/types_csi_cluster_driver.go
JoelSpeed: Ok, I'd expect the comment here to mention that it's on the Infrastructure object, gives user a hint as to where they can find the defaults

operator/v1/types_csi_cluster_driver.go
```diff
@@ -96,6 +96,37 @@ type ClusterCSIDriverSpec struct {
...
+}
+
+// CSIDriverConfigSpec defines configuration spec that can be
+// used to optionally configure a specific CSI Driver.
+// +union
+type CSIDriverConfigSpec struct {
+	// driverName indicates which CSI driver should be installed.
+	// +kubebuilder:validation:Required
+	// +unionDiscriminator
+	DriverName CSIDriverName `json:"driverName"`
```
JoelSpeed, line 114: This should be an enum, do we have a list of valid values for this?

operator/v1/types_csi_cluster_driver.go
```diff
@@ -96,6 +96,37 @@ type ClusterCSIDriverSpec struct {
...
+
+	// vsphere is used to configure the vsphere CSI driver.
+	// +optional
+	VSphere *VSphereCSIDriverConfigSpec `json:"vsphere,omitempty"`
```
JoelSpeed, line 118: What value must `DriverName` have for this configuration to be observed?

operator/v1/types_csi_cluster_driver.go
gnufied: I have renamed this to `CSIDriverType` and defined a new enum. @bertinatto raised a concern that we were overloading `CSIDriverName` type which is used for defining driver name.

operator/v1/types_csi_cluster_driver.go
gnufied: done. PTAL

operator/v1/types_csi_cluster_driver.go
gnufied: Added more details. PTAL

operator/v1/types_csi_cluster_driver.go
```diff
@@ -96,6 +96,55 @@ type ClusterCSIDriverSpec struct {
...
+	//
+	// Valid values are:
+	//
+	// * vsphere
+	//
+	// Allows configuration of vsphere CSI driver topology.
+	//
+	// +kubebuilder:validation:Required
+	// +unionDiscriminator
+	DriverType CSIDriverType `json:"driverType"`
```
JoelSpeed, line 130: As this enum is expected over time, you need to add a developer note to explain what consumers should do when they observe an unknown value, there's typically two options there
```suggestion
	// ---
	// + Consumers should treat unknown values as a no-op || Consumers should treat unknown values as an error and stop processing.
	DriverType CSIDriverType `json:"driverType"`
```

JoelSpeed (commented):
I think we are pretty close, a few more comments

operator/v1/types_csi_cluster_driver.go
```diff
@@ -96,6 +96,55 @@ type ClusterCSIDriverSpec struct {
...
+
+	// vsphere is used to configure the vsphere CSI driver.
+	// +optional
+	VSphere *VSphereCSIDriverConfigSpec `json:"vsphere,omitempty"`
```
jsafrane, line 134: If the right `CSIDriverType` is  `"VSphere"`, then this must be `json:"vSphere`

operator/v1/types_csi_cluster_driver.go
```diff
@@ -96,6 +96,55 @@ type ClusterCSIDriverSpec struct {
...
+
+// VSphereCSIDriverConfigSpec defines properties that
+// can be configured for vsphere CSI driver.
+type VSphereCSIDriverConfigSpec struct {
+	// topologyCategories indicates tag categories with which
+	// vcenter resources such as hostcluster or datacenter were tagged with.
+	// If unspecified CSI driver configuration will default to configuration
+	// specified in Infrastructure object. If both Infrastructure object and
+	// ClusterCSIDriver specify topology - then values in Infrastructure object
+	// will take precedence.
```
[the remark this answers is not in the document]
gnufied: Currently its bit like you would expect ebs provisioner as default in a storageclass deployed on AWS cluster not ceph provisioner. One can deploy ceph on OCP AWS cluster sure but it requires bunch of manual configuration. 

Similarly topology from `Infrastructure` object is guaranteed to be right and does not require any further manual configuration in vcenter,  so we want to start with somewhat stricter definition of `topologyCategories` if `Infra` object already defines one. 

Currently we are thinking - once `Infra` changes are merged, we will not allow setting these values at all by end users and any changes to `ClusterCSIDriver` topologies will be rejected by webhook, if `Infra` object already defines a topology(via failure-domains). I will *update* the wording here to be more clear on this. 

But we also realize that, compute and storage topologies can have two different values - so we don't want to close the door completely on that option, so we may choose to relax this in a future version of OCP.

operator/v1/types_csi_cluster_driver.go
```diff
@@ -96,6 +96,55 @@ type ClusterCSIDriverSpec struct {
...
+	// +optional
+	DriverConfig CSIDriverConfigSpec `json:"driverConfig,omitempty"`
+}
+
+// CSIDriverType indicates type of CSI driver being configured.
+// +kubebuilder:validation:Enum=vsphere
+type CSIDriverType string
+
+const (
+	VSphereDriverType CSIDriverType = "vsphere"
```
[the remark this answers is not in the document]
gnufied: fixed.

operator/v1/types_csi_cluster_driver.go
gnufied: fixed.

gnufied:
@JoelSpeed PTAL. Dropped `omitempty` from optional but non-pointer fields and addressed other review comments.

JoelSpeed (commented):
API looks good, have you tested it against a real cluster to make sure it behaves how you expect it should?

gnufied:
@JoelSpeed yep. The entire PR has been verified using linked changes into vmware-vpshere-csi-driver https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/107 and it works as expected.

gnufied:
Since this is just an API change.

/label docs-approved
/label px-approved

PR #1282, opened 2022-09-02
platform/v1alpha1: Update the cluster profile annotations

timflannagan:
Update the PlatformOperator CRD and add the required self-managed-high-availability cluster profile.

Signed-off-by: timflannagan <timflannagan@gmail.com>

### DIFF
--- a/platform/v1alpha1/platformoperators.crd.yaml
+++ b/platform/v1alpha1/platformoperators.crd.yaml
@@ -3,6 +3,9 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     api-approved.openshift.io: https://github.com/openshift/api/pull/1234
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
     release.openshift.io/feature-set: TechPreviewNoUpgrade
   creationTimestamp: null
   name: platformoperators.platform.openshift.io


platform/v1alpha1/platformoperators.crd.yaml
```diff
@@ -3,6 +3,7 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     release.openshift.io/feature-gate: TechPreviewNoUpgrade
+    include.release.openshift.io/self-managed-high-availability: "true"
```
bparees, line 6: is there a reason not to also include this in single node?  (and eventually i'd expect we have it in hypershift too, but we need to think about that one)

timflannagan: Is there any requirement on being "SNO-aware" for using that cluster profile, or are we fine with deploying the HA configuration regardless of cluster topology? I was just keeping the set of changes small, but I don't see any obvious reason we couldn't include the SNO cluster profile annotation too.

bparees: > Is there any requirement on being "SNO-aware" for using that cluster profile

The expectation for the most part is that you set your replica counts to 1 (not relevant to CRDs obviously).  There may be some components that also tune their resource requests down for the SNO version of their resource yaml, i'm not positive.  Any anti-affinity rules should also be configured to accomodate a SNO install (i.e. ensure that you won't end up with a bunch of unschedulable pods)

My understanding of how profiles work, if you do not add the SNO annotation, the resource will not be deployed on SNO installs at all.

So if you want a user to be able to enable this TP feature on a SNO install, you need to either tag this resource w/ it, or introduce a second yaml that is tagged for SNO.  (Since this is a CRD, i don't see why you'd have a second one.  If it was a deployment that was setting a replica count, i could see having two)

platform/v1alpha1/platformoperators.crd.yaml
```diff
@@ -3,6 +3,7 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
```
deads2k, line 4: This looks like a techpreview API.  TechPreview CRDs are not supposed to be installed by default.  Is your operator managing and gating the installation or using the CVO?  if the CVO, you need to add this annotation

```
    release.openshift.io/feature-set: TechPreviewNoUpgrade
```

Even if you're managing the installation, I'd still suggest the annotation to be clear about the intent.

timflannagan: Yep, updated in the latest force push. We're planning on removing the hacks in the o/platform-operators repository once the requisite cluster profile annotations have landed here.

platform/v1alpha1/platformoperators.crd.yaml
timflannagan: Updated in the latest force push and added the SNO cluster profile, and excluded the hypershift (?) cluster profile to match the rest of the platform operator CVO manifests housed in o/platform-operators.

timflannagan:
I wasn't able to easily run any of the Makefile update/verify related targets locally so I had to manually inspect the prowjob failures to determine what changes are needed. The last force push should (hopefully :tm:) fix that failing verification check.

timflannagan:
@bparees Do you have time to take another look at this? We're just removing the hacks that are in-tree for the o/platform-operators repository: https://github.com/openshift/platform-operators/blob/main/Makefile#L80-L83, and it would be nice to get this in before end of sprint 225.

timflannagan:
We're doing the TE slides so manually adding the px-approved label. See https://coreos.slack.com/archives/C030W4V9BHN/p1661875975344669 for more information.

/label px-approved

timflannagan:
Holding for QE approval. This is a no-op while we work towards removing the hacks in the o/platform-operators repository that already contain these cluster profile annotations.

/assign jianzhangbjz
/hold

PR #1283, opened 2022-09-06
[CNF-5647] add ImageRegistry capability

qJkee:
Capability manages the image registry which allows to distribute
Docker images

### DIFF
--- a/config/v1/0000_00_cluster-version-operator_01_clusterversion.crd.yaml
+++ b/config/v1/0000_00_cluster-version-operator_01_clusterversion.crd.yaml
@@ -70,6 +70,7 @@ spec:
                           - Insights
                           - Storage
                           - CSISnapshot
+                          - ImageRegistry
                       x-kubernetes-list-type: atomic
                     baselineCapabilitySet:
                       description: baselineCapabilitySet selects an initial set of optional capabilities to enable, which can be extended via additionalEnabledCapabilities.  If unset, the cluster will choose a default, and the default may change over time. The current default is vCurrent.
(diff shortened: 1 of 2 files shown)


dmage (changes requested):
We cannot make the image registry operator optional until we resolve https://issues.redhat.com/browse/IR-230. Otherwise customers will loose ability to use external registries with their internal CAs.

qJkee:
@dmage 
Can we merge this for now? I've put your epic as a blocker to my task.
The actual implementation would be added later

adambkaplan:
Pipelines Service / App Studio may be interested in this feature. An internal image registry makes less sense when compute is distributed across multiple clusters. Removing the registry restores some resources for compute, and reduces cluster footprint on the cloud provider.

adambkaplan:
cc @coreydaley

dmage:
@qJkee I guess you can merge it (I'm not an approver here), I would like to see documented what'll be removed from the cluster so that customers won't be surprised that the CA management for registries is also removed.

qJkee:
@dmage 
Usually, the maintainer or owner puts lgtm, and later on, it gets approved by David.
By default, this capability would remove all resources related to the image-registry component, such as deployments, namespace, etc.
This is just a name PR to API. Actual changes would be done when we put annotations to target components. And, for sure, we will document it =)

deads2k:
/lgtm cancel
/hold
/assign @dmage 

The name looks ok, but I leave a judgement about merge or not to @dmage.  @dmage your question is reasonable and my advice is that you don't tag this PR until you're satisfied that you know how your component will be made optional without having unexpected side-effects on the cluster.  CA management being a good example.

flavianmissi:
Hi @qJkee, the MCO team has picked up the work that will allow the image registry to become optional. They're targeting 4.14.
Would you be willing to pick this back up?
Thx!

PR #1309, opened 2022-09-27
config: add trustedServiceAccountIssuer to auth config

mfojtik:
`trustedServiceAccountIssuer` is the identifier of the bound service account token issuer the server will trust.

According to the Kubernetes documentation, starting from Kubernetes 1.22, the service-account-issuer flag can be specified multiple times. The first value is then used to generate new tokens and other values are accepted. Using this field can prevent cluster disruptions and allows for smoother reconfiguration of this field.

see: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection


config/v1/types_authentication.go
```diff
@@ -67,6 +67,14 @@ type AuthenticationSpec struct {
 	// duration.
 	// +optional
 	ServiceAccountIssuer string `json:"serviceAccountIssuer"`
+
+	// trustedServiceAccountIssuer is the identifier of the bound service account token issuer the server will trust.
```
stlaz, line 71: // trustedServiceAccountIssuer is an additional Service Account Issuer that the kube-apiserver will tolerate besides the issuer from the ServiceAccountIssuer.

stlaz:
LGTM

PR #1312, opened 2022-09-27
IR-269: Add manifests to images in ImageStreamLayers

dmage:
This allows image stream clients to understand what multi-arch images are consist of.

This is a follow-up to #1211. The apiserver changes are openshift/openshift-apiserver#309. The image-registry (consumer) changes are openshift/image-registry#347.


image/v1/types.go
```diff
@@ -605,6 +605,10 @@ type ImageBlobReferences struct {
 	// not have separate config blobs and this field will be set to nil if so.
 	// +optional
 	Config *string `json:"config" protobuf:"bytes,2,opt,name=config"`
+	// manifests is the list of manifests that compose this image. For a
+	// single architecure image, it is empty. For a multi-arch image, it
+	// consists of single architecture images.
+	Manifests []string `json:"manifests,omitempty" protobuf:"bytes,4,rep,name=manifests"`
```
flavianmissi, line 611: are we gonna need any validation for this?

flavianmissi:
left a question, but /lgtm

image/v1/types.go
dmage: it's a read-only API, so I doubt we need any validation for this

deads2k: > it's a read-only API, so I doubt we need any validation for this

There should still be validation to provide guarantees about the data returned by the openshift-apiserver.  You can wire validation before the return from the REST layer.

image/v1/types.go
```diff
@@ -605,6 +605,10 @@ type ImageBlobReferences struct {
 	// not have separate config blobs and this field will be set to nil if so.
 	// +optional
 	Config *string `json:"config" protobuf:"bytes,2,opt,name=config"`
+	// manifests is the list of manifests that compose this image. For a
+	// single architecure image, it is empty. For a multi-arch image, it
```
deads2k, line 609: This seems weird to me.  Why doesn't a single architecture image contain a single manifest?

image/v1/types.go
```diff
@@ -605,6 +605,10 @@ type ImageBlobReferences struct {
 	// not have separate config blobs and this field will be set to nil if so.
 	// +optional
 	Config *string `json:"config" protobuf:"bytes,2,opt,name=config"`
+	// manifests is the list of manifests that compose this image. For a
```
deads2k, line 608: What is the format of a manifest that can be returned?  Please provide an example.

image/v1/types.go
deads2k, line 611: is this field mutually exclusive to any other fields?  Layers comes to mind.

dmage: what if the generated response doesn't pass validation, should the apiserver panic and crash? as it'll be an indication that there is a bug

image/v1/types.go
flavianmissi: The reason for manifest lists existing is for one image to support multiple architectures. A manifest list could in theory only list a single architecture, but when this is the case people use a single manifest (as opposed to a manifest list). Single manifests do not point to other manifests (instead of manifests they have layers and a config for example), thus should not use this field.

Does that help? Perhaps there's something we can improve in the comment 🤔

image/v1/types.go
dmage: maybe, it doesn't matter for clients

image/v1/types.go
dmage: indeed, I rephrased it make it more clear.

image/v1/types.go
dmage: rephrased it, it's a list of image names.

image/v1/types.go
dmage: it can write:

> For a multi-arch image, it consists of single architecture images, and such images shouldn't have layers nor config.

dmage:
@deads2k please take a look again. Validation is added at https://github.com/openshift/openshift-apiserver/pull/309.

image/v1/types.go
deads2k: The doc now makes sense, but the validation is not yet present in https://github.com/openshift/openshift-apiserver/pull/309/files#diff-4e280f22beaa1bd7ae3747b41a8f5797ad011d86d61adbf82500c73151daeddfR806

deads2k:
reminder for future me:

> The doc now makes sense, but the validation is not yet present in https://github.com/openshift/openshift-apiserver/pull/309/files#diff-4e280f22beaa1bd7ae3747b41a8f5797ad011d86d61adbf82500c73151daeddfR806

deads2k:
/approve

/assign @flavianmissi 

@flavianmissi has lgtm

dmage:
/label px-approved

/assign @stevsmit 
for docs approval

/assign @xiuwang 
for qe approval

PR #1315, opened 2022-09-28
NE-1084: Add additional cache ttl time units

brandisher:
* Adds additional time units to DNSCache to avoid time serialization/deserialization issues like those in https://issues.redhat.com/browse/OCPBUGS-236
* Adds a note about the behavior of fractional second units and units less than `1s`.


jerpeter1:
As this is an API PR, docs & px engagement isn't really needed yet.

/label docs-approved
/label px-approved

Miciah:
For cross-referencing purposes, I'll note that this is a follow-up for <https://github.com/openshift/api/pull/1214>.

PR #1321, opened 2022-10-06
openstack: BGP Load Balancer configuration

pierreprinetti:


config/v1/types_infrastructure.go
```diff
@@ -505,9 +505,100 @@ type BareMetalPlatformStatus struct {
...
+}
+
+// ControlPlaneBGPSpeaker describes the BGP speaker configuration that will
+// be used for the control plane.
+type ControlPlaneBGPSpeaker struct {
+	// subnetID is the subnet which this BGP configuration configuration
+	// applies to.
+	//
+	// +kubebuilder:validation:Required
+	SubnetID string `json:"subnetID"`
```
EmilienM, line 536: I think this is too opinionated to be in the ControlPlaneBGPSpeaker and I think this should be up to each platform where the BGP configuration applies to (here a subnet, but for VMware? Baremetal?).

config/v1/types_infrastructure.go
```diff
@@ -505,9 +505,100 @@ type BareMetalPlatformStatus struct {
...
+	// password for BGP authentication against the peer.
+	//
+	// +optional
+	Password string `json:"password,omitempty"`
+}
+
+// ControlPlaneBGPSpeaker describes the BGP speaker configuration that will
+// be used for the control plane.
+type ControlPlaneBGPSpeaker struct {
+	// subnetID is the subnet which this BGP configuration configuration
```
EmilienM, line 532: configuration x2

config/v1/types_infrastructure.go
mdbooth: @EmilienM I think you're suggesting this should be reusable across multiple providers, which sounds like a good goal. However, for OpenStack we are going to want this reference to be by SubnetID. Where does the SubnetID go?

Off the top of my head I can think of 2 general approaches:

1. There's already an idiom in OpenShift of cloud-specific discriminated unions. In this case the discriminator would be the platform type in Infrastructure, but we'd still end up with an entry here for every supported cloud provider, and we'd match based on whichever was in use.

2. We turn this into a generic reference, probably a name. We then need a mapping elsewhere of subnetID to name. Where would that be?

config/v1/types_infrastructure.go
```diff
@@ -505,9 +505,81 @@ type BareMetalPlatformStatus struct {
...
+	//
+	// +kubebuilder:validation:Required
+	// + ---
+	// + {IPv4 or IPv6} validation in crd.yaml-patch
+	IP string `json:"ip"`
+
+	// password for BGP authentication against the peer.
+	//
+	// +optional
+	Password string `json:"password,omitempty"`
```
deads2k, line 526: no confidential information in CRs. This must reference a secret in ns/openshift-config.

config/v1/types_infrastructure.go
```diff
@@ -505,9 +505,81 @@ type BareMetalPlatformStatus struct {
 	NodeDNSIP string `json:"nodeDNSIP,omitempty"`
 }
 
+// BGPPeer describes a BGP peer.
+type BGPPeer struct {
+	// asn is the Autonomous System Number of the peer.
```
deads2k, line 510: Example please.

config/v1/types_infrastructure.go
```diff
@@ -505,9 +505,81 @@ type BareMetalPlatformStatus struct {
...
+	//
+	// +optional
+	ASN string `json:"asn,omitempty"`
```
deads2k, line 513: no formatting rules for this?

config/v1/types_infrastructure.go
```diff
@@ -505,9 +505,81 @@ type BareMetalPlatformStatus struct {
...
+	// +kubebuilder:validation:MinItems:=1
+	// +listType=map
+	// +listMapKey=subnetID
+	Speakers []OpenStackBGPSpeaker `json:"speakers"`
+}
+
 // OpenStackPlatformSpec holds the desired state of the OpenStack infrastructure provider.
 // This only includes fields that can be modified in the cluster.
-type OpenStackPlatformSpec struct{}
+type OpenStackPlatformSpec struct {
```
deads2k, line 563: if this information about how to write and modify a BGP API s supposed to function for other platforms at some future date (as indicated in the enhancement), please open another PR showing us what that would look like.  This doesn't immediately look so much like infrastructure versus networking configuration of some kind.  Especially if the information collected is actually generically used by other platforms.

config/v1/types_infrastructure.go
EmilienM: done

config/v1/types_infrastructure.go
EmilienM: ack, I'll work on this next. Thanks

config/v1/types_infrastructure.go
```diff
@@ -505,9 +505,73 @@ type BareMetalPlatformStatus struct {
 	NodeDNSIP string `json:"nodeDNSIP,omitempty"`
 }
 
+// BGPPeer describes a BGP peer.
+type BGPPeer struct {
+	// asn is the Autonomous System Number of the peer.
+	//
+	// +optional
+	ASN string `json:"asn,omitempty"`
```
EmilienM, line 513: @pierreprinetti @mdbooth I wonder if we can remove this one, since we already have the ASN in the speaker configuration. I don't think a node would peer with more than one ASNs (not sure it's even possible)...

config/v1/types_infrastructure.go
EmilienM: done in next patchset.

config/v1/types_infrastructure.go
EmilienM: done in next patchset.

config/v1/types_infrastructure.go
EmilienM: I'm not sure yet how to address this; since this whole thing is being consumed by a static pod which has no access to ConfigMaps or secrets for now. I'll investigate with @mdbooth on how we can achieve this.

PR #1323, opened 2022-10-10
OCPBUGS-2163: [release-4.10] operator: add service account issuers into kubernetes apiserver operator

mfojtik:
backports https://github.com/openshift/api/pull/1313


mfojtik:
adding lgtm back after fixed the protoc diff using go 1.17

PR #1326, opened 2022-10-13
operator: fix typo in kubeapiserver serviceaccountissuer field

mfojtik:


mfojtik:
This does not need a valid-bug as it is a trivial typo fix with no functional change, as the PR that introduces this to KAS operator has not merged yet. Skipping the paperwork process here as it does not make sense, and it was followed in the original PR.

PR #1333, opened 2022-10-17
OCPBUGS-2297: Empty string should be valid for i18n loadType

jhadvig:
/assign @JoelSpeed


console/v1/types_console_plugin.go
```diff
@@ -50,6 +50,9 @@ const (
 	// Lazy wont preload any plugin's localization resources, instead
 	// will leave thier loading to runtime's lazy-loading.
 	Lazy LoadType = "Lazy"
+	// Empty is the default value of the LoadType field and it's
+	// purpose is to improve discoverability of the field.
+	Empty LoadType = ""
```
JoelSpeed, line 55: What is the behaviour when this is empty?

Just noticed that the docs on LoadType are a little light, can we expand those?

```
// loadType indicates how the plugin's localization resource should be loaded.
// Valid values are Preload, Lazy and the empty string.
// When set to Preload, all localization resources are fetched when the plugin is loaded.
// When set to Lazy, localization resources are lazily loaded as and when they are required by ...
// When omitted or set to the empty string, the behaviour is equivalent to ...
```

JoelSpeed:
We are making these changes for two reasons, one is to improve discoverability, secondly because of complications in the way the conversion webhooks are built.

It was found that the conversion webhook was initializing the loadType field as an empty string, this was then stored and became an invalid object (ie not passing the validation). This meant the garbage collector couldn't remove the object.

By adding the empty string as a valid value and removing `omitempty` from the parent struct, this will mean the value is marshalled even when empty, improving discoverability, and removing this GC bug

JoelSpeed (commented):
/lgtm

JoelSpeed:
/lgtm cancel

Verify failed because the openapi hasn't been updated, please run `make update-openapi`

JoelSpeed:
/lgtm

/hold

Could we get QE to pre-merge verify this by building a branch for them to test that includes these updates?

jhadvig:
Can confirm that the fix is solving the issue when trying to delete the CR on the foreground

Screenshot 2022-10-19 at 15 27 57

/hold cancel

PR #1339, opened 2022-11-01
config/v1/types_cluster_version: Add Architecture to DesiredUpdate

jottofar:
field of ClusterVersion spec to request target architecture per [OTA-817](https://issues.redhat.com/browse/OTA-817).


config/v1/types_cluster_version.go
```diff
@@ -42,11 +42,16 @@ type ClusterVersionSpec struct {
...
-	// an update exists with that version in the availableUpdates or history.
+	// the current architecture or version does not match the desired
+	// architecture or version). The set of recommended update values is
+	// listed as part of available updates in status, and setting values
+	// outside that range may cause the upgrade to fail. You may specify the
+	// version field without setting architecture or image if an update exists
+	// with that version in the availableUpdates or history. Upgrading from a
+	// homogeneous to a heterogeneous cluster architecture at the same version
+	// can be done by setting architecture to "multi". Image can be used to
+	// upgrade from a homogeneous to a heterogeneous cluster architecture at a
```
wking, line 53: I'd rather drop this sentence, because it's a "this is possible" safety valve, and not a "this is encouraged" UX.

config/v1/types_cluster_version.go
```diff
@@ -224,6 +229,17 @@ type UpdateHistory struct {
 // ClusterID is string RFC4122 uuid.
 type ClusterID string
 
+// ClusterVersionArchitecture enumerates valid cluster architectures.
+// +kubebuilder:validation:Enum=multi
+type ClusterVersionArchitecture string
+
+const (
+	// ClusterVersionArchitectureMulti identifies a heterogeneous architecture.
+	// A heterogeneous cluster is a cluster with nodes running on different
```
wking, line 238: Maybe:

> A heterogeneous cluster is capable of running nodes with different architectures.

or some such to distinguish between "is capable of running nodes with different architectures" (all clusters with heterogeneous-arch release images) and "is actually running nodes with different architectures" (most clusters with heterogeneous-arch release images)?

config/v1/types_cluster_version.go
```diff
@@ -408,6 +424,15 @@ type URL string
 // Update represents an administrator update request.
 // +k8s:deepcopy-gen=true
 type Update struct {
+	// architecture is an optional field that indicates the desired
+	// value of the cluster architecture. In this context cluster
+	// architecture means either homogeneous or heterogeneous. When
+	// this field is part of spec, architecture is ignored if image is
```
wking, line 430: > When this field is part of spec...

We can drop this, because "Update represents an administrator update request." scoping on the type a few lines up, and:

```console
$ git --no-pager grep 'Update `json:' origin/pr/1339 config/v1/types_cluster_version.go
origin/pr/1339:config/v1/types_cluster_version.go:      DesiredUpdate *Update `json:"desiredUpdate,omitempty"`
origin/pr/1339:config/v1/types_cluster_version.go:      ConditionalUpdates []ConditionalUpdate `json:"conditionalUpdates,omitempty"`
```

mean that `Update` will only ever be part of the `spec` or other locations that represent admin requests.

config/v1/types_cluster_version.go
```diff
@@ -408,6 +422,15 @@ type URL string
 // Update represents an administrator update request.
 // +k8s:deepcopy-gen=true
 type Update struct {
+	// architecture is an optional field that indicates the desired
+	// value of the cluster architecture. In this context cluster
+	// architecture means either homogeneous or heterogeneous.
+	// Architecture is ignored if image is specified. The empty
+	// value indicates that the field is not set.
```
deads2k, line 429: What happens if it is not set?

config/v1/types_cluster_version.go
```diff
@@ -408,6 +422,15 @@ type URL string
...
+	//
+	// +optional
+	Architecture ClusterVersionArchitecture `json:"architecture"`
```
deads2k, line 432: Why does this need to be specified for every update request?

wking: it does not need to be specified, which is why it is `+optional`

config/v1/types_cluster_version.go
wking: if `image` is set, both `version` and `architecture` are ignored.  Which is covered here with "Architecture is ignored if image is specified".

If `image` is not set, `version` and `architecture` are still both optional when looking up targets in `availableUpdates`, and both default to "the release that I'm currently reconciling".  Maybe we can cover that with wording in `desiredUpdate` like:

> desiredUpdate is an optional field that indicates the desired value of the cluster version. Setting this value will request an update if the requested target does not match the current target release.  The set of recommended update values is listed as part of available updates in status, and setting values outside that range may cause the upgrade to fail.  You may specify an image pullspec to request updates regardless of the recommended available updates.  In the absence of an image pullspec, you may set either or both of version and architecture to manage lookup in available updates, with both properties defaulting to the value for the current target release.

or some such...

jottofar: Right. I generally followed the pattern of `version` but can add something explicit.

deads2k: why should we allow an ignored field to be set instead of rejecting ignored and potentially conflicting data and forcing the caller to write a guaranteed non-conflicting manifest?

config/v1/types_cluster_version.go
deads2k: Cute, but doesn't address the problem.  If I want an architecture for my cluster, why is the architecture not "sticky" across whatever updates I may choose?  it seems more natural to say, "my cluster has this architecture" and separately, "now please update to this version".

config/v1/types_cluster_version.go
jottofar: > What happens if it is not set?
The short answer is we use the current value of architecture which CVO knows. 

> why should we allow an ignored field to be set instead of rejecting ignored
Assuming this is, if `architecture` is set why should we allow `version` to be set? If so, I'm not sure we do. When they make the first update to heter they'll set arch to `multi` and version should not be set because they can only upgrade to current version. After that there's no reason to ever set `arch` again because it can only be `multi' or "not set" so everything they need to do can be done by just setting version (or image).

config/v1/types_cluster_version.go
jottofar: Ties into my above comment as well. Arch is "sitcky" in that once it's set to `multi` it doesn't need to be set agai since it can only be set to `multi` anyway.

deads2k: > Ties into my above comment as well. Arch is "sitcky" in that once it's set to `multi` it doesn't need to be set agai since it can only be set to `multi` anyway.

Where is this other arch field?  The DesiredUpdate field where I remember this being used is set every time that I want to issue an update and can be cleared to nil as I recall.

jottofar: CVO knows the current architecture by extracting it from the current payload, i.e. if it's "multi" otherwise it sets it using GOARCH as before. Since we couldn't come up with any other consumer of current Architecture we see no need to have it in ClusterVersion status.

config/v1/types_cluster_version.go
```diff
@@ -224,6 +227,17 @@ type UpdateHistory struct {
 // ClusterID is string RFC4122 uuid.
 type ClusterID string
 
+// ClusterVersionArchitecture enumerates valid cluster architectures.
+// +kubebuilder:validation:Enum=multi
+type ClusterVersionArchitecture string
+
+const (
+	// ClusterVersionArchitectureMulti identifies a heterogeneous architecture.
+	// A heterogeneous cluster is capable of running nodes with different
```
wking, line 236: nit: maybe `different` -> `multiple`, to tie into `multi` more clearly?

config/v1/types_cluster_version.go
```diff
@@ -224,6 +227,17 @@ type UpdateHistory struct {
...
 
+// ClusterVersionArchitecture enumerates valid cluster architectures.
+// +kubebuilder:validation:Enum=multi
+type ClusterVersionArchitecture string
+
+const (
+	// ClusterVersionArchitectureMulti identifies a multi architecture. A
+	// multi architecture cluster, also known as a heterogeneous cluster,
+	// is capable of running nodes with multiple architectures.
+	ClusterVersionArchitectureMulti ClusterVersionArchitecture = "multi"
```
deads2k, line 238: enumerations are CamelCase, so Multi

config/v1/types_cluster_version.go
```diff
@@ -406,8 +420,19 @@ type ComponentOverride struct {
...
 // Update represents an administrator update request.
+// +kubebuilder:validation:XValidation:rule="has(self.architecture) && has(self.version) ? (size(self.architecture) == 0 || size(self.version) == 0) : true",message="cannot set both Architecture and Version"
 // +k8s:deepcopy-gen=true
 type Update struct {
+	// architecture is an optional field that indicates the desired
+	// value of the cluster architecture. In this context cluster
+	// architecture means either a single architecture (homogeneous)
+	// or a multi architecture (heterogeneous). Architecture is
+	// ignored if image is specified. The empty value indicates that
+	// the field is not set.
```
deads2k, line 431: What happens if this is not set?  Does the CVO inspect the nodes in my cluster and make the best choice it can?

config/v1/types_cluster_version.go
```diff
@@ -406,8 +420,19 @@ type ComponentOverride struct {
 type URL string
 
 // Update represents an administrator update request.
+// +kubebuilder:validation:XValidation:rule="has(self.architecture) && has(self.version) ? (size(self.architecture) == 0 || size(self.version) == 0) : true",message="cannot set both Architecture and Version"
```
deads2k, line 423: the godoc below talks about architecture and image not working properly together, but doesn't indicate that architecture and version cannot be set together.  Which is correct, the godoc or the CEL?

deads2k, line 423: this seems to be missing a test. Have a look at how to create a testing yaml in the example api group.

config/v1/types_cluster_version.go
jottofar: Can we leave it as "multi" since that's what's being used in the payload metadata which CVO checks to set architecture? A big part of the reason to use multi rather than hetero was to avoid a disconnect here.

config/v1/types_cluster_version.go
jottofar: updated godoc

jottofar: added test

config/v1/types_cluster_version.go
deads2k: > Can we leave it as "multi" since that's what's being used in the payload metadata which CVO checks to set architecture?  A big part of the reason to use multi rather than hetero was to avoid a disconnect here.

No.  Our API is consistent with kube standards to work with kube tools and match kube expectations.  There's more detail here, but we follow kube guidance regardless of the API this closely related to: https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#prefer-consistency-with-kubernetes-style-apis

config/v1/types_cluster_version.go
```diff
@@ -41,12 +41,39 @@ type ClusterVersionSpec struct {
...
+	// version is optional if image is specified. version is not allowed
+	// to be specified if architecture is specified and will default to
+	// the current target release's version.
+	//
+	// image is a container image location that contains the update.
+	// image is optional if version is specified and an update exists
+	// with that version in the availableUpdates or history. image is
+	// also optional if architecture is specified.
+	//
+	// force allows an administrator to update to an image that has failed
```
wking, line 71: `force` is orthogonal to the other `Update` properties, so I don't think we need to discuss it here.

jottofar: Based on your google doc updates you removed the other property descriptions and left `force` in so that's what I've done.

config/v1/types_cluster_version.go
```diff
@@ -406,17 +432,34 @@ type ComponentOverride struct {
...
+	// only allowing updates from single to multi architecture. If
+	// architecture is set image cannot be set and version must be
+	// set. The empty value indicates that architecture is not set
+	// in which case architecture defaults to the current target
+	// release's architecture.
+	//
+	// +optional
+	Architecture ClusterVersionArchitecture `json:"architecture"`
+
+	// version is a semantic versioning identifying the update version.
```
wking, line 452: nit: "versioning" -> "version"?

config/v1/types_cluster_version.go
```diff
@@ -406,17 +432,34 @@ type ComponentOverride struct {
 type URL string
 
 // Update represents an administrator update request.
+// +kubebuilder:validation:XValidation:rule="has(self.architecture) && has(self.image) ? (size(self.architecture) == 0 || size(self.image) == 0) : true",message="cannot set both Architecture and Image"
+// +kubebuilder:validation:XValidation:rule="has(self.architecture) && size(self.architecture) != 0 ? size(self.version) != 0 : true",message="Version must be set if Architecture is set"
```
wking, line 436: Can we simplify (and use cluster-user-facing downcased JSON/YAML property names) with:

```go
// +kubebuilder:validation:XValidation:rule="size(self.architecture) > 0 ? size(self.image) == 0,message="cannot set both architecture and image"
// +kubebuilder:validation:XValidation:rule="size(self.architecture) > 0 ? size(self.version) > 0 : true",message="version must be set if architecture is set"
```

config/v1/stable.clusterversion.testsuite.yaml
```diff
@@ -14,3 +14,108 @@ tests:
...
+        desiredUpdate:
+          image: bar
+    updated: |
+      apiVersion: config.openshift.io/v1
+      kind: ClusterVersion
+      spec:
+        clusterID: foo
+        desiredUpdate:
+          architecture: Multi
+          version: 4.11.1
```
wking, line 119: I'd expect us to leave off `version` for the `Should not allow architecture to be set if image set` test-case.  And I don't understand why we're using `initial`/`updated` for this test-case, but `initial`/`expected` for the earlier test-cases.  Can we continue with the `initial`/`expected` pattern, since none of our Common Expression Language (CEL) evaluation rules depend on prior state?

config/v1/types_cluster_version.go
```diff
@@ -406,17 +432,34 @@ type ComponentOverride struct {
...
+	// set. The empty value indicates that architecture is not set
+	// in which case architecture defaults to the current target
+	// release's architecture.
+	//
+	// +optional
+	Architecture ClusterVersionArchitecture `json:"architecture"`
+
+	// version is a semantic versioning identifying the update version.
+	// version is optional if image is specified and required if
+	// architecture is specified.
```
deads2k, line 454:
```suggestion
	// version is ignored if image is specified and required if
	// architecture is specified.
	// If version is set, image should be empty.
```

config/v1/types_cluster_version.go
```diff
@@ -406,17 +432,34 @@ type ComponentOverride struct {
...
 	// +optional
 	Version string `json:"version"`
 
-	// image is a container image location that contains the update. When this
-	// field is part of spec, image is optional if version is specified and the
-	// availableUpdates field contains a matching version.
+	// image is a container image location that contains the update.
+	// image is optional if version is specified and an update exists
+	// with that version in the availableUpdates or history. image
+	// cannot be specified if architecture is specified.
```
deads2k, line 462:
```suggestion
	// image is a container image location that contains the update.
	// image should be used when the desired version does not exist in availableUpdates or history.
	// When image is set, version is ignored.  When image is set, version should be empty.
	// When image is set, architecture cannot be specified.
```

config/v1/types_cluster_version.go
```diff
@@ -40,13 +40,29 @@ type ClusterVersionSpec struct {
...
+	// Several update properties are involved in selecting a target
+	// release image. The set of recommended update values is listed as
+	// part of available updates in status, and setting values outside
+	// that range may cause the upgrade to fail. In some cases where
+	// the set of available updates does not include the desired target
+	// release image, setting the 'image' property extends the set of
+	// available updates to include the provided image. The
+	// cluster-version operator then filters the possibly-extended set
+	// of available updates to find an entry that matches all of the
+	// provided desired properties. Currently 'image', 'version', and
```
deads2k, line 56: This isn't what actually happens and we discussed the prereqs needed before we make this change.

Right now, the goal of this struct is produce an image.  The cases are.

1. image is specified, version is specified, architecture is specified.  API validation error.
2. image is specified, version is specified, architecture is not specified.   You should not do this.  version is silently ignored and image is used.
3. image is specified, version is not specified, architecture is specified.  API validation error
4. image is specified, version is not specified, architecture is not specified.  image is used.
5. image is not specified, version is specified, architecture is specified.  version and desired architecture are used to select an image.
6. image is not specified, version is specified, architecture is not specified.  version and current architecture are used to select an image.
7. image is not specified, version is not specified, architecture is specified.  API validation error
8. image is not specified, version is not specified, architecture is not specified.  API validation error

config/v1/types_cluster_version.go
jottofar: If you don't check that the property exists with `has` first and it's not there `size` gets `no such key` errors.

config/v1/types_cluster_version.go
jottofar: >When image is set, version is ignored.  When image is set, version should be empty.

So this just screams, then why in the heck are "you" forcing me to set it?
>`...message="Version must be set if Architecture is set"`

config/v1/types_cluster_version.go
wking: These are not pointer properties.  What does "it's not there" mean?  Is that "`self` is `nil`" when the whole `desiredUpdate` is unset?

config/v1/types_cluster_version.go
```diff
@@ -40,13 +40,29 @@ type ClusterVersionSpec struct {
...
+	// 'architecture'. If architecture is set image cannot be set and
+	// version must be set.
+	//
+	// force allows an administrator to update to an image that has failed
+	// verification or upgradeable checks. This option should only
+	// be used when the authenticity of the provided image has been verified out
+	// of band because the provided image will run with full administrative access
+	// to the cluster. Do not use this flag with images that comes from unknown
+	// or potentially malicious sources.
```
deads2k, line 65:
```suggestion
	// desiredUpdate is an optional field that indicates the desired value of
	// the cluster version. Setting this value will trigger an upgrade (if
	// the current version does not match the desired version). The set of
	// recommended update values is listed as part of available updates in
	// status, and setting values outside that range may cause the upgrade
	// to fail.
	
	// Some of the fields are inter-related with restrictions and meanings described here.
	// 1. image is specified, version is specified, architecture is specified. API validation error.
	// 2. image is specified, version is specified, architecture is not specified. You should not do this. version is silently ignored and image is used.
	// 3. image is specified, version is not specified, architecture is specified. API validation error
	// 4. image is specified, version is not specified, architecture is not specified. image is used.
	// 5. image is not specified, version is specified, architecture is specified. version and desired architecture are used to select an image.
	// 6. image is not specified, version is specified, architecture is not specified. version and current architecture are used to select an image.
	// 7. image is not specified, version is not specified, architecture is specified. API validation error
	// 8. image is not specified, version is not specified, architecture is not specified. API validation error
```

config/v1/types_cluster_version.go
JoelSpeed: When CEL validation occurs, it uses an unstructured object during the validation so it only knows about fields that are represented in the `yaml` that is sent to the API. That means that there is a possibility that fields are not present, which then causes the validation to fail.

Having the `has` check is the correct way to prevent errors for fields that are optional in our APIs.

jottofar: I know it fails the test cases when not there. My thought is that it's because they're optional, and therefore not there, checking them with anything like `size` causes a failure.

config/v1/types_cluster_version.go
jottofar: done

config/v1/types_cluster_version.go
jottofar: done

config/v1/types_cluster_version.go
jottofar: done

config/v1/types_cluster_version.go
jottofar: done

config/v1/types_cluster_version.go
jottofar: removed `force`

config/v1/types_cluster_version.go
```diff
@@ -224,6 +233,16 @@ type UpdateHistory struct {
 // ClusterID is string RFC4122 uuid.
 type ClusterID string
 
+// ClusterVersionArchitecture enumerates valid cluster architectures.
+// +kubebuilder:validation:Enum=Multi
```
deads2k, line 237: Enum has to allow "" in order to handle a golang json serialization.

config/v1/types_cluster_version.go
```diff
@@ -406,17 +425,32 @@ type ComponentOverride struct {
...
+// +kubebuilder:validation:XValidation:rule="has(self.architecture) && has(self.image) ? (size(self.architecture) == 0 || size(self.image) == 0) : true",message="cannot set both Architecture and Image"
+// +kubebuilder:validation:XValidation:rule="has(self.architecture) && size(self.architecture) != 0 ? size(self.version) != 0 : true",message="Version must be set if Architecture is set"
 // +k8s:deepcopy-gen=true
 type Update struct {
-	// version is a semantic versioning identifying the update version. When this
-	// field is part of spec, version is optional if image is specified.
+	// architecture is an optional field that indicates the desired
+	// value of the cluster architecture. In this context cluster
+	// architecture means either a single architecture or a multi
+	// architecture. architecture can only be set to multi thereby
```
deads2k, line 435:
```suggestion
	// architecture. architecture can only be set to Multi thereby
```

LalatenduMohanty: Fixed in https://github.com/openshift/api/commit/b941ee060e410404a9ea36293eeed4a776ce7c1f

config/v1/types_cluster_version.go
LalatenduMohanty: Fixed in https://github.com/openshift/api/commit/32939f536c1b41128f58f87a26534f3bca036359

config/v1/types_cluster_version.go
```diff
@@ -406,17 +425,32 @@ type ComponentOverride struct {
 type URL string
 
 // Update represents an administrator update request.
+// +kubebuilder:validation:XValidation:rule="has(self.architecture) && has(self.image) ? (size(self.architecture) == 0 || size(self.image) == 0) : true",message="cannot set both Architecture and Image"
+// +kubebuilder:validation:XValidation:rule="has(self.architecture) && size(self.architecture) != 0 ? size(self.version) != 0 : true",message="Version must be set if Architecture is set"
```
JoelSpeed, line 429: Do you need to use size here? I would have expected a string comparison to be used here
```suggestion
// +kubebuilder:validation:XValidation:rule="has(self.architecture) && self.architecture != '' ? self.version != '' : true",message="Version must be set if Architecture is set"
```

LalatenduMohanty: No, I do not think we need to use size here.

JoelSpeed (commented):
Couple of questions

Also, I broke the verify job slightly so I think your CRD is out of date, verify job is fixed now

/test verify

config/v1/types_cluster_version.go
```diff
@@ -406,17 +425,32 @@ type ComponentOverride struct {
 type URL string
 
 // Update represents an administrator update request.
+// +kubebuilder:validation:XValidation:rule="has(self.architecture) && has(self.image) ? (self.architecture != '' || self.image != '') : true",message="cannot set both Architecture and Image"
```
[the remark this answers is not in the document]
LalatenduMohanty: Yes, that's correct. Thats why we have `message="cannot set both Architecture and Image"`

config/v1/stable.clusterversion.testsuite.yaml
```diff
@@ -14,3 +14,108 @@ tests:
...
+  - name: Version must be set if architecture is set
+    initial: |
+      apiVersion: config.openshift.io/v1
+      kind: ClusterVersion
+      spec:
+        clusterID: foo
+        desiredUpdate:
+          architecture: Multi
+    expectedError: "Version must be set if Architecture is set"
+  onUpdate:
```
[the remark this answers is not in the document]
LalatenduMohanty: Ok, I am going to remove this test.

config/v1/stable.clusterversion.testsuite.yaml
```diff
@@ -14,3 +14,108 @@ tests:
...
+        desiredUpdate:
+          architecture: Multi
+          version: 4.11.1
+    expected: |
+      apiVersion: config.openshift.io/v1
+      kind: ClusterVersion
+      spec:
+        clusterID: foo
+        desiredUpdate:
+          architecture: Multi
```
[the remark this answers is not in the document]
LalatenduMohanty: Fixed in https://github.com/openshift/api/commit/c9e0d99bbe4f6368cedcee287fc299ff6bb89711

config/v1/stable.clusterversion.testsuite.yaml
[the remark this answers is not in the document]
JoelSpeed: I had meant the entire set of onUpdate, none of them seem necessary since they all duplicate previous test cases IIUC. But I see now why you've got two different ones, it's the transitions from image to both and from architecture to both, makes sense, we should add an onCreate version though

config/v1/types_cluster_version.go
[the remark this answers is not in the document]
JoelSpeed: Yes, but the validation itself is wrong, and you can see this if you look at the integration test failures which are reporting `expected an error, but got nil`. 

Your current check passes if either value is non-empty. You want it to only pass when one value is empty.

I created https://github.com/openshift/api/commit/31656002d08a594d4a37cb5be893f1d115ca5870 which demonstrates the diff you need, can you cherry pick that into this branch please

```suggestion
// +kubebuilder:validation:XValidation:rule="has(self.architecture) && has(self.image) ? (self.architecture == '' || self.image == '') : true",message="cannot set both Architecture and Image"
```

config/v1/stable.clusterversion.testsuite.yaml
```diff
@@ -14,3 +14,114 @@ tests:
...
+        desiredUpdate:
+          version: 4.11.1
+  - name: Should allow architecture to be empty
+    initial: |
+      apiVersion: config.openshift.io/v1
+      kind: ClusterVersion
+      spec:
+        clusterID: foo
+        desiredUpdate:
+          architecture:
```
JoelSpeed, line 54: I would prefer to see this explicitly with the empty string
```suggestion
          architecture: ""
```

config/v1/types_cluster_version.go
```diff
@@ -406,17 +425,32 @@ type ComponentOverride struct {
...
+// +kubebuilder:validation:XValidation:rule='has(self.architecture) && has(self.image) ? (self.architecture == "" || self.image == "") : true',message="cannot set both Architecture and Image"
+// +kubebuilder:validation:XValidation:rule='has(self.architecture) && self.architecture != "" ? self.version != "" : true',message="Version must be set if Architecture is set"
 // +k8s:deepcopy-gen=true
 type Update struct {
-	// version is a semantic versioning identifying the update version. When this
-	// field is part of spec, version is optional if image is specified.
+	// architecture is an optional field that indicates the desired
+	// value of the cluster architecture. In this context cluster
+	// architecture means either a single architecture or a multi
+	// architecture. architecture can only be set to Multi thereby
```
JoelSpeed, line 435: So there's no way to transition back to the a single arch cluster?

I wonder if we need to explicitly say, "on future updates, if the existing cluster image is multi architecture, then the downloaded image will also be multi architecture. There is no way to convert a cluster from multi to single architecture."

I think I missed this while reviewing before but it seems odd that on future updates, having the field omitted/empty relies on the existing architecture

> There is no way to convert a cluster from multi to single architecture

Thinking about this, which I got from trying to understand this API, that's not actually true is it, if I forced an image into the update that was single arch I could force the cluster back to single arch, is that ok?

config/v1/types_cluster_version.go
```diff
@@ -406,17 +425,32 @@ type ComponentOverride struct {
...
+	// only allowing updates from single to multi architecture. If
+	// architecture is set image cannot be set and version must be
+	// set.
```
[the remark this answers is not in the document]
LalatenduMohanty: Agree, will fix it.

config/v1/types_cluster_version.go
```diff
@@ -406,17 +425,32 @@ type ComponentOverride struct {
 type URL string
 
 // Update represents an administrator update request.
+// +kubebuilder:validation:XValidation:rule='has(self.architecture) && has(self.image) ? (self.architecture == "" || self.image == "") : true',message="cannot set both Architecture and Image"
+// +kubebuilder:validation:XValidation:rule='has(self.architecture) && self.architecture != "" ? self.version != "" : true',message="Version must be set if Architecture is set"
```
JoelSpeed, line 429: After spending a little time in the debugger, I found out you can't use single quotes to surround these, you have to use double quotes, can you switch the single for double and vice versa in this line

config/v1/types_cluster_version.go
LalatenduMohanty: > Thinking about this, which I got from trying to understand this API, that's not actually true is it, if I forced an image into the update that was single arch I could force the cluster back to single arch, is that ok?

That is right. Even if the current ask is only one time migration from single arch to multi arch , we do not want to add any more code to restrict the opposite. Officially we will only document that single to multi is supported. But there will be room for future change of scope.

config/v1/types_cluster_version.go
LalatenduMohanty: yeah, will fix it.

config/v1/types_cluster_version.go
```diff
@@ -406,17 +425,32 @@ type ComponentOverride struct {
...
 // +k8s:deepcopy-gen=true
 type Update struct {
-	// version is a semantic versioning identifying the update version. When this
-	// field is part of spec, version is optional if image is specified.
+	// architecture is an optional field that indicates the desired
+	// value of the cluster architecture. In this context cluster
+	// architecture means either a single architecture or a multi
+	// architecture. architecture can only be set to Multi thereby
+	// only allowing updates from single to multi architecture. If
+	// architecture is set image cannot be set and version must be
```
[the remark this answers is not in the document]
LalatenduMohanty: Fixed in https://github.com/openshift/api/commit/08f0ef0f775bfe622dee42df15a94c9ee648f189

config/v1/types_cluster_version.go
```diff
@@ -406,17 +425,33 @@ type ComponentOverride struct {
 type URL string
 
 // Update represents an administrator update request.
+// +kubebuilder:validation:XValidation:rule="has(self.architecture) && has(self.image) ? (self.architecture == "" || self.image == "") : true",message="cannot set both Architecture and Image"
+// +kubebuilder:validation:XValidation:rule="has(self.architecture) && self.architecture != "" ? self.version != "" : true",message="Version must be set if Architecture is set"
```
JoelSpeed, line 429: This still won't work as you're unquoting mid way through, either we move back to the length based or use single quotes, I don't mind either way tbh
```suggestion
// +kubebuilder:validation:XValidation:rule="has(self.architecture) && has(self.image) ? (self.architecture == '' || self.image == '') : true",message="cannot set both Architecture and Image"
// +kubebuilder:validation:XValidation:rule="has(self.architecture) && self.architecture != '' ? self.version != '' : true",message="Version must be set if Architecture is set"
```
or
```suggestion
// +kubebuilder:validation:XValidation:rule="has(self.architecture) && has(self.image) ? (size(self.architecture) == 0 || size(self.image) == 0) : true",message="cannot set both Architecture and Image"
// +kubebuilder:validation:XValidation:rule="has(self.architecture) && size(self.architecture) != 0 ? size(self.version) != 0 : true",message="Version must be set if Architecture is set"
```

config/v1/stable.clusterversion.testsuite.yaml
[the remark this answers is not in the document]
LalatenduMohanty: Both onUpdate tests are covered by the following onCreate test
```
- name: Should not allow image and architecture to be set
    initial: |
      kind: ClusterVersion
      spec:
        clusterID: foo
        desiredUpdate:
          architecture: Multi
          version: 4.11.1
          image: bar
```

config/v1/types_cluster_version.go
LalatenduMohanty: Fixed in https://github.com/openshift/api/commit/354e2fb75e59cc5c396e1665eb8c778e801ed32b

config/v1/stable.clusterversion.testsuite.yaml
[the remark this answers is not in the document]
jottofar: We have onCreate `Should not allow image and architecture to be set` which checks that both cannot be set so I'm not sure what can be added to onCreate to replicate the two onUpdates since each of them can start with only one of the values but the onCreate cannot (AFAIK).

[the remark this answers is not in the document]
jottofar: And I originally added the `onUpdate` tests to cover a user doing an `oc patch` for example.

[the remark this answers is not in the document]
LalatenduMohanty: Also I can work on this even after the PR is merged. So lets not block the PR to get merged because of this.

PR #1345, opened 2022-11-12
OCPBUGS-3635: Set default for spec.to.weight in Route CRD.

benluddy:
Setting the default value for spec.to.weight is typically performed by openshift-apiserver, but it's missing when Route is CRD-backed.


benluddy:
Seeing some errors from `make update`, so the verify job will probably fail. I'll look into it on Monday.

PR #1351, opened 2022-11-17
ODC-7182: Add pinned resources customization

debsmita1:
**Story:**
https://issues.redhat.com/browse/ODC-7182

**Solution Description:**
Add pinned resources customization API based on https://github.com/openshift/enhancements/pull/1285

**GIF:**
https://user-images.githubusercontent.com/22490998/206478836-20a28adf-aff4-4764-86b8-049b5b3683d7.mov


operator/v1/types_console.go
```diff
@@ -298,6 +298,12 @@ type Perspective struct {
 	// visibility defines the state of perspective along with access review checks if needed for that perspective.
 	// +kubebuilder:validation:Required
 	Visibility PerspectiveVisibility `json:"visibility"`
+	// pinnedResources defines the list of default pinned resources that will show up on the perspective navigation.
+	// Example:"apps~v1~Deployment", "core~v1~ConfigMap".
```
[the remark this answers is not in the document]
debsmita1: updated the API, instead of this format user would need to pass the `Group`, `Version`, and `Kind` of the resource. PTAL

config/v1/types.go
```diff
@@ -398,3 +398,13 @@ const (
...
+
+// GroupVersionResource includes the group, version and type of resource
+type GroupVersionResource struct {
+	// Group is the API Group of the Resource.
+	Group string `json:"group" protobuf:"bytes,3,opt,name=group"`
+	// Version is the API Version of the Resource.
+	Version string `json:"version" protobuf:"bytes,4,opt,name=version"`
+	// Resource is the type that is being referenced
+	Resource string `json:"resource" protobuf:"bytes,2,opt,name=resource"`
+}
```
jerolimov, line 410: We don't define the protobuf in this file yet, so I believe its not necessary here. Or?

```suggestion
// GroupVersionResource includes the group, version and type of resource
type GroupVersionResource struct {
	// Group is the API Group of the Resource.
	Group string `json:"group"`
	// Version is the API Version of the Resource.
	Version string `json:"version"`
	// Resource is the type that is being referenced
	Resource string `json:"resource"`
}
```

operator/v1/types_console.go
```diff
@@ -293,11 +293,18 @@ type Perspective struct {
 	// Example: "dev", "admin".
 	// The available perspective ids can be found in the code snippet section next to the yaml editor.
 	// Incorrect or unknown ids will be ignored.
+	// +kubebuilder:validation:XValidation:rule="self.id == 'admin'",message="pinnedResources are not supported for id "admin""
```
jerolimov, line 296: Can we move this validation rule to the `PinnedResource` or to the type to have more context?

The validation rule should also warn only when `PinnedResource` is defined or has length > 1, right?

```suggestion
	// +kubebuilder:validation:XValidation:rule="self.id == 'admin' && !has(self.pinnedResources) : true",message="pinnedResources are not supported for id "admin""
```

jerolimov (changes requested):
Two small things:

operator/v1/types_console.go
JoelSpeed: Does this work where it is currently, typically we would by adding the validation rule to the `Perspective` struct itself rather than fields within it

config/v1/types.go
```diff
@@ -398,3 +398,13 @@ const (
 	// NoOpinionIncludeSubDomains means HSTS "includeSubDomains" doesn't matter to the RequiredHSTSPolicy
 	NoOpinionIncludeSubDomains IncludeSubDomainsPolicy = "NoOpinion"
 )
+
+// GroupVersionResource includes the group, version and type of resource
+type GroupVersionResource struct {
```
JoelSpeed, line 403: You should move this to the operator package, we don't allow cross API group references like this

Please also rename it `PinnedResourceReference`

I'd also like to see field validations each of the fields, required vs optional, examples of valid values, patterns for the values where appropriate.

Please review the [API conventions doc](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#write-user-readable-documentation-in-godoc) and review the section about good godocs to help improve the context here

config/v1/types.go
```diff
@@ -398,3 +398,13 @@ const (
...
+	// Group is the API Group of the Resource.
+	Group string `json:"group" protobuf:"bytes,3,opt,name=group"`
+	// Version is the API Version of the Resource.
+	Version string `json:"version" protobuf:"bytes,4,opt,name=version"`
+	// Resource is the type that is being referenced
+	Resource string `json:"resource" protobuf:"bytes,2,opt,name=resource"`
```
JoelSpeed, line 409: I don't think we need the protobuf tags

SHAKY404 (approved):

SHAKY404 (approved):

imageregistry/v1/00_imageregistry.crd.yaml
```diff
@@ -30,6 +30,7 @@ spec:
             metadata:
               type: object
             spec:
+              $ref: '#/definitions/github.com~1openshift~1api~1operator~1v1~0OperatorSpec'
```
JoelSpeed, line 33: You need to regenerate this as it's not correct at the moment

operator/v1/types_console.go
```diff
@@ -298,6 +301,28 @@ type Perspective struct {
...
+// PinnedResourceReference includes the group, version and type of resource
+type PinnedResourceReference struct {
+	// Group is the API Group of the Resource.
+	// Example: "apps", "build.openshift.io", etc.
+	// +optional
+	Group string `json:"group"`
+	// Version is the API Version of the Resource.
+	// Example: "v1", "v1beta1", etc.
+	// +kubebuilder:validation:Required
+	Version string `json:"version"`
```
JoelSpeed, line 321: CRD versions must fit a particular regex right? We could validate that here

operator/v1/types_console.go
```diff
@@ -298,6 +301,28 @@ type Perspective struct {
 	// visibility defines the state of perspective along with access review checks if needed for that perspective.
 	// +kubebuilder:validation:Required
 	Visibility PerspectiveVisibility `json:"visibility"`
+	// pinnedResources defines the list of default pinned resources that users will see on the perspective navigation if they have not customized these pinned resources themselves.
+	// The list of available Kubernetes resources could be read via `kubectl api-resources`.
+	// The console will also provide a configuration UI and a YAML snippet that will list the available resources that can be pinned to the navigation.
+	// Incorrect or unknown resources will be ignored.
+	// +optional
+	PinnedResources []PinnedResourceReference `json:"pinnedResources"`
```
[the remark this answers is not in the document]
debsmita1: there is no maximum length, a vertical scrollbar appears on the nav when there are a lot of items pinned
cc @jerolimov

operator/v1/types_console.go
```diff
@@ -298,6 +301,28 @@ type Perspective struct {
...
+	// Group is the API Group of the Resource.
+	// Example: "apps", "build.openshift.io", etc.
+	// +optional
+	Group string `json:"group"`
+	// Version is the API Version of the Resource.
+	// Example: "v1", "v1beta1", etc.
+	// +kubebuilder:validation:Required
+	Version string `json:"version"`
+	// Resource is the type that is being referenced
+	// Example: "Deployment", "DeploymentConfig", "Pod", etc.
```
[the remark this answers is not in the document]
debsmita1: So, should I change 'Resource' to 'Kind' ?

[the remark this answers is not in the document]
debsmita1: The Group , Version and Kind of the Resource will be used in our frontend code to identify if the Resource exists and then display it on the Navigation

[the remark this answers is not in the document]
JoelSpeed: Can you link me to where you're doing the mapping from Kind to the resource? A code example would help my understanding

Normally, in Go, you would use a restmapper, which fetches information from the API server about available API resources, to map the resource to a kind, your suggestion suggests you're doing this the other way around which is not necessarily guaranteed to produce the correct result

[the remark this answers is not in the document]
debsmita1: This is how we are passing the the default Pins in the extension plugin
```
"defaultPins": [
        {
          "group": "",
          "version": "v1",
          "kind": "ConfigMap"
        },
        {
          "group": "",
          "version": "v1",
          "kind": "Secret"
        }
      ],
```

and this the logic:

```
export const referenceForExtensionModel = (model: ExtensionK8sGroupModel): GroupVersionKind =>
  referenceForGroupVersionKind(model?.group || 'core')(model?.version)(model?.kind);
 
const defaultPins = useMemo(
    () =>
      perspectiveExtensions.reduce(
        (acc, e) => ({
          ...acc,
          [e.properties.id]: (e.properties.defaultPins || []).map((gvk) =>
            referenceForExtensionModel(gvk),
          ),
        }),
        {},
      ),
    [perspectiveExtensions],
  );
```

and this is how it is saved in the user settings config map
```
"dev":["core~v1~ConfigMap","core~v1~Secret"]}
```

operator/v1/types_console.go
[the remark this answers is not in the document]
jerolimov: Yeah, 100 is fine ;)  I would not expect any limit here.

operator/v1/types_console.go
```diff
@@ -298,6 +301,28 @@ type Perspective struct {
...
+	// +optional
+	PinnedResources []PinnedResourceReference `json:"pinnedResources"`
+}
+
+// PinnedResourceReference includes the group, version and type of resource
+type PinnedResourceReference struct {
+	// Group is the API Group of the Resource.
+	// Example: "apps", "build.openshift.io", etc.
+	// +optional
+	Group string `json:"group"`
```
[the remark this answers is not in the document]
jerolimov: So you suggesting an empty string for core group like "Pods", right?

Debsmita, please test if we can pass `group: ""` if we say this is required. This make it maybe more clear.

And we can also add an example (`	// Example: "", "apps", "build.openshift.io", etc.`) for that.

@JoelSpeed Do you mean we should expect an format here like (pseudo regex): `^(|[a-z]+[a-z.]+[a-z]+)$` ??? I would only apply this if we found a common used pattern for a group somewhere. Do you know one?

operator/v1/types_console.go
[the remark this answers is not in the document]
jerolimov: @JoelSpeed Your lowercase "deployments" confuses me as well. You said a GVR refers exactly and directly a resource. Isn't "deployments" one of the aliases?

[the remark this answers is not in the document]
JoelSpeed: > Your lowercase "deployments" confuses me as well. You said a GVR refers exactly and directly a resource. Isn't "deployments" one of the aliases?

No, `deployments` is the correct and canonical resource name, `Deployment` is an alias to `deployments`. 

For example, if you look at the `kubectl api-resources` for `deployments`, note the Name field (the unique identifier) is `deployments` and the `Kind` field, is later, because it's an alias
```
NAME                              SHORTNAMES   APIVERSION                             NAMESPACED   KIND
deployments                       deploy       apps/v1                                true         Deployment
```

Somewhere in the code you are doing, you are converting `Deployment` to `deployments`, except you're not guaranteed that that's correct. You need this conversion to form a valid API request, so you may not be explicitly conscious of it, this is why I asked where the code was, so that I can have a look and point out where that conversion is happening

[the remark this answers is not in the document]
JoelSpeed: In https://github.com/openshift/console/blob/3ea52056b9f8e486599353fc044336eae4c2a6c0/frontend/public/module/k8s/get-resources.ts#L139-L200 you are performing API discovery to look up the CRDs and API types that exist within the cluster.

You're creating a model map here which maps the name of the resources (ie what we reference to as `resource` here) as the `plural` of the model. Note how this logic is all based on the plural version being the name of the resource, as that's the unique identifier.

When you lookup the models from your cache, you should look them up by the GVR as this is guaranteed to be unique, looking at the current implementation, you are looking them up by GVK, but that's not guaranteed to be unique

operator/v1/types_console.go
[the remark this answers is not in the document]
JoelSpeed: The API group must be a valid domain, the `name` of the CRD must pass https://github.com/kubernetes/apimachinery/blob/4fbe8e4db39b17a53c0756caf58466baad871871/pkg/util/validation/validation.go#L209 , the name of the CRD is `<plural>.<group>` so the validation of `<group>` is pretty much the same as the DNS1123 subdomain validation, so I think the following should be valid
```
[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*
```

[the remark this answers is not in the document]
debsmita1: This regex is accepting strings like this `ab.jhkjh.`

JoelSpeed (commented):
This is looking good I think, I'd like you to look at the validations we have added here and add integration tests for these, https://github.com/openshift/api/tree/master/tests#readme, in particular the validations of regexes, required vs optional fields and the CEL expression you've added should be tested please

operator/v1/types_console.go
```diff
@@ -298,6 +301,34 @@ type Perspective struct {
...
+	// +kubebuilder:validation:Pattern:=^$|[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*
+	// +kubebuilder:validation:Required
+	Group string `json:"group"`
+	// Version is the API Version of the Resource.
+	// Example: "v1", "v1beta1", etc.
+	// +kubebuilder:validation:Pattern:=^[a-z0-9]+$
+	// +kubebuilder:validation:Required
+	Version string `json:"version"`
+	// Resource is the type that is being referenced.
+	// It is the plural form of the resource kind in lowercase.
```
JoelSpeed, line 327: This isn't always true, can you update this to say `It is normally the plural form the resource kind in lowercase`. There are occasions where they don't align

operator/v1/types_console.go
```diff
@@ -298,6 +301,34 @@ type Perspective struct {
...
+	// +kubebuilder:validation:MaxItems=100
+	// +optional
+	PinnedResources []PinnedResourceReference `json:"pinnedResources"`
+}
+
+// PinnedResourceReference includes the group, version and type of resource
+type PinnedResourceReference struct {
+	// Group is the API Group of the Resource.
+	// Enter empty string for the core group.
+	// Example: "", "apps", "build.openshift.io", etc.
```
JoelSpeed, line 317: We should add `// This value should consist of only lowercase alphanumeric characters, hyphens and periods.`

operator/v1/types_console.go
```diff
@@ -298,6 +301,34 @@ type Perspective struct {
...
+	Group string `json:"group"`
+	// Version is the API Version of the Resource.
+	// Example: "v1", "v1beta1", etc.
+	// +kubebuilder:validation:Pattern:=^[a-z0-9]+$
+	// +kubebuilder:validation:Required
+	Version string `json:"version"`
+	// Resource is the type that is being referenced.
+	// It is the plural form of the resource kind in lowercase.
+	// Example: "deployments", "deploymentconfigs", "pods", etc.
+	// +kubebuilder:validation:Pattern:=[a-z0-9]([-a-z0-9]*[a-z0-9])?
```
JoelSpeed, line 329: We should add `// This value should consist of only lowercase alphanumeric characters and hyphens.`

operator/v1/types_console.go
```diff
@@ -298,6 +301,34 @@ type Perspective struct {
...
+type PinnedResourceReference struct {
+	// Group is the API Group of the Resource.
+	// Enter empty string for the core group.
+	// Example: "", "apps", "build.openshift.io", etc.
+	// +kubebuilder:validation:Pattern:=^$|[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*
+	// +kubebuilder:validation:Required
+	Group string `json:"group"`
+	// Version is the API Version of the Resource.
+	// Example: "v1", "v1beta1", etc.
+	// +kubebuilder:validation:Pattern:=^[a-z0-9]+$
```
JoelSpeed, line 323: We should add `// This value should consist of only lowercase alphanumeric characters.`

operator/v1/types_console.go
```diff
@@ -298,6 +301,34 @@ type Perspective struct {
...
+	// The console will also provide a configuration UI and a YAML snippet that will list the available resources that can be pinned to the navigation.
+	// Incorrect or unknown resources will be ignored.
+	// +kubebuilder:validation:MaxItems=100
+	// +optional
+	PinnedResources []PinnedResourceReference `json:"pinnedResources"`
+}
+
+// PinnedResourceReference includes the group, version and type of resource
+type PinnedResourceReference struct {
+	// Group is the API Group of the Resource.
```
JoelSpeed, line 315: Nit, the first word of the comment on each field should match the json tag, so this should be `group` not `Group`, please fix version and resource too

operator/v1/stable.console.testsuite.yaml
```diff
@@ -14,3 +14,114 @@ tests:
...
+        customization:
+          perspectives:
+          - id: dev
+            visibility: 
+              state: Enabled
+            pinnedResources:
+            - resource: configmaps
+              version: v1
+    expectedError: "spec.customization.perspectives[0].pinnedResources[0].group: Required value"
+  - name: Should throw an error if the value of "version" in the pinnedResources doesnot match the required regex
```
JoelSpeed, line 113: You need validation of the regexes for the group and resource fields too please

debsmita1: done

debsmita1:
Added GIF

JoelSpeed:
Changes look good, please make sure the enhancement is updated to reflect what we've agreed upon here and make sure that we get some SME review from the console team and the other reviewers on the enhancement to make sure they're ok with this API

Given recent activity on the enhancement, I'd expect to see an LGTM from @spadgett before we merge

jerolimov (approved):
Tested this together with this PRs on a cluster bot instance and it works really well already :+1:

1. https://github.com/openshift/console-operator/pull/697
2. https://github.com/openshift/console/pull/12314
3. https://github.com/openshift/console/pull/12354

debsmita1:
Propagating docs and px acks from the Epic
/label docs-approved
/label px-approved

spadgett (approved):
/lgtm

jerolimov:
As we have tested this on a cluster bot (see https://github.com/openshift/api/pull/1351#pullrequestreview-1220587190) and in agreement with @sanketpathak we add QE-approved for this change:

/label qe-approved

PR #1376, opened 2022-12-13
RFE-3360: add httpRequestTimeout to IngressController CRD

EmmanuelKasper:
This helps to better tune slowloris defensive actions.


candita:
@EmmanuelKasper https://issues.redhat.com//browse/RFE-3360 is currently on our backlog.  In order for us to consider accepting it, we kindly ask you to demonstrate that this change 1) solves the original problem for this customer, and 2) doesn't cause any new performance issues with the router.  Benchmark a series of documented performance tests without the change, and then compare it to the same tests with the change.

PR #1379, opened 2022-12-20
CSIMigrationvSphere feature is GA in kube 1.26

sanchezl:


sanchezl:
Dup of https://github.com/openshift/api/pull/1370

PR #1381, opened 2023-01-03
CFE-677 : Update the type of Tag field in AzureMachineProviderSpec model

bharath-b-rh:
azure-sdk-for-go defines Tags as of the type `map[string]*string` and updating the `Tags` field definition in `AzureMachineProviderSpec` model to same type eliminates traversing the map and having to convert the value to `*string` type.


bharath-b-rh:
It's not a common practice to use pointer types in APIs and since the change is not a must have closing it.

PR #1386, opened 2023-01-16
STOR-1133: set RetroactiveDefaultStorageClass feature as TechPreview

RomanBednar:
We want to release RetroactiveDefaultStorageClass as TechPreview in 4.13: https://issues.redhat.com/browse/STOR-1133

KEP: https://github.com/kubernetes/enhancements/issues/3333

cc @openshift/storage


jsafrane:
lgtm from storage POV, we want a Beta feature turned off by default, because it changes behavior of StorageClasses and there is no other API that would enable the new behavior while the feature is still beta / tech preview.

RomanBednar:
Doc and QE trackers are approved: 

https://issues.redhat.com/browse/STOR-1103
https://issues.redhat.com/browse/STOR-1105

/label docs-approved
/label qe-approved

Also this can not be tested before merge, adding px label.

/label px-approved

PR #1390, opened 2023-01-20
Add optional new fields to NutanixMachineProviderConfig

yanhua121:
Add some optional new fields to NutanixMachineProviderConfig, for Windows support, etc.


machine/v1/types_nutanixprovider.go
```diff
@@ -56,6 +56,22 @@ type NutanixMachineProviderConfig struct {
...
 	SystemDiskSize resource.Quantity `json:"systemDiskSize"`
 
+	// bootType indicates the boot type (SECURE_BOOT, UEFI or LEGACY) the VM should use to boot.
+	// If this field is empty, the default boot type LEGACY is used for the VM to boot.
+	// SECURE_BOOT depends on UEFI boot, i.e., enabling SECURE_BOOT means that UEFI boot is also enabled.
+	// +kubebuilder:validation:Enum="";LEGACY;UEFI;SECURE_BOOT
+	// +optional
+	BootType NutanixBootType `json:"bootType,omitempty"`
+
+	// Project optionally identifies a Prism Central project for the Machine's VM to associate with.
```
JoelSpeed, line 66: Can you extend this comment to explain what happens when this is omitted please

machine/v1/types_nutanixprovider.go
```diff
@@ -56,6 +56,22 @@ type NutanixMachineProviderConfig struct {
...
+	// +optional
+	Project NutanixResourceIdentifier `json:"project,omitempty"`
+
+	// Categories optionally defines one or more categories (each with name and value) for
+	// the Machine's VM to associate with.
+	// +optional
+	Categories map[string]string `json:"categories,omitempty"`
```
JoelSpeed, line 73: We don't use maps within Kube APIs, instead you need to create a struct and use a list type

```
type NutanixCategory strict {
  Name string `json:"name"`

  Value string `json:"value"`
}
```
```
// categories ...
// +listType=map
// +listMapKey=name
// +optional
Categories []NutanixCategory `json:"categories"`
```
The map type and key will ensure that each entry in the list has a unique name

machine/v1/types_nutanixprovider.go
```diff
@@ -66,6 +82,20 @@ type NutanixMachineProviderConfig struct {
...
+
+const (
+	// NutanixLegacyBoot is the legacy BIOS boot type
+	NutanixLegacyBoot NutanixBootType = "LEGACY"
+
+	// NutanixUEFIBoot is the UEFI boot type
+	NutanixUEFIBoot NutanixBootType = "UEFI"
+
+	// NutanixSecureBoot is the Secure boot type
+	NutanixSecureBoot NutanixBootType = "SECURE_BOOT"
```
JoelSpeed, line 96: These should be PascalCase to match the kube convention, not any upstream constants
```suggestion
	// NutanixLegacyBoot is the legacy BIOS boot type
	NutanixLegacyBoot NutanixBootType = "Legacy"

	// NutanixUEFIBoot is the UEFI boot type
	NutanixUEFIBoot NutanixBootType = "UEFI"

	// NutanixSecureBoot is the Secure boot type
	NutanixSecureBoot NutanixBootType = "SecureBoot"
```

machine/v1/types_nutanixprovider.go
yanhua121: Note that L60 explains:

// If this field is empty, the default boot type LEGACY is used for the VM to boot.

yanhua121: done

machine/v1/types_nutanixprovider.go
yanhua121: Fixed.

machine/v1/types_nutanixprovider.go
```diff
@@ -66,6 +84,29 @@ type NutanixMachineProviderConfig struct {
 	CredentialsSecret *corev1.LocalObjectReference `json:"credentialsSecret"`
 }
 
+// NutanixCategory identifies a pair of prism category key and value
+type NutanixCategory struct {
+	// +kubebuilder:validation:Required
+	Key string `json:"key"`
+
+	// +kubebuilder:validation:Required
+	Value string `json:"value"`
```
JoelSpeed, line 93: Can you please add godoc to explain these two fields

Also, are there any validations we can put on, for example, a maximum length or a set of valid characters?

yanhua121: Addressed. Only found that both the category key and value requires <= 64 characters, but not found any restriction on the characters pattern. Also updated the godoc of the requirement that all the categories specified in the Machine spec need to already exist (defined) in the prism central.

elmiko:
the `bootType`  field can be tested with this https://github.com/openshift/machine-api-provider-nutanix/pull/37

elmiko:
the `project` and `categories` fields can be tested with this https://github.com/openshift/machine-api-provider-nutanix/pull/38

yanhua121:
The Machine webhook validation for these new fields can be tested with https://github.com/openshift/machine-api-operator/pull/1117.

JoelSpeed (commented):
Quick question, do we need the `omitempty` on these fields? If we remove it, the fields will show in a serialized version with empty values, which then hints to users "oh, that's an option I can use", so it improves the discoverability, which is generally something we advise for configuration APIs such as these

yanhua121:
> Quick question, do we need the `omitempty` on these fields? If we remove it, the fields will show in a serialized version with empty values, which then hints to users "oh, that's an option I can use", so it improves the discoverability, which is generally something we advise for configuration APIs such as these

@JoelSpeed The only concern I have is if we remove the 'omitempty' on these optional fields, the Machine/MachineSet yaml used in a 4.13 cluster then cannot use to a OCP cluster with earlier version 4.11/4.12. In some scenario, customer may have clusters with different OCP versions. They may want to deploy the Machine/MachineSet yamls to all the cluster.

JoelSpeed:
Unknown fields are just dropped on older versions so it would be ok to have the new fields with empty values and be applied to the older versions.

I don't think changing omitempty changes that scenario really. We can add it back if we find issues but I think for now it should be removed for discoverability

yanhua121:
> Unknown fields are just dropped on older versions so it would be ok to have the new fields with empty values and be applied to the older versions.
> 
> I don't think changing omitempty changes that scenario really. We can add it back if we find issues but I think for now it should be removed for discoverability

@JoelSpeed Okay, then I will go ahead to remove 'omitempty' for these optional fields.

machine/v1/types_nutanixprovider.go
```diff
@@ -56,6 +56,25 @@ type NutanixMachineProviderConfig struct {
 	// +kubebuilder:validation:Required
 	SystemDiskSize resource.Quantity `json:"systemDiskSize"`
 
+	// bootType indicates the boot type (Legacy, UEFI or SecureBoot) the Machine's VM uses to boot.
+	// If this field is empty or omitted, the VM will use the default boot type "Legacy" to boot.
+	// "SecureBoot" depends on "UEFI" boot, i.e., enabling "SecureBoot" means that "UEFI" boot is also enabled.
+	// +kubebuilder:validation:Enum="";Legacy;UEFI;SecureBoot
+	// +optional
+	BootType NutanixBootType `json:"bootType,omitempty"`
```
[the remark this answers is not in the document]
yanhua121: Since the UEFI and SecureBoot boot-type usually only used by Windows Machine vm, and it has the default value "Legacy", I think it is better to have the "omitempty".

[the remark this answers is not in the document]
yanhua121: Removed

elmiko (commented):
/lgtm

PR #1392, opened 2023-01-28
IR-342: Add OS/arch information to image stream status

werled:
https://issues.redhat.com/browse/IR-342

is a dependency for https://github.com/openshift/openshift-apiserver/pull/347


werled:
There are currently concerns wrt. to increased object size (see https://issues.redhat.com/browse/IR-342), which have to be addressed before proceeding with this PR.

PR #1397, opened 2023-02-03
Add source of DNS configuration to AWSPlatformStatus

sadasu:
When and AWS customer wishes to use their own DNS solution instead of Route53, they are expected to create the API and API-Int Load Balancers and use their LB names to configure their custom DNS prior to cluster installation. The customer created API and API-Int LB information is provided via install-config.

AWSPlatformStatus is updated with a special "provider" if the user does not want OpenShift to configure the default cloud DNS solution.

Enhancement Proposal : https://github.com/openshift/enhancements/pull/1400


config/v1/types_infrastructure.go
```diff
@@ -435,6 +435,11 @@ type AWSPlatformStatus struct {
 	// +kubebuilder:validation:MaxItems=25
 	// +optional
 	ResourceTags []AWSResourceTag `json:"resourceTags,omitempty"`
+
+	// AWSClusterDNSConfig contains all the DNS config required to configure a custom DNS solution.
+	// This field will be set only when the userConfiguredDNS feature is enabled.
+	// +optional
+	AWSClusterDNSConfig *ClusterDNSConfig `json:"awsClusterDNSConfig,omitempty"`
```
JoelSpeed, line 442: Given this is already in an AWS struct, do you think we need AWS in the name of this struct?

I'd like to expand the `userConfiguredDNS` feature comment, I'd like to make it obvious that this is an installer feature that is at install time, else it could lead users to think, ok, how do I go and turn that on

config/v1/types_infrastructure.go
```diff
@@ -457,6 +462,32 @@ type AWSResourceTag struct {
 	Value string `json:"value"`
 }
 
+type ClusterDNSConfig struct {
```
JoelSpeed, line 465: Structs need godoc too please

config/v1/types_infrastructure.go
```diff
@@ -457,6 +462,32 @@ type AWSResourceTag struct {
...
+	// APIServerDNSConfig contains DNS configuration for the API Server.
```
JoelSpeed, line 466: What happens when this is omitted? What's the expectation?

config/v1/types_infrastructure.go
```diff
@@ -457,6 +462,32 @@ type AWSResourceTag struct {
...
+	// +optional
+	APIServerDNSConfig []DNSConfig `json:"apiServerDNSConfig,omitempty"`
+
+	// InternalAPIServerDNSConfig contains DNS configuration for the Internal API Server.
```
JoelSpeed, line 470: What happens when this is omitted? What's the expectation?

config/v1/types_infrastructure.go
```diff
@@ -457,6 +462,32 @@ type AWSResourceTag struct {
...
+	// +optional
+	InternalAPIServerDNSConfig []DNSConfig `json:"internalAPIServerDNSConfig,omitempty"`
+
+	// IngressDNSConfig contains DNS configuration for cluster services.
```
JoelSpeed, line 474: What happens when this is omitted? What's the expectation?

config/v1/types_infrastructure.go
```diff
@@ -457,6 +462,32 @@ type AWSResourceTag struct {
...
+	// +optional
+	IngressDNSConfig []DNSConfig `json:"ingressDNSConfig,omitempty"`
+}
+
+// DNSConfig contains the LB IP and DNS Record type to configure a DNS entry.
+type DNSConfig struct {
+	// recordType is the DNS record type.
```
JoelSpeed, line 481: Are there a subset that we support here? Would we expect an MX for example? Could this be an enum of the ones we support?

config/v1/types_infrastructure.go
```diff
@@ -457,6 +462,32 @@ type AWSResourceTag struct {
...
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern=`^[A-Za-z]+$`
+	RecordType string `json:"recordType"`
+	// lBIPAddress is the Load Balancer IP address for DNS config
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Format=ip
+	LBIPAddress string `json:"lbIPAddress"`
```
JoelSpeed, line 488: I would avoid the abbreviation here, `loadBalancerIPAddress` would be better IMO, WDYT?

> // +kubebuilder:validation:Format=ip

Have you tested this? I'm not sure this works, you might need a patch for this. Can you add some integration tests that check that these work (see the yaml file for examples, `.testsuite.yaml` for the appropriate resource)

config/v1/types_infrastructure.go
```diff
@@ -457,6 +462,32 @@ type AWSResourceTag struct {
 	Value string `json:"value"`
 }
 
+type ClusterDNSConfig struct {
+	// APIServerDNSConfig contains DNS configuration for the API Server.
+	// +optional
+	APIServerDNSConfig []DNSConfig `json:"apiServerDNSConfig,omitempty"`
```
JoelSpeed, line 468: For this and the others, this is a slice, what does it mean for there to be more than one entry? Can there be an infinite number of entries within here?

config/v1/types_infrastructure.go
sadasu: Not sure how I missed that!

config/v1/types_infrastructure.go
sadasu: When the userConfiguredDNS feature is enabled (via install-config), the Installer and not the user is responsible for populating these values. Even with this feature enabled, the Installer is responsible for configuring the external and internal Load Balancers for API and API-Int servers respectively and utilizes the LB DNS names to derive the DNS configuration.

So, these values will be omitted only when the feature is Disabled (which is the default). If they are empty when the userConfiguredDNS is Enabled, then cluster cannot be created and it is a failure scenario.

Does any of this need to go into the godoc?

config/v1/types_infrastructure.go
sadasu: Same as the DNS config for API server. See https://github.com/openshift/api/pull/1397#discussion_r1131541743

config/v1/types_infrastructure.go
sadasu: Same as with API and API-Int but the Ingress controller is responsible for providing the Load Balancer configuration that is used to generate the Ingress DNS config. 

When userConfiguredDNS is Enabled and this config is not present, cluster installation will fail.

config/v1/types_infrastructure.go
sadasu: Yes, it would be useful to make this an enum type. Updating.

config/v1/types_infrastructure.go
JoelSpeed: So, if this field is provided, then we expect all three types of DNS to be specified? Just trying to work out if it's possible to have 
```
apiServerDNSConfig: []
internalAPIServerDNSConfig:
- ....
ingressDNSConfig: []
```
It's making me thinking that actually, these fields shouldn't be optional and that each one needs an entry when the struct is specified.

JoelSpeed: Some of this context would be useful in the parent, where the ClusterDNSConfig is included, we could have godoc there that explains this field is expected to be omitted when the user configured DNS feature wasn't used in the installation process, and mention what it means for these values to be present there

JoelSpeed:
Heads up, I'm on PTO for the next 2 weeks, so if you need additional reviews in that time, I'd recommend poking in #forum-api-review

config/v1/types_infrastructure.go
sadasu: We expect one LB IP address for each availability zone. If dualstack is supported that number would be doubled; one per address family for each LB in availability zone.

JoelSpeed: ok, so perhaps we can set a sensible limit on this and the other slices, maybe a limit of 256 would be sufficient for most cases?

config/v1/types_infrastructure.go
```diff
@@ -478,6 +478,20 @@ type AWSPlatformStatus struct {
 	// +kubebuilder:validation:MaxItems=25
 	// +optional
 	ResourceTags []AWSResourceTag `json:"resourceTags,omitempty"`
+
+	// userDNS indicates if the cluster is using a DNS solution configured
+	// by the user in place of Route53.
```
JoelSpeed, line 483: What is the default value for this? Can this be modified at all? Is there any additional configuration needed now or maybe needed in the future that will warrant this being a struct with a type field rather than just a string?

config/v1/types_infrastructure.go
```diff
@@ -478,6 +478,20 @@ type AWSPlatformStatus struct {
...
+	// +optional
+	UserDNS string `json:"userDNS,omitempty"`
+
+	// apiLBName contains the name of the API LB
+	// +optional
+	ApiLBName string `json:"apiLBName,omitempty"`
+
+	// apiIntLBName contains the name of the API-Int LB
+	// +optional
+	ApiIntLBName string `json:"apiIntLBName,omitempty"`
```
JoelSpeed, line 494: I can see us wanted to add more load balancer options in the future, why not create something structured
```
apiLoadBalancers:
  internal:
    name: blah
  external:
    name: blah
```
This would be more future proof and allow us to add additional configuration in the future

JoelSpeed (commented):
Is there any relation between the load balancer name fields and any other fields specified? Eg they must be populated when the UserDNS is Enabled?

config/v1/types_infrastructure.go
sadasu: Default value is Disabled and cannot be modified once the cluster installation is complete. Hence the choice to add it to `Status` vs `Spec`. 
Yes, IMHO making this a type would make it future proof and I was toggling between this option (simple) vs the other (future proof).

sadasu:
> Is there any relation between the load balancer name fields and any other fields specified? Eg they must be populated when the UserDNS is Enabled?

Yes, they will be populated when UserDNS is Enabled. So, it would be helpful to keep them together in a struct vs as individual values.

config/v1/types_infrastructure.go
```diff
@@ -478,6 +478,11 @@ type AWSPlatformStatus struct {
 	// +kubebuilder:validation:MaxItems=25
 	// +optional
 	ResourceTags []AWSResourceTag `json:"resourceTags,omitempty"`
+
+	// userDNS contains user provided configuration for API and API-Int
+	// Load Balancers when user configured DNS is used in place of Route53.
+	// +optional
+	UserDNS UserConfiguredDNS `json:"userDNS,omitempty"`
```
JoelSpeed, line 485: I think we want to change this a little, so it's less specific, I think the end result should look like this in a normal cluster
```
dns:
  provider: Default
```
or like below for the user configured version.
```
dns:
  provider: User
  user:
    apiLoadBalancer:
      name: foo
    apiInternalLoadBalancer:
       name: bar
```
This would implement a discriminated union for the configuration, example [here](https://github.com/openshift/api/blob/391aead809e1374e9978feeb1305a3479523aadc/example/v1/types_stable.go#L88-L105)

This gives us the flexibility in the future to expand configuration for the default DNS provider should we need to

JoelSpeed: The additional layer of `user` is required for it to be a [discriminated union](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#discriminated-unions), KEP for reference: https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/1027-api-unions/README.md,

sadasu: Just making sure we want the discriminated union even when the additions to the API are in the "Status" and not the "Spec". (Double checking because all examples are for the Spec.)

JoelSpeed: I think we want it in the status too, makes it clear. Spec and status are equally editable by end users these days, even though that's not that desirable

config/v1/types_infrastructure.go
```diff
@@ -500,6 +506,43 @@ type AWSResourceTag struct {
...
+// DNSAndLBConfiguration contains information about API and API-Int Load Balancers
+// and who is responsible for that configuration.
+type DNSAndLBConfiguration struct {
+	// When the user has pre-created the Load Balancers for API and API-Int and the
+	// the DNS solution has already been configured, this field is set to "UserConfig".
+	// When the cluster is resposible for its own DNS configuration and LB creation,
+	// this field is set to "Default".
+	// +default="Default"
+	// +kubebuilder:default:="Default"
+	// +kubebuilder:validation:Enum="UserConfig";"Default"
```
JoelSpeed, line 518: I think I suggested it, though I'm not 100% convinced on the name `Default`, what else could we have here to suggest this is the installer created DNS/load balancers

config/v1/types_infrastructure.go
```diff
@@ -500,6 +506,43 @@ type AWSResourceTag struct {
...
+	// +unionDiscriminator
+	// +optional
+	Provider string `json:"provider,omitempty"`
```
JoelSpeed, line 521: Provider should be required
```suggestion
	// +kubebuilder:validation:Required
	Provider string `json:"provider"`
```
You should also use a type alias with constants for the different options for the enum

config/v1/types_infrastructure.go
```diff
@@ -500,6 +506,43 @@ type AWSResourceTag struct {
...
+
+	// userConfig is the user provided LoadBlancer configuration.
+	// These values are expected to be set only when provider is set to "UserConfig"
+	UserConfig UserLBConfig `json:"userConfig,omitempty"`
```
JoelSpeed, line 525: This should be a pointer as it's part of a discriminated union

config/v1/types_infrastructure.go
```diff
@@ -500,6 +506,43 @@ type AWSResourceTag struct {
 	Value string `json:"value"`
 }
 
+// DNSAndLBConfiguration contains information about API and API-Int Load Balancers
+// and who is responsible for that configuration.
+type DNSAndLBConfiguration struct {
```
JoelSpeed, line 511: Check the example in openshift/api/example/v1. There's some CEL validation that will be required for the userConfig option for this union

This should also have a `+union` marker on it

JoelSpeed (commented):
Reviewing this and thinking about the name default, I'm curious, how does this relate to IPI vs UPI? Does this feature affect UPI clusters at all?

config/v1/types_infrastructure.go
```diff
@@ -500,6 +506,43 @@ type AWSResourceTag struct {
...
+	// When the user has pre-created the Load Balancers for API and API-Int and the
+	// the DNS solution has already been configured, this field is set to "UserConfig".
+	// When the cluster is resposible for its own DNS configuration and LB creation,
+	// this field is set to "Default".
```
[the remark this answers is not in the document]
sadasu: I knew that! Not sure how I missed this.

config/v1/types_infrastructure.go
sadasu: How about "userProvided" and "clusterProvided"?

JoelSpeed: Yeah lets go with that for now

config/v1/0000_10_config-operator_01_infrastructure-CustomNoUpgrade.crd.yaml
```diff
@@ -466,6 +466,55 @@ spec:
...
+                                      description: name is the name of the LB created by the user.
+                                      type: string
+                                  type: object
+                              type: object
+                            provider:
+                              default: ClusterProvided
+                              description: provider determines which DNS solution is in use for this cluster. When the user has pre-created the Load Balancers for API and API-Int and the the DNS solution has already been configured, this field is set to "UserProvided". When the cluster is resposible for its own DNS configuration and LB creation, this field is set to "ClusterProvided".
+                              enum:
+                                - UserProvided
+                                - ClusterProvided
```
sadasu, line 495: When Load Balancers are created by the cluster (which is the default case), the installer could set the 'provider` to "ClusterProvided" and populate the values for the LB Names. But, there is no real use for it.

So, what if the enum have 1 value i.e "USerProvided" ?

JoelSpeed: Interesting idea, so that would be along the lines of `When omitted, this means no opinion and the platform is left to choose a reasonable default` that we have elsewhere. Typically though that information is normally tagged to one of the enum values, eg, in this case, `The default is ClusterProvided` would be the next sentence.

IMO I think it's better to have a value though. Since this is a configuration API we like to have all options populated/displayed. I think it's better to have a value that explains it's ClusterProvided rather than just the empty string

config/v1/types_infrastructure.go
```diff
@@ -478,6 +478,11 @@ type AWSPlatformStatus struct {
 	// +kubebuilder:validation:MaxItems=25
 	// +optional
 	ResourceTags []AWSResourceTag `json:"resourceTags,omitempty"`
+
+	// dnsConfig contains information about API and API-Int Load Balancer
+	// and DNS configuration.
+	// +optional
+	DNSConfig DNSAndLBConfiguration `json:"dnsConfig,omitempty"`
```
JoelSpeed, line 485: Is this a tech preview feature? If so, we need to add the appropriate tag to make it only show up on tech preview/custom feature sets

JoelSpeed, line 485: I don't think this needs an omitempty tag

config/v1/types_infrastructure.go
```diff
@@ -500,6 +505,58 @@ type AWSResourceTag struct {
...
+	// provider determines which DNS solution is in use for this cluster.
+	// When the user has pre-created the Load Balancers for API and API-Int and the
+	// the DNS solution has already been configured, this field is set to "UserProvided".
+	// When the cluster is resposible for its own DNS configuration and LB creation,
+	// this field is set to "ClusterProvided".
+	// +default="ClusterProvided"
+	// +kubebuilder:default:="ClusterProvided"
+	// +kubebuilder:validation:Enum="UserProvided";"ClusterProvided"
+	// +unionDiscriminator
+	// +kubebuilder:validation:Required
```
JoelSpeed, line 521: Might have to test this, can you add some integration tests that check that this can be defaulted correctly

config/v1/types_infrastructure.go
```diff
@@ -500,6 +505,58 @@ type AWSResourceTag struct {
 	Value string `json:"value"`
 }
 
+// DNSAndLBConfiguration contains information about API and API-Int Load Balancers
+// and who is responsible for that configuration.
+// +union
+type DNSAndLBConfiguration struct {
+	// provider determines which DNS solution is in use for this cluster.
+	// When the user has pre-created the Load Balancers for API and API-Int and the
+	// the DNS solution has already been configured, this field is set to "UserProvided".
```
JoelSpeed, line 514: `and the the` double the in the middle of this

config/v1/types_infrastructure.go
```diff
@@ -500,6 +505,58 @@ type AWSResourceTag struct {
 	Value string `json:"value"`
 }
 
+// DNSAndLBConfiguration contains information about API and API-Int Load Balancers
+// and who is responsible for that configuration.
+// +union
+type DNSAndLBConfiguration struct {
+	// provider determines which DNS solution is in use for this cluster.
+	// When the user has pre-created the Load Balancers for API and API-Int and the
```
JoelSpeed, line 513: I'm concerned this might be ambiguous for things like UPI, would you expect for UPI that this would be UserProvided or ClusterProvided?

config/v1/types_infrastructure.go
```diff
@@ -500,6 +505,58 @@ type AWSResourceTag struct {
...
+	// +unionDiscriminator
+	// +kubebuilder:validation:Required
+	Provider DNSAndLBProviderType `json:"provider,omitempty"`
+
+	// userProvided contains information about the user created LoadBalancers.
+	// These values are expected to be set only when provider is set to "UserProvided"
+	UserProvided *LBConfigType `json:"userProvided,omitempty"`
+
+	// clusterProvided contains information about the OpenShift cluster created LoadBalancers.
+	ClusterProvided *LBConfigType `json:"clusterProvided,omitempty"`
```
JoelSpeed, line 529: If you don't want to have config for ClusterProvided we don't need to have an option here

config/v1/types_infrastructure.go
JoelSpeed: This still needs the CEL validation from the example to make sure the members of the union are only set when the discriminator has the correct value

config/v1/types_infrastructure.go
```diff
@@ -500,6 +505,58 @@ type AWSResourceTag struct {
...
+	ApiIntLB LBConfig `json:"apiIntLB,omitempty"`
+
+	// apiLB contains configuration for the API Load Balancer
+	// +optional
+	ApiLB LBConfig `json:"apiLB,omitempty"`
+}
+
+// LBConfig contains information about the user created Load Balancer.
+type LBConfig struct {
+	// name is the name of the LB created by the user.
```
JoelSpeed, line 556: I think this comment needs more context, what does it mean by name? Is this the AWS name of the resource? Can we accept other formats such as the ARN or an ID?

config/v1/0000_10_config-operator_01_infrastructure-CustomNoUpgrade.crd.yaml
sadasu: Ok. Let us keep it the way it is. It also has the advantage of the default behavior being less ambiguous. Comment updated to reflect that.

config/v1/types_infrastructure.go
```diff
@@ -500,6 +505,58 @@ type AWSResourceTag struct {
...
+type DNSAndLBConfiguration struct {
+	// provider determines which DNS solution is in use for this cluster.
+	// When the user has pre-created the Load Balancers for API and API-Int and the
+	// the DNS solution has already been configured, this field is set to "UserProvided".
+	// When the cluster is resposible for its own DNS configuration and LB creation,
+	// this field is set to "ClusterProvided".
+	// +default="ClusterProvided"
+	// +kubebuilder:default:="ClusterProvided"
+	// +kubebuilder:validation:Enum="UserProvided";"ClusterProvided"
+	// +unionDiscriminator
```
patrickdillon, line 520: It's not clear to me why we would use a diiscriminated union here. The mutually-exclusive fields in the union are the same type `LBConfigType`. Why not just have a single field for LBConfig which includes an enum field to distinguish cluster vs user-provided. The main benefit I see of a union discriminator is when you have mutually-exclusive fields of different types.

config/v1/types_infrastructure.go
```diff
@@ -500,6 +505,58 @@ type AWSResourceTag struct {
...
+	// clusterProvided contains information about the OpenShift cluster created LoadBalancers.
+	ClusterProvided *LBConfigType `json:"clusterProvided,omitempty"`
+}
+
+// DNSAndLBProviderType defines the source of the DNS and LB configuration.
+type DNSAndLBProviderType string
+
+const (
+	// DNSAndLBUserProvided indicates that the user provides the LB and DNS for API and API-Int.
+	DNSAndLBUserProvided DNSAndLBProviderType = "UserProvided"
```
patrickdillon, line 537: Coupling LB & DNS in this manner is problematic. It excludes a possible valid configuration of user-provided load balancer with installer-provisioned DNS.

patrickdillon (commented):
Trying to summarize/capture the outcome of the meeting @sadasu and I had earlier today.

I feel like the scope of this API design is expanding beyond the needs of enabling custom-DNS solutions; which is fine but I am losing track of the goal. 

I believe our sole need for enabling custom-DNS is a flag to indicate to the MCO we need to run CoreDNS pods and to the cluster-ingress operator/runtimecfg to write the LB dns names to the right config. 

As part of the installer implementation for this feature, we will need to accept user-provided load balancers (like we do with subnets), but there is no need to write them to the cluster API AFAIK.

config/v1/types_infrastructure.go
sadasu: Done.

config/v1/types_infrastructure.go
sadasu: Yes, this is the AWS name of the resource for AWS. I didn't use the platform name here because I am expecting to use the same type for other cloud platforms later.

Updated the comment to make it clearer hopefully.

config/v1/types_infrastructure.go
sadasu: Updated the type name and the names of the string constants.

config/v1/types_infrastructure.go
sadasu: When the union discriminator is of the default type "ClusterProvided", we don't need to save the LB names. But, in the case of "UserProvided", we do. So, the union represents this case best IMO.

patrickdillon: We don't need to save LB name in either case right? These values will not be used. 

To keep track of user-provided LBs, we should tag them to associate them with the cluster, similar to the way we do with user-provided subnets. Installer-provisioned LBs are also tagged, so this would be consistent across all clusters.

Unless we have a reason for these fields, I think it's better to remove them and defer the design of the API to when we have a need or reason to include them.

patrickdillon: > It's not clear to me why we would use a diiscriminated union here. 

Looking at [LoadBalancerType ](https://github.com/openshift/api/pull/1391)  a model for this API, I see that the PlatformLoadBalancer structs are unions but with only union discriminators--no member fields. So I'm assuming that the reason to use a union there is future proofing, which is fine with me and something I would leave up to the discretion of more knowledgeable api reviewers.

sadasu: In the on-prem BYO LoadBalancer scenario, the customer is completely responsible for managing the loadbalancers so they do not provide the name of the Load Balancers to the cluster to being with. 
In the custom DNS case, the user needs to provide the LB names so that the Installer can configure them. So, not exactly the same data is provided via install-config.
I am unclear about the best practices around this. If the goal is to accurately represent that state of the cluster, then the LB names need also make it into the Infrastructure Status. If the goal is to only add items that other components in the cluster read to take action, then the LB Names are not necessary as @patrickdillon pointed out.
I am Ok with adding the LB names at a later point but don't see it adding too much of an overhead now.  I'll let @JoelSpeed weigh in before I go ahead and remove the LB names.

JoelSpeed: If the fields aren't going to be consumed, lets not add them, no point adding them until we have something that is a concrete consumer.

And yes, by having a union we have the option to expand this later and diverge from being the same across both. If we don't think that's going to be a future we see, flipping this round to have the type inside the same struct as the names isn't an unreasonable suggestion IMO

sadasu: Updated implementation to match discussion in this thread.

config/v1/types_infrastructure.go
```diff
@@ -500,6 +507,37 @@ type AWSResourceTag struct {
...
+
+// DNSProviderType defines the source of the DNS and LB configuration.
+type DNSProviderType string
+
+const (
+	// DNSUserProvided indicates that the user provides the LB and DNS for API and API-Int.
+	DNSUserProvided DNSProviderType = "UserProvided"
+
+	// DNSClusterProvided indicates that the cluster provides the LB and DNS for API and API-Int.
+	DNSClusterProvided DNSProviderType = "ClusterProvided"
```
patrickdillon, line 538: These values seem a little squishy to me. I am having trouble keeping track of all the discussions around this feature, so I'm not sure if these values have been gamed out. 

My understanding of the intention here is that when the value is `UserProvided` that would indicate that we need to run `CoreDNS`. That to me has some cognitive dissonance. 

For the sake of discussion, what about something along the lines of
`CloudProvider` - this is standard IPI, DNS is in the cloud provider
`ClusterProvider` - uses on-cluster baremetal dns provider (coredns)
`UserProvider` - cluster/ipi does not provide DNS, this would be for something like ROSA where they are running dnsmasq in pods

Do these values better capture the current/future use cases we are aware of?

config/v1/types_infrastructure.go
```diff
@@ -500,6 +507,37 @@ type AWSResourceTag struct {
 	Value string `json:"value"`
 }
 
+// DNSConfigurationType contains information about API and API-Int Load Balancers
+// and who is responsible for that configuration.
+// +union
+type DNSConfigurationType struct {
```
patrickdillon, line 513: To enable external DNS, I think we are only concerned about using coreDNS to run ingress. Should that be taken into account for the shape of this API? Any concerns there?

patrickdillon (commented):
Left some more comments for discussion. There are still quite a few references to load balancers in the code comments.

config/v1/types_infrastructure.go
```diff
@@ -478,6 +478,13 @@ type AWSPlatformStatus struct {
 	// +kubebuilder:validation:MaxItems=25
 	// +optional
 	ResourceTags []AWSResourceTag `json:"resourceTags,omitempty"`
+
+	// dnsConfig contains information about the type of DNS solution in use
+	// for the cluster and its related configuration.
+	// +default={"provider": "ClusterProvided"}
+	// +kubebuilder:default={"provider": "ClusterProvided"}
+	// +openshift:enable:FeatureSets=TechPreviewNoUpgrade
+	DNSConfig *DNSConfigurationType `json:"dnsConfig"`
```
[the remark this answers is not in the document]
sadasu: Kubebuilder will automatically add the default value of provider="ClusterProvided". There is a integration test to that effect.

config/v1/types_infrastructure.go
```diff
@@ -500,6 +507,37 @@ type AWSResourceTag struct {
...
+// DNSConfigurationType contains information about API and API-Int Load Balancers
+// and who is responsible for that configuration.
+// +union
+type DNSConfigurationType struct {
+	// provider determines which DNS solution is in use for this cluster.
+	// When the user wants to use their own DNS solution, the `provider` is set to
+	// "UserProvided". In this state, the user is expected to pre-create the Load
+	// Balancers for API and API-Int.
+	// When the cluster is resposible for its own DNS configuration, this field is
+	// set to "ClusterProvided". The Load Balancers are also created by the cluster.
```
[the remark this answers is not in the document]
sadasu: Removed any mention of Load Balancers.

config/v1/types_infrastructure.go
sadasu: I had a version that seperated 3 cases. But looking at it a different way, the cluster is taking care of DNS for both the CloudProvider and ClusterProvider case. I though I would get push back for superfluous cases and hence removed it.

config/v1/types_infrastructure.go
patrickdillon: I think this comment was unclear. My concern is that this shape may give the impression that this is the DNS config for everything (API & Ingress) but I believe our current direction is that this will only treat ingress.

sadasu: I think you are looking at this config as "who provides DNS for Ingress?" Whereas, I have designed this API to indicate "who provides DNS for the cluster?". I think the 2nd question is the most generic form of the question.

We need DNS for Ingress to be functional during cluster bringup so we have a CoreDNS pod that provides that during cluster bring up. When the cluster is up, we do want the user to update their DNS solution for the *.apss entry too. When all steps are completed, the customer's DNS solution resolves API, API-Int and Ingress.

The enhancement proposal https://github.com/openshift/enhancements/pull/1400 talks about this in greater detail.

config/v1/types_infrastructure.go
```diff
@@ -500,6 +507,37 @@ type AWSResourceTag struct {
...
+	// +default="ClusterProvided"
+	// +kubebuilder:default:="ClusterProvided"
+	// +kubebuilder:validation:Enum="UserProvided";"ClusterProvided"
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="oldSelf == '' || self == oldSelf",message="type is immutable once set"
+	// +unionDiscriminator
+	Provider DNSProviderType `json:"provider,omitempty"`
+}
+
+// DNSProviderType defines the source of the DNS and LB configuration.
```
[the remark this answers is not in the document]
sadasu: The LB comment was added to provide a complete picture of the feature i.e not only should the user configure their DNS, they should also have pre-created the LB. I am OK with removing it.

config/v1/types_infrastructure.go
```diff
@@ -500,6 +508,36 @@ type AWSResourceTag struct {
...
+// +union
+type DNSConfigurationType struct {
+	// provider determines which DNS solution is in use for this cluster.
+	// When the user wants to use their own DNS solution, the `provider` is set to
+	// "UserProvided".
+	// When the cluster configures the cloud DNS, this field is set to
+	// "CloudProvided". The default is "CloudProvided".
+	// +default="CloudProvided"
+	// +kubebuilder:default:="CloudProvided"
+	// +kubebuilder:validation:Enum="UserProvided";"CloudProvided"
```
sadasu, line 522: After discussion with @patrickdillon , we came to the conclusion that "UserProvided" does not give an accurate complete picture. So, proposal is to update this string to "UserAndInClusterProvided". 
@JoelSpeed so I am currently working on updating the PR.

config/v1/types_infrastructure.go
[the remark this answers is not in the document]
patrickdillon: > Kubebuilder will automatically add the default value of provider="ClusterProvided". There is a integration test to that effect.

Interesting. I didn't know you could do that! That's cool. 

I'm having a hard time wrapping my head around the idea of something being both _required_ and having a _default_ value; to me they seem mutually exclusive, or at least to mutually diminish one another. Just a personal observation/opinion.

Joel helpfullly brought up UPI installs--we don't technically know that all existing clusters are using IPI-style cloud DNS. Users could opt for UPI in order to establish their own non-route53 DNS solutions. I don't know how many that would be, but I'm pretty confident it's non-zero. So if we default this value, I think the values would likely be incorrect in some cases.

config/v1/types_infrastructure.go
patrickdillon: The original UPI comment was about load balancers, but I think this is still a good question regarding DNS. For UPI, the installer simply doesn't know.

config/v1/types_infrastructure.go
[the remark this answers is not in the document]
sadasu: With the current implementation, if DNSConfig.Provider is left unset by the user, it will be given the default of "CloudProvided". No further action will be taken within OpenShift when Provider is "CloudProvided". Afaik, this should work for UPI too. Are there some functionality/semantics issues that I am missing?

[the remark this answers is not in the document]
sadasu: I replaced "CloudProvided" with "Default" to indicate that that the right defaults for UPI or IPI are in use.

config/v1/types_infrastructure.go
sadasu: The 2 Provider values are "Default" and "UserAndClusterProvided" to represent the default and hybrid (where user provides DNS for API and API-Int and the cluster has to provide DNS for *.apps during cluster bring up) cases  respectively.

config/v1/types_infrastructure.go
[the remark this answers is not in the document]
sadasu: Replaced "Default" with "".

sadasu:
@JoelSpeed @patrickdillon could you please take another look?

sadasu:
> > @JoelSpeed @patrickdillon could you please take another look?
> 
> Our [enhancement discussion has revealed a path](https://github.com/openshift/enhancements/pull/1400/files#r1267306024) that may very well make these changes unneeded, so we should probably resolve that before iterating much further.

@patrickdillon is there a desire for having the cluster have specific knowledge about whether external DNS is in use? If yes, then we would need this.

sadasu:
This PR is no longer needed since the design for this epic has changed.

PR #1400, opened 2023-02-08
config: enable secretbox encryption

dgrisonnet:


dgrisonnet:
@tkashem @deads2k opening this PR to start a discussion about secretbox.

The initial code supporting encryption at rest in library-go was written to work with both `aes-cbc` and `secretbox` but only the former was made available to the users:
- https://github.com/openshift/library-go/blob/master/pkg/operator/encryption/state/types.go#L60

Now that we want to support `aes-gcm` should we also consider supporting `secretbox` if the feature is already there and we only need to test it?

Note that I haven't tested anything yet, but I wanted to start an early discussion to avoid wasting time. `aes-gcm` is still the priority, but depending on what we decide we may want to create a ticket that would be a follow-up to the aes-gcm effort.

/hold

PR #1401, opened 2023-02-08
Update openshift/api to include encryption keys

jhixson74:
https://issues.redhat.com/browse/CORS-2089


operator/v1/types_csi_cluster_driver.go
```diff
@@ -118,25 +121,136 @@ const (
...
 	// +optional
 	VSphere *VSphereCSIDriverConfigSpec `json:"vSphere,omitempty"`
 }
 
+// AWSCSIDriverConfigSpec defines properties that can be configured for the AWS CSI driver.
+type AWSCSIDriverConfigSpec struct {
+	// kmsKey sets the cluster default storage class to encrypt volumes with a user-defined KMS key,
+	// rather than the default KMS key used by AWS.
+	// +optional
+	KMSKey AWSKMSKey `json:"kmsKey,omitempty"`
```
patrickdillon, line 152: After flipping around quite a bit in the enhancement, we ultimately landed on having the ARN as a single string rather than structured:
https://github.com/openshift/enhancements/blob/master/enhancements/installer/storage-class-encrypted.md#aws

operator/v1/types_csi_cluster_driver.go
```diff
@@ -118,25 +121,136 @@ const (
...
+}
+
+// AWSKMSKey contains the component parts that comprise an ARN for a KMS key.
+// According to https://docs.aws.amazon.com/service-authorization/latest/reference/list_awskeymanagementservice.html
+// the format for a KMS key ARN is:
+// arn:${Partition}:kms:${Region}:${Account}:key/${KeyId}
+// When constructing the ARN, the partition can be determined from the region.
+type AWSKMSKey struct {
```
dobsonj, line 160: There was [some discussion](https://github.com/openshift/enhancements/pull/1163#discussion_r1087055672) on splitting up the ARN string in the enhancement, but it raised some problems so we agreed on [using the ARN string](https://github.com/openshift/enhancements/blob/ac2729f61cb593a82cfe8a7bb771e7c1d5043fcb/enhancements/installer/storage-class-encrypted.md#aws) like we use for [root volumes](https://access.redhat.com/solutions/4827341) today.

operator/v1/types_csi_cluster_driver.go
```diff
@@ -105,10 +105,13 @@ type ClusterCSIDriverSpec struct {
 }
 
 // CSIDriverType indicates type of CSI driver being configured.
-// +kubebuilder:validation:Enum="";vSphere
+// +kubebuilder:validation:Enum=AWS;Azure;GCP;vSphere
```
dobsonj, line 108: Isn't `""` still valid?

operator/v1/types_csi_cluster_driver.go
```diff
@@ -118,25 +121,136 @@ const (
...
+	// +kubebuilder:validation:MaxLength:=30
+	// +optional
+	ProjectID string `json:"projectID,omitempty"`
+
+	// location is the GCP location in which the Key Ring exists.
+	// The input must match an existing GCP location, or "global".
+	// Defaults to global, if not set.
+	// +kubebuilder:validation:Pattern:=[a-zA-Z0-9_-]
+	// +optional
+	Location string `json:"location,omitempty"`
```
dobsonj, line 243: Can these two fields be required too? For essentially the [same reason](https://github.com/openshift/enhancements/pull/1163#discussion_r1070147787) as the Azure fields. (cc @patrickdillon)

dobsonj:
Related enhancement for future reference: https://github.com/openshift/enhancements/blob/master/enhancements/installer/storage-class-encrypted.md

operator/v1/types_csi_cluster_driver.go
gnufied: Yeah - we should still keep `""` until we can fix the defaulting for upgraded clusters.

operator/v1/types_csi_cluster_driver.go
jhixson74: Fixed.

operator/v1/types_csi_cluster_driver.go
patrickdillon: I created [a PR](https://github.com/openshift/enhancements/pull/1343) to bump the project id to required. 

Location is a static default, if not supplied, the value should be "global". The driver/operator does not need to look up any values. In that case do you still think it should be required or can the driver/operator supply the static default?

operator/v1/types_csi_cluster_driver.go
```diff
@@ -118,25 +121,116 @@ const (
...
 }
 
+// AWSCSIDriverConfigSpec defines properties that can be configured for the AWS CSI driver.
+type AWSCSIDriverConfigSpec struct {
+	// kmsKeyARN sets the cluster default storage class to encrypt volumes with a user-defined KMS key,
+	// rather than the default KMS key used by AWS.
+	// The value may be either the ARN or Alias ARN of a KMS key.
+	// +kubebuilder:validation:Pattern:=arn:(aws|aws-cn|aws-us-gov):kms:[a-z0-9]+(-[a-z0-9]+)*:[0-9]{12}:(key|alias)/.*
+	// +optional
+	KMSKeyARN string `json:"kmsKeyARN,omitempty"`
```
JoelSpeed, line 154: What happens in terms of validation if you remove the omitempty here? I assume an empty string doesn't pass the validation and it errors out?

operator/v1/types_csi_cluster_driver.go
```diff
@@ -118,25 +121,116 @@ const (
...
+}
+
+// AzureDiskEncryptionSet defines the configuration for a disk encryption set.
+type AzureDiskEncryptionSet struct {
+	// subscriptionID defines the Azure subscription that contains the disk encryption set.
+	// When omitted, the subscription from the authenticating credentials of the CSI driver
+	// will be used.
```
JoelSpeed, line 161: Each of the fields in this struct need to explain the validations/patterns within the godoc, as they aren't presented in the generated docs otherwise

operator/v1/types_csi_cluster_driver.go
```diff
@@ -118,25 +121,116 @@ const (
...
+	// It must start with a letter. Trailing hyphens are prohibited.
+	// Defaults to the cluster project defined in the cluster infrastructure object, if not set.
+	// +kubebuilder:validation:Pattern:=^[a-z][a-z0-9-]+[a-z0-9]$
+	// +kubebuilder:validation:MinLength:=6
+	// +kubebuilder:validation:MaxLength:=30
+	// +optional
+	ProjectID string `json:"projectID,omitempty"`
+
+	// location is the GCP location in which the Key Ring exists.
+	// The input must match an existing GCP location, or "global".
```
JoelSpeed, line 219:
```suggestion
	// The value must match an existing GCP location, or "global".
	// The value should consist of only alphanumeric characters, hyphens (-) and underscores (_).
```
Any length validations on the location?

operator/v1/types_csi_cluster_driver.go
jhixson74: Fixed.

operator/v1/types_csi_cluster_driver.go
jhixson74: Fixed.

operator/v1/types_csi_cluster_driver.go
```diff
@@ -118,25 +121,116 @@ const (
...
+	// +optional
+	AWS *AWSCSIDriverConfigSpec `json:"aws,omitempty"`
+
+	// Azure is used to configure the Azure CSI driver.
+	// +optional
+	Azure *AzureCSIDriverConfigSpec `json:"azure,omitempty"`
+
+	// GCP is used to configure the GCP CSI driver.
+	// +optional
+	GCP *GCPCSIDriverConfigSpec `json:"gcp,omitempty"`
```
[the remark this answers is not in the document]
jhixson74: Fixed.

operator/v1/types_csi_cluster_driver.go
jhixson74: I've updated the explanations for the validations as well as some of the patterns that didn't quite match up. Feel free to let me know if I can word them better.

operator/v1/types_csi_cluster_driver.go
```diff
@@ -118,25 +121,116 @@ const (
...
+	// customer-managed encryption set, rather than the default platform-managed keys.
+	// +optional
+	DiskEncryptionSet *AzureDiskEncryptionSet `json:"diskEncryptionSet,omitempty"`
+}
+
+// GCPKMSKeyReference gathers required fields for looking up a GCP KMS Key
+type GCPKMSKeyReference struct {
+	// name is the name of the customer-managed encryption key to be used for disk encryption.
+	// The value should correspond to an existing KMS key
+	// and input should match the regular expression: [a-zA-Z0-9_-]{1,63}.
```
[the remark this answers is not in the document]
jhixson74: Fixed.

operator/v1/types_csi_cluster_driver.go
```diff
@@ -118,25 +121,116 @@ const (
...
+	// +kubebuilder:validation:Pattern:=^[a-zA-Z0-9_-]$
+	// +kubebuilder:validation:MinLength:=1
+	// +kubebuilder:validation:MaxLength:=63
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+
+	// keyRing is the name of the KMS Key Ring which the KMS Key belongs to.
+	// The value should correspond to an existing KMS key ring
+	// and input should match the regular expression: [a-zA-Z0-9_-]{1,63}.
```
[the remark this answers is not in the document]
jhixson74: Fixed.

operator/v1/types_csi_cluster_driver.go
jhixson74: Fixed.

operator/v1/types_csi_cluster_driver.go
```diff
@@ -118,25 +121,116 @@ const (
...
+	SubscriptionID string `json:"subscriptionID"`
+
+	// resourceGroup defines the Azure resource group that contains the disk encryption set.
+	// When omitted, the cluster resource group from the cluster infrastructure object
+	// will be used.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern=^[-\w\._\(\)]+$
+	ResourceGroup string `json:"resourceGroup"`
+
+	//name is the name of the disk encryption set that will be set on the default storage class.
```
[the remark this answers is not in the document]
jhixson74: Fixed.

operator/v1/types_csi_cluster_driver.go
jhixson74: > What happens in terms of validation if you remove the omitempty here? I assume an empty string doesn't pass the validation and it errors out?

That would be my assumption also.  Should we remove omitempty?

operator/v1/types_csi_cluster_driver.go
```diff
@@ -118,25 +121,116 @@ const (
...
+	// +kubebuilder:validation:Pattern:=[a-zA-Z0-9_-]
+	Name string `json:"name"`
+}
+
+// AzureCSIDriverConfigSpec defines properties that can be configured for the Azure CSI driver.
+type AzureCSIDriverConfigSpec struct {
+	// diskEncryptionSet sets the cluster default storage class to encrypt volumes with a
+	// customer-managed encryption set, rather than the default platform-managed keys.
+	// +optional
+	DiskEncryptionSet *AzureDiskEncryptionSet `json:"diskEncryptionSet,omitempty"`
```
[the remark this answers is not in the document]
jhixson74: I don't know. I will look into this.

operator/v1/types_csi_cluster_driver.go
```diff
@@ -118,25 +121,116 @@ const (
...
 	// driverConfig is being applied to.
-	//
-	// Valid values are:
-	//
-	// * vSphere
-	//
-	// Allows configuration of vsphere CSI driver topology.
-	//
-	// ---
+	// Valid values are: AWS, Azure, GCP, vSphere
```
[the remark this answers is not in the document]
jhixson74: I don't know. I will look into this.

operator/v1/types_csi_cluster_driver.go
```diff
@@ -118,25 +121,116 @@ const (
...
+
+	// projectID is the ID of the Project in which the KMS Key Ring exists.
+	// It must be 6 to 30 lowercase letters, digits, or hyphens.
+	// It must start with a letter. Trailing hyphens are prohibited.
+	// Defaults to the cluster project defined in the cluster infrastructure object, if not set.
+	// +kubebuilder:validation:Pattern:=^[a-z][a-z0-9-]+[a-z0-9]$
+	// +kubebuilder:validation:MinLength:=6
+	// +kubebuilder:validation:MaxLength:=30
+	// +optional
+	ProjectID string `json:"projectID,omitempty"`
```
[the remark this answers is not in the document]
jhixson74: I have removed omitempty and make it required.

operator/v1/types_csi_cluster_driver.go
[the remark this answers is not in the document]
dobsonj: If DriverType is "" then the contents of CSIDriverConfigSpec are ignored. It's a no-op.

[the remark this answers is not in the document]
dobsonj: I think the way it's worded is correct:
```
	// Valid values are: AWS, Azure, GCP, vSphere
	// Consumers should treat unknown values as a NO-OP.
```
If you say `AWS, Azure, GCP, vSphere and omitted` instead then it sounds like the string "omitted" is also valid.

operator/v1/types_csi_cluster_driver.go
JoelSpeed: It depends, for discoverability, removing omitempty would mean we get the `kmsKeyARN: ""` represented in the struct , but you'd need to adjust the pattern to allow an empty value or the ARN representation, prepending with `^$|` might work

We still want this field optional IIUC so maybe try that?

operator/v1/types_csi_cluster_driver.go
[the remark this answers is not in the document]
JoelSpeed: We use `omitted` frequently through our API documentation to mean that the value isn't present. I would prefer we are consistent with our other optioanl Enums and keep the `and omitted` and then explain`, `When omitted, this means...` verbiage.

The Consumers note you've pointed out seems like it's aimed at developers, rather than end users? Right? 
If it's a developer note, prefix it with a `+` to remove it from the generated customer docs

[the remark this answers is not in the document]
dobsonj: > We use omitted frequently through our API documentation to mean that the value isn't present.

Ok, then I have no objection.

> The Consumers note you've pointed out seems like it's aimed at developers, rather than end users? Right?
If it's a developer note, prefix it with a + to remove it from the generated customer docs

Yes, and with the validator, I can't set an unknown or invalid value as a user anyway.

So based on your suggestions, how about this?
```
	// Valid values are: AWS, Azure, GCP, vSphere and omitted.
	// When this value is omitted, the CSI driver config spec is not used.
	// +Consumers should treat unknown values as a NO-OP.
```

dobsonj:
@jhixson74 I made some changes in my own branch to [fix validation patterns](https://github.com/dobsonj/api/commit/eab90248766fef920ddf4685a42065a3507bcb1c) and run [make update](https://github.com/dobsonj/api/commit/f4ecd2ef3195438943bf20ea4dbb04743848d0d0) so that I could test my operator changes with the new CRD. If you pull those commits into your branch, it should fix the verify failures.

dobsonj:
Er, [one more commit](https://github.com/dobsonj/api/commit/b4c2779b70c7982e339712d36b99efe205bbdf3e) now, I just noticed some of those patterns didn't have a quantifier.

dobsonj:
@JoelSpeed @jhixson74 I think this looks good from the storage side. What is left from your perspective? I'd like to get this merged soon so we can wrap up the dependent PR's.

jhixson74:
> @JoelSpeed @jhixson74 I think this looks good from the storage side. What is left from your perspective? I'd like to get this merged soon so we can wrap up the dependent PR's.

I believe everything has been addressed. If there is anything more please let me know!

dobsonj:
Thanks @jhixson74 :)
/lgtm
/assign @JoelSpeed 
for approval

operator/v1/types_csi_cluster_driver.go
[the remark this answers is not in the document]
JoelSpeed: Yeah I think that's good, but what is used when it is omitted, some defaults?

operator/v1/types_csi_cluster_driver.go
```diff
@@ -118,25 +121,129 @@ const (
...
 	VSphere *VSphereCSIDriverConfigSpec `json:"vSphere,omitempty"`
 }
 
+// AWSCSIDriverConfigSpec defines properties that can be configured for the AWS CSI driver.
+type AWSCSIDriverConfigSpec struct {
+	// kmsKeyARN sets the cluster default storage class to encrypt volumes with a user-defined KMS key,
+	// rather than the default KMS key used by AWS.
+	// The value may be either the ARN or Alias ARN of a KMS key.
+	// +kubebuilder:validation:Pattern:=`^arn:(aws|aws-cn|aws-us-gov):kms:[a-z0-9-]+:[0-9]{12}:(key|alias)\/.*$`
+	// +optional
```
JoelSpeed, line 153: What happens if this isn't set?

operator/v1/types_csi_cluster_driver.go
```diff
@@ -118,25 +121,129 @@ const (
...
+	// +kubebuilder:validation:MaxLength:=80
+	// +kubebuilder:validation:Pattern:=`^[a-zA-Z0-9\_-]+$`
+	Name string `json:"name"`
+}
+
+// AzureCSIDriverConfigSpec defines properties that can be configured for the Azure CSI driver.
+type AzureCSIDriverConfigSpec struct {
+	// diskEncryptionSet sets the cluster default storage class to encrypt volumes with a
+	// customer-managed encryption set, rather than the default platform-managed keys.
+	// +optional
```
JoelSpeed, line 196: What happens if this isn't set

operator/v1/types_csi_cluster_driver.go
```diff
@@ -118,25 +121,129 @@ const (
...
+	// +kubebuilder:validation:Pattern:=`^[a-zA-Z0-9\_-]+$`
+	// +optional
+	Location string `json:"location,omitempty"`
+}
+
+// GCPCSIDriverConfigSpec defines properties that can be configured for the GCP CSI driver.
+type GCPCSIDriverConfigSpec struct {
+	// kmsKey sets the cluster default storage class to encrypt volumes with customer-supplied
+	// encryption keys, rather than the default keys managed by GCP.
+	// +optional
```
JoelSpeed, line 243: What happens if this isn't set?

operator/v1/types_csi_cluster_driver.go
dobsonj: For each of these specs, if this field isn't set then we don't use a customer-managed key in the default StorageClass and it behaves the same as previous releases without this field.

dobsonj: If we ever add more fields to `AWSCSIDriverConfigSpec` and the others, then it would be good for this to be an optional field so the the new fields can be set without requiring a customer-managed key.

patrickdillon: The comment indicates what happens when the field isn’t set: the default cloud managed key is used.

operator/v1/types_csi_cluster_driver.go
patrickdillon: The comment indicates what happens when the field isn’t set: the default cloud managed key is used.

operator/v1/types_csi_cluster_driver.go
patrickdillon: The comment indicates what happens when the field isn’t set: the default cloud managed key is used.

JoelSpeed:
/approve

Please make sure the appropriate feature PRs are merged before the 4.13 deadline, if they do not make it, let me know as we should revert this PR if they don't make it

PR #1406, opened 2023-02-15
MON-2849: Graduate alertingrules and alertrelabelconfigs to v1

jan--f:
This PR is mostly about $title.
I'm not sure if https://github.com/openshift/api/commit/46d922b15972209d1127c528bc844bf67004c158 is the correct usage of `.codegen.yaml`.
Additionally I added a drive-by typo fix. Please let me know if this is appropriate, happy to drop this or propose separately.


jan--f:
Seems like having the file `monitoring/v1alpha1/.codegen.yaml` was not the correct choice.

monitoring/v1/0000_50_monitoring_01_alertingrules.crd.yaml
```diff
@@ -0,0 +1,121 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    api-approved.openshift.io: https://github.com/openshift/api/pull/1179
```
JoelSpeed, line 5: 1406 now?

monitoring/v1/0000_50_monitoring_02_alertrelabelconfigs.crd.yaml
```diff
@@ -0,0 +1,139 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    api-approved.openshift.io: https://github.com/openshift/api/pull/1179
```
JoelSpeed, line 5: 1406 now?

JoelSpeed (commented):
The `.codegen.yaml` file needs to stay where it is, please add else I don't think you'll get any code generation for the techpreview types, try [copying the values from the example API](https://github.com/openshift/api/blob/42edf4f1d9052cd332f9a5d5c6fd5dcffdd16449/example/.codegen.yaml#L9-L12)

monitoring/v1/types.go
```diff
@@ -0,0 +1,337 @@
...
+// +genclient
+// +k8s:openapi-gen=true
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:subresource:status
+type AlertingRule struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// spec describes the desired state of this AlertingRule object.
+	Spec AlertingRuleSpec `json:"spec"`
```
JoelSpeed, line 40: This should have a required tag right? Can we make it explicit? It is already required in the CRD but better to have an explicit tag for readability down the line
```suggestion
	// spec describes the desired state of this AlertingRule object.
	// +kubebuilder:validation:Required
	Spec AlertingRuleSpec `json:"spec"`
```

monitoring/v1/types.go
```diff
@@ -0,0 +1,337 @@
...
+//
+// Compatibility level 1: Stable within a major release for a minimum of 12 months or 3 minor releases (whichever is longer).
+// +openshift:compatibility-gen:level=1
+// +k8s:openapi-gen=true
+type AlertingRuleList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+
+	// items is a list of AlertingRule objects.
+	Items []AlertingRule `json:"items"`
```
JoelSpeed, line 60: This should also have a required tag please
```suggestion
	// +kubebuilder:validation:Required
	Items []AlertingRule `json:"items"`
```

monitoring/v1/types.go
```diff
@@ -0,0 +1,337 @@
...
+	// to create AlertingRule resources with only a single rule group, but be
+	// aware that this can have a performance impact on Prometheus if the group is
+	// extremely large or has very complex query expressions to evaluate.
+	// Spreading very complex rules across multiple groups to allow them to be
+	// processed in parallel is also a common use-case.
+	//
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MinItems:=1
+	Groups []RuleGroup `json:"groups"`
```
JoelSpeed, line 85: Explicit required tag please
```suggestion
	// +kubebuilder:validation:Required
	Groups []RuleGroup `json:"groups"`
```

monitoring/v1/types.go
```diff
@@ -0,0 +1,337 @@
...
+}
+
+// RuleGroup is a list of sequentially evaluated alerting rules.
+//
+// +k8s:openapi-gen=true
+type RuleGroup struct {
+	// name is the name of the group.
+	//
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
```
JoelSpeed, line 95: Did we conclude there was any way to restrict this by character set?

monitoring/v1/types.go
```diff
@@ -0,0 +1,337 @@
...
+	// +kubebuilder:validation:Pattern:="^(([0-9]+)y)?(([0-9]+)w)?(([0-9]+)d)?(([0-9]+)h)?(([0-9]+)m)?(([0-9]+)s)?(([0-9]+)ms)?$"
+	// +optional
+	Interval string `json:"interval,omitempty"`
+
+	// rules is a list of sequentially evaluated alerting rules.  Prometheus may
+	// process rule groups in parallel, but rules within a single group are always
+	// processed sequentially, and all rules are processed.
+	//
+	// +kubebuilder:validation:MinItems:=1
+	Rules []Rule `json:"rules"`
```
JoelSpeed, line 121: Required tag please
```suggestion
	// +kubebuilder:validation:Required	
	Rules []Rule `json:"rules"`
```

monitoring/v1/types.go
```diff
@@ -0,0 +1,337 @@
...
+// See Prometheus documentation:
+// - https://www.prometheus.io/docs/prometheus/latest/configuration/alerting_rules
+//
+// +k8s:openapi-gen=true
+type Rule struct {
+	// alert is the name of the alert. Must be a valid label value, i.e. only
+	// contain ASCII letters, numbers, and underscores.
+	//
+	// +kubebuilder:validation:Pattern:="^[a-zA-Z_][a-zA-Z0-9_]*$"
+	// +required
```
JoelSpeed, line 134: This isn't the right format
```suggestion
	// +kubebuilder:validation:Required
```

monitoring/v1/types.go
```diff
@@ -0,0 +1,337 @@
...
+	// expression, e.g.:
+	//
+	//   mapi_current_pending_csr > mapi_max_pending_csr
+	//
+	// In rare cases this could be a simple integer, e.g. a simple "1" if the
+	// intent is to create an alert that is always firing.  This is sometimes used
+	// to create an always-firing "Watchdog" alert in order to ensure the alerting
+	// pipeline is functional.
+	//
+	// +required
```
JoelSpeed, line 149:
```suggestion
	// +kubeuilder:validation:Required
```

monitoring/v1/types.go
```diff
@@ -0,0 +1,337 @@
...
+	// labels to add or overwrite for each alert.  The results of the PromQL
+	// expression for the alert will result in an existing set of labels for the
+	// alert, after evaluating the expression, for any label specified here with
+	// the same name as a label in that set, the label here wins and overwrites
+	// the previous value.  These should typically be short identifying values
+	// that may be useful to query against.  A common example is the alert
+	// severity:
+	//
+	//   labels:
+	//     severity: warning
```
JoelSpeed, line 171: Have you tried an oc describe on this to see how it renders? I'm concerned it's not going to render well as I don't think the spacing is preserved in the openapi schema

monitoring/v1/types.go
```diff
@@ -0,0 +1,337 @@
...
+// PrometheusRuleRef is a reference to an existing PrometheusRule object.  Each
+// AlertingRule instance results in a generated PrometheusRule object in the same
+// namespace, which is always the openshift-monitoring namespace.  This is used to
+// point to the generated PrometheusRule object in the AlertingRule status.
+type PrometheusRuleRef struct {
+	// This is a struct so that we can support future expansion of fields within
+	// the reference should we ever need to.
+
+	// name of the referenced PrometheusRule.
+	Name string `json:"name"`
```
JoelSpeed, line 215: This needs a required tag
```suggestion
	// +kubebuilder:validation:Required
	Name string `json:"name"`
```

monitoring/v1/types.go
```diff
@@ -0,0 +1,337 @@
...
+//
+// Compatibility level 1: Stable within a major release for a minimum of 12 months or 3 minor releases (whichever is longer).
+// +openshift:compatibility-gen:level=1
+// +k8s:openapi-gen=true
+type AlertRelabelConfig struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// spec describes the desired state of this AlertRelabelConfig object.
+	Spec AlertRelabelConfigSpec `json:"spec"`
```
JoelSpeed, line 232:
```suggestion
	// +kubebuilder:validation:Required
	Spec AlertRelabelConfigSpec `json:"spec"`
```

monitoring/v1/types.go
```diff
@@ -0,0 +1,337 @@
...
+}
+
+// AlertRelabelConfigsSpec is the desired state of an AlertRelabelConfig resource.
+//
+// +k8s:openapi-gen=true
+type AlertRelabelConfigSpec struct {
+	// configs is a list of sequentially evaluated alert relabel configs.
+	//
+	// +kubebuilder:validation:MinItems:=1
+	Configs []RelabelConfig `json:"configs"`
```
JoelSpeed, line 247:
```suggestion
	// +kubebuiler:validation:Required
	Configs []RelabelConfig `json:"configs"`
```

monitoring/v1/types.go
```diff
@@ -0,0 +1,337 @@
...
+// Compatibility level 1: Stable within a major release for a minimum of 12 months or 3 minor releases (whichever is longer).
+// +openshift:compatibility-gen:level=1
+// +k8s:openapi-gen=true
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type AlertRelabelConfigList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+
+	// items is a list of AlertRelabelConfigs.
+	Items []*AlertRelabelConfig `json:"items"`
```
JoelSpeed, line 275:
```suggestion
	// +kubebuilder:validation:Required
	Items []*AlertRelabelConfig `json:"items"`
```

monitoring/v1/types.go
```diff
@@ -0,0 +1,337 @@
...
+	SourceLabels []LabelName `json:"sourceLabels,omitempty"`
+
+	// separator placed between concatenated source label values. When omitted,
+	// Prometheus will use its default value of ';'.
+	//
+	// +optional
+	Separator string `json:"separator,omitempty"`
+
+	// targetLabel to which the resulting value is written in a 'Replace' action.
+	// It is mandatory for 'Replace' and 'HashMod' actions. Regex capture groups
```
JoelSpeed, line 305: Since this field is required when action has a certain value, we could use a CEL validation to make sure this is set, add this to the parent struct
```
// +kubebuilder:validation:XValidation:rule="self.action != 'Replace' && self.action != 'HashMod' || has(self.targetLabel)",message="targetLabel is required when action is Replace or HashMod"
```
The logic here is either action isn't HashMod/Replace, or we have the targetLabel

monitoring/v1/types.go
```diff
@@ -0,0 +1,337 @@
...
+// RelabelConfig allows dynamic rewriting of label sets for alerts.
+// See Prometheus documentation:
+// - https://prometheus.io/docs/prometheus/latest/configuration/configuration/#alert_relabel_configs
+// - https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config
+//
+// +k8s:openapi-gen=true
+type RelabelConfig struct {
+	// sourceLabels select values from existing labels. Their content is
+	// concatenated using the configured separator and matched against the
+	// configured regular expression for the Replace, Keep, and Drop actions.
```
JoelSpeed, line 293: This is only valid with certain actions?

monitoring/v1/types.go
```diff
@@ -0,0 +1,337 @@
...
+	// +optional
+	TargetLabel string `json:"targetLabel,omitempty"`
+
+	// regex against which the extracted value is matched. Default is: '(.*)'
+	//
+	// +optional
+	Regex string `json:"regex,omitempty"`
+
+	// modulus to take of the hash of the source label values.  This can be
+	// combined with the 'HashMod' action to set 'target_label' to the 'modulus'
```
JoelSpeed, line 317: Only valid with the `HashMod` action?

monitoring/v1/types.go
```diff
@@ -0,0 +1,337 @@
...
+	// of a hash of the concatenated 'source_labels'.
+	//
+	// +optional
+	Modulus uint64 `json:"modulus,omitempty"`
+
+	// replacement value against which a regex replace is performed if the regular
+	// expression matches. This is required if the action is 'Replace' or
```
JoelSpeed, line 324: We can use a CEL validation on the parent struct to make sure this is required when the action is Replace or LabelMap right?

monitoring/v1/types.go
jan--f: I rather think this is only valid (or effective) if `sourceLabels` is not empty.

JoelSpeed: Do we need to update the documentation here then? Likewise, if this does have some dependency on another field, lets add some validation using CEL for that

jan--f:
@JoelSpeed Thanks a lot, great input. I added a bunch of cross validations based on whats described in https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config and the subset of what we implement in this API. I also added the suggested `requires` in order to be explicit.
cc @simonpasquier

monitoring/v1/types.go
```diff
@@ -0,0 +1,349 @@
...
+	// combined with the 'HashMod' action to set 'target_label' to the 'modulus'
+	// of a hash of the concatenated 'source_labels'. This is only valid of sourceLabels is not empty
+	//
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="!has(self.sourceLabels) || has(self.modulus) && size(self.sourceLabels) > 0",message="modulus requires sourceLabels to be present"
+	Modulus uint64 `json:"modulus,omitempty"`
+
+	// replacement value against which a regex replace is performed if the regular
+	// expression matches. This is required if the action is 'Replace' or
+	// 'LabelMap'. Regex capture groups are available. Default is: '$1'
```
simonpasquier, line 336: also valid for Keep, Drop, LabelDrop and LabelKeep actions?

jan--f: Valid yes, but its not required for those actions iiuc.

monitoring/v1/types.go
```diff
@@ -0,0 +1,349 @@
...
+// See Prometheus documentation:
+// - https://www.prometheus.io/docs/prometheus/latest/configuration/alerting_rules
+//
+// +k8s:openapi-gen=true
+type Rule struct {
+	// alert is the name of the alert. Must be a valid label value, i.e. only
+	// contain ASCII letters, numbers, and underscores.
+	//
+	// +kubebuilder:validation:Pattern:="^[a-zA-Z_][a-zA-Z0-9_]*$"
+	// +kubebuilder:validation:Required
```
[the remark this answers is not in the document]
jan--f: Prometheus docs actually say it must be a valid label name: https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/

+1 on `MinLength` though.

monitoring/v1/types.go
simonpasquier: Same as for "replace" and "labelmap" actions I believe:  Prometheus will default to "$1" so users can omit the "replacement" value if the default value is ok for them.

monitoring/v1/types.go
```diff
@@ -0,0 +1,350 @@
...
+	//
+	// +optional
+	Modulus uint64 `json:"modulus,omitempty"`
+
+	// replacement value against which a regex replace is performed if the regular
+	// expression matches. This is required if the action is 'Replace' or
+	// 'LabelMap'. Regex capture groups are available. Default is: '$1'
+	//
+	// +optional
+	// +kubebuilder:default=$1
```
simonpasquier, line 340: This could cause an issue if an action (e.g. "labelkeep") doesn't permit the replacement parameter?

monitoring/v1/types.go
JoelSpeed: Where is this being passed to? Does it end up in logs? Does it end up in any prometheus configuration? Is there anywhere where this could be injected that might mean that a user could do something unintended (thinking injection attacks here) if we allow any character?

Have you tested that required is the same as `minLength: 1`, because I suspect that required will pass if `name: ""` (you could use our integration testing suite to verify this

monitoring/v1/types.go
JoelSpeed: So, this only works with a subset of the actions? And in some it's required, some it's defaulted? Or are they all behaving the same?

JoelSpeed:
> expensive by the validation checker

What does this mean? Where were you seeing an error?

monitoring/v1/types.go
JoelSpeed: We probably want to default this server side if there are certain actions where it is valid and certain actions where they aren't valid

jan--f:
> > expensive by the validation checker
> 
> What does this mean? Where were you seeing an error?

@JoelSpeed When I change the rule like so:
``` diff
diff --git a/monitoring/v1/0000_50_monitoring_02_alertrelabelconfigs.crd.yaml b/monitoring/v1/0000_50_monitoring_02_alertrelabelconfigs.crd.yaml
index ed3b1784..45b64f91 100644
--- a/monitoring/v1/0000_50_monitoring_02_alertrelabelconfigs.crd.yaml
+++ b/monitoring/v1/0000_50_monitoring_02_alertrelabelconfigs.crd.yaml
@@ -82,7 +82,7 @@ spec:
                         description: targetLabel to which the resulting value is written in a 'Replace' action. It is mandatory for 'Replace' and 'HashMod' actions. Regex capture groups are available.
                         type: string
                     x-kubernetes-validations:
-                      - rule: (self.action == 'LabelDrop' || self.action == 'LabelKeep') || has(self.sourceLabels)
+                      - rule: (self.action == 'LabelDrop' || self.action == 'LabelKeep') || size(self.sourceLabels) > 0
                         message: sourceLabels is required for actions Replace, Keep, Drop, HashMod and LabelMap
                       - rule: (self.action != 'Replace' && self.action != 'HashMod') || has(self.targetLabel)
                         message: targetLabel is required when action is Replace or HashMod
diff --git a/monitoring/v1/types.go b/monitoring/v1/types.go
index 6cd2215f..d3000025 100644
--- a/monitoring/v1/types.go
+++ b/monitoring/v1/types.go
@@ -293,7 +293,7 @@ type LabelName string
 // - https://prometheus.io/docs/prometheus/latest/configuration/configuration/#alert_relabel_configs
 // - https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config
 //
-// +kubebuilder:validation:XValidation:rule="(self.action == 'LabelDrop' || self.action == 'LabelKeep') || has(self.sourceLabels)",message="sourceLabels is required for actions Replace, Keep, Drop, HashMod and LabelMap"
+// +kubebuilder:validation:XValidation:rule="(self.action == 'LabelDrop' || self.action == 'LabelKeep') || size(self.sourceLabels) > 0",message="sourceLabels is required for actions Replace, Keep, Drop, HashMod and LabelMap"
 // +kubebuilder:validation:XValidation:rule="(self.action != 'Replace' && self.action != 'HashMod') || has(self.targetLabel)",message="targetLabel is required when action is Replace or HashMod"
 // +kubebuilder:validation:XValidation:rule="!has(self.modulus) || (has(self.modulus) && size(self.sourceLabels) > 0)",message="modulus requires sourceLabels to be present"
 // +k8s:openapi-gen=true
@@ -303,6 +303,7 @@ type RelabelConfig struct {
        // configured regular expression for the Replace, Keep, and Drop actions.
        //
        // +optional
+       // +kubebuilder:validation:maxItems:=10
        SourceLabels []LabelName `json:"sourceLabels,omitempty"`

        // separator placed between concatenated source label values. When omitted,
 ```
and run `make update-codegen-crds integration` the rule causes this error (abbreviated message):
```
unable to create CRD instances: unable to create CRD "alertrelabelconfigs.monitoring.openshift.io": CustomResourceDefinition.apiextensions.k8s.io "alertrelabelconfigs.monitoring.openshift.io" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[configs].items.x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of 1.048576x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)
```

File: monitoring/v1/types.go (hunk 20 above)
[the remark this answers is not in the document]
Comment (jan--f) [reply monitoring/v1/types.go]: Can you link a source for that rule? I was going by the alerting rule syntax documentation.
2023-02-27-200346_497x105

File: monitoring/v1/types.go [hunk 22]
```diff
@@ -0,0 +1,349 @@
...
+	// 'LabelMap'. Regex capture groups are available. Default is: '$1'
+	//
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self.action != 'Replace' && self.action != 'LabelMap' || has(self.replacement)",message="replacement is required when action is Replace or LabelMap"
+	Replacement string `json:"replacement,omitempty"`
+
+	// action to perform based on regex matching. Must be one of: Replace, Keep,
+	// Drop, HashMod, LabelMap, LabelDrop, or LabelKeep.  Default is: 'Replace'
+	//
+	// +kubebuilder:validation:Enum=Replace;Keep;Drop;HashMod;LabelMap;LabelDrop;LabelKeep
```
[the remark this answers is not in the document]
jan--f: :+1: on taking those at a later stage. The validation for those is also fairly complex.

monitoring/v1/types.go
jan--f: I dropped the kubebuilder default annotation since for some actions this and other fields are required to be empty. For the actions that don't require a `replacement` prometheus itself will set the default if needed. For the actions that require `replacement` I added a cross validation rule.

monitoring/v1/types.go
jan--f: Yes this would have indeed caused an issue. I dropped the default annotation and added a cross validation for cases where `regex` is required.

monitoring/v1/types.go
simonpasquier: > Where is this being passed to? Does it end up in logs? Does it end up in any prometheus configuration? Is there anywhere where this could be injected that might mean that a user could do something unintended (thinking injection attacks here) if we allow any character?

This ends up in the Prometheus rules configuration file. I can't think of any way this could trigger a vulnerability. If one exists then it is on the Prometheus side.

> Have you tested that required is the same as minLength: 1, because I suspect that required will pass if name: "" (you could use our integration testing suite to verify this)

Pretty sure that required ≠ minLengh=1.

monitoring/v1/types.go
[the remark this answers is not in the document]
simonpasquier: It says "label value" not "label name" :)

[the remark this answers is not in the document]
simonpasquier: And sorry for the confusion as I wrote "label names" in my earlier comment :facepalm:

jan--f:
@JoelSpeed @simonpasquier Thanks a lot for the input, I fixed a couple of issues that you pointed out:

1. I dropped the default value for `replacement` since actions `LabelKeep` and `LabelDrop` don't allow that field.
2. I added additional cross validation rules. I mostly implemented the rules that Prometheus itself uses when [unmarshaling a relabel rule](https://github.com/prometheus/prometheus/blob/8f3b2fc9aeaf1bd3f839e820e7f9ffc571006e89/model/relabel/relabel.go#L100-L152). I have however skipped rules that call functions in the Prometheus code (like !relabelTarget.MatchString(c.Replacement), which validates a value against a regex). The [first seven rules](https://github.com/openshift/api/pull/1406/files#diff-296785711e0969d01c7cd355a5ccbb002f71b86b3b39db815bf0a6d400c03d21R296-R302) are a result of this. The [remaining 3 rules](https://github.com/openshift/api/pull/1406/files#diff-296785711e0969d01c7cd355a5ccbb002f71b86b3b39db815bf0a6d400c03d21R303-R3) are based on the [relabel documentation](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config) where some actions require certain field to be present.

I think the last remaining question is whether alert names can be UTF-8 or if they have to be a valid label name. See https://github.com/openshift/api/pull/1406#discussion_r1114480193

monitoring/v1/types.go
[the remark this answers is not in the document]
jan--f: :facepalm: indeed, reading is hard. Fix is coming up.

monitoring/v1/types.go
jan--f: +1 on having minLength and required. The group name must not be empty. [Prometheus validates this](https://github.com/prometheus/prometheus/blob/8f3b2fc9aeaf1bd3f839e820e7f9ffc571006e89/model/rulefmt/rulefmt.go#L101-L109) as well. I don't think we can validate the "unique rule group names per file" requirement here.

I agree with Simon on the vulnerability question. The group name is and should only ever be treated as a string. Prometheus-operators [PrometheusRule CRD also uses a string](https://github.com/prometheus-operator/prometheus-operator/blob/6c18db1083546ae6950f5fda33bd6629c10487e2/pkg/apis/monitoring/v1/prometheusrule_types.go#L63) here.

monitoring/v1/types.go
jan--f: The spacing is not preserved indeed. The contains `A common example is the alert severity: \n labels: severity: warning`. I suppose that should render as
``` 
common example is the alert severity: 
 labels: severity: warning
```
Wdyt?

simonpasquier:
> I think the last remaining question is whether alert names can be UTF-8 or if they have to be a valid label name. See [#1406 (comment)](https://github.com/openshift/api/pull/1406#discussion_r1114480193)

Alert names can be any UTF-8 strings (see Prometheus [validation code](https://github.com/prometheus/prometheus/blob/8f3b2fc9aeaf1bd3f839e820e7f9ffc571006e89/model/rulefmt/rulefmt.go#L167-L254)). The alert name ends up as a label *value* (`alertname`) of the ALERTS metric:

image

monitoring/v1/types.go
```diff
@@ -0,0 +1,357 @@
...
+	// aware that this can have a performance impact on Prometheus if the group is
+	// extremely large or has very complex query expressions to evaluate.
+	// Spreading very complex rules across multiple groups to allow them to be
+	// processed in parallel is also a common use-case.
+	//
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MinItems:=1
+	// +kubebuilder:validation:Required
+	Groups []RuleGroup `json:"groups"`
```
simonpasquier, line 88: I don't remember why we said `+kubebuilder:validation:MinItems:=1` . An object without any group could be useful if you want to "remove" rules without deleting the object.

monitoring/v1/types.go
simonpasquier: > I don't think we can validate the "unique rule group names per file" requirement here

it will be handled by `+listType=map` and `+listMapKey=name` on the `groups` field.

monitoring/v1/types.go
JoelSpeed: Why would you want to keep an empty object around? That's not the expected way of interacting with kube, normally you'd just remove the object

monitoring/v1/types.go
```diff
@@ -0,0 +1,357 @@
...
+
+	// interval is how often rules in the group are evaluated.  If not specified,
+	// it defaults to the global.evaluation_interval configured in Prometheus,
+	// which itself defaults to 30 seconds.  You can check if this value has been
+	// modified from the default on your cluster by inspecting the platform
+	// Prometheus configuration:
+	//
+	// $ oc -n openshift-monitoring describe prometheus k8s
+	//
+	// The relevant field in that resource is: spec.evaluationInterval
```
JoelSpeed, line 115: Might want to review how this is rendered in `oc describe`

monitoring/v1/types.go
```diff
@@ -0,0 +1,357 @@
...
+// See Prometheus documentation:
+// - https://www.prometheus.io/docs/prometheus/latest/configuration/alerting_rules
+//
+// +k8s:openapi-gen=true
+type Rule struct {
+	// alert is the name of the alert. Must be a valid label value, i.e. may
+	// contain any Unicode character.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
```
JoelSpeed, line 139: Should we set a max length? Even if it is pretty large (1024?), just to be safe

monitoring/v1/types.go
```diff
@@ -0,0 +1,357 @@
...
+	// are available.
+	//
+	// +optional
+	TargetLabel string `json:"targetLabel,omitempty"`
+
+	// regex against which the extracted value is matched. Default is: '(.*)'
+	// regex is required for all actions except 'HashMod'
+	//
+	// +optional
+	// +kubebuilder:default=(.*)
```
JoelSpeed, line 333: Is the regex valid for all action types? Do we really want to default it?

monitoring/v1/types.go
JoelSpeed: Is it optional or not allowed for others? We should specify within the godoc

Either `and optional for other actions`, or `and forbidden for other actions`

monitoring/v1/types.go
jan--f: I revised the text. Not sure an example `oc` command adds that much. This is well documented in our official docs and might change in the future.

monitoring/v1/types.go
jan--f: Prose it is.

monitoring/v1/types.go
jan--f: I think yes. Its the [first thing prometheus does after unmarshaling the yaml](https://github.com/prometheus/prometheus/blob/8f3b2fc9aeaf1bd3f839e820e7f9ffc571006e89/model/relabel/relabel.go#L107) and its valid for all actions.

monitoring/v1/types.go
```diff
@@ -0,0 +1,357 @@
...
+	// +listMapKey=name
+	// +kubebuilder:validation:MinItems:=1
+	// +kubebuilder:validation:Required
+	Groups []RuleGroup `json:"groups"`
+}
+
+// Duration is a valid prometheus time duration.
+// Supported units: y, w, d, h, m, s, ms
+// Examples: `30s`, `1m`, `1h20m15s`, `15d`
+// +kubebuilder:validation:Pattern:="^(0|(([0-9]+)y)?(([0-9]+)w)?(([0-9]+)d)?(([0-9]+)h)?(([0-9]+)m)?(([0-9]+)s)?(([0-9]+)ms)?)$"
```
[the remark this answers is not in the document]
jan--f: I like this but unfortnately this rule evaluation is much to costly:
`spec.validation.openAPIV3Schema.properties[spec].properties[groups].items.properties[interval].x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of more than 100x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)`

monitoring/v1/types.go
JoelSpeed: I'm ok with that, having unbounded strings though could be bad, we should have some sensible but high limit on most things that have length

monitoring/v1/types.go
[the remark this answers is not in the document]
JoelSpeed: That's a shame 😅 Ok, keep it as it is then I guess

monitoring/v1/types.go
JoelSpeed: Prometheus is initialising a blank regex though?

monitoring/.codegen.yaml
```diff
@@ -1,3 +1,5 @@
 schemapatch:
   requiredFeatureSets:
+  - ""
+  - "Default"
   - "TechPreviewNoUpgrade"
```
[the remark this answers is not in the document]
jan--f: This produces
```
error running generator swaggerdocs on group monitoring.openshift.io:
	error generating swagger docs for monitoring.openshift.io/v1alpha1:
		found 4 missing comments:
In struct: AlertRelabelConfig, field documentation is missing: metadata
In struct: AlertRelabelConfigList, field documentation is missing: metadata
In struct: AlertingRule, field documentation is missing: metadata
In struct: AlertingRuleList, field documentation is missing: metadata
```
Same for `v1` of course. Seems like other APIs have the same "issue"?

monitoring/v1/types.go
jan--f: Prometheus initializes the default that we have here, `(.*)`. https://github.com/prometheus/prometheus/blob/8f3b2fc9aeaf1bd3f839e820e7f9ffc571006e89/model/relabel/relabel.go#L33

monitoring/v1/types.go
jan--f: :+1: for string types.
For lists and maps I think we need to be more careful with this limit. A string length limitation can possibly be worked around, not so for lists and maps.

monitoring/.codegen.yaml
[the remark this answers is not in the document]
JoelSpeed: Yes, you're right, I was expecting this, the fix for each of those warnings if to find the `metadata` field and add a comment, which you can copy/paste from here onto each one in both APIs to fix
```
```
> Seems like other APIs have the same "issue"?

They do indeed, not only on metadata, but in a wider scope too, my plan is to fix these up in time, so that no warnings are produced

monitoring/v1/types.go
JoelSpeed: Ack ok, SGTM then

monitoring/.codegen.yaml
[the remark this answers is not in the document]
jan--f: Ack! I checked the example api, but had no luck.

monitoring/v1/types.go
JoelSpeed: May contain any unicode character? Really? Are we sure we want to support that? Is there any way could validate this with a regex?

monitoring/v1/types.go
```diff
@@ -0,0 +1,362 @@
...
+// point to the generated PrometheusRule object in the AlertingRule status.
+type PrometheusRuleRef struct {
+	// This is a struct so that we can support future expansion of fields within
+	// the reference should we ever need to.
+
+	// name of the referenced PrometheusRule.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=2048
+	Name string `json:"name"`
```
JoelSpeed, line 211: What's the valid character set for a name?

monitoring/v1/types.go
```diff
@@ -0,0 +1,362 @@
...
+// +kubebuilder:validation:XValidation:rule="(self.action != 'LabelDrop' && self.action != 'LabelKeep') || !has(self.replacement)",message="LabelKeep and LabelDrop actions require only 'regex', and no other fields (found replacement)"
+// +kubebuilder:validation:XValidation:rule="!has(self.modulus) || (has(self.modulus) && size(self.sourceLabels) > 0)",message="modulus requires sourceLabels to be present"
+// +kubebuilder:validation:XValidation:rule="(self.action == 'LabelDrop' || self.action == 'LabelKeep') || has(self.sourceLabels)",message="sourceLabels is required for actions Replace, Keep, Drop, HashMod and LabelMap"
+// +kubebuilder:validation:XValidation:rule="(self.action != 'Replace' && self.action != 'LabelMap') || has(self.replacement)",message="replacement is required for actions Replace and LabelMap"
+// +k8s:openapi-gen=true
+type RelabelConfig struct {
+	// sourceLabels select values from existing labels. Their content is
+	// concatenated using the configured separator and matched against the
+	// configured regular expression for the Replace, Keep, and Drop actions.
+	// Not allowed for actions 'LabelKeep' and 'LabelDrop'.
```
JoelSpeed, line 309: You're inconsistent in how you refer to the action types, sometimes quoting, sometimes not, can we standardise please?

JoelSpeed (commented):
Can we check that each string value in this API has mentioned in the godoc what the valid characters are and, where possible, introduce a validation for this

Eg if we can only have alphanumeric characters, we can easily validate that, if you can have anything thats not a control character, we can regex that too (`[^[:cntrl:]]`) for example

monitoring/v1/types.go
simonpasquier: Any valid Go string.

monitoring/v1/types.go
simonpasquier: I'm not sure why we want to put restrictions on which characters are supported knowing that Prometheus supports any valid Go string.

monitoring/v1/types.go
simonpasquier: This is done.

monitoring/v1/types.go
JoelSpeed: In that case, fine, just wanted to ask the question a few times in case it triggered some realisation that actually, there are some characters we don't want to support, if you are certain any valid go string is good, then this is ok to proceed as is

monitoring/v1/types.go
JoelSpeed: What does an empty group achieve in prometheus? Wondering why you wouldn't just remove it? (Which is the sentiment we would expect here)

JoelSpeed:
@simonpasquier @jan--f Are you happy with this now? I think if we are good to go, lets squash this down and get it merged, I have no further feedback

monitoring/v1/types.go
jan--f: I think I'm in favor of allowing empty rule groups. Besides @simonpasquier's argument about it being a valid prometheus config, I can see future use cases using empty groups even if they arguably rather hypothetical at this point.
Say we'd like to add flexibility in terms of how often rules are evaluated. We could provide predefined rule groups (with their own eval interval) and let a user move rule between them.
@JoelSpeed I have removed the `MinItems` requirement. Let me know if that is objectionable.

jan--f:
> @simonpasquier @jan--f Are you happy with this now? I think if we are good to go, lets squash this down and get it merged, I have no further feedback

I think one final ack of the drop of `MinItems` for `Groups` above and then we're good to go from my perspective.

monitoring/v1/types.go
simonpasquier: As I said before, I've got no strong feeling about this. Sticking with the usual Kubernetes practices is a good argument for having `MinItems > 0`.

> Do you ever see a future where there is a sibling to groups within the spec? Meaning that groups would be optional and not necessarily need an item? ie, a valid configuration that needs no groups?

I can't think of anything concrete on top of my head. If such situation arises, I guess that we could lift the requirement and allow empty groups?

JoelSpeed: > I can't think of anything concrete on top of my head. If such situation arises, I guess that we could lift the requirement and allow empty groups?

Yes, we can relax this in the future provided we know that we can update all consumers to understand the change, luckily, we should be the owners of the consumers here. You couldn't make it tighter in the future as that could mean there are stored objects that no longer make the spec. For now I'd be in favour of including it, then relaxing in the future if we find a need to

JoelSpeed:
PR is reporting a merge conflict, can we make sure it's rebased please

JoelSpeed:
/approve

/hold

@jan--f Have you considered getting QE to review this before merge? Is there any way they can pre-merge validate this in combination with an operator PR to make sure it's working as expected?

jan--f:
> Have you considered getting QE to review this before merge? Is there any way they can pre-merge validate this in combination with an operator PR to make sure it's working as expected?

Good questions. I can certainly raise the related CMO PR. @juzhao wdyt?

jan--f:
@JoelSpeed I raised https://github.com/openshift/cluster-monitoring-operator/pull/1945, that pulls in this PR and the related openshift/client-go wip PR.
This will run our existing [e2e test](https://github.com/openshift/cluster-monitoring-operator/blob/e154bc6e83de52e6e6e3f3b1cc04d3cb94d81ee5/test/e2e/alert_relabel_config_test.go) for this API. Wdyt?

juzhao:
> Good questions. I can certainly raise the related CMO PR. @juzhao wdyt?
will review and test the PR

juzhao:
failed to launch cluster-bot cluster
`launch openshift/api#1406 gcp`
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp-modern/1645977177813946368
```
ERRO[2023-04-12T02:31:12Z] Failed to load arguments.                     error=invalid configuration: you must define at least one test or image build in 'tests' or 'images'
```

juzhao:
tested along with the CMO PR
`launch 4.14-ci,openshift/api#1406,openshift/cluster-monitoring-operator#1945 gcp`
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp-modern/1646048977654648832
still the same error
```
{  invalid configuration: you must define at least one test or image build in 'tests' or 'images'}
```
as it's not easy to do pre-merge testing for openshift/api PR, I suggest we merge the openshift/api PR first, then test the CMO PR

juzhao:
the code is LGTM
/lgtm

jan--f:
@JoelSpeed PTAL.
tests in https://github.com/openshift/cluster-monitoring-operator/pull/1945 have also passed.

JoelSpeed:
@juzhao OpenShift API cannot be tested directly as you have tried with cluster bot, and may only be tested by inclusion in PRs in other repositories, so the CMO PR has included these latest changes, please try directly with CMO only and it should show the updated v1 API from this PR

juzhao:
@jan--f see above comment, do we have any change not in PR: https://github.com/openshift/cluster-monitoring-operator/pull/1945, if no changes for CMO PR, I will test the PR with CMO PR only

juzhao:
launched CMO PR alone, 
`launch 4.14-ci,openshift/cluster-monitoring-operator#1945 no-spot`
1. did not enable TechPreviewNoUpgrade featureSet, no "alertingrules.monitoring.openshift.io" and "alertrelabelconfigs.monitoring.openshift.io" CRDs, so can't create AlertingRule with "apiVersion: monitoring.openshift.io/v1", error is
```
error: resource mapping not found for name: "example" namespace: "openshift-monitoring" from "STDIN": no matches for kind "AlertingRule" in version "monitoring.openshift.io/v1"
ensure CRDs are installed first
```
according to @jan--f , should have these CRDs in default cluster, no need to enable TechPreviewNoUpgrade

2. enabled TechPreviewNoUpgrade featureSet, there are "alertingrules.monitoring.openshift.io" and "alertrelabelconfigs.monitoring.openshift.io" CRDs, but still can't create AlertingRule with "apiVersion: monitoring.openshift.io/v1", error is
```
EOF
error: resource mapping not found for name: "example" namespace: "openshift-monitoring" from "STDIN": no matches for kind "AlertingRule" in version "monitoring.openshift.io/v1"
ensure CRDs are installed first
```
API version is still v1alpha1 for alertingrules from "oc explain" and CRD, full info please see from http://pastebin.test.redhat.com/1097396

is this a bug for CMO or we need to merge openshift/api PR first?

juzhao:
should be CMO bug  https://issues.redhat.com/browse/OCPBUGS-11804

juzhao:
[OCPBUGS-11804](https://issues.redhat.com/browse/OCPBUGS-11804) is fixed, https://github.com/openshift/cluster-monitoring-operator/pull/1945 is LGTM

PR #1415, opened 2023-03-01
Add loadbalancer field to Power VS machine provider spec

Karthik-K-N:
This PR adds LoadBalancers filed to Power VS spec.
Usage:
1. Installer will set the loadbalaner name in machine spec for control-plane machines
2. Power VS machine api provider will use it to add or remove created/deleted instance to loadbalancer pool
3. This will help us in providing support for [cluster-control-plane-machine-set-operator](https://github.com/openshift/cluster-control-plane-machine-set-operator)


Karthik-K-N:
@mkumatag @JoelSpeed could you please take a look and help in reviewing this.

machine/v1/types_powervsprovider.go
```diff
@@ -190,3 +204,9 @@ type PowerVSSecretReference struct {
 	// +optional
 	Name string `json:"name,omitempty"`
 }
+
+// LoadBalancerReference is a reference to a load balancer on IBM Cloud virtual private cloud.
+type LoadBalancerReference struct {
+	Name string                 `json:"name"`
+	Type IBMVPCLoadBalancerType `json:"type"`
+}
```
JoelSpeed, line 212: These are both required fields? Can you please add some godoc to them, following the guidance for [writing user readable documentation](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#write-user-readable-documentation-in-godoc)

machine/v1/types_powervsprovider.go
```diff
@@ -121,6 +130,11 @@ type PowerVSMachineProviderConfig struct {
 	// default, which is subject to change over time. The current default is 32.
 	// +optional
 	MemoryGiB int32 `json:"memoryGiB,omitempty"`
+
+	// LoadBalancers is the set of load balancers to which the new instance
+	// should be added once it is created.
+	// +optional
+	LoadBalancers []LoadBalancerReference `json:"loadBalancers,omitempty"`
```
JoelSpeed, line 137: What will happen if this is set on a worker machine? Will it still work?
Is there any difference between internat and external LBs in IBM?

machine/v1/types_powervsprovider.go
```diff
@@ -11,6 +11,15 @@ type PowerVSResourceType string
 // PowerVSProcessorType enum attribute to identify the PowerVS instance processor type
 type PowerVSProcessorType string
 
+// IBMVPCLoadBalancerType is the type of LoadBalancer to use when registering
+// an instance with load balancers specified in LoadBalancerNames
+type IBMVPCLoadBalancerType string
+
+// ApplicationLoadBalancerType is possible values for IBMVPCLoadBalancerType.
+const (
+	ApplicationLoadBalancerType IBMVPCLoadBalancerType = "application" // Application Load Balancer for VPC (ALB)
```
JoelSpeed, line 20: Kube conventions dictate that enums should use PascalCase, I would expect this to be `Application` and to be converted to the IBM version within the controller

machine/v1/types_powervsprovider.go
Karthik-K-N: We have planned to set it to only on control plane machines and even in MAPI we can have check to add loadbalancer for only Master VMs.
In IBM we have public loadbalancer for serving api server traffic which will be served from 6443 port and private for internal communications , it serves on port 6443 and 22623 for machine-config traffic.

machine/v1/types_powervsprovider.go
Karthik-K-N: Updated to use it as Application.

machine/v1/types_powervsprovider.go
Karthik-K-N: Both are required fields and added some documentation. Please take a look.

machine/v1/types_powervsprovider.go
```diff
@@ -190,3 +204,18 @@ type PowerVSSecretReference struct {
 	// +optional
 	Name string `json:"name,omitempty"`
 }
+
+// LoadBalancerReference is a reference to a load balancer on IBM Cloud virtual private cloud(VPC).
+type LoadBalancerReference struct {
+	// Name of the LoadBalancer in IBM Cloud VPC
+	// It is a reference to existing LoadBalancer created by openshift installer component.
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
```
JoelSpeed, line 213: Should start with the json tag name not the Go field name, is there any way to restrict this? For example can we explain what character set, minimum/maximum lengths are required for this field?
```suggestion
	// name of the LoadBalancer in IBM Cloud VPC
	// It is a reference to existing LoadBalancer created by openshift installer component.
	// +kubebuilder:validation:Required
	Name string `json:"name"`
```

machine/v1/types_powervsprovider.go
```diff
@@ -190,3 +204,18 @@ type PowerVSSecretReference struct {
...
+	// Type of the LoadBalancer service supported by IBM Cloud VPC
```
JoelSpeed, line 214:
```suggestion
	// type of the LoadBalancer service supported by IBM Cloud VPC
```

machine/v1/types_powervsprovider.go
Karthik-K-N: sure , not sure how can we restrict it, Currently the loadbalancer name will be of format "{cluster-id}-loadbalancer" for public and "{cluster-id}-loadbalancer-int" for internal loadbalancer

machine/v1/types_powervsprovider.go
Karthik-K-N: Updated.

machine/v1/types_powervsprovider.go
```diff
@@ -190,3 +204,18 @@ type PowerVSSecretReference struct {
...
+	// Currently only Application LoadBalancer is supported
+	// More details about Application LoadBalancer
+	// https://cloud.ibm.com/docs/vpc?topic=vpc-load-balancers-about&interface=ui
```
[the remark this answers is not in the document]
Karthik-K-N: Updated.

machine/v1/types_powervsprovider.go
JoelSpeed: Is cluster-id restricted at all by the installer? We can probably restrict this to alphanumeric characters, hyphens and underscores, but we should check what restrictions are imposed by the installer

Karthik-K-N: I could see these restrictions: https://github.com/Karthik-K-N/installer/blob/e91df626c10e569e7613249053b7b9b264db42df/pkg/asset/installconfig/clusterid.go#L37-L50

Karthik-K-N: Apologies we are using infraID not clusterid, Infra ID is generated like this: https://github.com/Karthik-K-N/installer/blob/e91df626c10e569e7613249053b7b9b264db42df/pkg/asset/installconfig/clusterid.go#L60-L79

I think we can add validation for alphanumeric with hyphen,

JoelSpeed: Yeah looks like it can be limited to alphanum and hyphen, great!

machine/v1/types_powervsprovider.go
```diff
@@ -190,3 +204,19 @@ type PowerVSSecretReference struct {
 	// +optional
 	Name string `json:"name,omitempty"`
 }
+
+// LoadBalancerReference is a reference to a load balancer on IBM Cloud virtual private cloud(VPC).
+type LoadBalancerReference struct {
+	// name of the LoadBalancer in IBM Cloud VPC.
+	// It is a reference to existing LoadBalancer created by openshift installer component.
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
```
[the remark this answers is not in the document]
JoelSpeed: I found the following in the docs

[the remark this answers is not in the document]
Karthik-K-N: Loadbalancer name may contain hyphens, Currently this is how the installer creates the loadbalancer `rdr-kn08031125-vhwtg-loadbalancer`.
From [api doc](https://cloud.ibm.com/apidocs/vpc/latest#list-load-balancer-profiles) I can see this description for name
```
The globally unique name for this load balancer profile

Possible values: 1 ≤ length ≤ 63, Value must match regular expression ^([a-z]|[a-z][-a-z0-9]*[a-z0-9]|[0-9][-a-z0-9]*([a-z]|[-a-z][-a-z0-9]*[a-z0-9]))$

Example: network-fixed
```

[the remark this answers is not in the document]
JoelSpeed: Great, thanks for finding that, can you explain this in prose within the godoc please? We need to say something along the lines of
```
// The name should be between 1 and 63 characters long and may consist of lowercase alphanumeric characters and hyphens only.
// The value must not end with a hyphen and integer values are not allowed.
```
I think I've understood that regex correctly anyway

JoelSpeed (commented):
If we can just add a touch more documentation to this name field as per my suggestion I think this is good to go

machine/v1/types_powervsprovider.go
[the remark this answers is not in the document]
Karthik-K-N: > If we can just add a touch more documentation to this name field as per my suggestion I think this is good to go

Updated accordingly, Please take a look. Thank you.

PR #1418, opened 2023-03-03
api_operator_v1_docs: downstream doc update request for unsupportedCo…

dfitzmau:
Some Red Hat OCP content is pulled from API docs: this content is auto-generated from parts of binaries that cannot be modified in the openshift-docs repo. 

Red Hat customer requested a doc changed based on the following feedback:

> The problem is that if a user has set something in 'unsupportedConfigOverrides' previously, those changes are also visible in the observedConfig status, but those parts of the observedConfig are not overlayed with a newer unsupportedConfigOverrides.

> In other words:
1. I set a feature-gate in unsupportedConfigOverrides
2. This feature-gate then appears in observedConfig
3. At this point, the documentation implies that if I wanted to add a second feature-gate, I could set unsupporedConfigOverrides to contain only the new option and it would be combined with the previous value I see in observedConfig, resulting in both flags being set.  However, I actually need to set unsupportedConfigOverrides to be the value I set in (1) plus the additional feature-gate I want, or the previous value set in (1) gets dropped.

>  We should come up with a clearer way to explain this fact both in the description of the field and the online documentation.

See links (link point to private networks that require credentials):

* https://bugzilla.redhat.com/show_bug.cgi?id=2061949
* https://github.com/openshift/openshift-docs/pull/55319

Downstream PR in the openshift-docs repository contains an approved solution for the content. Please let me know your thoughts. If this content gets approved in the API repo, the content can be pulled into the openshift-docs repo.


dfitzmau:
Hi @soltysh . The automated message stated that I need to tag you for approval. I'm unsure of the messages, because I am changing the existing text for an API definition. I've not contributed to this project before, so I'm not 100 sure on what guidelines I needed to apply to change text, but leave all other API criteria intact.

dfitzmau:
Hi @JoelSpeed or @deads2k . Sorry to ask: I work in the docs team for OpenShift. I'm not fully sure how to steer this PR in the right direction. I'm new to the API space and I do not come here from an engineer's perspective. Would you know anyone that can help me through the process to make a content update for a parameter?

JoelSpeed:
Couple of points for the godoc, we have a convention that every field must start with the json tag name, so your current iteration will need to be reworded.

Secondly, any whitespacing you are adding for effect, I don't think renders correctly when it's generated into the product docs, so thinks like lists might not work, so may want to check on that

operator/v1/types.go
```diff
@@ -70,11 +70,11 @@ type OperatorSpec struct {
...
-	// unsupportedConfigOverrides holds a sparse config that will override any previously set options.  It only needs to be the fields to override
-	// it will end up overlaying in the following order:
-	// 1. hardcoded defaults
-	// 2. observedConfig
-	// 3. unsupportedConfigOverrides
+	// You can use `unsupportedConfigOverrides` to add configuration changes that are not supported by an operator. 
+	// Consider seeking guidance from the Red Hat support team before you use this property, 
+	// because using this property without knowledge of its capabilities could lead to unexpected behavior or conflict with other configuration options.
+	//
+	// Additionally, using this property blocks cluster upgrades, so remove this setting before you upgrade your cluster.
```
JoelSpeed, line 77: Do we allow people to remove this value? Would have to find an SME for this as I don't actually know

JoelSpeed:
Once we've agreed on what the text should say, you'll need to update the generated code.

I think if you use `PROTO_OPTIONAL=true GOPATH= make update` you should be good

operator/v1/types.go
dfitzmau: Thanks, @JoelSpeed . I'll ask an SME to confirm.

operator/v1/types.go
```diff
@@ -70,11 +70,11 @@ type OperatorSpec struct {
 	// +kubebuilder:default=Normal
 	OperatorLogLevel LogLevel `json:"operatorLogLevel,omitempty"`
 
-	// unsupportedConfigOverrides holds a sparse config that will override any previously set options.  It only needs to be the fields to override
-	// it will end up overlaying in the following order:
-	// 1. hardcoded defaults
-	// 2. observedConfig
-	// 3. unsupportedConfigOverrides
+	// UnsupportedConfigOverrides adds configuration changes that are not supported by an operator.
```
JoelSpeed, line 73: Must be lower case `u` on this so that the name matches the json tag, which is what the user would see

operator/v1/types.go
sanchezl: Yes, the field can be removed.

operator/v1/types.go
```diff
@@ -70,11 +70,11 @@ type OperatorSpec struct {
 	// +kubebuilder:default=Normal
 	OperatorLogLevel LogLevel `json:"operatorLogLevel,omitempty"`
 
-	// unsupportedConfigOverrides holds a sparse config that will override any previously set options.  It only needs to be the fields to override
-	// it will end up overlaying in the following order:
-	// 1. hardcoded defaults
-	// 2. observedConfig
-	// 3. unsupportedConfigOverrides
```
JoelSpeed, line 77: So I think the new content is good, useful context, but I feel like we are also losing some context by dropping this content, any way they can be combined?

dfitzmau: Thanks for the quick review and reply. I'll need to ask a few people.

dfitzmau: For context, here is the original comment for removing the original context: https://github.com/openshift/openshift-docs/pull/55319#discussion_r1114817459

I can see the reasoning to remove the original text, because of the unsupported nature of the resource.

JoelSpeed: I think it can possibly be reworded, but pointing out that this overrides the `observedConfig` is worthwhile context IMO

operator/v1/types.go
```diff
@@ -70,11 +70,11 @@ type OperatorSpec struct {
...
+	// unsupportedConfigOverrides adds configuration changes that are not supported by an operator.
+	// Consider seeking guidance from the Red Hat support team before you use this property,
+	// because by using this property without knowledge of its capabilities could lead to unexpected behavior or conflict with other configuration options.
+	//
+	// Additionally, using this property blocks cluster upgrades, so remove this setting before you upgrade your cluster.
```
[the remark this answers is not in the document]
dfitzmau: Thanks, @JoelSpeed . Super helpful!

I'll need to discuss the changes with the SME, because they wanted to remove certain content and replace it with specific (more generalized text). I'll update you straight away when I get an answer to your queries and suggestions.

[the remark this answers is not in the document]
sanchezl: How about something like this for the first line:

> `unsupportedConfigOverrides` overrides the final configuration computed by the operator, enabling configuration properties to be overwritten or added in a manner not supported by Red Hat.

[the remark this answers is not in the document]
dfitzmau: Thanks, @sanchezl . Your suggestion aligns the text well with the feedback. I've implemented your suggested with some slight tweaks.

operator/v1/types.go
dfitzmau: Thanks, @JoelSpeed . The text is reworded now.

operator/v1/types.go
dfitzmau: Thanks, @sanchezl and @JoelSpeed . I see an `optional` flag with this property, so I do not need to explicitly state that one can remove the property if they so wish?

operator/v1/types.go
```diff
@@ -70,11 +70,12 @@ type OperatorSpec struct {
 	// +kubebuilder:default=Normal
 	OperatorLogLevel LogLevel `json:"operatorLogLevel,omitempty"`
 
-	// unsupportedConfigOverrides holds a sparse config that will override any previously set options.  It only needs to be the fields to override
-	// it will end up overlaying in the following order:
-	// 1. hardcoded defaults
-	// 2. observedConfig
-	// 3. unsupportedConfigOverrides
+	// unsupportedConfigOverrides overrides the final configuration that was computed by the operator.
+	// By using this property, you create an environment where configuration properties can be overwritten or added in a way that is not supported by Red Hat.
```
JoelSpeed, line 74: This sentence is a bit long, do we need it to be so verbose? I think all we need to say here is that using this is unsupported right?
```suggestion
	// Use of this field is not supported by Red Hat.  
```

JoelSpeed (commented):
Just reviewing this again, it seems all of my previous review suggestions have been reverted, can we restore those? 

In particular, "use of this property without knowledge" should come early as it is important. We should remove filler like "because" and "additionally". Keep the context here precise and with purpose. The current mood of this comment reads more like a conversation than we would normally expect for API field documentation.

s/Consider seeking guidance/Seek guidance/
s/because by using this property/Use of this field/

Trying to be more direct with the tone here if that makes sense?

operator/v1/types.go
dfitzmau: Agreed, @JoelSpeed . I updated the content to make it shorter.

operator/v1/types.go
```diff
@@ -70,11 +70,12 @@ type OperatorSpec struct {
...
-	// it will end up overlaying in the following order:
-	// 1. hardcoded defaults
-	// 2. observedConfig
-	// 3. unsupportedConfigOverrides
+	// unsupportedConfigOverrides overrides the final configuration that was computed by the operator.
+	// Red Hat does not support the use of this property.
+	// Consider seeking guidance from the Red Hat support team before you use this property,
+	// because misuse of this property could lead to unexpected behavior or conflict with other configuration options.
+	//
+	// Additionally, using this property blocks cluster upgrades, so remove this setting before you upgrade your cluster.
```
JoelSpeed, line 78: I had suggested most of this before but it got reverted, I think this paragraph should read as
```suggestion
	// unsupportedConfigOverrides overrides the final configuration that was computed by the operator.
	// Red Hat does not support the use of this field.
	// Misuse of this field could lead to unexpected behavior or conflict with other configuration options.
	// Seek guidance from the Red Hat support before using this field.
	// Use of this property blocks cluster upgrades, it must be removed before upgrading your cluster.
```
Just trying to be concise and direct with the intention of what we are trying to portray here

dfitzmau: Yes, concise is the best way forward. Less maintenance and chance of causing confusion. My apologies for missing some of those suggested changes. I've implemented them now.

JoelSpeed:
I'm happy with the current version if SME are

dfitzmau:
> I'm happy with the current version if SME are

I've requested these reviews.

sanchezl (approved):
/lgtm

dfitzmau:
Thanks, Joel and Luis. I'll request a QE review to clsoe the review circle.

dfitzmau:
Hi @wangke19 . I had to migrate the doc update from https://github.com/openshift/openshift-docs/pull/55319 (openshift-docs) repo to the api repo (where the content resides). Would you be OK to review the changes and provide a LGTM? After the PR is merged, the OCP builds can pull the API updated content into the docs set.

PR #1423, opened 2023-03-15
STOR-1266: Add CSI migration option to Storage operator spec

dobsonj:
https://issues.redhat.com/browse/STOR-1266
/cc @openshift/storage

Dependent PR's:
- https://github.com/openshift/kubernetes/pull/1514
- https://github.com/openshift/cluster-kube-apiserver-operator/pull/1466
- https://github.com/openshift/cluster-kube-controller-manager-operator/pull/710
- https://github.com/openshift/cluster-kube-scheduler-operator/pull/469
- https://github.com/openshift/machine-config-operator/pull/3614
- https://github.com/openshift/cluster-storage-operator/pull/351
- https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/145

### DIFF
--- a/operator/v1/0000_50_cluster_storage_operator_01_crd.yaml
+++ b/operator/v1/0000_50_cluster_storage_operator_01_crd.yaml
@@ -69,6 +69,19 @@ spec:
                   type: object
                   nullable: true
                   x-kubernetes-preserve-unknown-fields: true
+                vsphereStorageDriver:
+                  description: 'VSphereStorageDriver indicates the storage driver to use on VSphere clusters. Once this field is set to CSIWithMigrationDriver, it can not be changed. If this is empty, the platform will choose a good default, which may change over time without notice. DEPRECATED: This field will be removed in a future release.'
+                  type: string
+                  enum:
+                    - ""
+                    - LegacyDeprecatedInTreeDriver
+                    - CSIWithMigrationDriver
+                  x-kubernetes-validations:
+                    - rule: self == oldSelf || oldSelf == "" || self == "CSIWithMigrationDriver"
+                      message: VSphereStorageDriver can not be changed once it is set to CSIWithMigrationDriver
+              x-kubernetes-validations:
+                - rule: '!has(oldSelf.vsphereStorageDriver) || has(self.vsphereStorageDriver)'
+                  message: VSphereStorageDriver is required once set
             status:
               description: status holds observed values from the cluster. They may not be overridden.
               type: object
--- a/operator/v1/types_storage.go
+++ b/operator/v1/types_storage.go
@@ -26,9 +26,28 @@ type Storage struct {
 	Status StorageStatus `json:"status"`
 }
 
+// StorageDriverType indicates whether CSI migration should be enabled for drivers where it is optional.
+// +kubebuilder:validation:Enum="";LegacyDeprecatedInTreeDriver;CSIWithMigrationDriver
+type StorageDriverType string
+
+const (
+	LegacyDeprecatedInTreeDriver StorageDriverType = "LegacyDeprecatedInTreeDriver"
+	CSIWithMigrationDriver       StorageDriverType = "CSIWithMigrationDriver"
+)
+
 // StorageSpec is the specification of the desired behavior of the cluster storage operator.
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.vsphereStorageDriver) || has(self.vsphereStorageDriver)", message="VSphereStorageDriver is required once set"
 type StorageSpec struct {
 	OperatorSpec `json:",inline"`
+
+	// VSphereStorageDriver indicates the storage driver to use on VSphere clusters.
+	// Once this field is set to CSIWithMigrationDriver, it can not be changed.
+	// If this is empty, the platform will choose a good default,
+	// which may change over time without notice.
+	// DEPRECATED: This field will be removed in a future release.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf || oldSelf == \"\" || self == \"CSIWithMigrationDriver\"",message="VSphereStorageDriver can not be changed once it is set to CSIWithMigrationDriver"
+	// +optional
+	VSphereStorageDriver StorageDriverType `json:"vsphereStorageDriver"`
 }
 
 // StorageStatus defines the observed status of the cluster storage operator.
--- a/openapi/generated_openapi/zz_generated.openapi.go
+++ b/openapi/generated_openapi/zz_generated.openapi.go
@@ -46544,6 +46544,14 @@ func schema_openshift_api_operator_v1_StorageSpec(ref common.ReferenceCallback) 
 							Ref:         ref("k8s.io/apimachinery/pkg/runtime.RawExtension"),
 						},
 					},
+					"vsphereStorageDriver": {
+						SchemaProps: spec.SchemaProps{
+							Description: "VSphereStorageDriver indicates the storage driver to use on VSphere clusters. Once this field is set to CSIWithMigrationDriver, it can not be changed. If this is empty, the platform will choose a good default, which may change over time without notice. DEPRECATED: This field will be removed in a future release.",
+							Default:     "",
+							Type:        []string{"string"},
+							Format:      "",
+						},
+					},
 				},
 				Required: []string{"managementState"},
 			},
--- a/openapi/openapi.json
+++ b/openapi/openapi.json
@@ -27274,6 +27274,11 @@
           "description": "unsupportedConfigOverrides overrides the final configuration that was computed by the operator. Red Hat does not support the use of this field. Misuse of this field could lead to unexpected behavior or conflict with other configuration options. Seek guidance from the Red Hat support before using this field. Use of this property blocks cluster upgrades, it must be removed before upgrading your cluster.",
           "default": {},
           "$ref": "#/definitions/io.k8s.apimachinery.pkg.runtime.RawExtension"
+        },
+        "vsphereStorageDriver": {
+          "description": "VSphereStorageDriver indicates the storage driver to use on VSphere clusters. Once this field is set to CSIWithMigrationDriver, it can not be changed. If this is empty, the platform will choose a good default, which may change over time without notice. DEPRECATED: This field will be removed in a future release.",
+          "type": "string",
+          "default": ""
         }
       }
     },
--- a/operator/v1/stable.storage.testsuite.yaml
+++ b/operator/v1/stable.storage.testsuite.yaml
@@ -14,3 +14,91 @@ tests:
       spec:
         logLevel: Normal
         operatorLogLevel: Normal
+  onUpdate:
+  - name: Should allow enabling CSI migration for vSphere
+    initial: |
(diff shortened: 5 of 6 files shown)


elmiko:
i've created a library-go change to help with the CCM reversion process, it will need this to work properly.

https://github.com/openshift/library-go/pull/1490

config/v1/types_feature.go
```diff
@@ -40,6 +40,9 @@ var (
 
 	// TopologyManager enables ToplogyManager support. Upgrades are enabled with this feature.
 	LatencySensitive FeatureSet = "LatencySensitive"
+
+	// InTreeVSphereVolumes disables CSI Migration for in-tree volumes on vSphere.
+	InTreeVSphereVolumes FeatureSet = "InTreeVSphereVolumes"
```
JoelSpeed, line 45: If I understand correctly, in 4.12 this would be harmless and do nothing, so a `""`->`InTreeVSphereVolumes`->`""` transition is ok, in 4.13, we don't have the CCM tested for uninstalling (eg CCM -> KCM) migration, as this was something we never intended to support, so should we, in 4.13, prevent the transition `""`->`InTreeVSphereVolumes`, as this would be an unsupported scenario?

CC @deads2k IIUC this should be ok as a CEL validation, can you think of some reason introducing a transition prevention would be a breaking change here?

dobsonj:
Updated to implement the latest proposal
/hold for testing

operator/v1/types_storage.go
```diff
@@ -26,9 +26,24 @@ type Storage struct {
...
 // StorageSpec is the specification of the desired behavior of the cluster storage operator.
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.csiMigration) || has(self.csiMigration)", message="CSIMigration is required once set"
 type StorageSpec struct {
 	OperatorSpec `json:",inline"`
+
+	// CSIMigration enables CSI migration for drivers where it is optional.
+	// This field is immutable once it is set and can not be undone.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="CSIMigration is immutable"
+	// +optional
+	CSIMigration CSIMigrationToggle `json:"csiMigration,omitempty"`
```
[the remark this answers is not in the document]
deads2k: > And second thing, it should be a pointer.

We prefer not-pointer for configuration APIs.  Missing or "" is clear as "no opinion"

operator/v1/types_storage.go
```diff
@@ -26,9 +26,24 @@ type Storage struct {
...
+const (
+	CSIMigrationEnabled CSIMigrationToggle = "Enabled"
+)
+
 // StorageSpec is the specification of the desired behavior of the cluster storage operator.
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.csiMigration) || has(self.csiMigration)", message="CSIMigration is required once set"
 type StorageSpec struct {
 	OperatorSpec `json:",inline"`
+
+	// CSIMigration enables CSI migration for drivers where it is optional.
```
deads2k, line 42: Add the text for "If empty, the platform will choose a good default, which may change over time without notice."

operator/v1/types_storage.go
deads2k, line 46: indicate that this is deprecated in the docs and will be removed in a future release.

operator/v1/types_storage.go
```diff
@@ -26,9 +26,24 @@ type Storage struct {
...
+	// This field is immutable once it is set and can not be undone.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="CSIMigration is immutable"
+	// +optional
```
deads2k, line 45: rather than allow empty, let's allow "".  That is in keeping with configuration APIs and allows "no opinion" to be expressed.

operator/v1/types_storage.go
```diff
@@ -26,9 +26,24 @@ type Storage struct {
...
+)
+
 // StorageSpec is the specification of the desired behavior of the cluster storage operator.
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.csiMigration) || has(self.csiMigration)", message="CSIMigration is required once set"
 type StorageSpec struct {
 	OperatorSpec `json:",inline"`
+
+	// CSIMigration enables CSI migration for drivers where it is optional.
+	// This field is immutable once it is set and can not be undone.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="CSIMigration is immutable"
```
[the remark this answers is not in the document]
dobsonj: These rules came from the example [here](https://kubernetes.io/blog/2022/09/29/enforce-immutability-using-cel/#immutablility-after-first-modification). It should allow setting a value exactly once, but only if the field is optional in the first place. After that, the value can't change.

operator/v1/types_storage.go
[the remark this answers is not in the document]
dobsonj: Done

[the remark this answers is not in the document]
dobsonj: I'd like to stick with "" instead of a pointer, since that is consistent with other API's including [CSIDriverType](https://github.com/openshift/api/blob/fefb3487546079495fb80ca0f1155ecd7417b9d8/operator/v1/types_csi_cluster_driver.go#L110-L131). I recall some previous discussions around `""` vs `nil` during API review and we settled on `""`.
I was trying to avoid introducing a specific "Disabled" value, because there is not a way to honor that value in future releases. If you set it to Disabled, and then upgrade to 4.14... well it's still enabled then. IMO the only values we can honor are "" (no opinion, behavior may change) and "Enabled" (will be locked to this in the future).

operator/v1/types_storage.go
dobsonj: Done

operator/v1/types_storage.go
dobsonj: Done

operator/v1/types_storage.go
dobsonj: Done

operator/v1/types_storage.go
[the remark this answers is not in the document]
dobsonj: From the [docs](https://kubernetes.io/blog/2022/09/23/crd-validation-rules-beta/#crd-transition-rules), regarding `self == oldSelf`: "For an optional field, only allow transitioning from unset to set, or from set to unset."
And the rule for StorageSpec prevents transitioning from set to unset. I think this will work, but needs to be tested.

operator/v1/types_storage.go
```diff
@@ -26,9 +26,27 @@ type Storage struct {
 	Status StorageStatus `json:"status"`
 }
 
+// CSIMigrationToggle indicates whether CSI migration should be enabled for drivers where it is optional.
+// +kubebuilder:validation:Enum="";Enabled
+type CSIMigrationToggle string
+
+const (
+	CSIMigrationEnabled CSIMigrationToggle = "Enabled"
```
deads2k, line 34: Enum values like `KeepDeprecatedLegacyVSphereStorageDriver` and `MigrateToModernVSphere` are descriptive when viewing.  Selecting `KeepDeprecatedLegacyVSphereStorageDriver` should be possible so you can mark those closers as upgradeable=false since they expressed a specific opinion.

dobsonj: I thought we had decided not to mark clusters upgradeable=false? At least that was my understanding, also see @jsafrane 's comment in [STOR-1269](https://issues.redhat.com/browse/STOR-1269).

If I set `KeepDeprecatedLegacyVSphereStorageDriver` on my cluster, and the cluster is not upgradeable until I change that value, and also we don't allow changing the value of `CSIMigrationVSphere` once it is set... how do I get to a state where the cluster is upgradeable? We could change the validation to explicitly allow transitioning from disabled to enabled, but I thought we were trying to avoid introducing an extra step before upgrading.

dobsonj: For the enum value, `MigrateToModernVSphere` seems a little redundant IMO now that the field itself is named `CSIMigrationVSphere`.

operator/v1/0000_50_cluster_storage_operator_01_crd.yaml
```diff
@@ -69,6 +78,9 @@ spec:
                   type: object
                   nullable: true
                   x-kubernetes-preserve-unknown-fields: true
+              x-kubernetes-validations:
```
JoelSpeed, line 81: We should add integration tests for validations into https://github.com/openshift/api/blob/master/operator/v1/stable.storage.testsuite.yaml

The documentation for how the tests work is https://github.com/openshift/api/tree/master/tests#readme

For these, I would expect to see some "onupdate" style tests, so, something like:

```
onUpdate:
- name: Should allow enabling the vSphere migration for vSphere
  initial: |
    kind: Storage
    spec: {} # No spec is required for a Storage
   updated: |
     kind: Storage
     spec:
       csiMigrationVSphere: Enabled
   expected: |
    kind: Storage
    spec:
       csiMigrationVSphere: Enabled
       logLevel: Normal
       operatorLogLevel: Normal
- name: Should not allow disabling the vSphere migration for vSphere
  initial: |
    kind: Storage
    spec:
      csiMigrationVSphere: Enabled
   updated: |
     kind: Storage
     spec:
       csiMigrationVSphere:""
   expectedError: "csiMigrationVSphere may not be unset once Enabled"
```

You probably want another case for trying to unset the field instead of changing to empty string as well, I'll leave that and any others to you

operator/v1/types_storage.go
JoelSpeed: I agree that there is value in having the `KeepDeprecated...` option, you have three different scenarios right:
- People who don't care, who will leave the value unset
- People who care about the bug and won't want to upgrade until the issue is resolved, they want to set `KeepDeprecated...`
- People who care about having external CCM, they want to set "MigrateTo...`
For the second group here, having this explicitly means that they do need to make a more conscious decision to upgrade, it would almost act as a reminder to them that they knew there was some reason they weren't migrating, perhaps they then see the upgrade blocker, and review the issues before they decide whether to upgrade or not.

If they have no upgrade blocker, they have no reminder for this 

As for the naming, you could always rename the field to something like `vsphereStorageDriver: LegacyDeprecatedInTreeDriver | CSIWithMigrationDriver`

operator/v1/0000_50_cluster_storage_operator_01_crd.yaml
dobsonj: Done, thanks for the references.

deads2k:
test for 
```
- name: Should not allow disabling the vSphere migration for vSphere
  initial: |
    kind: Storage
    spec:
      csiMigrationVSphere: Enabled
   updated: |
     kind: Storage
     spec:
       csiMigrationVSphere:""
   expectedError: "csiMigrationVSphere may not be unset once Enabled"
```

appears missing.  there is a test for missing, but not for empty (near as I see)

dobsonj:
> there is a test for missing, but not for empty (near as I see)

Added in latest push here:
https://github.com/dobsonj/api/blob/426d59473937ba3aaec1e6ef768051090d1c78ac/operator/v1/stable.storage.testsuite.yaml#L82-L93

PR #1432, opened 2023-03-28
[release-4.13] Enable ValidatingAdmissionPolicy in tech preview.

benluddy:


deads2k:
I would only do this if an external product or customer says they want to test it.  

/hold

PR #1442, opened 2023-04-14
OSASINFRA-3061 - OpenStackFailureDomain for CPMS

EmilienM:
This implements the failure domain information for the
OpenStack platform.


EmilienM:
/cc mdbooth pierreprinetti
for initial review.

machine/v1/types_controlplanemachineset.go
```diff
@@ -284,6 +289,17 @@ type GCPFailureDomain struct {
...
 
+// OpenStackFailureDomain configures failure domain information for the OpenStack platform.
+type OpenStackFailureDomain struct {
+	// ComputeAvailabilityZone is the nova availability zone in which the OpenStack machine provider will create the VM.
+	// +optional
+	ComputeAvailabilityZone string `json:"computeAvailabilityZone,omitempty"`
+
+	// StorageAvailabilityZone is the zone in which the OpenStack machine provider will create the root volume attached to the VM.
+	// +optional
+	StorageAvailabilityZone string `json:"storageAvailabilityZone,omitempty"`
```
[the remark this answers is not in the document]
mdbooth: @stephenfin 👆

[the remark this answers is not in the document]
stephenfin: > I guess the question is, @mdbooth: does it make any sense (or is it even possible) to attach volumes from different Cinder zones, in a storage-flexibility context?

That depends on the value of `[cinder] cross_az_attach` in `nova.conf`. From the [docs](https://docs.openstack.org/nova/latest/configuration/config.html#cinder.cross_az_attach):

> <dl>
> <dt><code>cross_az_attach</code></dt>
> <dd>
> Allow attach between instance and volume in different availability zones.
> 
> If False, volumes attached to an instance must be in the same availability zone in Cinder as the instance availability zone in Nova.
> 
> This also means care should be taken when booting an instance from a volume where source is not “volume” because Nova will attempt to create a volume using the same availability zone as what is assigned to the instance.
> 
> If that AZ is not in Cinder (or `allow_availability_zone_fallback=False` in cinder.conf), the volume create request will fail and the instance will fail the build request.
> 
> By default there is no availability zone restriction on volume attach.
> </dd>
> </dl>

Since that defaults to `False`, it is possible. I don't know if it would make _sense_ though.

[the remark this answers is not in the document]
stephenfin: Also, reproducing from Slack, which covered a different area:

>Speaking with my Nova hat on, we'd usually expect to see `set(compute_azs) == set(volume_azs)` (with the idea being that an AZ would correspond to e.g. a rack). iirc, this is encoded into nova and doing a boot from volume request with a compute AZ of e.g. foo will result in nova requesting a cinder volume in the foo volume AZ.
>
> Couple of caveats, however. In environments with Ceph but without DCN, it's apparently common enough to have multiple nova AZs but a single cinder AZ. Cinder also provides a mechanism to ignore invalid AZ requests so you'd need to make sure that what you request is what you get.

The encoded assumption I referred to is documented [here](https://docs.openstack.org/nova/latest/admin/availability-zones.html#implications-for-moving-servers):

> With respect to availability zones, a server is restricted to a zone if:
>
> 1. The server was created in a specific zone with the `POST /servers` request containing the `availability_zone` parameter.
> 1. If the server create request did not contain the `availability_zone` parameter but the API service is configured for [`default_schedule_zone`](https://docs.openstack.org/nova/latest/configuration/config.html#DEFAULT.default_schedule_zone) then by default the server will be scheduled to that zone.
> 1. The shelved offloaded server was unshelved by specifying the `availability_zone` with the `POST /servers/{server_id}/action` request using microversion 2.77 or greater.
> 1. **[`cinder.cross_az_attach`](https://docs.openstack.org/nova/latest/configuration/config.html#cinder.cross_az_attach) is False, [`default_schedule_zone`](https://docs.openstack.org/nova/latest/configuration/config.html#DEFAULT.default_schedule_zone) is None, the server is created without an explicit zone but with pre-existing volume block device mappings. In that case the server will be created in the same zone as the volume(s) if the volume zone is not the same as [`default_availability_zone`](https://docs.openstack.org/nova/latest/configuration/config.html#DEFAULT.default_availability_zone). See [Resource affinity](https://docs.openstack.org/nova/latest/admin/availability-zones.html#resource-affinity) for details.**

(emphasis mine)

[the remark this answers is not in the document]
pierreprinetti: In our out-of-band discussion on the matter, we decided to use a general `storageAvailabilityZone` that initially applies to the root volume, and that may just be the default for further attached volumes.

Essentially: even though additional volumes, in future feature developments, might need to be placed in separate availability zones, we deem `storageAvailabilityZone` to have the highest potential of being compatible with future APIs.

mdbooth:
For reference: we intend to add a volume type to this struct. However, as support that isn't implemented yet and it will be a backwards compatible change, there is no reason to delay merging this.

EmilienM:
/assign deads2k 
for review, thanks a lot.

machine/v1/types_controlplanemachineset.go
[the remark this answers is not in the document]
ingvagabund: https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#use-json-field-names-in-godoc will be tricky to conform to:
```
Ensure that the godoc for a field name matches the JSON name, not the Go name, in Go definitions for API objects
```
@JoelSpeed should we advocate for this convention only for the newly created APIs? Requiring it for new fields for already existing APIs will make a mess without renaming already existing fields.

https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#my-proposed-design-looks-like-an-existing-api-we-have-why-am-i-being-told-that-it-must-be-changed gives yes as an answer. Nevertheless, we should fix the naming everywhere then. Starting 4.14?

machine/v1/types_controlplanemachineset.go
```diff
@@ -284,6 +289,17 @@ type GCPFailureDomain struct {
 	Zone string `json:"zone"`
 }
 
+// OpenStackFailureDomain configures failure domain information for the OpenStack platform.
+type OpenStackFailureDomain struct {
+	// ComputeAvailabilityZone is the nova availability zone in which the OpenStack machine provider will create the VM.
+	// +optional
```
ingvagabund, line 295: Can you also mention what happens when the field is not specified?

What if `ComputeAvailabilityZone` is empty but `StorageAvailabilityZone` is not?

machine/v1/types_controlplanemachineset.go
```diff
@@ -284,6 +289,17 @@ type GCPFailureDomain struct {
 	Zone string `json:"zone"`
 }
 
+// OpenStackFailureDomain configures failure domain information for the OpenStack platform.
+type OpenStackFailureDomain struct {
```
ingvagabund, line 293: https://www.fir3net.com/Cloud/Openstack/openstack-partitioning-regions-availability-zones-and-host-aggregates.html mentions there are three OpenStack partitions: Regions, Availability Zones & Host Aggregates. Any chance host aggregates might be added later? Are there other failure domain information that might change the shape of `OpenStackFailureDomain`? `AWSFailureDomain` includes subnet part which moves zones under a new `AWSFailureDomainPlacement` type. Something that might be better to do right now to avoid growing a list of distinct information (e.g. placement, networking) on the same level.

machine/v1/types_controlplanemachineset.go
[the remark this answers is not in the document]
JoelSpeed: > @JoelSpeed should we advocate for this convention only for the newly created APIs? Requiring it for new fields for already existing APIs will make a mess without renaming already existing fields.

I would argue this is a newly created API and that this convention should be followed, if it isn't already being followed by machine v1 then that's on me and we should definitely fix it!

I think we should at some point at some programatic linting for this rule and start enforcing it. Since it's a minor documentation change I don't see any harm in updating old APIs to follow that rule

machine/v1/types_controlplanemachineset.go
ingvagabund, line 293: What if both `ComputeAvailabilityZone` and `StorageAvailabilityZone` are empty? What will be the default behavior for this case?

EmilienM: > Any chance host aggregates might be added later? Are there other failure domain information that might change the shape of `OpenStackFailureDomain`?

Yes, and they'll be added later but they'll always be optional and default to empty.

> `AWSFailureDomain` includes subnet part which moves zones under a new `AWSFailureDomainPlacement` type. Something that might be better to do right now to avoid growing a list of distinct information (e.g. placement, networking) on the same level.

I guess we could, we could have:

```golang
type OpenStackFailureDomain struct {
	// Placement configures the placement information for this instance.
	// +optional
	Placement AWSFailureDomainPlacement `json:"placement,omitempty"`

	// Storage is a reference to the storage to use for this instance.
	// +optional
	Storage *OpenStackFailureDomainStorage `json:"storage,omitempty"`

	// Network is a reference to the network to use for this instance.
	// +optional
	Network *OpenStackFailureDomainNetwork `json:"network,omitempty"`

}

type OpenStackFailureDomainPlacement struct {
	// AvailabilityZone is the availability zone of the instance.
	AvailabilityZone string `json:"availabilityZone"`
}

type OpenStackFailureDomainStorage struct {
	// AvailabilityZone is the availability zone of the volume.
	AvailabilityZone string `json:"availabilityZone"`

        // Later we could have other fields for Type, and additional volumes.
}

// Later we could have another type for Networking
type OpenStackFailureDomainNetwork struct {
	// Subnet(s) where the instance is attached to
}
```

WDYT?

EmilienM: * one or two fields set: valid
* both empty: valid, it means no failure domain. However in CCPMSO we check that if no failure domain was matched for a machine, the other masters can't match a failure domain. It's either we have 1+ failure domain or 0.

machine/v1/types_controlplanemachineset.go
EmilienM: I think I answered that above.

machine/v1/types_controlplanemachineset.go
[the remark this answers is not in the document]
EmilienM: I'll look into it in the next patchset. Thank you.

machine/v1/types_controlplanemachineset.go
stephenfin: > [fir3net.com/Cloud/Openstack/openstack-partitioning-regions-availability-zones-and-host-aggregates.html](https://www.fir3net.com/Cloud/Openstack/openstack-partitioning-regions-availability-zones-and-host-aggregates.html) mentions there are three OpenStack partitions: Regions, Availability Zones & Host Aggregates. Any chance host aggregates might be added later?

From a nova (compute) perspective, availability zones are just a special type of host aggregate. A host can belong to multiple host aggregates and this information is only exposed to admins, not end-users, so I couldn't imagine anyone would ever use them for failure domain purposes. I'm not sure it ever makes sense to consider them.

Regions are purely a keystone (identity) thing. They allow you to share a single keystone instance across multiple OpenStack deployments. Because it's not possible to e.g. migrate instances between separate deployments, I don't think it would be reasonable to think of these as a failure domain either.

> Are there other failure domain information that might change the shape of `OpenStackFailureDomain`?

Based on the above, I would think not. We already cover compute, storage (both cinder and manila), and networking. This should be more than sufficient and is how any HA configuration I've seen has been handled.

> Besides already mentioned [volume type](https://github.com/openshift/api/pull/1442#issuecomment-1525584951). `AWSFailureDomain` includes subnet part which moves zones under a new `AWSFailureDomainPlacement` type. Something that might be better to do right now to avoid growing a list of distinct information (e.g. placement, networking) on the same level.

pierreprinetti: @ingvagabund 

> What if both ComputeAvailabilityZone and StorageAvailabilityZone are empty?

The fields [are evaluated separately](https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/195/files#diff-ae5bff491dfb3714be91613a81c7756db1ced45ca9698505b585b4a68ac346dbR38-R50).

If both fields are empty, then the Machine will not contain placement information, leaving Nova to act by its defaults.

pierreprinetti: Note that as @EmilienM points out, CCPMSO will refuse to generate empty failureDomains. Those are however accepted if added by the user.

machine/v1/types_controlplanemachineset.go
```diff
@@ -284,6 +289,17 @@ type GCPFailureDomain struct {
...
+	// ComputeAvailabilityZone is the nova availability zone in which the OpenStack machine provider will create the VM.
+	// +optional
+	ComputeAvailabilityZone string `json:"computeAvailabilityZone,omitempty"`
+
+	// StorageAvailabilityZone is the cinder availability zone in which the OpenStack machine provider will create the root volume attached to the VM.
```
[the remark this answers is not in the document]
EmilienM: Let me know if you prefer me to write `block storage` instead.

machine/v1/types_controlplanemachineset.go
mdbooth: In addition to what @stephenfin wrote, we have explicitly pushed Networking out of scope of our initial failure domains. We have several problems here:
* Neutron doesn't have a comprehensive AZ concept in the way Nova and Cinder do
* Unlike in public cloud, it is common in on-prem clusters to have multiple networks.
* There is no requirement for a user to segregate all possible network interfaces by failure domain
* The definition of a network interface is substantially more complex than simply what subnet it is connected to (e.g. hw offload, trunking, etc), which we don't want to add to the failure domain

Also, almost nobody is going to use it in OpenStack. The use case for networking like this is strict spine-leaf deployments and the one customer who was asking for this decided it wasn't important as they originally thought.

In short, it was hard and we punted it so we could land the rest.

JoelSpeed: In AWS, we have the additional step in the API to add structure because this matches the `providerSpec`. The idea is that the failure domain itself represents a subset of the `providerSpec`, so to a user, it's obvious which fields from the failure domain map where in the provider spec. So I think we should keep the same pattern for this as well

machine/v1/types_controlplanemachineset.go
```diff
@@ -284,6 +289,17 @@ type GCPFailureDomain struct {
...
+	// +optional
```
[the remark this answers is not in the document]
EmilienM: Thanks for your suggestion, I added more documentation.

machine/v1/types_controlplanemachineset.go
EmilienM: I'll add validation to only accept a failure domain that has at least one of the fields to be set.

EmilienM: that makes sense Joel, I'll implement that.

EmilienM: @JoelSpeed @ingvagabund I followed your advice, what do you think of this proposal?

machine/v1/types_controlplanemachineset.go
```diff
@@ -284,6 +290,24 @@ type GCPFailureDomain struct {
...
+// +kubebuilder:validation:MinProperties:=1
+type OpenStackFailureDomain struct {
+	// availabilityZone is the nova availability zone in which the OpenStack machine provider will create the VM.
+	// If not specified, the VM will be created in the default availability zone specified in the nova configuration.
+	// +optional
+	AvailabilityZone string `json:"availabilityZone,omitempty"`
+
+	// rootVolume contains settings that will be used by the OpenStack machine provider to create the root volume attached to the VM.
+	// If not specified, no root volume will be created.
+	// If specified without a Zone, the root volume will be created in the availability specified by the volume type. If the volume type does not specify an
```
ingvagabund, line 303: `availability specified` -> `availability zone specified`

ingvagabund: > If the volume type does not specify an availability zone

Can you more elaborate on what it means to specify AZ in the volume type? E.g. "volumetype:availabilityzone" form?

EmilienM: I added a few words and also a link to the doc. I hope this helps.

machine/v1/types_controlplanemachineset.go
```diff
@@ -284,6 +290,24 @@ type GCPFailureDomain struct {
...
+	// availability zone, the root volume will be created in the default availability zone specified in the cinder configuration.
+	// If the OpenStack cluster is deployed with the cross_az_attach configuration option set to false, the root volume will have to be in the same
+	// availability zone as the VM (defined by availabilityZone).
+	// +optional
+	RootVolume *machinev1alpha1.RootVolume `json:"rootVolume,omitempty"`
```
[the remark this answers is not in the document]
EmilienM: Oh I didn't know it wasn't good practice. I'll implement what you proposed then.

machine/v1/types_controlplanemachineset.go
```diff
@@ -284,6 +289,40 @@ type GCPFailureDomain struct {
...
+	RootVolume *RootVolume `json:"rootVolume,omitempty"`
+}
+
+// RootVolume represents the volume metadata to boot from.
+// The original RootVolume struct is defined in the v1alpha1 but it's not best practice to use it directly here so we define a new one
+// that should stay in sync with the original one.
+// +kubebuilder:validation:MinProperties:=1
+type RootVolume struct {
+	// availabilityZone specifies the Cinder availability where the root volume will be created.
+	// If not specifified, the root volume will be created in the availability zone specified by the volume type. If the volume type
```
EmilienM, line 317: To reviewers: I put this comments here before in the very near future @pierreprinetti will add the `Type` field too, which will make the Zone field not required.

EmilienM:
@ingvagabund if you can have another look please, I addressed your comments I think.

@pierreprinetti @mdbooth you as well, please.

machine/v1/types_controlplanemachineset.go
ingvagabund: What about `If not specifified, the root volume will be created in the availability zone specified by the volume type.` -> `If not specifified, the root volume will be created in the availability zone specified by the volume type in the cinder configuration.`? To stress the volume type configuration is currently available only through cinder? To avoid anyone reading the description from searching for the volume type field.

machine/v1/types_controlplanemachineset.go
```diff
@@ -284,6 +289,40 @@ type GCPFailureDomain struct {
...
+// +kubebuilder:validation:MinProperties:=1
+type OpenStackFailureDomain struct {
+	// availabilityZone is the nova availability zone in which the OpenStack machine provider will create the VM.
+	// If not specified, the VM will be created in the default availability zone specified in the nova configuration.
+	// +optional
+	AvailabilityZone string `json:"availabilityZone,omitempty"`
+
+	// rootVolume contains settings that will be used by the OpenStack machine provider to create the root volume attached to the VM.
+	// If not specified, no root volume will be created.
+	// If specified without a Zone, the root volume will be created in the availability zone specified by the volume type. If the volume type
```
ingvagabund, line 302: I wonder whether mentioning this comment twice (already as part of `RootVolume.Zone`) has any additional value for a reader. If the goal here is to inform a reader about a possibility to set the volume availability zone alongside the VM AZ, worth putting a short sentence about it into the `AvailabilityZone` description and refer to the `RootVolume` type.

machine/v1/types_controlplanemachineset.go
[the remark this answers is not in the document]
mdbooth: @EmilienM Lets review what's left before effectively declaring it stable. It's probably fine as we already deprecated the nonsense, but lets take the time to check deliberately.

machine/v1/types_controlplanemachineset.go
EmilienM: that was a mistake.

machine/v1/types_controlplanemachineset.go
EmilienM: ok, I'll change the text.

machine/v1/types_controlplanemachineset.go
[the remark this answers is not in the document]
EmilienM: @mdbooth I agree. Do I understand that you're fine with the proposed interface?

machine/v1/types_controlplanemachineset.go
```diff
@@ -284,6 +289,40 @@ type GCPFailureDomain struct {
...
+// +kubebuilder:validation:MinProperties:=1
+type RootVolume struct {
+	// availabilityZone specifies the Cinder availability where the root volume will be created.
+	// If not specifified, the root volume will be created in the availability zone specified by the volume type. If the volume type
+	// (configured in the OpenStack cluster) does not specify an availability zone, the root volume will be created in the default availability
+	// zone specified in the cinder configuration. See https://docs.openstack.org/cinder/latest/admin/availability-zone-type.html for more details.
+	// If the OpenStack cluster is deployed with the cross_az_attach configuration option set to false, the root volume will have to be in the same
+	// availability zone as the VM (defined by OpenStackFailureDomain.AvailabilityZone).
+	// +optional
+	Zone string `json:"availabilityZone,omitempty"`
```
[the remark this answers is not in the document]
mdbooth: It's also called `availabilityZone` upstream, so I think this *must* be called `availabilityZone`. I'd recommending updating the Go field name to match, though.

mdbooth (commented):
/lgtm

I'd prefer we updated `Zone` to `AvailabilityZone` because it's weird they don't match, but as that's not an external API change I assume it's not incredibly important. The upstream field has been updated to match the json field name.

machine/v1/types_controlplanemachineset.go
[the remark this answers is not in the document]
EmilienM: > I'd recommend updating the Go field name to match, though.

yeah it's not elegant, but this is how it's defined in `v1alpha1`...

EmilienM:
> /lgtm
> 
> I'd prefer we updated `Zone` to `AvailabilityZone` because it's weird they don't match, but as that's not an external API change I assume it's not incredibly important. The upstream field has been updated to match the json field name.

This was addressed.

machine/v1/types_controlplanemachineset.go
```diff
@@ -284,6 +289,35 @@ type GCPFailureDomain struct {
...
+	// rootVolume contains settings that will be used by the OpenStack machine provider to create the root volume attached to the VM.
+	// If not specified, no root volume will be created.
+	// +optional
+	RootVolume *RootVolume `json:"rootVolume,omitempty"`
+}
+
+// RootVolume represents the volume metadata to boot from.
+// The original RootVolume struct is defined in the v1alpha1 but it's not best practice to use it directly here so we define a new one
+// that should stay in sync with the original one.
+// +kubebuilder:validation:MinProperties:=1
```
ingvagabund, line 309: `kubebuilder:validation:MinProperties:=1` will force the `AvailabilityZone` field to be always set when specified. Which is equivalent to `+required`. You probably want to remove the validation line.

machine/v1/types_controlplanemachineset.go
```diff
@@ -284,6 +289,35 @@ type GCPFailureDomain struct {
 	Zone string `json:"zone"`
 }
 
+// OpenStackFailureDomain configures failure domain information for the OpenStack platform.
+// +kubebuilder:validation:MinProperties:=1
```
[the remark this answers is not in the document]
EmilienM: yes, when a FailureDomain is found by CPMS, it can't be empty. Otherwise there is no FailureDomain that was found by CCPMSO.

machine/v1/types_controlplanemachineset.go
EmilienM: but since `RootVolume` is optional, isn't it fine that we require at least one field in `RootVolume`?

EmilienM: I just realized that there is a testsuite for CPMS (e.g. machine/v1/stable.controlplanemachineset.gcp.testsuite.yaml), I'll add one for OpenStack.

Sorry for that.

machine/v1/stable.controlplanemachineset.openstack.testsuite.yaml
```diff
@@ -0,0 +1,277 @@
+apiVersion: apiextensions.k8s.io/v1 # Hack because controller-gen complains if we don't have this
```
EmilienM, line 1: @ingvagabund I added all the test cases that I thought would be useful. Please let me know what you think.

@JoelSpeed for a quick review on the tests, if their results make sense to you from a CPMS point of view.

machine/v1/types_controlplanemachineset.go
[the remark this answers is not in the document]
EmilienM: BTW this was added to the testsuite to make sure it works that way.

machine/v1/types_controlplanemachineset.go
EmilienM: Here I also added a case to the testsuite to make sure it works that way, when a rootVolume is specified, it can't be empty.

mdbooth (commented):
Test coverage lgtm

/lgtm

machine/v1/stable.controlplanemachineset.openstack.testsuite.yaml
```diff
@@ -0,0 +1,277 @@
...
+                machine.openshift.io/cluster-api-machine-role: master
+                machine.openshift.io/cluster-api-machine-type: master
+                machine.openshift.io/cluster-api-cluster: cluster
+            spec:
+              providerSpec: {}
+            failureDomains:
+              platform: OpenStack
+              openstack:
+              - availabilityZone: foo
+                rootVolume: {}
```
JoelSpeed, line 251: Do you have omitempty on this? If you don't, this could cause issues for a Go type trying to serialise the CPMS

machine/v1/types_controlplanemachineset.go
JoelSpeed: Looks better to me now yes, looks like it's closer to matching the providerSpec structure. One nit though, we don't use pointers in APIs in general (bar a few exceptions, eg Discriminated Unions), so, can we change the rootVolume to not be a pointer, will make coding against it easier.

Unless I've missed some reason that this _must_ be a pointer, but I can't think of one at the moment

machine/v1/types_controlplanemachineset.go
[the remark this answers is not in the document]
JoelSpeed: You can match the "API" from a user perspective without matching it in a Go perspective, I would update the Go field name to match the (correctly cased version of the) json tag name

machine/v1/stable.controlplanemachineset.openstack.testsuite.yaml
mdbooth: rootVolume has `omitempty`, as do `availabilityZone` and the containing `openstack` field.

machine/v1/types_controlplanemachineset.go
ingvagabund: > // If not specified, no root volume will be created.
> // +optional
> RootVolume *RootVolume `json:"rootVolume,omitempty"`

Turning the field into non-pointer would change the semantics. I.g. `"rootVolume": {"availabilityZone": ""}` would mean "create a volume without the az specified (random)" as opposed to "do not create any root volume".

JoelSpeed: I had thought we had a requirement in the failure domain that the AZ fields were not permitted to be empty, so I would expect if it were a non-pointer we would say
```
if failureDomain.RootVolume.AvailabilityZone != "" {
  // User has an opinion about the Az here, so lets make sure we set it on the providerSpec
  ...
}
```

JoelSpeed: Perhaps a clarifying comment would be good for these fields, and perhaps a min-length to make sure when it is provided, it isn't empty?
```
When omitted, this means no opinion and the value will be taken from any existing value present in the machine template providerSpec.
```

ingvagabund: > I had thought we had a requirement in the failure domain that the AZ fields were not permitted to be empty

I see. Can it be always guaranteed? I mean `RootVolume` had two fields originally (zone and volume type). Now there's only a single one. In case another field gets introduced, will the `AvailabilityZone` field be always required to be non-empty? Or, will the requirement be to have at least one field in the `RootVolume` struct non-empty?

ingvagabund: What I am asking is: Is the `RootVolume` struct always going to have at least one AZ field set? Either the current `AvailabilityZone` field or maybe-to-be-added `VolumeZone` field? And if not (i.e. both fields are empty), will it mean "no root volume will be created"? Currently, the `RootVolume` struct has `+kubebuilder:validation:MinProperties:=1` which means `If not specified, no root volume will be created.` is not a valid case.

ingvagabund: > I had thought we had a requirement in the failure domain that the AZ fields were not permitted to be empty,

Is this requirement documented? If this is a universal requirement, the `AvailabilityZone` field under the `OpenStackFailureDomain` struct should be required as well.

ingvagabund:
Questions to resolve before approving the API
- `RootVolume` is a pointer to `RootVolume` vs. non-pointer and requiring AZs to be always non-empty (https://github.com/openshift/api/pull/1442#discussion_r1200541903)
- AZ naming: https://github.com/openshift/api/pull/1442#discussion_r1200544730

machine/v1/types_controlplanemachineset.go
JoelSpeed: So I think this boils down to what is valid when we look at the structure
```
openstack:
- availabiltyZone: foo
- rootVolume:
    availabailityZone: foo
- availabilityZone: foo
  rootVolume:
    availabilityZone: foo
- rootVolume:
    someNewField: foo
```
IMO these are all valid
Whereas the following would not be valid:
```
openstack:
- {}
- rootVolume: {}
- availabilityZone: ""
- rootVolume:
    availabilityZone: ""
```

So I think this boils down to a few requirements:
* the failuredomain for openstack must have at least one, non-empty field
* the rootVolume, must have at least one non-empty field
* If specified, an availabilityZone must have length at least 1

Which makes me think that the fields for AZs should be optional, but using min-properties we can make the structure make sense

Perhaps some integration tests for these would help? 

Oh and, because this is all from the view of an unstructured client, whether this field is a pointer or not doesn't actually matter, what matters is the json serialization, so, do we have the right omitempty tags

mdbooth:
> Questions to resolve before approving the API
> 
> * `RootVolume` is a pointer to `RootVolume` vs. non-pointer and requiring AZs to be always non-empty ([OSASINFRA-3061 - OpenStackFailureDomain for CPMS #1442 (comment)](https://github.com/openshift/api/pull/1442#discussion_r1200541903))

@ingvagabund could you try linking this again, or quote the whole comment here? GH doesn't like this link and I can't see which comment you're referring to.

Trying to guess from context, though: we are requiring failure domains to be non-empty but we don't, and must not, require them to have a RootVolume. We could convert the field from a pointer to a non-pointer, but we must still allow the value to be empty.

> * AZ naming: [OSASINFRA-3061 - OpenStackFailureDomain for CPMS #1442 (comment)](https://github.com/openshift/api/pull/1442#discussion_r1200544730)

This is a mechanical change. @EmilienM: we should change the name of the go field from `Zone` to `AvailabilityZone`, and nothing else. The json tag name remains as it currently is. It doesn't matter that this doesn't matter that [it doesn't match what's in v1alpha1.RootVolume](https://github.com/openshift/api/blob/f6e7dffe5f27fe10a21940fcf90a48d56e5e0ab9/machine/v1alpha1/types_openstack.go#L351-L368), because that's not the public facing API that we care about.

We should probably separately update the v1alpha1 type too.

EmilienM:
@ingvagabund 

> Questions to resolve before approving the API

I think I've addressed it.

machine/v1/types_controlplanemachineset.go
ingvagabund:
```yaml
- rootVolume:
    availabilityZone: ""
```
```
// +kubebuilder:validation:MinProperties:=1
type RootVolume struct {
...
	// +optional
```
Common sense tells me `kubebuilder:validation:MinProperties:=1` together with `AvailabilityZone` as optional field does not work well together. Though, there might be a bigger picture of the validation mechanism I don't see. 
From https://book.kubebuilder.io/reference/markers/crd-validation.html:
```
// +kubebuilder:validation:MinProperties int
restricts the number of keys in an object
```
Meaning one needs to have at least one field present or set? When preset, the value can be nil. So `{RootVolume: {AvailabilityZone: ""}}` should pass.

Anyway, I agree the unit tests need to validate the mentioned cases.

EmilienM: @JoelSpeed  : 

> So I think this boils down to what is valid when we look at the structure
(...)

This is all covered by the tests; so we're good here.


> Whereas the following would not be valid:

Let me go line by line:

---
```
openstack:
- {}
```

A test was missing, I added one.

---
```
openstack:
- rootVolume: {}
```

This one is tested already:
```
  - name: Should reject an OpenStack failure domain with an empty rootVolume provided
    initial: |
      apiVersion: machine.openshift.io/v1
      kind: ControlPlaneMachineSet
      spec:
            [...]
            spec:
              providerSpec: {}
            failureDomains:
              platform: OpenStack
              openstack:
              - availabilityZone: foo
                rootVolume: {}
    expectedError: "spec.template.machines_v1beta1_machine_openshift_io.failureDomains.openstack[0].rootVolume in body should have at least 1 properties"
```

And it works as expected.

---
```
openstack:
- availabilityZone: ""
```

I added a test and I noticed that I had no error, so I added `+kubebuilder:validation:MinLength=1` to `OpenStackFailureDomain.AvailabilityZone`, and now the test fails as we expect.

---
```
openstack:
- rootVolume:
    availabilityZone: ""
```

I added a test and I noticed that I had no error, so I added `+kubebuilder:validation:MinLength=1` to `RootVolume.AvailabilityZone`, and now the test fails as we expect.

---

@ingvagabund :

> Common sense tells me kubebuilder:validation:MinProperties:=1 together with AvailabilityZone as optional field does not work well together. Though, there might be a bigger picture of the validation mechanism I don't see.

I'll let @JoelSpeed comment on that, but looking at the tests I think we now cover every possible scenario.

ingvagabund:
```
// +optional
```
combination of `+optional` and `+kubebuilder:validation:MinLength=1` makes me wonder about the same. They seem to complement each other.

machine/v1/stable.controlplanemachineset.openstack.testsuite.yaml
```diff
@@ -0,0 +1,349 @@
...
+                machine.openshift.io/cluster-api-machine-type: master
+                machine.openshift.io/cluster-api-cluster: cluster
+            spec:
+              providerSpec: {}
+            failureDomains:
+              platform: OpenStack
+              openstack:
+              - rootVolume:
+                  availabilityZone: ""
+    expectedError: "spec.template.machines_v1beta1_machine_openshift_io.failureDomains.openstack[0].rootVolume.availabilityZone in body should be at least 1 chars long"
```
JoelSpeed, line 349: Nit: Files should end with a newline character

JoelSpeed (commented):
For the availability zone fields, are there any character based validation we can add? Is there a maximal length for the strings?

Given these are passing into pre-existing fields in the v1alpha1 API, I expect not, but perhaps there is something in a webhook or other validation (eg in CAPO) that means there is some validation

Otherwise API LGTM

EmilienM:
all comments addressed @JoelSpeed

machine/v1/types_controlplanemachineset.go
ingvagabund: After detailed analysis of the kube-openapi code the `+optional` has no additional effect as long as `omitempty` is used. 
- With only `omitempty` specified neither `Default: ""` nor `Required:` fields in the schema definition are generated
- With both `+optional` and `omitempty` specified, neither `Default: ""` nor `Required:` fields in the schema definition are generated
- With only `+optional` specified the schema generates additional `Default: ""` field in the schema definition.
- Without both `+optional` and `omitempty` specified,  I get additional `Required: []string{...}` field generated in the schema definition

ingvagabund: Back to:
```
// +optional
```
This no longer complements each other.
- `+optional` means `AvailabilityZone` is not required
- `+kubebuilder:validation:MinLength=1` applies only when `AvailabilityZone` field is specified. If it is its minLength is 1.

Back to:
```
// +kubebuilder:validation:MinProperties:=1
type RootVolume struct {
...
	// +optional
```
This together with `+kubebuilder:validation:MinLength=1` for the `AvailabilityZone` addresses the `{RootVolume: {AvailabilityZone: ""}}` case.

ingvagabund:
No more objections on my side. This api review was quite educational. Thank you for all the input.

machine/v1/types_controlplanemachineset.go
EmilienM: Thanks @ingvagabund for this deep analysis.

stephenfin:
> > For the availability zone fields, are there any character based validation we can add? Is there a maximal length for the strings?
> 
> ```
> Availability zone names must NOT contain : since it is used by admin users to specify hosts where instances are launched in server creation
> ```
> 
> Source: [docs.openstack.org/nova/latest/admin/availability-zones.html](https://docs.openstack.org/nova/latest/admin/availability-zones.html)
> 
> I'll look if we can validate that.

*cloud-provider-openstack* also imposes its own limitation, which is probably relevant:

> **OpenStack availability zone must not contain blank**
>
> `topology.kubernetes.io/zone` is used to label node and its value comes from availability zone of the node, according to [label spec](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set) it does not support blank (' ') but OpenStack availability zone supports blank. So your OpenStack availability zone must not contain blank otherwise it will lead to node that belongs to this availability zone register failure, see https://github.com/kubernetes/cloud-provider-openstack/issues/1379 for further information.

From the [Cloud Controller Manager docs](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md#limitation).

EmilienM:
> otherwise it will lead to node that belongs to this availability zone register failure, see [kubernetes/cloud-provider-openstack#1379](https://github.com/kubernetes/cloud-provider-openstack/issues/1379) for further information.

I addressed this comment and added more validation!

machine/v1/types_controlplanemachineset.go
```diff
@@ -248,6 +249,10 @@ type FailureDomains struct {
 	// GCP configures failure domain information for the GCP platform.
 	// +optional
 	GCP *[]GCPFailureDomain `json:"gcp,omitempty"`
+
+	// OpenStack configures failure domain information for the OpenStack platform.
+	// +optional
+	OpenStack *[]OpenStackFailureDomain `json:"openstack,omitempty"`
```
JoelSpeed, line 255: While this goes against the other items in this list, do you think you could change this to not be a pointer, and see if that still works? I think that would have made our zero length failure domains list bug much more obvious. I need to double check the unions guidance but I think that should be ok, the idea is that it needs to be omitted when empty, so we can change nil checks for length checks

EmilienM: Sounds good. I'm testing it now.
I'll update the other PRs accordingly.

machine/v1/types_controlplanemachineset.go
```diff
@@ -248,6 +249,10 @@ type FailureDomains struct {
 	// GCP configures failure domain information for the GCP platform.
 	// +optional
 	GCP *[]GCPFailureDomain `json:"gcp,omitempty"`
+
+	// OpenStack configures failure domain information for the OpenStack platform.
+	// +optional
+	OpenStack []OpenStackFailureDomain `json:"openstack,omitempty"`
```
EmilienM, line 255: @JoelSpeed I've updated it here to not be a pointer, and updated CCPMSO as well.

EmilienM: I guess I should add a comment here explaining why we don't use a pointer.

JoelSpeed (commented):
I've reviewed this again and there's one point I'd like to make for my future self and other API approvers. In a discriminated union, my understanding is that we need to make sure that the empty union member is omitted, by using omitempty. For a slice, which as far as I can tell is a valid union member type (nothing says you can't in the KEP), this means it must be a reference and not a pointer. We found a bug in the logic where an empty list was not omitted and this caused issue since the empty list is not valid in the failure domains structure.

Also, we found that to enable `minProperties` to work correctly, the `rootVolume` must also be a pointer. Omitempty does not work on pointers to structs so by including this as a pointer, we prevented the `minProperties` validations from working correctly in that area of the API. I have a [PR](https://github.com/openshift/enhancements/pull/1411) open to the enhancements repo to update this in the guidelines.

/lgtm

/hold

@EmilienM Please link a CPMS E2E job that is green that includes the API changes from this test as proof that this API is working as expected, once you've done that, feel free to unhold

EmilienM:
https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/195 includes this API change and the e2e is passing on OpenStack:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-control-plane-machine-set-operator/195/pull-ci-openshift-cluster-control-plane-machine-set-operator-main-e2e-openstack-operator/1666068518879105024

/hold cancel

Thanks for the reviews @JoelSpeed and @ingvagabund , this was really interesting to learn all of this.

PR #1447, opened 2023-04-24
Remove ValidatingAdmissionPolicy from tech preview.

benluddy:
This requires additional work in cluster-kube-apiserver-operator to recognize the gate and enable related APIs accordingly.


deads2k:
/lgtm
/approve

thanks.

PR #1451, opened 2023-04-27
Add maxTotalStorage config for OVN-K Audit Logging

jluhrsen:
changes are for:
  https://github.com/openshift/cluster-network-operator/pull/1798
  https://issues.redhat.com/browse/OCPBUGS-10765


kyrtapz:
@jluhrsen have you considered using something like `maxFiles` instead? I think it is more clear and in the end serves the same purpose. Kubelet does the same for log rotation with `containerLogMaxSize` and `containerLogMaxFiles`

jluhrsen:
> @jluhrsen have you considered using something like `maxFiles` instead? I think it is more clear and in the end serves the same purpose. Kubelet does the same for log rotation with `containerLogMaxSize` and `containerLogMaxFiles`

ok, I've updated the two PRs to use MaxLogFiles instead. PTAL

trozet (approved):
/lgtm

PR #1060, opened 2021-11-18
moving mco api/crd to openshift/api

kikisdeliveryservice:
Successfully ran make generate-with-container on these changes.

PR #1040, opened 2021-10-22
[WIP] move mco api to openshift/api

kikisdeliveryservice:
Updated https://github.com/openshift/api/pull/301 

Co-authored-by: Christian Glombek <cglombek@redhat.com>

PR #620, opened 2020-04-06
WIP: Add machineconfiguration API group

LorbusChris:
This moves over the machineconfiguration API from the
github.com/openshift/machine-config-operator repo,
and adds swagger doc generation for it.

TODO:
- [ ] Re-add ControllerConfig
- [ ] Exclude RawExtension from OpenAPI Validation generation
- [ ] Address review here 

/cc @runcom

PR #301, opened 2019-04-29
WIP: Add machineconfiguration API

LorbusChris:
This PR moves over the type definitions for the machineconfiguration API from the MCO repo, except for those depending on Ignition, which will stay in the MCO repo.

**WIP**

cc @runcom @sttts

PR #1453, opened 2023-05-04
MCO-52: add MCO API into openshift/api 

jkyros:
This is an [existing API](https://github.com/openshift/machine-config-operator/tree/master/pkg/apis/machineconfiguration.openshift.io), so fortunately or unfortunately we're probably stuck with some of the...tech debt here :frowning_face: 

We've tried this before, and while there wasn't really anything we were "hard blocked" on, it never quite got finished. 

Iif you want to re-read previous attempts/reviews: 

- https://github.com/openshift/api/issues/216 (historical protobuf context, not an issue anymore because we use RawExtension for ignition now)
-  https://github.com/openshift/machine-config-operator/issues/588
- https://github.com/openshift/api/pull/301
-  https://github.com/openshift/api/pull/620
- https://github.com/openshift/api/pull/1040
-  https://github.com/openshift/api/pull/1060

This time is going to be the one though, I can feel it! :smile:  

After this is done, we'll need to do client-go and pull it back into the MCO: 
- https://github.com/openshift/client-go/pull/243
- https://github.com/openshift/machine-config-operator/pull/3747

What I did: 
- took the existing MCO API from https://github.com/openshift/machine-config-operator/tree/master/pkg/apis/machineconfiguration.openshift.io
- added protobuf tags 
- added an operator CRD (`machineconfigurations.operator.openshift.io`) for the MCO since it did not have one  (this should let us [unblock our move](https://github.com/openshift/machine-config-operator/pull/2882) away from our library-go fork, since we'll have a place to store our generations) 
- added compatibility level annotations
- added more detailed field descriptions in some cases


machineconfiguration/v1/helpers.go
```diff
@@ -0,0 +1,199 @@
...
+
+import (
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// NewMachineConfigPoolCondition creates a new MachineConfigPool condition.
+func NewMachineConfigPoolCondition(condType MachineConfigPoolConditionType, status corev1.ConditionStatus, reason, message string) *MachineConfigPoolCondition {
```
jkyros, line 11: These helpers aren't used anywhere but in the MCO to my knowledge. I assume you'll want them excluded, but I figured there was no harm in asking? :smile:

machineconfiguration/v1/types.go
```diff
@@ -0,0 +1,654 @@
...
+type ControllerConfig struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+	// TODO(jkyros): inconsistent historical generation resulted in the controllerconfig CRD being
+	// generated with all fields required, while everything else was generated with optional
+
+	// +kubebuilder:validation:Required
+	// +required
+	Spec ControllerConfigSpec `json:"spec" protobuf:"bytes,2,opt,name=spec"`
```
jkyros, line 38: Hmm that failed integration test made me poke through all the required fields. 

The MCO hasn't been historically great at required/non-required fields since most of our CRD updates had been done manually.  It looks like ControllerConfig was originally (years ago) generated with "required by default" at the package level and has been updated that way, while everything else was generated with "optional by default" at the package level and maintained that way.  

This PR initially defaulted to "fields required by default" at the package level. Looking at the rest of the APIs in the repo, it looks like there's a mix?

I've adjusted this now so that:
-  `// +kubebuilder:validation:Optional` is set at the package level 
- the `ControllerConfig` fields that were previously required are set explicitly to required 
- the tests should be fixed and now supply the required fields (I will clean the data up before merge) 

Generally though, what are your "best practice" thoughts on how required fields should be handled: 
- Required by default, mark everything we don't want required as optional or 
- Optional by default, mark everything we don't want optional as required 
- Mark literally every field explicitly so it doesn't matter? :smile:

deads2k:
/hold

We definitely want this, but we have some other churn in progress that must complete first.  Then we take this to reduce future debt in 4.14.

jkyros:
Rebased, set our runlevel to 80 to match existing MCO, adjusted approval annotation to match this PR, pulled out the helpers :smile:

machineconfiguration/v1/types.go
```diff
@@ -0,0 +1,654 @@
...
+	// +optional
+	Status ControllerConfigStatus `json:"status" protobuf:"bytes,3,opt,name=status"`
```
jkyros, line 40: The MCO itself does not care about the protobufs or use them, so I'm cool with removing them.

Previous attempts at migration seemed to really care a lot about getting them in there but the reason was unclear. It could have been either cargo-cult or for something that has since gone away? 

I just want to make sure we don't have some important thing somewhere that's needs protobufs that I don't know about. 

Should I take them out?

jkyros:
Rebased and trued up with recent MCO merges. I'll squash the extra commits out when we're cool with this.

operator/v1/register.go
```diff
@@ -54,6 +54,8 @@ func addKnownTypes(scheme *runtime.Scheme) error {
 		&KubeSchedulerList{},
 		&KubeStorageVersionMigrator{},
 		&KubeStorageVersionMigratorList{},
+		&MachineConfiguration{},
+		&MachineConfigurationList{},
```
jkyros, line 58: Just added MachineConfiguration to operator register.go so @cdoern  could get client-go to generate properly

machineconfiguration/v1/0000_80_containerruntimeconfig.crd.yaml
```diff
@@ -113,10 +113,6 @@ spec:
                   items:
                     description: ContainerRuntimeConfigCondition defines the state of the ContainerRuntimeConfig
                     type: object
-                    required:
-                      - lastTransitionTime
-                      - status
-                      - type
```
JoelSpeed: No longer required?

machineconfiguration/v1/doc.go
```diff
@@ -1,5 +1,6 @@
 // +k8s:deepcopy-gen=package,register
 // +groupName=machineconfiguration.openshift.io
 
+// +kubebuilder:validation:Optional
```
JoelSpeed, line 4: Why did you chose to add this? Moving fields from being required to optional could be a breaking change so we need to be careful here

machineconfiguration/v1/types.go
```diff
@@ -30,6 +30,10 @@ type ControllerConfig struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
 
+	// TODO(jkyros): inconsistent historical generation resulted in the controllerconfig CRD being
+	// generated with all fields required, while everything else was generated with optional
+
+	// +kubebuilder:validation:Required
 	// +required
```
JoelSpeed, line 37: This tag isn't correct, please drop the `+required` from everywhere, only use `+kubebuilder:validation:Required`

machineconfiguration/v1/stable.controllerconfig.testsuite.yaml
```diff
@@ -1,15 +1,101 @@
...
+        networkType: OVNKubernetes
+        osImageURL: example.com/example/openshift-release-dev@sha256:eacdc37aec78fdbf8caa9601e4012ab31453cf59b086474901900e853e803ea8
+        platform: none
+        proxy: null
+        pullSecret:
+          name: pull-secret
+          namespace: openshift-config
+        releaseImage: ""
+        rootCAData: Y2VydGlmaWNhdGUK
+
```
JoelSpeed, line 101: More red markers at the end of files

machineconfiguration/v1/doc.go
jkyros: Historically..."mistakes were made" https://github.com/openshift/api/pull/1453#discussion_r1186430524. Half of our stuff was generated one way, half of it was generated the other. If you'd rather I "invert" it and go through and tag everything optional that was optional instead, I can do that.

machineconfiguration/v1/0000_80_containerruntimeconfig.crd.yaml
jkyros: Unforunately yes, because in the MCO originally it wasn't: 

https://github.com/openshift/machine-config-operator/blob/886c5c3fc7a96e922a65d83ecd39d19f4921dd50/install/0000_80_machine-config-operator_01_containerruntimeconfig.crd.yaml#L145

JoelSpeed: Ahh right, so in the copy/paste some generation has made it required, and this reverts that? Would be good to squash that change somehow so it's obvious?

Perhaps if we do the copy/paste, then the next commit fixes all the optional vs required, then generate as the final commit, it will be more obvious that there's actually no diff?

machineconfiguration/v1/0000_80_containerruntimeconfig.crd.yaml
```diff
@@ -0,0 +1,141 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: containerruntimeconfigs.machineconfiguration.openshift.io
+  labels:
+    "openshift.io/operator-managed": ""
+  annotations:
+    api-approved.openshift.io: https://github.com/openshift/api/pull/1060
```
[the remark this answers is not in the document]
jkyros: nope, that was the previous attempt, I missed one, thanks!

operator/v1/0000_80_machine-config-operator_01_config.crd.yaml
```diff
@@ -0,0 +1,192 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    api-approved.openshift.io: https://github.com/openshift/api/pull/TODO(jkyros)
```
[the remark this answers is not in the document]
jkyros: fixed, thanks!

machineconfiguration/v1/stable.controllerconfig.testsuite.yaml
```diff
@@ -0,0 +1,15 @@
...
+  - name: Should be able to create a minimal ControllerConfig
+    initial: |
+      apiVersion: machineconfiguration.openshift.io/v1
+      kind: ControllerConfig
+      spec: {} # No spec is required for a ControllerConfig
+    expected: |
+      apiVersion: machineconfiguration.openshift.io/v1
+      kind: ControllerConfig
+      spec: {}
+
```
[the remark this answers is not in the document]
jkyros: fixed, thanks!

machineconfiguration/v1/0000_80_containerruntimeconfig.crd.yaml
jkyros: Thanks, I tried to squish the commits down to a more clear: 
-  add existing -> modify annotations -> generate

hopefully that's...at least a little better?

machineconfiguration/v1/stable.controllerconfig.testsuite.yaml
jkyros: fixed, thanks!

machineconfiguration/v1/types.go
jkyros: thanks, removed all `+required`

machineconfiguration/v1/doc.go
jkyros: Thanks, yep, I have now marked all the fields as optional or required. 

I did keep the `+kubebuilder:validation:Optional` for now because:

-  in the existing MCO we a [field that is marked required](https://github.com/openshift/machine-config-operator/blob/40571adccaa3ef49a823abe8d7494bcdc48f2021/install/0000_80_machine-config-operator_01_containerruntimeconfig.crd.yaml#L51) in our CRD but is `omitempty` [in types.go](https://github.com/openshift/machine-config-operator/blob/40571adccaa3ef49a823abe8d7494bcdc48f2021/pkg/apis/machineconfiguration.openshift.io/v1/types.go#L523) (maintaining CRDs by hand...) and that seemed to be the only way to keep parity. 
-  I did generate both ways (with global optional in doc.go and without) and that field was the only difference


(I don't think preserving it matters *that* much, but I figure we can revisit it next time we make an MCO API change and/or do some "opinionated cleanup" on our API?)

jkyros:
Changes are: 
  - rebased
  - squashed out some of the intermediate commits in an attempt to make things clearer
  - removed all usage of `+required` annotation in favor of `+kubebuilder:validation:Required`
  - grouped the machineconfiguration operator stuff together in the same commit 
  - fixed approval annotations to match this PR 
  - fixed test newlines 
  - fixed the capitalization of omitEmpty in StaticPodOperatorStatus in operator types (but we can take it out if we don't want to fix it as part of this)

machineconfiguration/v1/0000_80_containerruntimeconfig.crd.yaml
JoelSpeed: That's what I was hoping for, let me re-review and i'll let you know if anything else can be done :)

machineconfiguration/v1/doc.go
JoelSpeed: Any field that isn't explicitly marked as optional or required, but happens to have `omitempty`, becomes optional by the logic of the generators.

Since the CRD had it as required, it should remain as required

machineconfiguration/v1/types.go
```diff
@@ -167,6 +185,32 @@ type ControllerConfigStatus struct {
...
+	// +kubebuilder:validation:Required
+	Signer string `json:"signer"`
+
+	// notBefore is the lower boundary for validity
+	// +kubebuilder:validation:Required
+	NotBefore string `json:"notBefore"`
+
+	// notAfter is the upper boundary for validity
+	// +kubebuilder:validation:Required
+	NotAfter string `json:"notAfter"`
```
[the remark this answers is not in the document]
cdoern: Not sure that this matters, I added these and for sanity of output string just worked better. I can change if we need to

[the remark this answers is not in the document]
JoelSpeed: Is this status or spec?

Typically timestamps in kube are all `*metav1.Time` since it has special marshal/unmarshal rules and a standard representation of the time, not sure it's hugely important but had I been doing an API review on this, I would have suggested to be consistent with the other APIs

operator/v1/types_machineconfiguration.go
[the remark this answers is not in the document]
jkyros: whoops, yep, moved it into the operator type commit, thanks

jkyros:
that was weird. gateway timeouts? 
/retest

machineconfiguration/v1/types.go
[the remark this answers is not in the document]
JoelSpeed: Yeah I think that would be preferable if these fields were introduced this release

machineconfiguration/v1/types.go
JoelSpeed: Protobufs have no effect for CRD types, so they should be removed, please make that its own commit

machineconfiguration/v1/0000_80_containerruntimeconfig.crd.yaml
JoelSpeed: Ok so these required to optional transitions still exist even when the update comes at the end? Why are these fields switching from required to optional?

machineconfiguration/v1/0000_80_controllerconfig.crd.yaml
```diff
@@ -49,7 +49,6 @@ spec:
               required:
                 - additionalTrustBundle
                 - baseOSContainerImage
-                - baseOSExtensionsContainerImage
```
JoelSpeed: Removing options from the enum, why has that change come into this PR? Was the CRD out of date in MCO?

JoelSpeed (commented):
I have concern about the number of fields we are switching from required to optional, what's the justification for switching these fields over? Is it even intentional?

machineconfiguration/v1/0000_80_controllerconfig.crd.yaml
jkyros: These specifically in ControllerConfig I did adjust. 

So last time ControllerConfig got regenerated it was a one-off by a "3rd party" PR and those fields got set wrong...and we were apparently in a hurry.  It effectively didn't matter to the MCO because it always supplies the whole structure but we marked it as "probably not great, we should fix it when we migrate the API":  https://github.com/openshift/machine-config-operator/pull/3567/files#r1114291604

And osImageURL is optional now that we're moved to native container format images and that just never got adjusted (again, I think we're supplying the whole thing every time so I don't think it actually matters, but it was wrong).

machineconfiguration/v1/0000_80_kubeletconfig.crd.yaml
```diff
@@ -134,10 +134,6 @@ spec:
                   items:
                     description: KubeletConfigCondition defines the state of the KubeletConfig
                     type: object
-                    required:
-                      - lastTransitionTime
-                      - status
-                      - type
```
jkyros: this is currently not required in the MCO: https://github.com/openshift/machine-config-operator/blob/40571adccaa3ef49a823abe8d7494bcdc48f2021/install/0000_80_machine-config-operator_01_kubeletconfig.crd.yaml#L216C1-L216C1

machineconfiguration/v1/0000_80_machineconfigpool.crd.yaml
```diff
@@ -87,9 +87,6 @@ spec:
             spec:
               description: MachineConfigPoolSpec is the spec for MachineConfigPool resource.
               type: object
-              required:
-                - configuration
```
jkyros: this is currently not required in the MCO: https://github.com/openshift/machine-config-operator/blob/40571adccaa3ef49a823abe8d7494bcdc48f2021/install/0000_80_machine-config-operator_01_machineconfigpool.crd.yaml#L93C13-L93C13

machineconfiguration/v1/0000_80_machineconfigpool.crd.yaml
```diff
@@ -220,26 +217,33 @@ spec:
             status:
               description: MachineConfigPoolStatus is the status for MachineConfigPool resource.
               type: object
-              required:
-                - configuration
-                - degradedMachineCount
-                - machineCount
-                - readyMachineCount
-                - unavailableMachineCount
-                - updatedMachineCount
```
jkyros: this is currently not required in the MCO: https://github.com/openshift/machine-config-operator/blob/40571adccaa3ef49a823abe8d7494bcdc48f2021/install/0000_80_machine-config-operator_01_machineconfigpool.crd.yaml#L284

machineconfiguration/v1/0000_80_machineconfigpool.crd.yaml
```diff
@@ -220,26 +217,33 @@ spec:
...
                   type: array
                   items:
                     description: MachineConfigPoolCondition contains condition information for an MachineConfigPool.
                     type: object
-                    required:
-                      - lastTransitionTime
-                      - message
-                      - reason
-                      - status
-                      - type
```
jkyros: this is currently not required in the MCO https://github.com/openshift/machine-config-operator/blob/40571adccaa3ef49a823abe8d7494bcdc48f2021/install/0000_80_machine-config-operator_01_machineconfigpool.crd.yaml#L293

jkyros (commented):
For the skew between the base CRD here and the generation after the tags have been adjusted, I believe I can account for all of it

jkyros:
Protobufs have been removed :smile:

machineconfiguration/v1/0000_80_controllerconfig.crd.yaml
JoelSpeed: Right, so in this case, you made a breaking change in the 4.13 branch, this is undoing that breaking change, I'm ok with that

machineconfiguration/v1/0000_80_containerruntimeconfig.crd.yaml
JoelSpeed: Can you link to where this CRD currently gets installed into payload to show that these fields aren't correct there? If that's the case, at least we have a record of it then

JoelSpeed: Ahh I see, you've done that already nvm

machineconfiguration/v1/0000_80_machineconfig.crd.yaml
```diff
@@ -49,8 +49,6 @@ spec:
         openAPIV3Schema:
           description: "MachineConfig defines the configuration for a machine \n Compatibility level 1: Stable within a major release for a minimum of 12 months or 3 minor releases (whichever is longer)."
           type: object
-          required:
-            - spec
```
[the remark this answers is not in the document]
JoelSpeed: This isn't required in the existing CRD in payload, https://github.com/openshift/machine-config-operator/blob/beca7a6c8a28d83419dd2dfd74e256ca2b80d754/install/0000_80_machine-config-operator_01_machineconfig.crd.yaml#L50-L52C9

JoelSpeed:
Apart from https://github.com/openshift/api/pull/1453#discussion_r1288134482, I think everything else I've asked for changes on has been resolved

machineconfiguration/v1/types.go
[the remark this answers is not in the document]
jkyros: I've changed these time fields to *metav1.Time, I left it as a separate commit but I can squash it out if need be

jkyros:
The "certificate time fields" from https://github.com/openshift/api/pull/1453#discussion_r1288134482 have been changed as requested, so...maybe we've got it? :smile: :smile: 

those test failures look systemic and unreleated to what  I did. They passed locally. Let's try again ?
/retest

jkyros:
@JoelSpeed we've got green on: 
- https://github.com/openshift/client-go/pull/243
And mostly green on: 
- https://github.com/openshift/machine-config-operator/pull/3747 (My attempt) 
(We stepped on ourselves with the upgrade test because we made that breaking metav1.Time change to our API, but everything else checks out) 
- https://github.com/openshift/machine-config-operator/pull/3840 (Charlie's attempt) 

are we too late to get this in?

JoelSpeed:
/lgtm
/hold

When you are ready to merge, go ahead.

If you think this can be merged, with the remaining dependents before the branch gets cut for 4.15, then go ahead, though I'd advise a cautious approach here given how short we are on time

cgwalters:
I think ControllerConfig is an implementation detail of the MCO that should absolutely *not* move into openshift/api - we really should aim to remove it entirely because it duplicates lots of data (including secrets) that we really shouldn't be duplicating.

machineconfiguration/v1/types.go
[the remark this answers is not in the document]
cgwalters: It sounds like changing this is tricky for backwards compatibility?  Since controllerconfig is an implementation detail today, I'm not sure we should block on it.

jkyros:
> I think ControllerConfig is an implementation detail of the MCO that should absolutely not move into openshift/api - we really should aim to remove it entirely because it duplicates lots of data (including secrets) that we really shouldn't be duplicating.

I did mark it as "internal" at least :smile: 

Until we "get rid of it completely" though, we have to do...something with it. Like we can leave it in the MCO, but it's getting updated there poorly by hand. 

Those embedded objects get updated here, and then don't get vendored back into the MCO all the time, in some cases because folks don't know they need to do that (because the MCO's API is just off by itself) so we're perpetually out of sync.

cgwalters:
OK if it's internal only and changing it is hard [let's not change it](https://github.com/openshift/api/pull/1453#discussion_r1295009744) - we have tons of other higher priority tech debt.

JoelSpeed:
The new verify-crd-schema check is expected to fail on a PR that is bringing existing types into the API. The idea is that it checks for new violations of CRD rules, but, we accept in certain cases it will need to be overridden. Will override when we are ready to merge

jkyros:
Ok cool, thanks @JoelSpeed. 

Are you and/or Eads okay with this merging once the 4.15 branch opens, or do we need to have a longer discussion about ControllerConfig being marked "internal" before we do that?

JoelSpeed:
@jkyros Can you add a topic to next weeks control plane arch so that we can discuss there what's required to get this over the line? David had some concerns about the controller config CRD so I think we should make sure those are enumerated

machineconfiguration/v1/types.go
```diff
@@ -0,0 +1,747 @@
...
+	runtime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+// MachineConfigRoleLabelKey is metadata key in the MachineConfig. Specifies the node role that config should be applied to.
+// For example: `master` or `worker`
+const MachineConfigRoleLabelKey = "machineconfiguration.openshift.io/role"
+
+// KubeletConfigRoleLabelPrefix is the label that must be present in the KubeletConfig CR
+const KubeletConfigRoleLabelPrefix = "pools.operator.machineconfiguration.openshift.io/"
```
wking, line 17: nit: Perhaps at some point in the future, other custom resources besides KubeletConfigs would use this annotation to associate themselves with pools? Do we want a more generic name for the variable, like `PoolLabelPrefix`?

machineconfiguration/v1/types.go
```diff
@@ -0,0 +1,747 @@
...
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ControllerConfig describes configuration for MachineConfigController.
+// This is currently only used to drive the MachineConfig objects generated by the TemplateController.
+//
+// Compatibility level 4: No compatibility is provided, the API can change at any point for any reason. These capabilities should not be used by applications needing long term support.
```
wking, line 26: worth putting a "can change at any time for any reason" api into v1alpha1, instead of putting it in `machineconfiguration/v1/`?

machineconfiguration/v1/types.go
```diff
@@ -0,0 +1,747 @@
...
+	// +optional
+	NetworkType string `json:"networkType,omitempty"`
+
+	// Network contains additional network related information
+	// +kubebuilder:validation:Required
+	// +nullable
+	Network *NetworkInfo `json:"network"`
+}
+
+// ImageRegustryBundle contains information for writing image registry certificates
```
wking, line 151: "ImageRegustryBundle" -> "ImageRegistryBundle" (u -> i)

machineconfiguration/v1/types.go
```diff
@@ -0,0 +1,747 @@
...
+
+const (
+	IPFamiliesIPv4                 IPFamiliesType = "IPv4"
+	IPFamiliesIPv6                 IPFamiliesType = "IPv6"
+	IPFamiliesDualStack            IPFamiliesType = "DualStack"
+	IPFamiliesDualStackIPv6Primary IPFamiliesType = "DualStackIPv6Primary"
+)
+
+// Network contains network related configuration
+type NetworkInfo struct {
```
wking, line 172: Godocs talking about `Network` but the type name is `NetworkInfo`.  Not clear to me what `Info` adds to the name; all of these structures are designed to carry information.  But a common pattern for `*Spec` child types is to be `*Spec` themselves, so `ControllerConfigSpec` would have `NetworkSpec` or, if you wanted to namespace defensively, `ControllerConfigNetworkSpec`.

machineconfiguration/v1/types.go
jkyros: good catch, thanks :smile:

machineconfiguration/v1/types.go
jkyros: It probably would if we were just now introducing ControllerConfig. We talked about this in control plane lifecycle arch on 9/19/2023, and my use of level 4 was probably erroneous.  We're just going to mark it level 1 and keep it in v1 since it's already out there.

jkyros:
Alright, I've: 
- added the missing `InternalRegistryPullSecret` field that the MCO merged the same day we last generated this :smile: 
- fixed the comment typo Trevor found above 
- rebased and regenerated 
- clients have also been regenerated https://github.com/openshift/client-go/pull/243

I notice I'm failing `verify-crd-schema` with even more errors ( `ListsMustHaveSSATags` ) now.

-  Do I need to default those lists to `//+listType=atomic` and we can come back later for merge efficiency gains? Or do we want to merge this to unblock MCO stuff and then come back and address that all at once?

JoelSpeed:
> Do I need to default those lists to //+listType=atomic and we can come back later for "list merge efficiency gains"? Or do we want to merge this to unblock MCO stuff and then come back and address that all at once?

Adding the atomic tag is good, as that will show that the field is treated as atomic. But, you can't change this now. The API is out and changing the behaviour of the merge strategy for the list could cause backwards compatibility issues for clients.

If you can add the tags for all the list type failures, we can ignore the other failures as these are "pre-existing unfixable" issues

operator/v1/types_machineconfiguration.go
```diff
@@ -0,0 +1,59 @@
...
+type MachineConfiguration struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// metadata is the standard object's metadata.
+	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
+	metav1.ObjectMeta `json:"metadata"`
+
+	// spec is the specification of the desired behavior of the Machine Config Operator
+	// +kubebuilder:validation:Required
+	// +required
```
JoelSpeed, line 24: Nit, don't need this tag, the kubebuilder version is canonical

machineconfiguration/v1/types.go
```diff
@@ -29,7 +29,10 @@ type ControllerConfig struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
 
-	// +required
+	// TODO(jkyros): inconsistent historical generation resulted in the controllerconfig CRD being
+	// generated with all fields required, while everything else was generated with optional
```
JoelSpeed, line 33: Is this todo resolved?

machineconfiguration/v1/types.go
JoelSpeed: Does this need to be fixed pre-merge?

machineconfiguration/v1/types.go
JoelSpeed: Does this need to be fixed pre-merge?

machineconfiguration/v1/types.go
```diff
@@ -0,0 +1,752 @@
...
+
+	// TemplateControllerFailing means the template controller is failing.
+	TemplateControllerFailing ControllerConfigStatusConditionType = "TemplateControllerFailing"
+)
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ControllerConfigList is a list of ControllerConfig resources
+//
+// Compatibility level 4: No compatibility is provided, the API can change at any point for any reason. These capabilities should not be used by applications needing long term support.
```
JoelSpeed, line 264: Should be level 1 right?

JoelSpeed (commented):
Have a few nits and a few questions about outstanding questions, but I think we are looking very close to go, is that correct?

machineconfiguration/v1/types.go
jkyros: good catch, I got the main type but I missed the List

machineconfiguration/v1/types.go
jkyros: yes, everything is explicitly tagged now, comment removed

operator/v1/types_machineconfiguration.go
jkyros: removed, thanks

machineconfiguration/v1/0000_80_controllerconfig.crd.yaml
```diff
@@ -0,0 +1,1246 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  # name must match the spec fields below, and be in the form: <plural>.<group>
+  name: controllerconfigs.machineconfiguration.openshift.io
+  labels:
+    "openshift.io/operator-managed": ""
+  annotations:
+    api-approved.openshift.io: https://github.com/openshift/api/pull/1453
```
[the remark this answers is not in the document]
jkyros: So that one wasn't previously in the install manifests, the MCO historically applied it when it came up. 

I've added annotations to match the rest of the CRDs:
```
  include.release.openshift.io/single-node-developer: "true"`
```
I didn't see any harm...well...unless the CVO has some sort of "these are annotated but I didn't apply them" cleanup/pruning function I don't know about.

jkyros:
Alright so now I've: 
 - added inclusion annotations to the ControllerConfig CRD
 - fixed the ControllerConfigList to be level 1 
 - added `//+listType=atomic ` tags to the lists that we own that verify was crying about (but upon further inspection, a lot of them are from embedded objects and other types here we don't own) 
 - fixed the required tag on the operator field 
 - removed my TODO about the tagging 
 - regenerated clients

I'm thinking "not right now" on Trevor's naming changes -- not that I disagree with him on the rationale, I appreciate the input, but I'm unsure of the blast radius since those aren't new fields (they've been in there for years). If we're gonna improve the naming, I'd rather do it as part of some type of holistic approach (and if they're safe to change now, they should be just as safe to change later).

machineconfiguration/v1/types.go
jkyros: I don't disagree with Trevor philosophically, but I vote no for now, since the fields have been in the MCO for years, and  if they're safe to change now, they're safe to change later once we've thought about it more.

Maybe we can do a pass after this merges of "things we can fix in the MCO API that won't break stuff" and clean this up some? (Yeah, I know we could version the API but ugh, that is unpleasant and I don't want to do it just for icky names  :smile: )

machineconfiguration/v1/types.go
jkyros: I vote no, it's already been in for a year: https://github.com/openshift/machine-config-operator/pull/3015 and was put in specifically for consumption by the kublet config controller

jkyros: okay, I say that, but I went and looked and...nothing references it right now. I wonder if it ever did?

I still think we should handle this later...but...yeah. Cleanup. :smile:

JoelSpeed:
I can't see any issues in net new fields, however, this type has a number of pre-existing failures causing the verify to fail. The verify compares new to old so with an override now should be ok in the future
/lgtm

jkyros:
@JoelSpeed are you cool if I release the hold? It's clean, I can safely pull it back into the MCO: https://github.com/openshift/machine-config-operator/pull/3747

PR #1455, opened 2023-05-05
OCPBUGS-8356: config/v1/types_cluster_version: Make update architecture omitempty

wking:
Avoid a cosmetic:

    Warning: unknown field "spec.desiredUpdate.architecture"

when a 4.13 or later `oc adm upgrade --to ...` or similar talks to a 4.12 or earlier ClusterVersion.  Before this commit, `oc` would format a patch body that included an explicit `"architecture": ""`, which the 4.12 ClusterVersion didn't understand.  With this change, `oc` will leave the `architecture` field unset unless it actually has architecture opinions.

### DIFF
--- a/config/v1/types_cluster_version.go
+++ b/config/v1/types_cluster_version.go
@@ -466,7 +466,7 @@ type Update struct {
 	// Valid values are 'Multi' and empty.
 	//
 	// +optional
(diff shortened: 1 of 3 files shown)


deads2k:
Server-side apply in the command sounds like a better solution.

PR #1460, opened 2023-05-16
DNS: Add awsPrivateHostedZoneRole

patrickdillon:
Adds a field to the DNS config to hold a role to assume when performing cross- account installs in an AWS shared VPC
environment.

https://github.com/openshift/enhancements/pull/1397


config/v1/types_dns.go
```diff
@@ -53,6 +53,14 @@ type DNSSpec struct {
...
 	// +optional
 	PrivateZone *DNSZone `json:"privateZone,omitempty"`
+	// awsPrivateHostedZoneRole contains the ARN of a role that should be assumed when performing
+	// operations on the cluster's private hosted zone specified in the cluster DNS config.
+	//
+	// When left empty, no role should be assumed.
+	//
+	// +kubebuilder:validation:Pattern:=`^arn:(aws|aws-cn|aws-us-gov):iam:[a-z0-9-]+:[0-9]{12}:role\/.*$`
+	// +optional
+	AWSPrivateHostedZoneRole string `json:"awsPrivateHostedZoneRole,omitempty"`
```
gcs278, line 63: Should this be made immutable if we don't support Day2? Could this just be in the `.status` field if it's immutable? Is the intent that we may support Day2 operations in the future, hence `.spec` makes sense?

config/v1/types_dns.go
```diff
@@ -96,3 +103,31 @@ type DNSList struct {
...
+	Type PlatformType `json:"type"`
+
+	// aws contains DNS configuration specific to the Amazon Web Services cloud provider.
+	// +optional
+	AWS *AWSDNSSpec `json:"aws,omitempty"`
+}
+
+// AWSDNSSpec contains DNS configuration specific to the Amazon Web Services cloud provider.
+type AWSDNSSpec struct {
+	// privateZoneIAMRole contains the ARN of a role that should be assumed when performing
```
gcs278, line 127: Minor nit pick:
```suggestion
	// privateZoneIAMRole contains the ARN of an IAM role that should be assumed when performing
```

config/v1/types_dns.go
```diff
@@ -53,6 +53,13 @@ type DNSSpec struct {
 	//
 	// +optional
 	PrivateZone *DNSZone `json:"privateZone,omitempty"`
+	// platform holds configuration specific to the underlying
+	// infrastructure provider for DNS.
+	// When omitted, this means the user has no opinion and the platform is left
+	// to choose reasonable defaults. These defaults are subject to change over time.
+	//
+	// +optional
+	Platform DNSPlatformSpec `json:"platform,omitempty"`
```
patrickdillon, line 62: oops, this should be a pointer.

mrunalp:
@deads2k ptal.

config/v1/types_dns.go
gcs278: Bump

config/v1/types_dns.go
gcs278: Not a big deal

config/v1/types_dns.go
```diff
@@ -96,3 +103,31 @@ type DNSList struct {
 
 	Items []DNS `json:"items"`
 }
+
+// DNSPlatformSpec holds cloud-provider-specific configuration
+// for DNS administration.
+// +union
```
ingvagabund, line 109: Worth adding `+kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'AWS' ?  has(self.aws) : !has(self.aws)",message="aws configuration is required when platform is AWS, and forbidden otherwise"` validation as in https://github.com/openshift/api/blob/master/machine/v1/types_controlplanemachineset.go#L230-L232. Even thought there's only `AWS` field for now, it helps to remind the validation is needed for new platforms as well.

ingvagabund: Also worth introducing integration tests. E.g. https://github.com/openshift/api/blob/master/machine/v1/stable.controlplanemachineset.aws.testsuite.yaml

config/v1/types_dns.go
```diff
@@ -96,3 +103,31 @@ type DNSList struct {
...
+type DNSPlatformSpec struct {
+	// type is the underlying infrastructure provider for the cluster.
+	// Allowed values are "AWS", "Azure", "BareMetal", "GCP", "Libvirt",
```
ingvagabund, line 112: FYI: when the list of platforms this description becomes outdated. Thus, better to mention which values are currently supported instead for the new struct. E.g. `// Currently supported value is AWS.`. Extend it later once new platform specific fields are added.

config/v1/types_dns.go
```diff
@@ -53,6 +53,13 @@ type DNSSpec struct {
 	//
 	// +optional
 	PrivateZone *DNSZone `json:"privateZone,omitempty"`
+	// platform holds configuration specific to the underlying
+	// infrastructure provider for DNS.
+	// When omitted, this means the user has no opinion and the platform is left
+	// to choose reasonable defaults. These defaults are subject to change over time.
+	//
+	// +optional
```
deads2k, line 61: For faster merge as we re-open the queue, make this field contingent on TechPreview like https://github.com/openshift/api/blob/master/example/v1/types_stable.go#L36.

You'll need to pre-create the CRD manifest as we did for infrastructure https://github.com/openshift/api/blob/master/config/v1/0000_10_config-operator_01_infrastructure-TechPreviewNoUpgrade.crd.yaml

config/v1/types_dns.go
```diff
@@ -96,3 +103,31 @@ type DNSList struct {
...
+// +union
+type DNSPlatformSpec struct {
+	// type is the underlying infrastructure provider for the cluster.
+	// Allowed values are "AWS", "Azure", "BareMetal", "GCP", "Libvirt",
+	// "OpenStack", "VSphere", "oVirt", "KubeVirt", "EquinixMetal", "PowerVS",
+	// "AlibabaCloud", "Nutanix" and "None". Individual components may not support all platforms,
+	// and must handle unrecognized platforms as None if they do not support that platform.
+	//
+	// +unionDiscriminator
+	Type PlatformType `json:"type"`
```
deads2k, line 118: Please restrict this field to "" and the values that we currently support (I think just AWS).  The discriminated union was exactly the right choice, but not everything is valid.

Can this be different than the infrastructure?  If not, indicate that skewing infrastructure and this field is unsupported and be sure to code logic into the operator to indicate Degraded=true when they don't match.

config/v1/types_dns.go
```diff
@@ -53,6 +53,13 @@ type DNSSpec struct {
 	//
 	// +optional
 	PrivateZone *DNSZone `json:"privateZone,omitempty"`
+	// platform holds configuration specific to the underlying
+	// infrastructure provider for DNS.
+	// When omitted, this means the user has no opinion and the platform is left
+	// to choose reasonable defaults. These defaults are subject to change over time.
+	//
+	// +optional
+	Platform *DNSPlatformSpec `json:"platform,omitempty"`
```
deads2k, line 62: For our configuration APIs, we avoid pointers for discoverability and make the zero value act as "no opinion, do good things"

config/v1/types_dns.go
```diff
@@ -96,3 +103,31 @@ type DNSList struct {
...
+}
+
+// AWSDNSSpec contains DNS configuration specific to the Amazon Web Services cloud provider.
+type AWSDNSSpec struct {
+	// privateZoneIAMRole contains the ARN of a role that should be assumed when performing
+	// operations on the cluster's private hosted zone specified in the cluster DNS config.
+	// When left empty, no role should be assumed.
+	// +kubebuilder:validation:Pattern:=`^arn:(aws|aws-cn|aws-us-gov):iam:[a-z0-9-]+:[0-9]{12}:role\/.*$`
+	// +optional
+	PrivateZoneIAMRole string `json:"privateZoneIAMRole,omitempty"`
```
deads2k, line 132: avoid omit empty (optional is fine) and allow empty string.

deads2k (changes requested):
a few additional changes please.

config/v1/types_dns.go
patrickdillon: > aws configuration is required when platform is AWS, and forbidden otherwise

Question about this: my intent was not to make aws configuration required on aws. Is that preferred? If we did that, how does the DNS config get updated when upgrading AWS clusters?

My intent was that the AWS platform would be an optional field, so DNS config in existing AWS clusters would not need to be updated.

patrickdillon: David answered my question. On upgrade, platform type  for aws clusters will be "". This logic makes sense to me now.

config/v1/types_dns.go
patrickdillon: > You'll need to pre-create the CRD manifest as we did for infrastructure https://github.com/openshift/api/blob/master/config/v1/0000_10_config-operator_01_infrastructure-TechPreviewNoUpgrade.crd.yaml

How should I pre-create this? I tried seeding the tech preview manifest by `cp`ing the default CRD and appending `TechPreviewNoUpgrade` to the filename, hoping that `make update` would pick it up, but it didn't seem to work (6e1722fb0a52c0569bca3ac52769db433da2e0df).

config/v1/types_dns.go
```diff
@@ -96,3 +103,44 @@ type DNSList struct {
...
+	// Individual components may not support all platforms,
+	// and must handle unrecognized platforms with best-effort defaults.
+	//
+	// +unionDiscriminator
+	// +kubebuilder:validation:Required
+	Type DNSPlatformType `json:"type"`
+
+	// aws contains DNS configuration specific to the Amazon Web Services cloud provider.
+	// +optional
+	AWS *AWSDNSSpec `json:"aws"`
```
patrickdillon, line 135: Should this be kept as a pointer (because it is in the discriminated union)?

config/v1/types_dns.go
```diff
@@ -96,3 +103,44 @@ type DNSList struct {
 
 	Items []DNS `json:"items"`
 }
+
+// DNSPlatformType is a cloud infrastructure provider that supports
+// platform specific DNS configuration.
+// +kubebuilder:validation:Enum=AWS
+type DNSPlatformType PlatformType
```
patrickdillon, line 110: Is this ok? Or would you prefer
```suggestion
type DNSPlatformType string
```
I was hoping this would allow 

```go
	AWSDNSPlatformType DNSPlatformType = AWSPlatformType
```
but that does not seem to work

config/v1/types_dns.go
```diff
@@ -96,3 +103,44 @@ type DNSList struct {
...
+	AWSDNSPlatformType DNSPlatformType = "AWS"
+)
+
+// DNSPlatformSpec holds cloud-provider-specific configuration
+// for DNS administration.
+// +union
+// +kubebuilder:validation:XValidation:rule="has(self.platform) && self.platform == 'AWS' ?  has(self.aws) : !has(self.aws)",message="aws configuration is required when platform is AWS, and forbidden otherwise"
+type DNSPlatformSpec struct {
+	// type is the underlying infrastructure provider for the cluster.
+	// Allowed values: "AWS".
```
patrickdillon, line 123: @deads2k suggested allowed values be `AWS` and `""`, but that seems to contrast with the validation suggested [here](https://github.com/openshift/api/pull/1460/files#r1202637837), which is what I went with. 

The intent being: `dns.spec.platform` is optional, but if it is provided `dns.spec.platform.type` is required

config/v1/types_dns.go
patrickdillon: Question about this [here](https://github.com/openshift/api/pull/1460#discussion_r1203324260)

patrickdillon:
@ingvagabund and @deads2k thanks for the review. I have addressed the feedback (except I still need to add the integration tests), but have a few resulting questions. Thanks!

config/v1/types_dns.go
deads2k: missing the `release.openshift.io/feature-set: TechPreviewNoUpgrade` annotation.

config/v1/types_dns.go
deads2k: Empty needs to be allowed to handle upgrade cases.

config/v1/types_dns.go
deads2k: PlatformType is fine.  I think you can add regex validation directly on your field.  If not, CEL would work ok in this instance.

config/v1/types_dns.go
deads2k: > Should this be kept as a pointer (because it is in the discriminated union)?

Yes.

config/v1/types_dns.go
patrickdillon: I have handled this in the code, but we couldn't figure out a test case (which you mentioned in slack). None of the fields in this config are mutable (although I'm not certain that is enforced). 

Was trying a test like this:

```yaml
  onUpdate:
  - name: Should allow empty platform on upgrade
    initial: |
      kind: DNS
      spec:
        zone:
          id: foo
    updated: |
      kind: DNS
      spec:
        zone:
          id: bar
    expected: |
      kind: DNS
      spec:
        zone:
          id: bar
        platform:
          type: ""
```

But that is failing 
>  [FAILED] the following fields were expected to match but did not:
  [(spec.platform/spec.platform) (spec.zone/spec.zone)]

We're not really sure how to write the upgrade test

config/v1/techpreview.dns.testsuite.yaml
```diff
@@ -0,0 +1,89 @@
+apiVersion: apiextensions.k8s.io/v1 # Hack because controller-gen complains if we don't have this
+name: "[Stable] DNS"
```
gcs278, line 2: Should this be `[TechPreview]`? (note: i'm not sure myself what this `Stable` refers to.

config/v1/techpreview.dns.testsuite.yaml
```diff
@@ -0,0 +1,89 @@
...
+      metadata:
+        creationTimestamp: null
+        name: cluster
+      spec:
+        platform:
+          type: Azure
+          azure:
+            privateZoneIAMRole: arn:aws:iam:123456789012:role/foo
+    expectedError: "Invalid value: \"string\": allowed values are '' and 'AWS'"
+  - name: Should not be able to specify invalid arn
```
gcs278, line 51: *nit* I know I suggested this, but probably better to be more specific for the test name for debug-ablity.
```suggestion
  - name: Should not be able to specify invalid AWS role ARN
```

patrickdillon:
@ingvagabund @deads2k I think all feedback has been addressed (except [one question about tests](https://github.com/openshift/api/pull/1460#discussion_r1207212747))

config/v1/types_dns.go
```diff
@@ -96,3 +103,35 @@ type DNSList struct {
...
 }
+
+// DNSPlatformSpec holds cloud-provider-specific configuration
+// for DNS administration.
+// +union
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'AWS' ?  has(self.aws) : !has(self.aws)",message="aws configuration is required when platform is AWS, and forbidden otherwise"
+type DNSPlatformSpec struct {
+	// type is the underlying infrastructure provider for the cluster.
+	// Allowed values: "", "AWS".
+	// DNSPlatformType is a subset of Infrastructure PlatformType.
```
Miciah, line 114: It isn't clear what `DNSPlatformType` refers to.  Is this what you mean to say?  
```suggestion
	// The set of allowed values for this field is a subset of the allowed values for Infrastructure PlatformType.
```

config/v1/0000_10_config-operator_01_dns-TechPreviewNoUpgrade.crd.yaml
```diff
@@ -0,0 +1,114 @@
...
+                        - None
+                        - VSphere
+                        - oVirt
+                        - IBMCloud
+                        - KubeVirt
+                        - EquinixMetal
+                        - PowerVS
+                        - AlibabaCloud
+                        - Nutanix
+                        - External
```
Miciah, line 77: Is this going to cause confusing error messages, or does the `x-kubernetes-validations` rule take precedence and prevent the `enum` rule from causing unsupported values to be printed in error messages?

config/v1/techpreview.dns.testsuite.yaml
```diff
@@ -0,0 +1,114 @@
...
+  - name: Should not be able to specify different type and platform
+    initial: |
+      apiVersion: config.openshift.io/v1
+      kind: DNS
+      spec:
+        platform:
+          type: ""
+          aws:
+            privateZoneIAMRole: arn:aws:iam:123456789012:role/foo
+    expectedError: "Invalid value: \"object\": aws configuration is required when platform is AWS, and forbidden otherwise"
```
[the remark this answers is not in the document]
patrickdillon: fixed

config/v1/types_dns.go
patrickdillon: fixed (removed) -- leftover from a previous iteration

config/v1/0000_10_config-operator_01_dns-TechPreviewNoUpgrade.crd.yaml
patrickdillon: In the test, the `x-kubernetes-validations` message takes precedence, so I think we're ok.

PR #1466, opened 2023-05-23
CNF-9384: add ImageRegistry capability

qJkee:
Capability manages the image registry which allows to distribute Docker images


bparees:
/approve

if you don't land the remaining PRs for this feature by feature complete/branch cut(august 17th), we will need to revert this api addition

PR #1468, opened 2023-05-24
OCPBUGS-3176: Add config knob for enabling ip forwarding

trozet:
With https://github.com/ovn-org/ovn-kubernetes/pull/3374 and https://github.com/openshift/machine-config-operator/pull/3676 we stopped enabling IP forwarding globally in the OCP host. It is now only enabled on interfaces that OVNK manages. Additionally iptables rules are put in place to limit what traffic is forwarded. This was to resolve an issue where our node by default acts as a router: https://issues.redhat.com/browse/OCPBUGS-3176

With these changes, by default in OCP we will no longer forward IP traffic. Most users should welcome this change as it means the OCP node will not be forwarding traffic unintentionally on their network. However, in case there is some corner case where a user expects traffic to be forwarded we will expose this new config knob to allow them to go back to the old behavior.

### DIFF
--- a/operator/v1/types_network.go
+++ b/operator/v1/types_network.go
@@ -489,6 +489,13 @@ type GatewayConfig struct {
 	// +kubebuilder:default:=false
 	// +optional
 	RoutingViaHost bool `json:"routingViaHost,omitempty"`
+	// EnableIPForwarding enables IP forwarding for all traffic on OVN-Kubernetes managed interfaces (such as br-ex).
+	// By default this is disabled, and Kubernetes related traffic is still forwarded appropriately. This setting
+	// is only useful if there is a desire to for the node to act as a router and forward traffic between interfaces
+	// on the host.
+	// +kubebuilder:default:=false
+	// +optional
+	EnableIPForwarding bool `json:"enableIPForwarding,omitempty"`
 }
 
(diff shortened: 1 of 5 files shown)


operator/v1/types_network.go
```diff
@@ -489,6 +489,13 @@ type GatewayConfig struct {
 	// +kubebuilder:default:=false
 	// +optional
 	RoutingViaHost bool `json:"routingViaHost,omitempty"`
+	// EnableIPForwarding enables IP forwarding for all traffic on OVN-Kubernetes managed interfaces (such as br-ex).
+	// By default this is disabled, and Kubernetes related traffic is still forwarded appropriately. This setting
+	// is only useful if there is a desire to for the node to act as a router and forward traffic between interfaces
+	// on the host.
+	// +kubebuilder:default:=false
+	// +optional
+	EnableIPForwarding bool `json:enableIPForwarding,omitempty`
```
tssurya, line 498:
```suggestion
	EnableIPForwarding bool `json:"enableIPForwarding,omitempty"`
```

tssurya (changes requested):
almost lgtm, this should fix the CI failure

tssurya (approved):
/lgtm

PR #1472, opened 2023-05-25
OCPBUGS-40906: Add encapsulation flag to IPsecConfig struct

mccv1r0:
Add encapsulation flag to IPsecConfig struct.

This flag allows UDP encapsulation of IPsec related packets for things like firewall and NAT traversal

See RFC  3948 for specific uses


yocum137:
Please re-open this PR for IPSec encapsulation on IBM VPC cloud.

operator/v1/types_network.go
```diff
@@ -562,6 +562,12 @@ type IPsecConfig struct {
 	// +kubebuilder:validation:Enum=Disabled;External;Full
 	// +optional
 	Mode IPsecMode `json:"mode,omitempty"`
+	// encapsulation uses rfc3949 encapsulation of IKE and IPsec encryption packets
+	// This is applicable only when mode is set to Full.
+	// Default is false.
+	// +optional
+	// +kubebuilder:default:=false
+	Encapsulation bool `json:"encapsulation,omitempty"`
```
JoelSpeed, line 570: This sounds like we are creating fields that only apply to certain modes, so, this struct should evolve to become a discriminated union.

It can't be a true union because mode is optional, but we should move any "full" related fields into a struct called `Full` to group them together

We can then use our standard union validation to ensure that the field `full` is specified only when `mode: Full`.

Given this is added after the mode field, that will be an optional union member

JoelSpeed, line 570: Bools are no longer permitted in our APIs, they do not evolve well, review our API conventions for further explanation.

You should come up with a meaningful enum instead in most cases

operator/v1/types_network.go
```diff
@@ -562,6 +562,31 @@ type IPsecConfig struct {
...
+	EnableEncapsulation = "yes"
+	// DisableEncapsulation disables RFC-3948 encapsulation on the IPsec tunnels
+	// configured by OVN and OpenvSwitch.
+	DisableEncapsulation = "no"
+)
+
+type IPsecFullModeConfig struct {
+	// encapsulation option to enabled or disable rfc3949 encapsulation of IKE and IPsec
+	// encryption packets
+	// +kubebuilder:validation:Enum:=yes;no;""
```
JoelSpeed, line 587: Yes and No are as bad as true and false. What does this actually do for an end user? What kind of encapsulation is this? Think about this from an end user perspective.

What difference does it make to me, as an end user, if I do, or do not have this field set to yes or no?

operator/v1/types_network.go
```diff
@@ -562,6 +562,31 @@ type IPsecConfig struct {
 	// +kubebuilder:validation:Enum=Disabled;External;Full
 	// +optional
 	Mode IPsecMode `json:"mode,omitempty"`
+
+	// full defines configuration parameters for the IPsec `Full` mode.
+	// This will be effective only when mode is configured with `Full`,
+	// otherwise it will be ignored.
+	// +optional
+	Full IPsecFullModeConfig `json:"full,omitempty"`
```
JoelSpeed, line 570: This needs to be a pointer, you can't have an omitempty with a non-pointer struct

JoelSpeed (commented):
New user facing features should have an enhancement and a feature gate for their introduction. Do you have any design docs for this yet, or thoughts on feature gating this and what you'll need to do to be able to declare this stable?

operator/v1/types_network.go
pperiyasamy: This option is going to set into OVN/OVS something like this: https://docs.openvswitch.org/en/latest/tutorials/ipsec/#custom-options and finally going to be set as ipsec config option for an ipsec tunnel.


```
encapsulation

In some cases, for example when ESP packets are filtered or when a broken IPsec peer does not properly recognise NAT,
it can be useful to force RFC-3948 encapsulation. In other cases, where IKE is NAT'ed but ESP packets can or should
flow without encapsulation, it can be useful to ignore the NAT-Traversal auto-detection. encapsulation=yes forces
the NAT detection code to lie and tell the remote peer that RFC-3948 encapsulation (ESP in port 4500 packets) is
required. encapsulation=no ignores the NAT detection causing ESP packets to send send without encapsulation.
The default value of encapsulation=auto follows the regular outcome of the NAT auto-detection code performed
in IKE. This option replaced the obsoleted forceencaps option.
```
https://libreswan.org/man/ipsec.conf.5.html

So I don't see any other values to represent this. Please suggest.

operator/v1/types_network.go
```diff
@@ -562,6 +562,31 @@ type IPsecConfig struct {
 	// +kubebuilder:validation:Enum=Disabled;External;Full
 	// +optional
 	Mode IPsecMode `json:"mode,omitempty"`
+
+	// full defines configuration parameters for the IPsec `Full` mode.
+	// This will be effective only when mode is configured with `Full`,
+	// otherwise it will be ignored.
```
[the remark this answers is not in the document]
pperiyasamy: @JoelSpeed thanks for the reference. added CEL to validate full parameter, is that fine ?

operator/v1/types_network.go
pperiyasamy: done.

pperiyasamy:
> New user facing features should have an enhancement and a feature gate for their introduction. Do you have any design docs for this yet, or thoughts on feature gating this and what you'll need to do to be able to declare this stable?

This is a minor change w.r.t implementation and addressing this for a bug [OCPBUGS-40906](https://issues.redhat.com/browse/OCPBUGS-40906), do you still recommend to have feature gate and design doc update for this ?

operator/v1/types_network.go
pperiyasamy: @JoelSpeed here is how implementation looks like for this new API: https://github.com/openshift/cluster-network-operator/pull/2573.

operator/v1/types_network.go
```diff
@@ -561,7 +563,31 @@ type IPsecConfig struct {
...
 	// +kubebuilder:validation:Enum=Disabled;External;Full
 	// +optional
+	// +unionDiscriminator
 	Mode IPsecMode `json:"mode,omitempty"`
+
+	// full defines configuration parameters for the IPsec `Full` mode.
+	// This is permitted only when mode is configured with `Full`,
+	// and forbidden otherwise.
+	// +unionMember,optional
+	Full *IPsecFullModeConfig `json:"full,omitempty"`
```
JoelSpeed, line 573: Needs an explicit optional tag too please
```suggestion
	// +optional
	Full *IPsecFullModeConfig `json:"full,omitempty"`
```

operator/v1/types_network.go
JoelSpeed: Seems to me like there are three modes if I read this correctly? Force, Auto, Disable? With the default being Auto?

Could you briefly explain to me, what actually is encapsulation in this context? Does AH or ESP, transport or tunnel come into this in any way? https://support.huawei.com/enterprise/en/doc/EDOC1100301600/c5d5a9b/understanding-ipsec-encapsulation-modes

I'm still trying to understand how we explain this to an end user in terms that they can easily understand

rbbratta: if ESP (IP Protocol 50)  is not supported this uses UDP encapsulation.  This is required for IBMCloud for example.

rbbratta: If your network or firewall blocks less common IP protocols like IP Protocol 50, this setting allows the use of standard UDP packets on ports 500 and 4500 instead.

rbbratta: IBMCloud IPI already uses UDP encap, it's just hardcoded.  This supports `platform: none` or other edge cases.

We should probably add UDP port 4500 somewhere in descriptions. 

Other UPI Docs already indicate UDP port 4500 is required, so we should be covered.
e.g. AWS UPI:
| Name                          | Description            | Protocol | Port |
|-------------------------------|------------------------|----------|------|
| MasterIngressIpsecNat         | IPsec NAT-T packets    | udp      | 4500 |
| MasterIngressWorkerIpsecNat   | IPsec NAT-T packets    | udp      | 4500 |

operator/v1/types_network.go
pperiyasamy: done.

operator/v1/types_network.go
pperiyasamy: Thanks @rbbratta about firewall use case, The NAT-T is very much needed there.

@JoelSpeed The IPsec tunnel is configured between OCP nodes in tunnel mode on the Geneve overlay traffic which carries pod payload traffic. Once IPsec SAs are established between nodes, then Gevenve traffic is encrypted with ESP header. So NAT-T (UDP) encapsulation happens on ESP payload so that it can pass firewalls and also compatible with intermediate NAT devices.

updated with 3 encapsulation modes (yes, no and auto) now.

JoelSpeed: Can we please avoid the use of `yes` and `no` as enum values, they are as bad as true and false. Enum values should be descriptive where possible, I still think Force and Disable are better

Think of it this way:
* Yes encapsulation
* No encapsulation
vs
* Force encapsulation
* Disable encapsulation

The latter two flow better in a sentence at a minimum

Also, enum values should be PascalCase

pperiyasamy: yes @JoelSpeed , done.

operator/v1/types_network.go
```diff
@@ -561,7 +563,35 @@ type IPsecConfig struct {
...
+	// DisableEncapsulation disables UDP encapsulation even if NAT is present.
+	DisableEncapsulation = "Disable"
+	// AutoEncapsulation determines use of UDP encapsulation based on the detection of NAT.
+	AutoEncapsulation = "Auto"
+)
+
+type IPsecFullModeConfig struct {
+	// encapsulation option to configure libreswan on how inter-pod traffic across nodes
+	// are encapsulated to handle NAT traversal. When configured it uses UDP port 4500
+	// for the encapsulation.
```
everettraven, line 591: I think the purpose of this field is documented well in this GoDoc, but I would like to see more information outlined on:
- Limitations of this field. I would recommend explicitly calling out the allowed values in the GoDoc and describing how each setting influences the behavior
- As of writing, this field is optional and doesn't look to have a default - what does not setting this field mean for encapsulation?

operator/v1/types_network.go
```diff
@@ -561,7 +563,35 @@ type IPsecConfig struct {
...
+	// +kubebuilder:validation:Enum:=Force;Disable;Auto;""
+	// +optional
+	Encap Encapsulation `json:"encapsulation,omitempty"`
```
everettraven, line 594: Any particular reason why the struct field name is `Encap` and not `Encapsulation`? Generally, I would recommend having the Go struct field name match the JSON field name where possible.

operator/v1/types_network.go
JoelSpeed: As Bryce says, we need to add godoc explaining the options, this typically takes the form of

```
// Valid values are Force, Disable, Auto and omitted.
// Force means ... 
// Disable means ...
// Auto means ...
// The current default is ...
```

Also, given this is optional/omitempty you don't need `""` as a valid value in the enum

operator/v1/types_network.go
```diff
@@ -561,7 +563,35 @@ type IPsecConfig struct {
...
+	Full *IPsecFullModeConfig `json:"full,omitempty"`
+}
+
+type Encapsulation string
+
+const (
+	// ForceEncapsulation forces UDP encapsulation regardless of whether NAT is detected.
+	ForceEncapsulation = "Force"
+	// DisableEncapsulation disables UDP encapsulation even if NAT is present.
+	DisableEncapsulation = "Disable"
```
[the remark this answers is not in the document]
everettraven: Maybe something like Always, Never, Auto would help avoid "Disabled" and be more naturally descriptive? 

Thinking from a UX perspective, I _think_ Always and Never would be easier to remember as "always encapsulate" and "never encapsulate" than something like Force and Disabled.

Provided the additional context, I don't think Force and Disabled are bad options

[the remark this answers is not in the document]
JoelSpeed: I like the suggestion of `Always` and `Never`, what do you think @pperiyasamy ?

operator/v1/types_network.go
pperiyasamy: sure, updated godoc with above info, hope it's fine now.

operator/v1/types_network.go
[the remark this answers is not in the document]
pperiyasamy: +1 
updated it with `Always` and `Never` options now.

operator/v1/types_network.go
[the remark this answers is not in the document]
pperiyasamy: at present `full` option may not be configured at all for most of the clusters, so let it be compatible with OCP upgrades as well. 
when customer wants to configure NAT-T encapsulation option other than default value, then they are free to configure it via full struct parameter.
of course there can be future requirements to customize ipsec connection on `Full` mode.

pperiyasamy: ok, changed it.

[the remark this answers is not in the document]
pperiyasamy: yes @everettraven , I can understand it now and the suggestion makes sense. 
adding `+kubebuilder:validation:MinProperties:=1` on the struct to support:
> - Make it such that when ipsecConfig.full is specified, at least one field member of the struct must also be specified

pperiyasamy:
Thanks @JoelSpeed and @everettraven for the reviews and merging it finally ! 
It was a good exercise for me from API perspective.

PR #1473, opened 2023-05-25
OCPBUGS-3176: Rework IP Forwarding config

trozet:
We are not supposed to use booleans anymore for API, so modifying the config to use a new type.

https://github.com/openshift/crd-schema-checker/blob/master/pkg/manifestcomparators/comp_no_bools.go#L20


deads2k:
/lgtm

tagging so we get the types right for deserialization before we update cluster-config-operator to write the CRs.  I'll make a followup to tweak validation.

PR #1482, opened 2023-06-04
Add etcd backup API

hasbro17:
Adds the techpreview API for etcd backups as outlined in https://github.com/openshift/enhancements/pull/1370

Adds the following APIs:
- `backups.config.openshift.io/v1alpha1`
  - Used to configure the schedule and retention policy of periodic backups
- `etcdbackups.operator.openshift.io/v1alpha1`
  - Triggers a onetime backup of the etcd cluster

/cc @tjungblu @deads2k


hasbro17:
Still some TODOs:
- Some CEL validation rules
- Need to figure out why `make update-code-gen-crds` isn't generating the openapi schema.

config/v1alpha1/types_backup.go
```diff
@@ -0,0 +1,131 @@
...
+
+// EtcdBackupSpec provides configuration for automated etcd backups to the cluster-etcd-operator
+type EtcdBackupSpec struct {
+
+	// Schedule defines the recurring backup schedule in Cron format
+	// every 2 hours: 0 */2 * * *
+	// every day at 3am: 0 3 * * *
+	// Setting to an empty string "" means disabling scheduled backups
+	// Default: ""
+	// TODO: Define CEL validation for the cron format
```
hasbro17, line 46: Need to figure out CEL validation for cron format. Apparently there are a few variants but we need the POSIX one. Would prefer to reuse whatever the upstream CronJob is using but if not I can always modify and appropriate the following (slightly scary) expression to the POSIX variety:
```console
^(?#second)(\*|(?:\*|(?:[0-9]|(?:[1-5][0-9])))\/(?:[0-9]|(?:[1-5][0-9]))|
(?:[0-9]|(?:[1-5][0-9]))(?:(?:\-[0-9]|\-(?:[1-5][0-9]))?|(?:\,(?:[0-9]|(?:[1-5][0-9])))*)) 
(?#minute)(\*|(?:\*|(?:[0-9]|(?:[1-5][0-9])))\/(?:[0-9]|(?:[1-5][0-9]))|(?:[0-9]|
(?:[1-5][0-9]))(?:(?:\-[0-9]|\-(?:[1-5][0-9]))?|(?:\,(?:[0-9]|(?:[1-5][0-9])))*)) 
(?#hour)(\*|(?:\*|(?:\*|(?:[0-9]|1[0-9]|2[0-3])))\/(?:[0-9]|1[0-9]|2[0-3])|
(?:[0-9]|1[0-9]|2[0-3])(?:(?:\-(?:[0-9]|1[0-9]|2[0-3]))?|(?:\,(?:[0-9]|1[0-9]|2[0-3]))*)) 
(?#day_of_month)(\*|\?|L(?:W|\-(?:[1-9]|(?:[12][0-9])|3[01]))?|(?:[1-9]|(?:[12][0-9])|
3[01])(?:W|\/(?:[1-9]|(?:[12][0-9])|3[01]))?|(?:[1-9]|(?:[12][0-9])|3[01])
(?:(?:\-(?:[1-9]|(?:[12][0-9])|3[01]))?|(?:\,(?:[1-9]|(?:[12][0-9])|3[01]))*)) 
(?#month)(\*|(?:[1-9]|1[012]|JAN|FEB|MAR|APR|MAY|JUN|JUL|AUG|SEP|OCT|NOV|DEC)
(?:(?:\-(?:[1-9]|1[012]|JAN|FEB|MAR|APR|MAY|JUN|JUL|AUG|SEP|OCT|NOV|DEC))?|
(?:\,(?:[1-9]|1[012]|JAN|FEB|MAR|APR|MAY|JUN|JUL|AUG|SEP|OCT|NOV|DEC))*)) 
(?#day_of_week)(\*|\?|[0-6](?:L|\#[1-5])?|(?:[0-6]|SUN|MON|TUE|WED|THU|FRI|SAT)
(?:(?:\-(?:[0-6]|SUN|MON|TUE|WED|THU|FRI|SAT))?|(?:\,(?:[0-6]|SUN|MON|TUE|WED|THU|FRI|SAT))*)) 
(?#year)(\*|(?:[1-9][0-9]{3})(?:(?:\-[1-9][0-9]{3})?|(?:\,[1-9][0-9]{3})*))$
```
https://www.codeproject.com/Tips/5299523/Regex-for-Cron-Expressions has a nice breakdown.

config/v1alpha1/types_backup.go
```diff
@@ -0,0 +1,131 @@
...
+	// +optional
+	RetentionCount int `json: "retentionCount,omitempty"`
+
+	// RetentionSizeGb defines the total size in Gb of backups to retain.
+	// If the current total size backups exceeds RetentionSizeGb then
+	// the oldest backup will be removed before a new backup is initiated.
+	// +kubebuilder:validation:Minimum=0.1
+	// +kubebuilder:validation:Optional
+	// +optional
+	RetentionSizeGb float64 `json: "retentionSizeGb,omitempty"`
```
hasbro17, line 111: > make a struct containing this int please.
> make a struct containing a resource.quantity please.

@deads2k resurfacing your [earlier review](https://github.com/openshift/enhancements/pull/1370#discussion_r1179654101) of the API here again. I can make the change but I'm missing the rationale here so could you please elaborate if it's a convention or requirement to wrap these as structs. Or if it helps with compatibility if we change the API later?

config/v1alpha1/types_backup.go
tjungblu: I would stick with what the batch API has, I can see you also got the TZ :)

Seems it's using: https://github.com/robfig/cron (v2)

https://github.com/kubernetes/kubernetes/blob/eca1f9d2d5ecb06b80cbd114e1ca1631fc3ff100/pkg/apis/batch/validation/validation.go#L493-L503

Given we would need to make an API call for the PVC, I would just have it as a validating webhook in CEO? wdyt?

config/v1alpha1/types_backup.go
```diff
@@ -0,0 +1,131 @@
...
+	// +optional
+	Status BackupStatus `json:"status"`
+}
+
+type BackupSpec struct {
+	// +kubebuilder:validation:Required
+	EtcdBackupSpec EtcdBackupSpec `json:"etcdBackupSpec,omitempty"`
+}
+
+type BackupStatus struct {
```
tjungblu, line 35: do you want to type the status (eg successful run at X, backup located with filename X on PVC X) here too or just have it as a unstructured json?

config/v1alpha1/types_backup.go
```diff
@@ -0,0 +1,131 @@
...
+	// +kubebuilder:validation:Optional
+	// +optional
+	RetentionPolicy RetentionPolicy `json: "retentionPolicy"`
+
+	// PVCName specifies the name of the PersistentVolumeClaim which binds a PersistentVolume where the
+	// etcd backup files would be saved
+	// +kubebuilder:validation:Required
+	// +required
+	// TODO: Cannot be empty string
+	PVCName string `json: "pvcName"`
```
tjungblu, line 71: the PVC requires a namespace, we would create it in openshift-etcd-operator? Probably worth mentioning this in the comment, too.

config/v1alpha1/types_backup.go
```diff
@@ -0,0 +1,131 @@
...
+	// RetentionPolicy defines the retention policy for retaining and deleting existing backups.
+	// +kubebuilder:validation:Optional
+	// +optional
+	RetentionPolicy RetentionPolicy `json: "retentionPolicy"`
+
+	// PVCName specifies the name of the PersistentVolumeClaim which binds a PersistentVolume where the
+	// etcd backup files would be saved
+	// +kubebuilder:validation:Required
+	// +required
+	// TODO: Cannot be empty string
```
tjungblu, line 70: IIRC there's a minLength in kubebuilder validation rules

tjungblu:
looks pretty good already, @hasbro17! thank you :)

config/v1alpha1/types_backup.go
hasbro17: Good point I forgot about PVC being namespaced. I guess the admin would need to create it in the etcd-operator namespace so that the backup executing pod can use it. I'll add a comment.

config/v1alpha1/types_backup.go
hasbro17: I initially had that in mind but it seems like there is a convention to not have `config.openshift.io` config types display or require a status block.

That’s also kind of why we have the `backup.openshift.io`  `EtcdBackup` type which tracks a single backup attempt with a status on whether each backup attempt is successful and which PVC it’s saving to. 

But if there’s anything else that’s more cumulative, perhaps we can add that to the `etcd.operator.openshift.io`'s type status.

hasbro17:
Missing the test suite for the new types. Will update that next.

config/v1alpha1/types_backup.go
hasbro17: I think handling this validation as a webhook or operator side would be better, given the complexity of the regex.
@deads2k WDYT?

config/v1alpha1/types_backup.go
```diff
@@ -0,0 +1,131 @@
...
+	Spec BackupSpec `json:"spec"`
+	// status holds observed values from the cluster. They may not be overridden.
+	// +kubebuilder:validation:Optional
+	// +optional
+	Status BackupStatus `json:"status"`
+}
+
+type BackupSpec struct {
+	// +kubebuilder:validation:Required
+	EtcdBackupSpec EtcdBackupSpec `json:"etcdBackupSpec,omitempty"`
```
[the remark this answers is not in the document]
ingvagabund: Are there any other planned backups of other components?

backup/v1alpha1/register.go
```diff
@@ -0,0 +1,38 @@
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+var (
+	GroupName     = "backup.openshift.io"
```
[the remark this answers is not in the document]
hasbro17: Nothing comes to mind right now for what other components may need to backup. The current restore procedure just needs the etcd snapshot and static pod manifests.
I'd be okay with narrowing down the group to just `etcdbackup` but would prefer to get some input from @deads2k who originally suggested an overall backup group with etcd being a subsection.
https://github.com/openshift/enhancements/pull/1370#discussion_r1179649241

backup/v1alpha1/types_etcdbackup.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+type EtcdBackupSpec struct {
+
+	// PVCName specifies the name of the PersistentVolumeClaim which binds a PersistentVolume where the
+	// etcd backup file would be saved
+	// +kubebuilder:validation:Required
+	// +required
+	// TODO: Cannot be empty string
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="pvcName is immutable once set"
```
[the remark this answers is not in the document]
hasbro17: Added unit tests for this and EtcdBackupState validations.

config/v1/feature_gates.go
```diff
@@ -252,4 +252,13 @@ var (
 		ResponsiblePerson:   "mkowalsk",
 		OwningProduct:       kubernetes,
 	}
+	FeatureGateEtcdBackupAPI = FeatureGateName("EtcdBackupAPI")
+	etcdBackupAPI            = FeatureGateDescription{
+		FeatureGateAttributes: FeatureGateAttributes{
+			Name: FeatureGateInsightsConfigAPI,
```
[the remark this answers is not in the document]
hasbro17: Good catch. Updated.

config/v1alpha1/types_backup.go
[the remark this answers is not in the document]
hasbro17: As [mentioned above](https://github.com/openshift/api/pull/1482#discussion_r1232765123) nothing planned yet for backups of other components.

I've removed the `omitempty` since it doesn't make sense for a required field I think. But did you mean that we should have this as a pointer struct? I figured those were discouraged.
https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#pointers

config/v1alpha1/types_backup.go
```diff
@@ -0,0 +1,131 @@
...
+	PVCName string `json:"pvcName"`
+}
+
+// RetentionType is the enumeration of valid retention policy types
+// +enum
+// +kubebuilder:validation:Enum:="RetentionCount";"RetentionSizeGb"
+type RetentionType string
+
+const (
+	RetentionTypeCount RetentionType = "RetentionCount"
```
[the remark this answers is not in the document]
hasbro17: Updated

config/v1alpha1/types_backup.go
```diff
@@ -0,0 +1,131 @@
...
+// This struct is a discriminated union that allows users to select the type of retention policy from the supported types.
+// +union
+type RetentionPolicy struct {
+	// RetentionType sets the type of retention policy. The currently supported and valid values are "retentionCount"
+	// Currently, the only valid policies are retention by count (RetentionCount) and by size (RetentionSizeGb). More policies or types may be added in the future.
+	// +unionDiscriminator
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum:=RetentionCount;RetentionSizeGb
+	RetentionType RetentionType `json:"retentionType"`
```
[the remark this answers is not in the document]
hasbro17: I think this might still work based off the example of the controlplanemachine set union which also uses a type definition "alias".

https://github.com/openshift/api/blob/a6fbaf36017d235ff58a51ddc3962088dabbf6c3/machine/v1/types_controlplanemachineset.go#L122-L125

https://github.com/openshift/api/blob/a6fbaf36017d235ff58a51ddc3962088dabbf6c3/machine/v1/types_controlplanemachineset.go#L109-L114

But if that's not the convention or is ambiguous, I can update to a string alias declaration.

config/v1alpha1/types_backup.go
```diff
@@ -0,0 +1,131 @@
...
+
+const (
+	RetentionTypeCount RetentionType = "RetentionCount"
+	RetentionTypeSize  RetentionType = "RetentionSizeGb"
+)
+
+// RetentionPolicy defines the retention policy for retaining and deleting existing backups.
+// This struct is a discriminated union that allows users to select the type of retention policy from the supported types.
+// +union
+type RetentionPolicy struct {
```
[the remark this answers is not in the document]
hasbro17: Updated the union members to be optional pointers to structs.

config/v1alpha1/types_backup.go
hasbro17: Nevermind, found the API convention which points this out.
https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#writing-a-union-in-go

hasbro17:
Hmm that's a new one:
```console
Error running codegen: 
error running generator schemapatch on group backup.openshift.io: 
	could not run schemapatch generator for group/version backup.openshift.io/v1alpha1: 
		could not load generation contexts: 
			could not unmarshal YAML for type meta inspection: error converting YAML to JSON: yaml: line 35: did not find expected key
```

config/v1alpha1/types_backup.go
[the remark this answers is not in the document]
JoelSpeed: > `type RetentionType string`

Looks like a string alias to me? Was this updated since the original comment?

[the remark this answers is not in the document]
hasbro17: No I haven't changed it. I think `type RetentionType string` is what @ingvagabund was originally referring to as being "type definition" instead of being a "alias declaration" but I wasn't aware of the distinction before this.
https://go.dev/ref/spec#Type_declarations

In any case, it seems like a string type works as well and seems to be in accordance with the api convention guide:

> The discriminant MUST be a string (or string alias) type

hasbro17:
So I did have malformed yaml but confusingly not where I expected it from the error:
```
error running generator schemapatch on group backup.openshift.io: could not run schemapatch generator for group/version backup.openshift.io/v1alpha1
```

That made me think the issue was with `backup/v1alpha1/0000_10_01_etcdbackup-TechPreviewNoUpgrade.crd.yaml` but it was actually in the testsuite `backup/v1alpha1/techpreview.etcdbackup.testsuite.yaml`. Should be updated now.

config/v1alpha1/types_backup.go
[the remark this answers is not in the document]
JoelSpeed: Ok, I think it's fine as it is. The bit that matters here is how it looks to the OpenAPI schema. When validations are applied, everything is unstructured, it won't know what the underlying Go type is, so whether it's a type declaration or an alias, it looks the same to the OpenAPI schema.

backup/v1alpha1/register.go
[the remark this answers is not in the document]
hasbro17: Going to move this over to the group `operator.openshift.io` as `etcdbackups.operator.openshift.io/v1alpha1` based on @deads2k's suggestion.

hasbro17:
A couple of TODOs for the next update:
- Rename the group for the lower level `EtcdBackup`CRD from `backup.openshift.io`=> `etcdbackups.operator.openshift.io`.
- Depending on how we want to allow and handle multiple one time `EtcdBackup` requests, decide if we need an additional `EtcdBackup` state  e.g `Aborted` to mark all but the latest (or a singular) `EtcdBackup` as unfulfilled.
  - One alternative is to only allow a cluster singleton `EtcdBackup` CR, which necessitates only one ongoing backup at a time. The drawback of the cluster singleton would be losing the history of previous backup requests i.e `Completed` `EtcdBackup` CR objects.

tjungblu:
@hasbro17 could you please rebase as well? it's currently impossible for me to create a cluster:
> Cluster Bot APP  
> pull request openshift/api#1482 needs to be rebased to branch master

hasbro17:
Rebased and moved out the featuregate into https://github.com/openshift/api/pull/1511
Moved the lower level `EtcdBackup` CRD from `backup.openshift.io` to `operator.openshift.io`

Still adding the cron validation and its unit tests. Also need to sort out why the schema generation isn't running again.

operator/v1alpha1/0000_10_01_etcdbackup-TechPreviewNoUpgrade.crd.yaml
```diff
@@ -0,0 +1,24 @@
...
+metadata:
+  annotations:
+    api-approved.openshift.io: https://github.com/openshift/api/pull/1482
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    release.openshift.io/feature-set: TechPreviewNoUpgrade
+  name: etcdbackups.operator.openshift.io
+spec:
+  group: backup.openshift.io
```
tjungblu, line 12:
```suggestion
  group: operator.openshift.io
```

operator/.codegen.yaml
```diff
@@ -1,2 +1,7 @@
+schemapatch:
+  requiredFeatureSets:
+  - ""
+  - "Default"
+  - "TechPreviewNoUpgrade"
 swaggerdocs:
```
tjungblu, line 6: I don't really know whether this is related to that file, but I'm missing the generation for the knownTypes in the alpha1 register:

https://github.com/hasbro17/api/blob/add-etcd-backups-api/operator/v1alpha1/register.go#L34-L38

those should also include the backup types. Which are important for the test fake clients, they currently don't find them:

> 	panic: no kind is registered for the type v1alpha1.EtcdBackup in scheme "github.com/openshift/client-go/operator/clientset/versioned/fake/register.go:15"

config/v1alpha1/types_backup.go
deads2k: Agree. We largely don't have status on these.

config/v1alpha1/types_backup.go
```diff
@@ -0,0 +1,147 @@
...
+	Spec BackupSpec `json:"spec"`
+	// status holds observed values from the cluster. They may not be overridden.
+	// +kubebuilder:validation:Optional
+	// +optional
+	Status BackupStatus `json:"status"`
+}
+
+type BackupSpec struct {
+	// +kubebuilder:validation:Required
+	EtcdBackupSpec EtcdBackupSpec `json:"etcdBackupSpec"`
```
deads2k, line 32:
```suggestion
	EtcdBackupSpec EtcdBackupSpec `json:"etcd"`
```

deads2k, line 32: needs godoc

config/v1alpha1/types_backup.go
deads2k: I'd use whatever off-the-shelf regex you find that seems close and then have your operator go degraded when this value isn't legal.

config/v1alpha1/types_backup.go
```diff
@@ -0,0 +1,147 @@
...
+	// +kubebuilder:validation:Optional
+	// +optional
+	Schedule string `json:"schedule"`
+
+	// The time zone name for the given schedule, see https://en.wikipedia.org/wiki/List_of_tz_database_time_zones.
+	// If not specified, this will default to the time zone of the kube-controller-manager process.
+	// See https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#time-zones
+	// +kubebuilder:validation:Optional
+	// +optional
+	// TODO: CEL validation for timezone format
```
deads2k, line 58: Yeah, something.

config/v1alpha1/types_backup.go
```diff
@@ -0,0 +1,147 @@
...
+	PVCName string `json:"pvcName"`
+}
+
+// RetentionType is the enumeration of valid retention policy types
+// +enum
+// +kubebuilder:validation:Enum:="RetentionCount";"RetentionSize"
+type RetentionType string
+
+const (
+	// RetentionTypeCount sets the retention policy based on the count or number of backup files saved
```
deads2k, line 80: is it the count or the number?

config/v1alpha1/types_backup.go
```diff
@@ -0,0 +1,147 @@
...
+// This struct is a discriminated union that allows users to select the type of retention policy from the supported types.
+// +union
+type RetentionPolicy struct {
+	// RetentionType sets the type of retention policy. The currently supported and valid values are "retentionCount"
+	// Currently, the only valid policies are retention by count (RetentionCount) and by size (RetentionSize). More policies or types may be added in the future.
+	// +unionDiscriminator
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum:=RetentionCount;RetentionSize
+	// TODO: How to provide a "NoRetentionPolicy" option as a default? Should that be another enum
```
deads2k, line 96: just allow empty

config/v1alpha1/types_backup.go
```diff
@@ -0,0 +1,147 @@
...
+// RetentionPolicy defines the retention policy for retaining and deleting existing backups.
+// This struct is a discriminated union that allows users to select the type of retention policy from the supported types.
+// +union
+type RetentionPolicy struct {
+	// RetentionType sets the type of retention policy. The currently supported and valid values are "retentionCount"
+	// Currently, the only valid policies are retention by count (RetentionCount) and by size (RetentionSize). More policies or types may be added in the future.
+	// +unionDiscriminator
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum:=RetentionCount;RetentionSize
```
deads2k, line 95: allow empty string

config/v1alpha1/types_backup.go
```diff
@@ -0,0 +1,147 @@
...
+
+// RetentionSizeConfig specifies the configuration of the retention policy on the total size of backups
+type RetentionSizeConfig struct {
+	// MaxSizeOfBackupsMb defines the total size in Mb of backups to retain.
+	// If the current total size backups exceeds MaxSizeOfBackupsMb then
+	// the oldest backup will be removed before a new backup is initiated.
+	// +kubebuilder:validation:Minimum=100
+	// +kubebuilder:validation:Required
+	// +required
+	MaxSizeOfBackupsMb int `json:"maxSizeOfBackupsMb,omitempty"`
```
deads2k, line 131: no.  GB

operator/v1alpha1/types_etcdbackup.go
```diff
@@ -0,0 +1,102 @@
...
+	// +kubebuilder:validation:Required
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="pvcName is immutable once set"
+	PVCName string `json:"pvcName"`
+}
+
+const (
+	// Etcd backup is running
+	Running EtcdBackupState = "Running"
```
deads2k, line 42: consistent with pods, this is Phase and it is discouraged.  We found that the state machine ended up with too many tuples and conditions worked better.  conditions also work with `oc wait`.

operator/v1alpha1/types_etcdbackup.go
```diff
@@ -0,0 +1,102 @@
...
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	// +optional
+	Conditions []metav1.Condition `json:"conditions" patchStrategy:"merge" patchMergeKey:"type"`
+
+	// etcdBackupState reflects the current state of the etcd backup process.
+	// +kubebuilder:validation:Optional
+	// +optional
+	State EtcdBackupState `json:"etcdBackupState,omitempty"`
```
deads2k, line 77: Remove.  use a well known condition instead.  Perhaps Complete, Succeeded, and Failed.

operator/v1alpha1/types_etcdbackup.go
```diff
@@ -0,0 +1,102 @@
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="startTime is immutable once set"
+	// +optional
+	StartTime metav1.Time `json:"startTime,omitempty"`
+
+	// finishTime is the time when the etcd backup finished.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="finishTime is immutable once set"
+	// +optional
+	FinishTime metav1.Time `json:"finishTime,omitempty"`
```
deads2k, line 89: pointer

operator/v1alpha1/types_etcdbackup.go
```diff
@@ -0,0 +1,102 @@
...
+	// etcdBackupState reflects the current state of the etcd backup process.
+	// +kubebuilder:validation:Optional
+	// +optional
+	State EtcdBackupState `json:"etcdBackupState,omitempty"`
+
+	// startTime is the time when the etcd backup started.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="startTime is immutable once set"
+	// +optional
+	StartTime metav1.Time `json:"startTime,omitempty"`
```
deads2k, line 83: pointer

operator/v1alpha1/types_etcdbackup.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+type EtcdBackupSpec struct {
+	// PVCName specifies the name of the PersistentVolumeClaim which binds a PersistentVolume where the
+	// etcd backup file would be saved
+	// +kubebuilder:validation:Required
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="pvcName is immutable once set"
+	PVCName string `json:"pvcName"`
```
deads2k, line 37: Is this actually configurable in some sense? A cluster-admin shouldn't be allowed to create PVCs in your namespace, right?

operator/v1alpha1/types_etcdbackup.go
```diff
@@ -0,0 +1,102 @@
...
+// +kubebuilder:validation:XValidation:rule="!(oldSelf == 'Failed' && self == 'Pending')", message="etcdBackupState cannot transition from Failed to Pending"
+// +kubebuilder:validation:XValidation:rule="!(oldSelf == 'Completed' && self == 'Running')", message="etcdBackupState cannot transition from Completed to Running"
+// +kubebuilder:validation:XValidation:rule="!(oldSelf == 'Failed' && self == 'Running')", message="etcdBackupState cannot transition from Failed to Running"
+type EtcdBackupState string
+
+// +kubebuilder:validation:XValidation:rule="(!has(oldSelf.startTime) || has(self.startTime))",message="cannot remove startTime attribute from status"
+// +kubebuilder:validation:XValidation:rule="(!has(oldSelf.finishTime) || has(self.finishTime))",message="cannot remove finishTime attribute from status"
+// +kubebuilder:validation:XValidation:rule="(!has(oldSelf.etcdBackupState) || has(self.etcdBackupState))",message="cannot remove etcdBackupState attribute from status"
+// +kubebuilder:validation:Optional
+type EtcdBackupStatus struct {
```
deads2k, line 65: Do you need a link to the pod or job running this backup?

operator/v1alpha1/types_etcdbackup.go
tjungblu: I would entirely remove StartTime and FinishTime. It's possible to reconcile this from the job status conditions, but is there a reason we need this?

operator/v1alpha1/types_etcdbackup.go
tjungblu: in the current implementation I added a "back pointer" label on the job to the backup CRD name. The other direction is a status condition. Do you think it makes sense to have this as a typed field?

operator/.codegen.yaml
tjungblu: that's what it probably should look like to pass the tests, not sure if this would be generated

```
func addKnownTypes(scheme *runtime.Scheme) error {
	metav1.AddToGroupVersion(scheme, GroupVersion)

	scheme.AddKnownTypes(GroupVersion,
		&GenericOperatorConfig{},
		&ImageContentSourcePolicy{},
		&ImageContentSourcePolicyList{},
		&EtcdBackup{},
		&EtcdBackupList{},
	)

	return nil
}

```

hasbro17: Good catch, forgot to register the actual types under the group. Still working on the other updates but pushing this change in the meantime. Hopefully they show up now with the fake clients.

config/v1alpha1/types_backup.go
hasbro17: I was using them interchangeably but updating the type and related comments to be `RetentionTypeNumber` as this policy would be to prune based on the number of completed backups.

config/v1alpha1/types_backup.go
hasbro17: Just to clarify, since the discriminant is always required, that can't be empty, so I'm assuming you mean just allowing empty values for the union members right?
https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#discriminated-unions
> - The discriminant should be required
> - Empty union members (discriminant values without a paired union member) are also permitted

So we would interpret the following configs as "NoRetentionPolicy":
```yaml
spec:
  etcd:
    retentionPolicy:
      retentionType: RetentionNumber
      # No member specified
```
```yaml
spec:
  etcd:
    retentionPolicy:
      retentionType: RetentionSize
      # No member specified
```

The `retentionPolicy` field is also optional itself so leaving that empty would also be valid:
```yaml
spec:
  etcd:
    schedule: * 2 * * *
    # No retentionPolicy specified
```
Is that what you meant?

hasbro17: Ah nevermind, just saw [your comment](https://github.com/openshift/api/pull/1482#discussion_r1245699431) below for allowing an empty string for the discriminant.

operator/v1alpha1/types_etcdbackup.go
hasbro17: Going to remove the EtcdBackupState and add as conditions instead.

operator/v1alpha1/types_etcdbackup.go
hasbro17: The start and finish times aren't strictly necessary. This was just my attempt to provide more context in the status, but on second thought this info can be gathered from the condition LastTransitionTime as well.
Will remove them in that case.

operator/v1alpha1/types_etcdbackup.go
hasbro17: Forgot to mention that this PVC should be in the "cluster-etcd-operator" namespace.
And yes I'm assuming that the cluster-admin does have the permissions to create PVCs in our namespace (or any namespace I think).

But please clarify if that's not the case, or if it's a bad idea to allow PVCs in the etcd-operator namespace.

Or did you mean that we should allow the admin to choose their own namespace for the PVC so they don't have to use the "cluster-etcd-operator" namespace?

operator/v1alpha1/types_etcdbackup.go
hasbro17: @tjungblu For the "back pointer" from the Job=>Backup CRD we can set the `metadata.ownerReferences` field so that the Backup CRD is identified as the parent on the Job's metadata.

For the Backup CRD=>Job direction, I was also thinking that the Job name would be visible in the status condition once it is created or pending.
But @deads2k let us know if you think a typed status field would be better.

operator/v1alpha1/types_etcdbackup.go
hasbro17: Actually looking around I don't see many other APIs that have their conditions defined in their respective types.go.
E.g for the controlplanemachineset API, they're defined in the operator repo:
https://github.com/openshift/cluster-control-plane-machine-set-operator/blob/main/pkg/controllers/controlplanemachineset/consts.go

At the risk of over/under designing the conditions, I'll hold off from defining them here and we can add add them with the controllers in cluster-etcd-operator.
Once we have the controller implementation PR ready that should give us a better idea of whether we need something more or less than the default `Complete, Succeeded, Failed, Skipped` conditions.

@tjungblu I can see you already have some different conditions defined so let me know if that works better.

hasbro17:
Updated the following for the latest review comments:
- Removed `EtcdBackupState` from `EtcdBackup` status in favor of defining conditions in the operator
- Removed `StartTime`  and `FinishTime`  from `EtcdBackup` status
- Other renaming nits for the union members in the `Backup` config.
- Added CEL validation regexes for the Cron and timezone formats.
  - The cron regex was pared down from the extended version to fit the POSIX variant used by the cronjob. See https://regexr.com/7gcfo for unit tests
  - I had to construct the timezone regex from scratch but it's passing for all the variants in the tz timezone database that I've tested. See https://regexr.com/7gedh for unit tests.

TODO: Translating the online regex validator unit tests into the API testsuite unit tests.

operator/v1alpha1/types_etcdbackup.go
tjungblu: the backup job must run in the `openshift-etcd` namespace: https://github.com/openshift/cluster-etcd-operator/blob/master/bindata/etcd/cluster-backup-pod.yaml#L5

so the PVC would also need to go there. Otherwise I assume we need to create it?

operator/v1alpha1/types_etcdbackup.go
tjungblu: great idea, will update the other controller accordingly

operator/v1alpha1/types_etcdbackup.go
tjungblu: feel free to define them properly, I mostly copied the conditions from the upgradebackup controller

operator/v1alpha1/types_etcdbackup.go
deads2k: Sharing our operator or operand namespace with a user strikes me as unusual and rules like immutability seem even stranger.  What does letting a customer provide their PVC allow us to do?

Could they provide a PV (cluster scoped) and we provide the PVC?

hasbro17: So my thought process on why I have the admin providing their PVC here:
- The admin is expected to provide their own PV (or storage class for dynamic provisioning) so they have control over how to provision the storage
- The admin is also expected to create their own PVC so they can choose the pvc spec: storage class, access modes, resource requests, etc
- The operator can’t own the PVC lifecycle, since deleting the PVC also triggers the deletion or reclamation of the PV.
  - Plus if the PV is the same for every subsequent `EtcdBackup` request, the operator would have to reuse the same PVC, since you can’t bind different PVCs to a single PV. But if it’s a different PV name, or a dynamic provisioning storage class then the operator has to use a new PVC.
  - So for multiple reasons the PVC lifecycle seems best left to the admin.
- Since the backup execution job/pod has to reference the PVC in its own namespace that means the PVC has to be in the namespace of our choosing e.g `openshift-etcd`. So that’s not configurable.
- So the expected workflow then is to allow the admin to create their PVC and then point us to that in the `EtcdBackup` spec 
- The reason I have `spec.pvcName` as immutable is because the `EtcdBackup` CR represents a one-time backup execution, and changing the `pvcName` after the CR creation from `pvc1` to `pvc2` means we would have to redo an in progress or completed backup and save it again to wherever `pvc2` is bound.  And that goes against the “1 `EtcdBackup` == 1 backup attempt” design. 

So that’s why I have the `pvcName` as configurable on creation but immutable after.
But happy to hear different ideas on if there's something simpler or I’m missing something.

If sharing our namespace with the admin is an absolute no-go then I need to rethink how we could get the pvc and our backup pod in the same namespace without us owning the PVC lifecycle.

hasbro17: The backup pod running the backup script [needs to run as privileged](https://github.com/openshift/cluster-etcd-operator/blob/c7cc2709d2d5b81f55a72a8ddd839ef0115eac96/bindata/etcd/cluster-backup-pod.yaml#L44C5-L45C23) so it can grab the static pod manifests and certs from the host filesystem.

We could run it outside `openshift-etcd`, in a namespace of the admin's choice, provided it's setup to allow privileged pods, i.e the ns has the following labels like `openshift-etcd` does:
```yaml
pod-security.kubernetes.io/audit: privileged
pod-security.kubernetes.io/enforce: privileged
pod-security.kubernetes.io/warn: privileged
```
So that way the admin provides us with the name and namespace of their PVC in the `EtcdBackup` spec and we can use that to execute and save the backup.

@deads2k WDYT? Would that address the namespace sharing concern?

operator/v1alpha1/types_etcdbackup.go
hasbro17: I can add a typed status field to reference the Job associated with the EtcdBackup. Name or better yet some kind of object reference, like the owner ref has it.

hasbro17:
Made the following update:
- Added the integration tests for the regex validations for the schedule and timezone formats (waiting to see that those go green)
- Added a Job reference field `status.backupJob` in the EtcdBackup status so we can reference the backup job associated with that CR
- Added basic conditions for the EtcdBackup status: Pending, Completed, Failed, Skipped.

Only last thing remaining is finalizing whether we want the admin's PVC to be created in our "openshift-etcd" namespace or in a separate one https://github.com/openshift/api/pull/1482#discussion_r1245704258

Besides that, this should be good for a final review.

config/v1alpha1/types_backup.go
```diff
@@ -0,0 +1,164 @@
...
+type BackupStatus struct {
+}
+
+// EtcdBackupSpec provides configuration for automated etcd backups to the cluster-etcd-operator
+type EtcdBackupSpec struct {
+
+	// Schedule defines the recurring backup schedule in Cron format
+	// every 2 hours: 0 */2 * * *
+	// every day at 3am: 0 3 * * *
+	// Setting to an empty string "" means disabling scheduled backups
```
deads2k, line 45:
```suggestion
	// Empty string means no opinion and the platform is left to choose a reasonable default which is subject to change without notice.
	// The current default is "no backups", but will change in the future.
```

config/v1alpha1/types_backup.go
```diff
@@ -0,0 +1,164 @@
...
+	// +kubebuilder:validation:Optional
+	// +optional
+	RetentionPolicy RetentionPolicy `json:"retentionPolicy"`
+
+	// PVCName specifies the name of the PersistentVolumeClaim which binds a PersistentVolume where the
+	// etcd backup files would be saved
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	// +required
+	PVCName string `json:"pvcName"`
```
deads2k, line 90: Allow "" and have it mean a rotating local volume, with one local volume per master.  This will allow a "do nothing" cluster to have automated backups with each backup on a different master to distribute risk.

config/v1alpha1/types_backup.go
```diff
@@ -0,0 +1,164 @@
...
+	RetentionTypeSize RetentionType = "RetentionSize"
+)
+
+// RetentionPolicy defines the retention policy for retaining and deleting existing backups.
+// This struct is a discriminated union that allows users to select the type of retention policy from the supported types.
+// +union
+type RetentionPolicy struct {
+	// RetentionType sets the type of retention policy.
+	// Currently, the only valid policies are retention by number of backups (RetentionNumber), by the size of backups (RetentionSize). More policies or types may be added in the future.
+	// Specifying an empty string "" means no retention policy and no backups will be pruned.
```
deads2k, line 111:
```suggestion
	// Empty string means no opinion and the platform is left to choose a reasonable default which is subject to change without notice
	// The current default is RetentionNumber with 15 backups kept.
```

operator/v1alpha1/types_etcdbackup.go
deads2k: I had to go back and read the docs. Because of the double-opt-in for PV and PVC (justified decision), I'm convinced on the acceptability of requiring a PVC, **but** we need a zero-config option that will take backups and store them locally on the masters. I've suggested a mechanism in a comment above to handle this.

operator/v1alpha1/types_etcdbackup.go
```diff
@@ -0,0 +1,96 @@
...
+	BackupJob BackupJobReference `json:"backupJob"`
+}
+
+// BackupJobReference holds a reference to the batch/v1 Job created to run the etcd backup
+type BackupJobReference struct {
+
+	// namespace is the namespace of the Job.
+	// Required
+	// +kubebuilder:validation:Required
+	Namespace string `json:"namespace"`
```
deads2k, line 62: I'm in favor of this field, but we can require it to be `openshift-etcd`, right?

operator/v1alpha1/types_etcdbackup.go
```diff
@@ -0,0 +1,96 @@
...
+	// +required
+	Spec EtcdBackupSpec `json:"spec"`
+	// status holds observed values from the cluster. They may not be overridden.
+	// +kubebuilder:validation:Optional
+	// +optional
+	Status EtcdBackupStatus `json:"status"`
+}
+
+type EtcdBackupSpec struct {
+	// PVCName specifies the name of the PersistentVolumeClaim which binds a PersistentVolume where the
```
deads2k, line 31: indicate which namespace it must be in.

operator/v1alpha1/types_etcdbackup.go
```diff
@@ -0,0 +1,96 @@
...
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	// +optional
+	Conditions []metav1.Condition `json:"conditions" patchStrategy:"merge" patchMergeKey:"type"`
+
+	// backupJob is the reference to the Job that executes the backup.
+	// Optional
+	// +kubebuilder:validation:Optional
+	BackupJob BackupJobReference `json:"backupJob"`
```
deads2k, line 53:
```suggestion
	BackupJob *BackupJobReference `json:"backupJob"`
```

This is more like a workload API than a configuration API.  This doesn't always exist, so it should be a pointer.

config/v1alpha1/types_backup.go
```diff
@@ -0,0 +1,164 @@
...
+	//   (\*|(?:[0-6]|SUN|MON|TUE|WED|THU|FRI|SAT)(?:(?:\-(?:[0-6]|SUN|MON|TUE|WED|THU|FRI|SAT))?|(?:\,(?:[0-6]|SUN|MON|TUE|WED|THU|FRI|SAT))*))
+	//
+
+	// The time zone name for the given schedule, see https://en.wikipedia.org/wiki/List_of_tz_database_time_zones.
+	// If not specified, this will default to the time zone of the kube-controller-manager process.
+	// See https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#time-zones
+	// +kubebuilder:validation:Optional
+	// +optional
+	// +kubebuilder:validation:Pattern:=`^([A-Za-z_]+([+-]*0)*|[A-Za-z_]+(\/[A-Za-z_]+){1,2})(\/GMT[+-]\d{1,2})?$`
+	TimeZone string `json:"timeZone,omitempty"`
```
deads2k, line 72: don't omitempty.  Configuration APIs should list the options available on the "empty" object.

deads2k:
A few comments, but not many.  I think the hardest remaining thing is leaving it possible for a no-opinion zero-value resource to indicate: "take backups stored on a different master every four hours".  We can argue over frequency, but I'll open there.

We don't need that no-opinion behavior now, but we should plan to have it ready a release after we make this feature accessible-by-default.

config/v1alpha1/types_backup.go
tjungblu: I think most people running backups already have a SAN/NAS in place which they would mount via NFS for such purposes. Maybe the "" default would trigger it the way it would be when you run the backup pod today, just on plain host mounts. 

Besides making this somewhat difficult to implement retention, how would you implement the rotation? Have a different node selector every time the backup triggers?

operator/v1alpha1/types_etcdbackup.go
hasbro17: Given that we're okay with requiring the PVC to always be in the `openshift-etcd` namespace, yes this will always be set as `openshift-etcd`.
Did you mean to imply that's redundant and we should remove this field, or instead have validation to ensure this is always set to `openshift-etcd`?
I've added the latter but please correct me if you meant otherwise.

config/v1alpha1/types_backup.go
hasbro17: Good question. I had that issue pointed out in the initial proposal which is why we started with the simpler PVC approach.
Since we're targeting the no-opinion local backup option as a follow up, I will do a follow up on the enhancement to try and sketch out how we could do local backups and the associated scheduling and retention challenges.

hasbro17:
@deads2k Updated for review comments. All except for allowing the no-option `pvcName: ""` which we'll target as a follow up. (Unless you meant to add it to the API first).

config/v1alpha1/types_backup.go
deads2k: > Besides making this somewhat difficult to implement retention, how would you implement the rotation? Have a different node selector every time the backup triggers?

You have a history of objects in the etcdbackup.operator.openshift.io.  You can order that history, read the jobs, and force-assign to the next master alphabetically.  If there is no history, randomly choose. You can force assign the job in the podtemplatespec.

For retention, i think a similar approach is possible by simply dividing the count (that's why count over size) and having the job itself delete the old backups (I'm betting this is your default flow).

config/v1alpha1/types_backup.go
deads2k: This comment is still outstanding.  The default can change and we should plan on that happening

hasbro17:
Ah, forgot to update the unit tests after making the pvcName field optional. Will update that.

operator/v1alpha1/types_etcdbackup.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+type EtcdBackupSpec struct {
+	// PVCName specifies the name of the PersistentVolumeClaim (PVC) which binds a PersistentVolume where the
+	// etcd backup file would be saved
+	// The PVC itself must always be created in the "openshift-etcd" namespace
+	// If the PVC is left unspecified "" then the platform will choose a reasonable default location to save the backup.
+	// In the future this would be backups saved across the control-plane master nodes.
+	// +kubebuilder:validation:Optional
+	// +optional
```
deads2k, line 37: Sorry, caught this in the unit test. The config.openshift.io API must accept an empty PVCName, but this object needs one so that the pods are assigned well, right?

hasbro17: Depends on where we want to handle the no-option local backup signal.

My rough idea is to handle the `pvcName: ""` at the `operator.openshift.io EtcdBackup` object for each request.
That would signal to the etcdbackup controller to pick a PVC from its own set of pre-created PVCs that are bound to the local PVs (one for each node).
How the node selectors (or rather picking from the PVs assigned to each node) is decided based on the history of previous backups and existing spread.

But essentially in this model, the no-option `pvcName: ""` from the `config.openshift.io Backup` object is just carried over to the `operator.openshift.io EtcdBackup` object, which is interpreted as the default local backup option by the lower level controller.

So this object also needs to accept an empty `pvcName` in that case.

PR #1484, opened 2023-06-05
tls/MCO: Describe how `machine-config-server-tls` works

cgwalters:
I didn't write this code, but some quick code searches reveals that what is formerly called "mco mystery cert" here is really the "MCS certificate" - it's generated by openshift/installer today, and then just passed into the cluster and used by the MCO.

Agreed with the implication here that "root-ca" is a terrible name for that CA, and it should probably be moved into the `openshift-machine-config-operator` namespace in the cluster and renamed.

We also really need to add day 2 rotation for this.


deads2k:
This file is generated by https://github.com/openshift/cluster-debug-tools  Glad to see details coming here.

cgwalters:
xref https://github.com/openshift/installer/pull/7232

PR #1493, opened 2023-06-14
AUTH-372: security/v1: add required SCC annotation

s-urbaniak:
This adds the required SCC annotation so it can be used in https://github.com/openshift/apiserver-library-go/pull/109.

/cc @deads2k @stlaz


soltysh (approved):
/approve

PR #1494, opened 2023-06-14
OCPBUGS-14716: Add ROSA branding option in console-operator config

jhadvig:
/assign @spadgett 

@rhamilto FYI


rhamilto (approved):
/lgtm

JoelSpeed (commented):
/lgtm
Thanks for updating, we should make sure documentation and other sources are updated to use the PascalCase versions, old versions are left for supporting upgrades

PR #1501, opened 2023-06-21
introduce vSphere static IP feature gate

rvanderp3:
Introduces a new feature gate included with the TechPreviewNoUpgrade featureset which gates the enablement of static IP support for vSphere IPI.


rvanderp3:
combining with api#1338

PR #1503, opened 2023-06-21
ODC-7332: New ConsoleSample CRD

jerolimov:
**Fixes:**
Epic: [ODC-7241: Include Serverless Function Samples in our Sample Catalog](https://issues.redhat.com/browse/ODC-7241) 
Story: [ODC-7332: Create new ConsoleSample CRD in the API](https://issues.redhat.com/browse/ODC-7332)

**Related issues:**
* https://github.com/openshift/enhancements/pull/1429
  * Rendered preview: https://github.com/jerolimov/openshift-enhancements/blob/console-samples/enhancements/console/samples.md

**Changes**
Added a new Go type and generated a new CRD:

* https://github.com/jerolimov/api/blob/console-samples/console/v1/types_console_sample.go
* https://github.com/jerolimov/api/blob/console-samples/console/v1/0000_10_consolesample.crd.yaml

To test the CRD you can apply it with this command to your cluster:

```
oc apply -f https://raw.githubusercontent.com/jerolimov/api/console-samples/console/v1/0000_10_consolesample.crd.yaml
```

Some ConsoleSamples could be found here: https://gist.github.com/jerolimov/2e557df4ea95961dd64a5c32268daadf

```
oc apply -f https://gist.githubusercontent.com/jerolimov/2e557df4ea95961dd64a5c32268daadf/raw/1ab9d62e1ffb616e5c925ea0e731db7df0b897e4/nodeinfo-console-samples.yaml
oc apply -f https://gist.githubusercontent.com/jerolimov/2e557df4ea95961dd64a5c32268daadf/raw/1ab9d62e1ffb616e5c925ea0e731db7df0b897e4/knative-console-samples.yaml
```


console/v1/types_console_sample.go
```diff
@@ -0,0 +1,166 @@
...
+type ConsoleSample struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// metadata is the standard object's metadata.
+	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
+	metav1.ObjectMeta `json:"metadata"`
+
+	// +kubebuilder:validation:Required
+	// +required
+	Spec ConsoleSampleSpec `json:"spec"`
```
JoelSpeed, line 22: No status?

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,166 @@
...
+// ConsoleSampleSpec is the desired sample for the web console.
+// Samples will appear with their title, descriptions and a badge in a samples catalog.
+type ConsoleSampleSpec struct {
+	// title is the display name of the sample.
+	//
+	// It is required and must be no more than 50 characters in length.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=50
+	// +required
```
JoelSpeed, line 34: Just use `+kubebuilder:validation:Required`, no need for both

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,166 @@
...
+	Title string `json:"title"`
+
+	// abstract is a one-liner that describes the sample.
```
JoelSpeed, line 37:
```suggestion
	// abstract is a one line description that describes the sample.
```
Should also explain the max length in prose. Is 100 chars enough?

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,166 @@
...
+	Description string `json:"description"`
+
+	// icon is an optional base64 encoded image that will be displayed beside the sample title.
+	// The icon should be an vector image for easy scaling.
+	// It will be shown on a white background (also when the dark theme is used)
+	// and in different sizes (currently not bigger then 40x100px).
+	// Most icons will be quadratic but the web console ensures that all aspect ratios work.
+	// +kubebuilder:validation:Optional
+	// +optional
+	Icon string `json:"icon,omitempty"`
```
JoelSpeed, line 61: Maybe []byte rather than string for base64 encoded? This would match the secret API

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,166 @@
...
+	// so that it can contain line breaks, links, and other simple formatting.
+	// +kubebuilder:validation:Required
+	// +required
+	Description string `json:"description"`
+
+	// icon is an optional base64 encoded image that will be displayed beside the sample title.
+	// The icon should be an vector image for easy scaling.
+	// It will be shown on a white background (also when the dark theme is used)
+	// and in different sizes (currently not bigger then 40x100px).
+	// Most icons will be quadratic but the web console ensures that all aspect ratios work.
```
JoelSpeed, line 58: Not sure i follow this, can you expand?

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,166 @@
...
+	// +kubebuilder:validation:Optional
+	// +optional
+	Icon string `json:"icon,omitempty"`
+
+	// provider like "Red Hat" is optional and shown with the prefix "Provided by ".
```
JoelSpeed, line 63: Rather than starting with an example, would be better to explain what it is, and then provide examples

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,166 @@
...
+// +kubebuilder:validation:XValidation:rule="self.type == 'ContainerImport' ? has(self.containerImport) : !has(self.containerImport)",message="source.containerImport is required when source.type is ContainerImport, and forbidden otherwise"
+type ConsoleSampleSource struct {
+	// type of the sample, currently supported: "GitImport";"ContainerImport"
+	// +unionDiscriminator
+	// +kubebuilder:validation:Enum:="GitImport";"ContainerImport"
+	// +kubebuilder:validation:Required
+	Type ConsoleSampleSourceType `json:"type"`
+
+	// gitimport let the user import code from a git repository.
+	// +unionMember,optional
```
JoelSpeed, line 123: This has a different semantic to, please swap
```suggestion
	// +unionMember
	// +optional
```

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,166 @@
...
+	GitImport *ConsoleGitImportSample `json:"gitImport,omitempty"`
+
+	// containerimport let the user import a container image.
+	// +unionMember,optional
+	ContainerImport *ConsoleContainerImportSample `json:"containerImport,omitempty"`
+}
+
+// ConsoleGitImportSample let the user import code from a public git repository.
+type ConsoleGitImportSample struct {
+	// Git repository URL.
```
JoelSpeed, line 133: For other URLs we use a kubebuilder pattern to check it starts with `https://`, can we add here too?

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,166 @@
...
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	// +required
+	URL string `json:"url"`
+	// Optional branch, tag, or commit.
+	// +kubebuilder:validation:Optional
```
JoelSpeed, line 139: Only need `+optional`

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,166 @@
...
+	// +optional
+	Ref string `json:"ref"`
+}
+
+// ConsoleContainerImportSample let the user import a container image.
+type ConsoleContainerImportSample struct {
+	// Reference to a container image like openshift/hello-openshift,
+	// docker.io/openshift/hello-openshift or quay.io/<username>/<image-name>
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
```
JoelSpeed, line 149: Can we put in some arbitrarily high maximum just in case, eg 2048?

JoelSpeed (commented):
Structure wise I think this is looking really good. I'd like to work a bit on the validations and comments/godoc explaining the fields, left a bunch of comments in line, lets clear those and I'll take another pass

console/v1/types_console_sample.go
jerolimov: No status. No body other than the web console consumes or update this at the moment.

console/v1/0000_10_consolesample.crd.yaml
```diff
@@ -0,0 +1,121 @@
...
+                      properties:
+                        ref:
+                          description: Optional branch, tag, or commit.
+                          type: string
+                        url:
+                          description: Git repository URL.
+                          type: string
+                          minLength: 1
+                    type:
+                      description: 'type of the sample, currently supported: "GitImport";"ContainerImport"'
```
lance, line 96: Would it be possible to have a third option, something like `ProcessImport` so that templates could be generated dynamically? The `func` CLI - which is included as a binary in the Serverless distribution generates samples directly that are built into the binary. For example, `kn func create -l go -t cloudevents -p /source/fn` will create a new Golang function project that handles cloudevents in the `/source/fn` directory.

There are at least a couple of benefits to this approach. First, we are guaranteed to create a function that we know is compatible with the installed Serverless version. Also, this ensures that templates are kept up to date with library upgrades, etc. as this is already managed in the knative/func repository. Having duplicate templates in a separate examples repository would create additional maintenance work that duplicates that in the func repo.

console/v1/types_console_sample.go
jerolimov: Updated

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,166 @@
...
+	//
+	// It is optional and must be no more than 20 characters in length.
+	//
+	// Recommendation is a singular term like "Builder Image", "Devfile" or "Serverless Function".
+	//
+	// Currently, the badge is displayed on sample card tile in the top right corner.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MaxLength=20
+	// +optional
+	Badge string `json:"badge"`
```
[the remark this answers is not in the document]
jerolimov: Updated and changed to `type` to make it more clear. See also https://github.com/openshift/enhancements/pull/1429#discussion_r1251598999

console/v1/types_console_sample.go
jerolimov: Removed all `+kubebuilder:validation:Optional` and `+kubebuilder:validation:Required`

console/v1/types_console_sample.go
jerolimov: Because Git URLs can also start with ssh:// and git://. I would avoid this. I checked the BuildConfig CRD and the Shipwright.io Build CRD, and both didn't specify a pattern for their URL property. Let me know if you still think it's required.

See also
* https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols
* https://stackoverflow.com/questions/31801271/what-are-the-supported-git-url-formats

console/v1/types_console_sample.go
jerolimov: Added a MaxLength for the `source.git.url` and `source.image`.

console/v1/types_console_sample.go
jerolimov: Updated to:

4. `icon` is an optional base64 encoded image and shown beside the sample title.
   The format must follow the [data: URL format](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs).

   `data:[<mediatype>][;base64],<base64 encoded image>`

   For example:

   `data:image;base64,` plus the base64 encoded SVG image.

   Vector images can also be used. SVG icons must start with:

   `data:image/svg+xml;base64,` plus the base64 encoded SVG image.

   All sample catalog icons will be shown on a white background (also when the dark theme is used).
   The web console ensures that different aspect radios work correctly.
   Currently, the surface of the icon is at most 40x100px.

Wdyt?

console/v1/0000_10_consolesample.crd.yaml
jerolimov: I don't see a straightforward option to run a cli command from the web console. Esp. when the cli creates some files. In theory, we could start a Pod with a volume. But I don't get how this will help the user to understand how the console can import, build and deploy a kn func.

For more complex scenarios, it might be an option to use OpenShift Templates.

But the idea of our (console) samples is to explain to the users how the console works and, in our case, how the "Import a Serverless function" feature works.

jerolimov: We (the web console team) don't run `git clone` our own.

We "do this" by creating OpenShift `BuildConfigs` or Tekton `Pipelines` (and in the future hopefully Shipwright `Builds` for the user. They of course clone the repo, build a container image and push it into our internal image registry.

The web console also creates Deployments or a Knative `Service` to start these images for you.

Of course, we could create resources, that run `kn func && kn deploy` (as Pod, or Job) to archive something similar. But that's really different from the samples we have already.

console/v1/types_console_sample.go
JoelSpeed: `+kubebuilder:validation:Required` is the correct form, it's `+required` that should be removed

console/v1/types_console_sample.go
JoelSpeed: Yeah that's a fair argument, I'd like to see some examples of the different types of URLs within the godoc please so that others don't ask this question again

console/v1/types_console_sample.go
JoelSpeed: Yeah I think that's good. Seems like there's a fixed format here, can we add a regex to validate it at all?

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,183 @@
...
+	// +required
+	// +kubebuilder:validation:MaxLength=100
+	Abstract string `json:"abstract"`
+
+	// description of the sample.
+	//
+	// It is required and can have a maximum length of **4096** characters.
+	//
+	// It is a README.md-like content for additional information, links, pre-conditions, and other instructions.
+	// It will be rendered as Markdown so that it can contain line breaks, links, and other simple formatting.
```
JoelSpeed, line 50:
```suggestion
	// description is a long form explanation of the sample.
	//
	// It is required and can have a maximum length of **4096** characters.
	//
	// It is a README.md-like content for additional information, links, pre-conditions, and other instructions.
	// It will be rendered as Markdown so that it can contain line breaks, links, and other simple formatting.
```

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,183 @@
...
+	// +required
+	// +kubebuilder:validation:MaxLength=4096
+	Description string `json:"description"`
+
+	// icon is an optional base64 encoded image and shown beside the sample title.
```
JoelSpeed, line 55: Have you tried to `oc describe` this to see if it renders correctly?

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,183 @@
...
+	//
+	//   data:image/svg+xml;base64,     plus the base64 encoded SVG image.
+	//
+	// All sample catalog icons will be shown on a white background (also when the dark theme is used).
+	// The web console ensures that different aspect radios work correctly.
+	// Currently, the surface of the icon is at most 40x100px.
+	//
+	// [data: URL format]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs
+	// +optional
+	// +kubebuilder:validation:Pattern=`^data:`
```
JoelSpeed, line 75: The pattern should always include `base64,` right? Can we update the regex?
I had a play with `^data:([a-z/\.+0-9]*;(([-a-zA-Z0-9=])*;)?)?base64,` and think that could work reviewing the content in https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,183 @@
...
+	// Recommendation is a singular term like "Builder Image", "Devfile" or "Serverless Function".
+	//
+	// Currently, the type is shown a badge on the sample card tile in the top right corner.
+	// +optional
+	// +kubebuilder:validation:MaxLength=20
+	Type string `json:"type"`
+
+	// provider is an optional label to honor who provides the sample.
+	//
+	// It is optional and must be no more than 20 characters in length.
```
JoelSpeed, line 91: Is 20 a bit limiting? Perhaps 50 would be better or is that likely to cause issues? Do you have some data that would show the average length of a provider from the various options for samples?

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,183 @@
...
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ConsoleSample is an extension to customizing OpenShift web console by adding samples.
+//
+// Compatibility level 2: Stable within a major release for a minimum of 9 months or 3 minor releases (whichever is longer).
+// +openshift:compatibility-gen:level=2
```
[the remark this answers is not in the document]
spadgett: Currently `console.openshift.io/v1` is documented as tier 2...  I vaguely recall @bparees suggesting this, but don't remember. @bparees?

https://docs.openshift.com/container-platform/4.13/rest_api/understanding-api-support-tiers.html#mapping-support-tiers-to-openshift-api-groups_understanding-api-tiers

console/v1/types_console_sample.go
jerolimov: It prints the complete icon string, but it "works"

```yaml
Spec:
  Abstract:     Project to test OpenShift import container image flow
  Description:  # About this project
Project to test OpenShift import container image flow

  Icon:  data:image;base64,
iVBORw0KGgoAAAANSUhEUgAAAUAAAADECAYAAADuxvjQAAAABGdBTUEAALGPC/xhBQAAACBjSFJN
....
aWZ5ADIwMTctMDgtMjNUMDg6MDI6NDcrMDA6MDCFs9OnAAAAAElFTkSuQmCC

  Provider:  Red Hat
  Source:
    Container Import:
      Image:  jerolimov/nodeinfo
    Type:     ContainerImport
  Tags:
    JavaScript
    Node.js
    s2i
  Title:  Nodeinfo Image example (PNG icon)
  Type:   Container image
Events:   <none>
```

console/v1/types_console_sample.go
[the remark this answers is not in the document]
bparees: I don't remember how we arrived at that value, it seems like that's what it was set to when the docs were first introduced:

https://github.com/openshift/openshift-docs/pull/30286

this looks like the draft doc where tiers were discussed:
https://docs.google.com/document/d/1dkfqbGfi__Cs9UZMZrv1zGaJwXoHxO84YQkMlvrfAo4/edit#heading=h.s3nekardp0u9

but i don't see console explicitly called out there.

That said, as you note, it's documented as an exception, so unless the console team *wants* to change to tier 1, it seems fine that it this new resource would fall under that tier 2 exception.

console/v1/types_console_sample.go
jerolimov: `^data:([a-z/\.+0-9]*;(([-a-zA-Z0-9=])*;)?)?base64,` would work for the testcases I would test.

```
data:image/svg;base64,asd
data:image/svg+xml;base64,asd
data:image/png;base64,asd
data:text/plain;charset=US-ASCII;base64,asd
```

I will test it on a cluster.

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,183 @@
...
+	// Currently, the provider is only shown on the sample card tile below the title with the prefix "Provided by "
+	// +optional
+	// +kubebuilder:validation:MaxLength=20
+	Provider string `json:"provider"`
+
+	// tags like "Java", "Quarkus", etc.
+	// Tags are optional and can be used by the user to find samples in the samples catalog.
+	// They will be also displayed on the samples details page.
+	// +optional
+	Tags []string `json:"tags"`
```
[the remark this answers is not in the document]
jerolimov: Both make sense, I will limit it to 10 tags.

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,183 @@
...
+
+	// containerimport let the user import a container image.
+	// +unionMember
+	// +optional
+	ContainerImport *ConsoleSampleContainerImportSource `json:"containerImport,omitempty"`
+}
+
+// ConsoleSampleGitImportSource let the user import code from a public git repository.
+type ConsoleSampleGitImportSource struct {
+	// Git repository URL.
```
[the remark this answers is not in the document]
jerolimov: Okay, let us limit this to public (https) GitHub, GitLab, and Bitbucket URLs. The console allow other Git URLs, but only for these, the console supports a client site validation.

```
	// url of the Git repository. Only public HTTPS repositories on GitHub, GitLab
	// and Bitbucket are currently supported:
	//
	//   - https://github.com/<org>/<repository>
	//   - https://gitlab.com/<org>/<repository>
	//   - https://bitbucket.org/<org>/<repository>
	//
	// +kubebuilder:validation:MaxLength=2048
	// +kubebuilder:validation:Pattern=`^https:\/\/(github.com|gitlab.com|bitbucket.org)\/[a-zA-Z-]+\/[a-zA-Z-]+(.git)?$`
```

console/v1/types_console_sample.go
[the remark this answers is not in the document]
JoelSpeed: Curious why the console team would want tier 2? We are declaring this a v1 API (where tier 2 is equivalent to a beta API). Is it possible that it was tier 2 because some console stuff was introduced under a v1alpha1 API group?

Realistically, looking at APIs in openshift, while we have declared some tier 2 APIs, we never treat these APIs as tier 2. We only ever treat them as tier 1.

This _should_ be tier 1, and if it isn't ready to go to tier 1, then it should be tier 4, behind a v1alpha1 API group, and behind a feature gate.

Are the console team confident in their assertion that this should be v1 straight away? If they are, I don't know why they wouldn't accept the Tier 1 responsibilities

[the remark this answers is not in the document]
deads2k: We should update the docs.  Level 2 was created to match up with kubernetes beta APIs.  Since the guide was developed, several things have shifted
1. new kubernetes beta APIs are not enabled by default
2. TechPreview is now usable with new APIs and new fields
3. TechPreview jobs now allow CI to run against them to allow testing before delivery into the product

Level 2 is still reasonable for kubernetes APIs as they arrive, but for openshift specific APIs, the value of a level 2 for both customers and developers is unclear and we should avoid it.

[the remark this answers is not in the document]
JoelSpeed: > the benefit for developers seems pretty clear: they can remove the api in the future(without waiting for an openshift 5.x) if they choose to do so(over a 3 release schedule), which you can't do with a tier 1 api under our current rules.

Given the conversations I've had with management/PM/staff engineers over the last 2 or 3 years, I don't think there's any appetite to treat Tier 2 APIs as Tier 2. For example the Machine API, all declared beta and tier 2, yet, whenever I've spoken to anyone about this, they say removing it would be OpenShift 5.x. I would expect this and any other Tier 2 declared API to come under that same scrutiny.

I appreciate the sentiment and the supposed value, but I don't think our support requirements/management decisions for OpenShift 4 actually allow tier 2 to exist as it suggests it does.

console/v1/types_console_sample.go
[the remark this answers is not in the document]
jerolimov: @JoelSpeed the CI job fails when testing the CRD with `// +kubebuilder:validation:UniqueItems:=true`:

```
  [38;5;9m[FAILED] Unexpected error:
      <*fmt.wrapError | 0xc0019d1800>: {
          msg: "unable to create CRD instances: unable to create CRD \"consolesamples.console.openshift.io\": CustomResourceDefinition.apiextensions.k8s.io \"consolesamples.console.openshift.io\" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[tags].uniqueItems: Forbidden: uniqueItems cannot be set to true since the runtime complexity becomes quadratic",
          err: <*fmt.wrapError | 0xc0019d17e0>{
              msg: "unable to create CRD \"consolesamples.console.openshift.io\": CustomResourceDefinition.apiextensions.k8s.io \"consolesamples.console.openshift.io\" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[tags].uniqueItems: Forbidden: uniqueItems cannot be set to true since the runtime complexity becomes quadratic",
              err: <*errors.StatusError | 0xc00125b860>{
                  ErrStatus: {
                      TypeMeta: {Kind: "", APIVersion: ""},
                      ListMeta: {
                          SelfLink: "",
                          ResourceVersion: "",
                          Continue: "",
                          RemainingItemCount: nil,
                      },
                      Status: "Failure",
                      Message: "CustomResourceDefinition.apiextensions.k8s.io \"consolesamples.console.openshift.io\" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[tags].uniqueItems: Forbidden: uniqueItems cannot be set to true since the runtime complexity becomes quadratic",
                      Reason: "Invalid",
                      Details: {
                          Name: "consolesamples.console.openshift.io",
                          Group: "apiextensions.k8s.io",
                          Kind: "CustomResourceDefinition",
                          UID: "",
                          Causes: [
                              {
                                  Type: "FieldValueForbidden",
                                  Message: "Forbidden: uniqueItems cannot be set to true since the runtime complexity becomes quadratic",
                                  Field: "spec.validation.openAPIV3Schema.properties[spec].properties[tags].uniqueItems",
                              },
                          ],
                          RetryAfterSeconds: 0,
                      },
                      Code: 422,
                  },
              },
          },
      }
      unable to create CRD instances: unable to create CRD "consolesamples.console.openshift.io": CustomResourceDefinition.apiextensions.k8s.io "consolesamples.console.openshift.io" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[tags].uniqueItems: Forbidden: uniqueItems cannot be set to true since the runtime complexity becomes quadratic
  occurred[0m
  [38;5;9mIn [1m[BeforeEach][0m[38;5;9m at: [1m/go/src/github.com/openshift/api/tests/generator.go:136[0m [38;5;243m@ 07/10/23 14:55:09.023[0m
```

From https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_api/1503/pull-ci-openshift-api-master-integration/1678413731614167040

I replaced the UniqueItems with `// +listType=set`, we used this a lot in the API.

console/v1/types_console_sample.go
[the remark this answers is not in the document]
jerolimov: I changed it to tier 1 as we explicitly said that this will be released as GA.

console/v1/types_console_sample.go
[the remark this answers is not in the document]
JoelSpeed: Can you add an integration test to prove that this works as expected?

console/v1/types_console_sample.go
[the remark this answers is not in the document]
JoelSpeed: Is 100 characters going to be enough here? That feels quite limiting, maybe limit at 512?

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,193 @@
...
+	ContextDir string `json:"contextDir"`
+}
+
+// ConsoleSampleContainerImportSource let the user import a container image.
+type ConsoleSampleContainerImportSource struct {
+	// Reference to a container image like openshift/hello-openshift,
+	// docker.io/openshift/hello-openshift or quay.io/<username>/<image-name>
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=200
```
JoelSpeed, line 177: Is 200 characters enough here? How long is a standard SHA for referring to an image? Are there any limits on the number of namespaces or length of the namespaces in the URL?

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,193 @@
...
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=100
+	// +kubebuilder:validation:Pattern=`^https:\/\/(github.com|gitlab.com|bitbucket.org)\/[a-zA-Z0-9-]+\/[a-zA-Z0-9-]+(.git)?$`
+	URL string `json:"url"`
+	// revision to clone a specific branch, tag or commit SHA, etc.
+	// +optional
+	Revision string `json:"revision"`
+	// contextDir, a subfolder in the repo to build the component.
```
JoelSpeed, line 165: The validation pattern is not explained in the godoc

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,193 @@
...
+	//   - https://github.com/<org>/<repository>
+	//   - https://gitlab.com/<org>/<repository>
+	//   - https://bitbucket.org/<org>/<repository>
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=100
+	// +kubebuilder:validation:Pattern=`^https:\/\/(github.com|gitlab.com|bitbucket.org)\/[a-zA-Z0-9-]+\/[a-zA-Z0-9-]+(.git)?$`
+	URL string `json:"url"`
+	// revision to clone a specific branch, tag or commit SHA, etc.
```
JoelSpeed, line 162: Is there a maximum length we can apply to this one?

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,193 @@
...
+	// url of the Git repository. Only public HTTPS repositories on GitHub, GitLab
+	// and Bitbucket are currently supported:
+	//
+	//   - https://github.com/<org>/<repository>
+	//   - https://gitlab.com/<org>/<repository>
+	//   - https://bitbucket.org/<org>/<repository>
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=100
```
JoelSpeed, line 159: 100 seems short here, maybe 256?

console/v1/types_console_sample.go
JoelSpeed: How will you assess feedback that the length is too short? Is there any data we can use to inform this decision?

console/v1/types_console_sample.go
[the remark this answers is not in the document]
jerolimov: As discussed on slack, changing this to 256. :smile:

console/v1/types_console_sample.go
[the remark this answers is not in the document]
jerolimov: Added two tests:

  - name: Should decline a ConsoleSample when reusing a tag
  - name: Should decline a ConsoleSample with more then 10 tags

console/v1/types_console_sample.go
jerolimov: Added MaxLength=256

console/v1/types_console_sample.go
jerolimov: Changed to 256

console/v1/types_console_sample.go
jerolimov: Its enough for the common scenarios. Changed it to 256 since we have used this for the git URL as well which is normally shorter then a image ref with a SHA.

console/v1/types_console_sample.go
jerolimov: Changed it to 50. For me it's up to the developer to check if their sample look and work correctly in the console. It will work also with 50 characters and looks "okay":

image

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,198 @@
...
+	//   - https://github.com/<org>/<repository>
+	//   - https://gitlab.com/<org>/<repository>
+	//   - https://bitbucket.org/<org>/<repository>
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=256
+	// +kubebuilder:validation:Pattern=`^https:\/\/(github.com|gitlab.com|bitbucket.org)\/[a-zA-Z0-9-]+\/[a-zA-Z0-9-]+(.git)?$`
+	URL string `json:"url"`
+	// revision to clone a specific branch, tag or commit SHA, etc.
```
JoelSpeed, line 163:
```suggestion
	// revision is the git revision at which to clone the git repository
	// Can be used to clone a specific branch, tag or commit SHA.
	// Must be at most 256 characters in length.
	// When omitted the repository's default branch is used.
```

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,198 @@
...
+	// +optional
+	// +kubebuilder:validation:MaxLength=256
+	Revision string `json:"revision"`
+	// contextDir, a subfolder in the repo to build the component.
```
JoelSpeed, line 167:
```suggestion
	// contextDir is used to specify a directory within the repository to build the
	// component.
	// Must start with `/` and have a maximum length of 256 characters.
	// When omitted, the default value is to build from the root of the repository.
```
And drop the line below about starting with a slash

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,198 @@
...
+	// The context directory must start with a slash.
+	// +optional
+	// +kubebuilder:validation:MaxLength=256
+	// +kubebuilder:validation:Pattern=`^/`
+	ContextDir string `json:"contextDir"`
+}
+
+// ConsoleSampleContainerImportSource let the user import a container image.
+type ConsoleSampleContainerImportSource struct {
+	// Reference to a container image like openshift/hello-openshift,
```
JoelSpeed, line 178:
```suggestion
	// reference to a container image like openshift/hello-openshift,
```

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,198 @@
...
+	// docker.io/openshift/hello-openshift or quay.io/<username>/<image-name>
```
JoelSpeed, line 179: This doesn't tell me anything about what the expected content within the container image should be, any additional context you can add?

jerolimov: Any container that runs a service is fine. :thinking:

jerolimov: Updated based on our conversation on slack. And added a new field `port` as well. :smile:

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,261 @@
...
+// Compatibility level 1: Stable within a major release for a minimum of 12 months or 3 minor releases (whichever is longer).
+// +openshift:compatibility-gen:level=1
+type ConsoleSample struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// metadata is the standard object's metadata.
+	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
+	metav1.ObjectMeta `json:"metadata"`
+
+	// +kubebuilder:validation:Required
```
JoelSpeed, line 20: Should have a godoc for the spec
```suggestion
	// spec contains configuration for a console sample.
	// +kubebuilder:validation:Required
```

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,261 @@
...
+	// For example:
+	//
+	//   data:image;base64,             plus the base64 encoded image.
+	//
+	// Vector images can also be used. SVG icons must start with:
+	//
+	//   data:image/svg+xml;base64,     plus the base64 encoded SVG image.
+	//
+	// All sample catalog icons will be shown on a white background (also when the dark theme is used).
+	// The web console ensures that different aspect radios work correctly.
```
JoelSpeed, line 70:
```suggestion
	// The web console ensures that different aspect ratios work correctly.
```

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,261 @@
...
+	//
+	// It is a README.md-like content for additional information, links, pre-conditions, and other instructions.
+	// It will be rendered as Markdown so that it can contain line breaks, links, and other simple formatting.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength=4096
+	Description string `json:"description"`
+
+	// icon is an optional base64 encoded image and shown beside the sample title.
+	//
+	// The format must follow the [data: URL format] and can have a maximum size of **10 KB**.
```
JoelSpeed, line 57: This isn't markdown, so the linking you're doing here won't work really, it's still a nice indicator but I wonder if it's better to just have this written out, and then at the end have, `For more information on the data URL format, please visit <link>.`

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,261 @@
...
+	//
+	// A provider can be a company like "Red Hat" or an organization like "CNCF" or "Knative".
+	//
+	// Currently, the provider is only shown on the sample card tile below the title with the prefix "Provided by "
+	// +optional
+	// +kubebuilder:validation:MaxLength=50
+	Provider string `json:"provider"`
+
+	// tags like "Java", "Quarkus", etc.
+	// Tags are optional and can be used by the user to find samples in the samples catalog.
```
JoelSpeed, line 102: This should be reversed, introduce what a tag is, then give examples
```suggestion
	// tags are optional string values that can be used to find samples in the samples catalog.
	// Examples of common tags may be "Java", "Quarkus", etc.
```

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,261 @@
...
+	// They will be also displayed on the samples details page.
+	// +optional
+	// +listType=set
+	// +kubebuilder:validation:MaxItems:=10
+	Tags []string `json:"tags"`
+
+	// The actual sample source, like an external git repository or container image.
```
JoelSpeed, line 109:
```suggestion
	// source defines where to deploy the sample service from.
	// The sample may be sourced from an external git repository or container image.
```

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,261 @@
...
+// +kubebuilder:validation:XValidation:rule="self.type == 'GitImport' ? has(self.gitImport) : !has(self.gitImport)",message="source.gitImport is required when source.type is GitImport, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="self.type == 'ContainerImport' ? has(self.containerImport) : !has(self.containerImport)",message="source.containerImport is required when source.type is ContainerImport, and forbidden otherwise"
+type ConsoleSampleSource struct {
+	// type of the sample, currently supported: "GitImport";"ContainerImport"
+	// +unionDiscriminator
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum:="GitImport";"ContainerImport"
+	Type ConsoleSampleSourceType `json:"type"`
+
+	// gitimport let the user import code from a git repository.
```
JoelSpeed, line 138:
```suggestion
	// gitImport allows the user to import code from a git repository.
```

console/v1/types_console_sample.go
```diff
@@ -0,0 +1,261 @@
...
+	// +unionMember
+	// +optional
+	GitImport *ConsoleSampleGitImportSource `json:"gitImport,omitempty"`
+
+	// containerimport let the user import a container image.
```
JoelSpeed, line 143:
```suggestion
	// containerImport allows the user import a container image.
```

JoelSpeed (commented):
If you can apply the suggested wordings, and maybe think about that link I've highlighted, I think this is good to go

console/v1/types_console_sample.go
jerolimov: Yes, not markdown, it follows the godoc convention for links: https://tip.golang.org/doc/comment#links

But I started godoc now and noticed that it only convert struct docs into real doc and shows the complete struct as go code.

Updated.

PR #1514, opened 2023-07-03
Create list of FeatureGates for cloud controller

mkowalski:
This PR creates a publicly available list of FeatureGates that are valid to be used in the context of Cloud Controller. Currently the problem is that lots of features that are valid for kubelet, are invalid for the cloud provider. With this PR we are introducing a very simple way to filter out undesired features.


mkowalski:
/close

It's stale, it seems we don't need it anymore

PR #1515, opened 2023-07-04
Retract v3.9.0+incompatible

dmage:
Similar to openshift/client-go#246, this change retracts the only published version v3.9.0+incompatible.


soltysh (approved):
/lgtm
/approve

PR #1516, opened 2023-07-04
NE-1303: cluster version: add ingress capability

alebedev87:
Enhancement Proposal: https://github.com/openshift/enhancements/pull/1415.

This PR introduces a new `Ingress` cluster capability which allows to enable or disable the default ingress of the OpenShift cluster: the cluster ingress operator and its custom resources. As of now the `Ingress` capability is added to 4.14 capabilityset. This is a subject to change if [the underlying epic](https://issues.redhat.com/browse/NE-1129) changes the target version.
Also, 2 CRDs managed by the cluster ingress operator are annotated with `Ingress` capability: `IngressController`, `DNSRecord`.

### DIFF
--- a/config/v1/0000_00_cluster-version-operator_01_clusterversion.crd.yaml
+++ b/config/v1/0000_00_cluster-version-operator_01_clusterversion.crd.yaml
@@ -72,6 +72,7 @@ spec:
                           - CSISnapshot
                           - NodeTuning
                           - MachineAPI
+                          - Ingress
                       x-kubernetes-list-type: atomic
                     baselineCapabilitySet:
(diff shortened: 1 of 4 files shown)


Miciah:
I had a question on the operator PR about whether the `capability.openshift.io/name: Ingress` should be added in the openshift/api PR: https://github.com/openshift/cluster-ingress-operator/pull/950/files#r1268238815

alebedev87:
@Miciah : you are right. The ingress manifests should be updated in this repository. I created https://github.com/openshift/cluster-ingress-operator/pull/950 according to [how to implement a new capability](https://github.com/openshift/enhancements/blob/master/enhancements/installer/component-selection.md#how-to-implement-a-new-capability) to check the behavior of the CI presubmit jobs when `Ingress` capability is unknown. The same instructions misled me with an example which didn't need CRD updates in the api repository.

PR #1517, opened 2023-07-05
SO-112: To Install helmcharts through Samples operator

fbm3307:


samples/v1/types_config.go
```diff
@@ -64,6 +64,17 @@ type ConfigSpec struct {
...
+	// helmchart list specifies the list of helmcharts that the samples operator will install
+	HelmChartList []HelmChart `json:"helmChartList,omitempty" protobuf:"bytes,7,opt,name=helmChartList"`
+}
+
+type HelmChart struct {
+	URL string `json:"url,omitempty"`
+
+	Namespace string `json:"namespace,omitempty"`
+
+	ReleaseName string `json:"releaseName,omitempty"`
```
bparees, line 77: godoc these fields please.  Also is there any validation we can put in place?

should these be required fields?

fbm3307:
closing this in favor of https://github.com/openshift/api/pull/1636
 as this PR is very old and it is causing problems in testcases

PR #1538, opened 2023-07-26
Adding etcd tuning profiles api

dusk125:
Adds the techpreview API for etcd tuning profiles as will be outlined in https://github.com/openshift/enhancements/pull/1447

This will allow for limited user configuration of etcd's tolerated intervals/timeouts (heartbeat interval, leader election timeout)

Adds the API:
profiles.config.openshift.io/v1alpha1
Used to configure the tuning profiles in etcd

/cc @tjungblu @deads2k @hasbro17 @williamcaban

### DIFF
--- a/operator/v1/types_etcd.go
+++ b/operator/v1/types_etcd.go
@@ -28,11 +28,40 @@ type Etcd struct {
 
 type EtcdSpec struct {
 	StaticPodOperatorSpec `json:",inline"`
+	// HardwareSpeed allows user to change the etcd tuning profile which configures
+	// the latency parameters for heartbeat interval and leader election timeouts
+	// allowing the cluster to tolerate longer round-trip-times between etcd members.
+	// Valid values are "", "Standard" and "Slower".
+	//	"" means no opinion and the platform is left to choose a reasonable default
+	//	which is subject to change without notice.
+	// +kubebuilder:validation:Optional
+	// +openshift:enable:FeatureSets=CustomNoUpgrade;TechPreviewNoUpgrade
+	// +optional
+	HardwareSpeed ControlPlaneHardwareSpeed `json:"controlPlaneHardwareSpeed"`
 }
 
 type EtcdStatus struct {
 	StaticPodOperatorStatus `json:",inline"`
+	HardwareSpeed           ControlPlaneHardwareSpeed `json:"controlPlaneHardwareSpeed"`
 }
+
+const (
+	// StandardHardwareSpeed provides the normal tolerances for hardware speed and latency.
+	//	Currently sets (values subject to change at any time):
+	//		ETCD_HEARTBEAT_INTERVAL: 100ms
+	// 	ETCD_LEADER_ELECTION_TIMEOUT: 1000ms
+	StandardHardwareSpeed ControlPlaneHardwareSpeed = "Standard"
+	// SlowerHardwareSpeed provides more tolerance for slower hardware and/or higher latency networks.
+	// Sets (values subject to change):
+	//		ETCD_HEARTBEAT_INTERVAL: 5x Standard
+	// 	ETCD_LEADER_ELECTION_TIMEOUT: 2.5x Standard
+	SlowerHardwareSpeed ControlPlaneHardwareSpeed = "Slower"
+)
+
+// ControlPlaneHardwareSpeed declares valid hardware speed tolerance levels
+// +enum
+// +kubebuilder:validation:Enum:="";Standard;Slower
+type ControlPlaneHardwareSpeed string
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
--- a/openapi/generated_openapi/zz_generated.openapi.go
+++ b/openapi/generated_openapi/zz_generated.openapi.go
@@ -42752,6 +42752,14 @@ func schema_openshift_api_operator_v1_EtcdSpec(ref common.ReferenceCallback) com
 							Format:      "int32",
 						},
 					},
+					"controlPlaneHardwareSpeed": {
+						SchemaProps: spec.SchemaProps{
+							Description: "HardwareSpeed allows user to change the etcd tuning profile which configures the latency parameters for heartbeat interval and leader election timeouts allowing the cluster to tolerate longer round-trip-times between etcd members. Valid values are \"\", \"Standard\" and \"Slower\".\n\t\"\" means no opinion and the platform is left to choose a reasonable default\n\twhich is subject to change without notice.\n\nPossible enum values:\n - `\"Slower\"` provides more tolerance for slower hardware and/or higher latency networks. Sets (values subject to change): ETCD_HEARTBEAT_INTERVAL: 5x Standard ETCD_LEADER_ELECTION_TIMEOUT: 2.5x Standard\n - `\"Standard\"` provides the normal tolerances for hardware speed and latency. Currently sets (values subject to change at any time): ETCD_HEARTBEAT_INTERVAL: 100ms ETCD_LEADER_ELECTION_TIMEOUT: 1000ms",
+							Default:     "",
+							Type:        []string{"string"},
+							Format:      "",
+							Enum:        []interface{}{"Slower", "Standard"}},
+					},
 				},
 				Required: []string{"managementState", "forceRedeploymentReason"},
 			},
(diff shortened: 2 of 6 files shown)


operator/v1/types_etcd.go
```diff
@@ -28,12 +28,33 @@ type Etcd struct {
...
 	StaticPodOperatorSpec `json:",inline"`
+	// profile allows user to change the etcd tuning profile which configures
+	// the latency parameters for heartbeat interval and leader election timeouts
+	// allowing the cluster to tolerate longer round-trip-times between etcd members.
+	// Valid values are "Default" and "Slower".
+	// The current default is Default.
+	// +kubebuilder:validation:Optional
+	// +openshift:enable:FeatureSets=CustomNoUpgrade;TechPreviewNoUpgrade
+	// +optional
+	Profile ProfileValue `json:"profile,omitempty"`
```
ingvagabund, line 39: There's no `// +kubebuilder:default:=Default`. Is the default value managed by the etcd operator? Is the default value a subject to change?

ingvagabund: How will the switching between Default and Slower will be handled? Any danger to the etcd instances when the heartbeat interval and leader election timeouts are changed? Any danger when heartbeat interval and leader election timeouts are changed to quickly (i.e. malicious user/admin alternating the values)?

operator/v1/types_etcd.go
```diff
@@ -28,12 +28,33 @@ type Etcd struct {
...
 }
 
 type EtcdStatus struct {
 	StaticPodOperatorStatus `json:",inline"`
 }
 
+const (
+	// DefaultProfile sets Heartbeat interval: 100ms, Leader election timeout: 1000ms
```
ingvagabund, line 47: Will any of heartbeat interval/leader election timeout change in the future? I.e. can the default values for the Default profile change? E.g. based on some performance tuning through various OCP releases. The same question for the Slower profile.

ingvagabund: Are the values going to be identical for all providers? Or, is there a possibility the default values will differ per each provider?

operator/v1/types_etcd.go
```diff
@@ -28,12 +28,33 @@ type Etcd struct {
...
+	DefaultProfile ProfileValue = "Default"
+	// SlowerProfile sets Heartbeat interval: 500ms, Leader election timeout: 2500ms
+	SlowerProfile ProfileValue = "Slower"
+)
+
+// profileValue declares valid etcd tuning profiles
+// +enum
+// +kubebuilder:validation:Enum:="Default";"Slower"
+type ProfileValue string
```
ingvagabund, line 56: Would ProfileValue -> Profile work as well? A `Profile` itself can be used as a field name.

operator/v1/types_etcd.go
dusk125: I'll add the default with the kubebuilder.

The current plan is to require an etcd rollout to consume the changed value. A rollout will restart each etcd instance one-by-one so there's no risk hot-changing the value.

We're still considering not requiring the rollout, but that will be for a future change.

operator/v1/types_etcd.go
dusk125: I'm not sure I understand. Are you suggesting to rename the ProfileValue type to just Profile?

operator/v1/types_etcd.go
dusk125: The mappings for each profile should be static. We'll add additional profiles if we need different values.

The values will be identical for all providers. If a certain provider needs a different starting profile, we can work to have that set for that provider by default.

For example, if we find that Azure, for whatever reason, needs to be on the Slower profile out of the box, we can add logic to set the initial profile to Slower.

operator/v1/types_etcd.go
ingvagabund: yes

operator/v1/types_etcd.go
```diff
@@ -28,12 +28,34 @@ type Etcd struct {
...
 }
 
 type EtcdStatus struct {
 	StaticPodOperatorStatus `json:",inline"`
 }
 
+const (
+	// DefaultProfile sets Heartbeat interval: 100ms, Leader election timeout: 1000ms
+	DefaultProfile Profile = "Default"
+	// SlowerProfile sets Heartbeat interval: 500ms, Leader election timeout: 2500ms
```
deads2k, line 50: Describe the intent, not the values.  Perhaps as a multiple off the default.  That way values can be subject to change in the future.  You can say what they currently are, but they are subject to change without notice.

operator/v1/types_etcd.go
deads2k: YOu should still describe intent and indicate that the values are subject to change without notice.

operator/v1/types_etcd.go
```diff
@@ -28,12 +28,34 @@ type Etcd struct {
...
+	// profile allows user to change the etcd tuning profile which configures
+	// the latency parameters for heartbeat interval and leader election timeouts
+	// allowing the cluster to tolerate longer round-trip-times between etcd members.
+	// Valid values are "Default" and "Slower".
+	// The current default is Default.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=Default
+	// +openshift:enable:FeatureSets=CustomNoUpgrade;TechPreviewNoUpgrade
+	// +optional
+	Profile Profile `json:"profile,omitempty"`
```
deads2k, line 40: will this field be force set to Slower for Azure? or are you thinking that empty means select a default.  I'm happy with "a default is chosen", but you need specify in the godoc.  Choosing a default is what we "normally" do and it works extremely well.

operator/v1/types_etcd.go
```diff
@@ -28,12 +28,34 @@ type Etcd struct {
...
 }
 
 type EtcdStatus struct {
```
deads2k, line 43: You need a status field indicating which profile is actually being used so that "" can work as expected.

operator/v1/types_etcd.go
```diff
@@ -28,12 +28,34 @@ type Etcd struct {
 
 type EtcdSpec struct {
 	StaticPodOperatorSpec `json:",inline"`
+	// profile allows user to change the etcd tuning profile which configures
+	// the latency parameters for heartbeat interval and leader election timeouts
+	// allowing the cluster to tolerate longer round-trip-times between etcd members.
+	// Valid values are "Default" and "Slower".
+	// The current default is Default.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=Default
```
deads2k, line 37: don't set a default.  This would be wrong on azure for instance.

operator/v1/types_etcd.go
deads2k, line 40: don't omitempty.  Our configuration APIs show the empty so people know what options are available.

dusk125: I was thinking having the default be "Default" for all platforms, so they're consistent, but if we needed to start a platform, like Azure, at a different value, we would set its starting value to Slower.

operator/v1/types_etcd.go
```diff
@@ -28,12 +28,40 @@ type Etcd struct {
 
 type EtcdSpec struct {
 	StaticPodOperatorSpec `json:",inline"`
+	// HardwareSpeed allows user to change the etcd tuning profile which configures
+	// the latency parameters for heartbeat interval and leader election timeouts
+	// allowing the cluster to tolerate longer round-trip-times between etcd members.
+	// Valid values are "", "Standard" and "Slower".
+	//	On "", the system will choose the speed.
```
deads2k, line 35: without notice
```suggestion
	//	"" means no opinion and the platform is left to choose a reasonable default which is subject to change without notice.  
```

operator/v1/types_etcd.go
```diff
@@ -28,12 +28,40 @@ type Etcd struct {
...
+	// SlowerHardwareSpeed provides more tolerance for slower hardware and/or higher latency networks.
+	// Sets (values subject to change):
+	//		ETCD_HEARTBEAT_INTERVAL: 5x Standard
+	// 	ETCD_LEADER_ELECTION_TIMEOUT: 2.5x Standard
+	SlowerHardwareSpeed ControlPlaneHardwareSpeed = "Slower"
+)
+
+// ControlPlaneHardwareSpeed declares valid hardware speed tolerance levels
+// +enum
+// +kubebuilder:validation:Enum:="Standard";"Slower"
```
deads2k, line 62: need to allow "".

Add a test in this repo

PR #1544, opened 2023-08-03
helm/v1beta1: Rename files to use the default run-level for updates

wking:
I'd meant to do these in 9e3e820a70 (#1541), but flubbed my regexp.  Fixed now, with:

```console
$ rename 's/0000_[0-9][0-9][_-]/00_/' helm/*/*.yaml
$ sed -i 's/0000_[0-9][0-9][_-]/00_/' helm/*/*.testsuite.yaml
```

PR #1541, opened 2023-08-02
CONSOLE-3717: Console CRDs: Rename files to use the default run-level for updates

wking:
It's unclear what the motivation was for their existing runlevels, with the API filenames landing without much context:

```console
$ git --no-pager log --oneline operator/v1/0000_70_console-operator.crd.yaml | tail -n1
862dd2fe4 Add console-operator CRD and generate from types
$ git --no-pager log -1 --format='%h %B' --stat=80 862dd2fe4
862dd2fe4 Add console-operator CRD and generate from types
    
 operator/v1/0000_70_console-operator.crd.yaml | 195 ++++++++++++++++++++++++++
 1 file changed, 195 insertions(+)
$ git --no-pager log --oneline helm/*.crd.yaml | tail -n1
a4be43c2b Added ability to configure Helm chart repository accessible within cluster
$ git --no-pager log -1 --format='%h %B' --stat=80 a4be43c2b
a4be43c2b Added ability to configure Helm chart repository accessible within cluster
    
* Introduced `HelmChartRepository` top-level CR modelled according to chartrepo.Entry
  from https://github.com/helm/helm/blob/master/pkg/repo/chartrepo.go#L42
* The corresponding enchancement: openshift/enhancements/pull/175
    
 Makefile                                           |   3 +-
 hack/update-deepcopy.sh                            |   2 +-
 hack/verify-crds.sh                                |   1 +
 .../0000_10-helm-chart-repository.crd.yaml         | 138 +++++++++++++++++++
 helm/v1alpha1/doc.go                               |   8 ++
 helm/v1alpha1/register.go                          |  38 ++++++
 helm/v1alpha1/types_helm.go                        | 108 +++++++++++++++
 helm/v1alpha1/zz_generated.deepcopy.go             | 152 +++++++++++++++++++++
 8 files changed, 448 insertions(+), 2 deletions(-)
$ git --no-pager log --oneline helm/v1beta1/0000_10-project-helm-chart-repository.crd.yaml | tail -n1
8fe639fe2 helm: add a new namespaced CRD for helm
$ git --no-pager log -1 --format='%h %B' --stat=80 8fe639fe2
8fe639fe2 helm: add a new namespaced CRD for helm
    
As part of the process to support namespace-scoped Helm repositories for
non-admin users, this PR adds a new namespace-scoped CRD definition
named `projecthelmchartrepositories.helm.openshift.io`.
    
Closes: https://issues.redhat.com/browse/HELM-258
[...]

PR #1545, opened 2023-08-03
samples/v1: Rename files to use the default run-level for updates

wking:
Like 9e3e820a70 (#1541) and 9aa8b74279 (#1544), but now for [the CustomResourceDefinition deployed by the samples operator][1].  As in the previous changes, the outgoing filename failed to match [the
`0000_<runlevel>_<dash-separated-component>_<manifest_filename>` template][2], and there is no reason that the CRD needs bumping at run-level 10 during updates.  I'm using a `00_` prefix to sort before [the other samples manifests][3], so there's no relative change in update order between those resources; this commit just shifts update CRD reconciliation vs. other, non-samples resources managed by the cluster-version operator.

Generated with:

```console
$ mv samples/v1/0000_10_samplesconfig.crd.yaml samples/v1/00_samplesconfig.crd.yaml
$ sed -i 's/0000_10_/00_/' samples/v1/stable.config.testsuite.yaml
$ git add -A samples
```

[1]: https://github.com/openshift/cluster-samples-operator/blob/684b8c7f1cf6aaf9fb4f2bcffe91cde7eae07347/Dockerfile#L10
[2]: https://github.com/openshift/enhancements/blob/cafeb5c3cba7f8c9e261b2aabffa92e34dd76be6/dev-guide/cluster-version-operator/dev/operators.md#what-is-the-order-that-resources-get-createdupdated-in
[3]: https://github.com/openshift/cluster-samples-operator/tree/684b8c7f1cf6aaf9fb4f2bcffe91cde7eae07347/manifests


bparees:
/approve

@dperaza4dustbit for the lgtm

PR #1550, opened 2023-08-08
WRKLDS-727: add controller names

knelasevero:
Needed by:

- https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/291
- We can also use those contants in https://github.com/openshift/openshift-controller-manager/blob/80c4923e0b2014ec8b8a6845c21078e6cc036f91/pkg/cmd/controller/config.go#L3-L25

Ref: https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/291#discussion_r1285259744


knelasevero:
@JoelSpeed added comments based on slack conversation.

PR #1554, opened 2023-08-11
OCPBUGS-15654: Add MAPI to all capability sets

qJkee:
In order to make capability sets working properly we had to add MAPI capability everywhere where baremetal capability present. We also add api and installer validation to ensure that baremetal and MAPI enabled together, see
1. https://github.com/openshift/installer/pull/7394
2. https://github.com/openshift/api/pull/1537


JoelSpeed (commented):
/lgtm

This must not be backported to other releases, discussed in slack https://redhat-internal.slack.com/archives/C032HSVS71T/p1691055174729009

PR #1558, opened 2023-08-14
Introduce ValidatingAdmissionPolicy gate as disabled.

benluddy:
This feature is intentionally not enabled in an feature set since it depends on a change to the kube-apiserver operator. The kube-apiserver operator change can't merge (sans hack) until this feature is "known". So the intention is to introduce ValidatingAdmissionPolicy here, then merge the operator change, and finally add ValidatingAdmissionPolicy to the tech preview feature set.

### DIFF
--- a/config/v1/types_feature.go
+++ b/config/v1/types_feature.go
@@ -163,6 +163,7 @@ var FeatureSets = map[FeatureSet]*FeatureGateEnabledDisabled{
 		Disabled: []FeatureGateDescription{},
 	},
 	TechPreviewNoUpgrade: newDefaultFeatures().
+		without(validatingAdmissionPolicy).
 		with(externalCloudProvider).
 		with(externalCloudProviderGCP).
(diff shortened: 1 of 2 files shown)


config/v1/types_feature.go
```diff
@@ -200,6 +200,7 @@ var defaultFeatures = &FeatureGateEnabledDisabled{
 	},
 	Disabled: []FeatureGateDescription{
 		retroactiveDefaultStorageClass,
+		validatingAdmissionPolicy,
```
deads2k, line 203: this shouldn't be specified. don't want to keep it disabled once it is enabled by default.  You can specify it as disabled in techpreview, which is weird, but I think will do the same thing.  test in a fake PR to cluster-config-operator.

benluddy: Thanks, done and it verified that it shows up in the disabled list for Default as expected.

PR #1559, opened 2023-08-14
Enable ValidatingAdmissionPolicy for tech preview.

benluddy:


dinhxuanvu (commented):
/lgtm

benluddy:
Proof https://github.com/openshift/cluster-config-operator/pull/343 is green, including the tech preview E2E job. I'm in the process of doing a smoke test now to confirm that this feature is in fact usable on a tech preview cluster (the point of the whole exercise).

PR #1561, opened 2023-08-16
Remove tech preview feature set annotation from existing non-techpreview crd

dusk125:
I mistakenly added this while working on the tests for #1538, I think that it's causing this crd to not get laid down correctly which is causing etcd to not start properly.


operator/v1/0000_12_etcd-operator_01_config.crd.yaml
```diff
@@ -6,7 +6,6 @@ metadata:
     include.release.openshift.io/ibm-cloud-managed: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
-    release.openshift.io/feature-set: TechPreviewNoUpgrade
```
deads2k, line 9: need to specify Default

PR #1577, opened 2023-09-01
OCPBUGS-18454: Match restrictions on ConditionalUpdateRisk name with Condition reason

petr-muller:
Conditional risk names are used as a `Reason` field values when the risk
applies and CVO sets up the `Recommended` condition. Hence, the risk
names need to be valid [Reason field values](https://github.com/openshift/api/blob/8891815aa476232109dccf6c11b8611d209445d9/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L1515C4-L1520).

xref: https://github.com/openshift/cincinnati-graph-data/pull/4059

### DIFF
--- a/config/v1/types_cluster_version.go
+++ b/config/v1/types_cluster_version.go
@@ -644,9 +644,11 @@ type ConditionalUpdateRisk struct {
 
 	// name is the CamelCase reason for not recommending a
 	// conditional update, in the event that matchingRules match the
-	// cluster state.
+	// cluster state. Can be used as a reason in a Condition.
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength=1024
 	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Pattern=`^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?$`
 	// +required
 	Name string `json:"name"`
 
--- a/config/v1/0000_00_cluster-version-operator_01_clusterversion.crd.yaml
+++ b/config/v1/0000_00_cluster-version-operator_01_clusterversion.crd.yaml
@@ -341,9 +341,11 @@ spec:
                               type: string
                               minLength: 1
                             name:
-                              description: name is the CamelCase reason for not recommending a conditional update, in the event that matchingRules match the cluster state.
+                              description: name is the CamelCase reason for not recommending a conditional update, in the event that matchingRules match the cluster state. Can be used as a reason in a Condition.
                               type: string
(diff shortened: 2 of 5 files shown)


config/v1/types_cluster_version.go
```diff
@@ -644,9 +644,11 @@ type ConditionalUpdateRisk struct {
 
 	// name is the CamelCase reason for not recommending a
 	// conditional update, in the event that matchingRules match the
-	// cluster state.
+	// cluster state. Can be used as a reason in a Condition.
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength=1024
 	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Pattern=`^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?$`
```
wking, line 651: I'm conflicted about this.  If we accept that the update service can serve risks with non-compliant names, then we can:

a. Land this regexp.  Drop the risk (silently for now, eventually triggering [`CannotEvaluateConditionalUpdates`][1] to warn the admin).  You'd have to do this for all clusters, not just clusters where the risk applied.
b. Land this regexp.  Squash any non-compliant names into compliant names before writing the risk.
c. Drop this API change.  Accept the risk with the non-compliant name, and squash it into something compliant when we translate to the `Recommended` `reason` (if we end up wanting to involve the risk's name because we assess it as the only valid risk).

I'm personally leaning towards (c), because fewer clusters are impacted (only clusters where the non-compliant risk is the only risk assessed to apply), and because I like complaining in `status` instead of relying on alerts alone.  But this pull request's current (a) is more aligned with our existing `status` enum precedent, where the Kube API server doesn't trust the CVO to write solid data.  And (b) hides the "hey, we wish the update service was using a better name for this risk" condition while (c) makes that more obvious just from `status`.

I'm not terribly committed though, so I'm ok with what you have here if my cost/benefit breakdown isn't convincing ;)

[1]: https://issues.redhat.com/browse/OCPBUGS-9050

petr-muller: I agree, I currently like (c) the best too. I filed https://github.com/openshift/cluster-version-operator/pull/962 that follows that path (including some refactors so that the change is easier to test). That PR just operates in the same area like https://github.com/openshift/cluster-version-operator/pull/964 which is more important to merge and I will need to rebase #962 on top of it.

petr-muller:
/close

We'll probably don't want to do this and only merge https://github.com/openshift/cluster-version-operator/pull/962. Closing for now, I can always reopen.

PR #1580, opened 2023-09-06
IR-401: remove references to OPENSHIFT_DEFAULT_REGISTRY

flavianmissi:
`OPENSHIFT_DEFAULT_REGISTRY` is deprecated for while, and now is removed by
https://github.com/openshift/image-registry/pull/378


deads2k:
What injects this env var?  I see one change in the image registry itself.  This is likely ok since we own the deployment.  But does anything auto-inject this value elsewhere?

flavianmissi:
> What injects this env var? I see one change in the image registry itself. This is likely ok since we own the deployment. But does anything auto-inject this value elsewhere?

Env vars are managed by the registry operator - nothing should inject this env var.
It might be used when the registry is unmanaged, and in that case users have the option to use `REGISTRY_OPENSHIFT_SERVER_ADDR` instead.

PR #1584, opened 2023-09-13
WIP: CCO-412: jump azureWorkloadIdentity to the head of the line.

jstuever:
The azureWorkloadIdentity feature needs to be enabled by default in 4.14. However, lots of changes have happened in the API prior to the relevant PR merging that we do not want in 4.14. This PR re-orders all of those merge commits such that the azureWorkloadIdentity PR is at the head of the line. As a result, this change can now be brought into cluster-config-operator without bringing int he other things.


jstuever:
This might not be the right approach.

PR #1586, opened 2023-09-14
Drop removed k8s.io/api/apps/v1beta1, k8s.io/api/apps/v1beta2 and k8s…

ingvagabund:
….io/api/extensions/v1beta1.Scale removed in k8s 1.16

Related: https://issues.redhat.com/browse/OCPBUGS-3850

@stlaz FYI


dgrisonnet:
Nice, thank you for looking into that @ingvagabund

oauth/v1/legacy.go
```diff
@@ -2,14 +2,13 @@ package v1
...
 import (
 	corev1 "k8s.io/api/core/v1"
-	extensionsv1beta1 "k8s.io/api/extensions/v1beta1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
 var (
 	legacyGroupVersion            = schema.GroupVersion{Group: "", Version: "v1"}
-	legacySchemeBuilder           = runtime.NewSchemeBuilder(addLegacyKnownTypes, corev1.AddToScheme, extensionsv1beta1.AddToScheme)
```
stlaz, line 12: where does the v1 for the API live?

stlaz (commented):
split out vendor and code changes, please

oauth/v1/legacy.go
stlaz: /lgtm
looks like that turned into the apps API but I don't think we need it in the builder, hopefully

ingvagabund: @deads2k might shed more light here based on  https://github.com/openshift/api/pull/67/commits/2f6f4d85d80b3f39d66e7e1674aeff71f8a68b3d#diff-a72fa071a28b4a3a648c8e60bd0b288dadc7ff57d1b64ac1244116709f7b452cR12

apps/v1/legacy.go
```diff
@@ -21,7 +20,6 @@ func addLegacyKnownTypes(scheme *runtime.Scheme) error {
 		&DeploymentRequest{},
 		&DeploymentLog{},
 		&DeploymentLogOptions{},
-		&extensionsv1beta1.Scale{},
```
deads2k, line 24: The v1 deploymentconfig endpoint uses the extensionsv1beta1 scale, right? Does it also accept the autoscalingv1 scale?  And when this is removed, does the scale endpoint in openshift/apiserver still function properly for all inputs?

install.go
```diff
@@ -142,7 +137,6 @@ var (
 		kcoordinationv1beta1.AddToScheme,
 		keventsv1.AddToScheme,
 		keventsv1beta1.AddToScheme,
-		kextensionsv1beta1.AddToScheme,
```
deads2k, line 145: same problem with scale.  This will prevent the ability to read such input from disk for instance.  If you demonstrate workingness is all our client using it (oc and openshift-apiserver are probably the most notable), then this coudl be merged, but not before due to risk of wedging.

oauth/v1/legacy.go
deads2k, line 12: This one is probably ok since oauth doesn't use scale.

PR #1592, opened 2023-09-20
OCPCLOUD-2130: Allow missing failureDomains platform

RadekManak:
Currently, when in an environment without failure domains, the platform field still needs to be set.
```
machines_v1beta1_machine_openshift_io:
  failuredomains:
    platform: ""
```

We already support the platform being set to "". But when not setting it at all, it gets rejected by the controller.
`E0912 04:23:44.246534       1 controller.go:324]  "msg"="Reconciler error" "error"="error reconciling control plane machine set: unable to create control plane machine set: unable to create control plane machine set: ControlPlaneMachineSet.machine.openshift.io \"cluster\" is invalid: [spec.template.machines_v1beta1_machine_openshift_io.failureDomains.platform: Required value, <nil>: Invalid value: \"null\": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation]" "controller"="controlplanemachinesetgenerator" "reconcileID"="1f6615e7-703e-4c8b-b1f7-edf77b7112f0"
 `
 
 The installer does not have this check before submitting, and it gets defaulted to the above example because of the missing omitempty.


RadekManak:
We are going to make FailureDomains a pointer instead.

PR #1593, opened 2023-09-20
OPNET-351: Extend infra CR to store VIPs and MachineNetwork

mkowalski:
In order to enable work around mutable VIPs we are extending the Infrastructure CR to hold VIPs in the spec and not only in the status.

In order to be able to perform proper validations (which currently only happen in o/installer as VIPs are immutable) we are storing machine networks alongside.

Implements: [OPNET-351](https://issues.redhat.com//browse/OPNET-351)
Contributes-to: [OPNET-80](https://issues.redhat.com//browse/OPNET-80)
Contributes-to: [OPNET-340](https://issues.redhat.com//browse/OPNET-340)


mkowalski:
/cc @cybertron 
Baremetal

/cc @MaysaMacedo 
OpenStack

/cc @rvanderp3 
VSphere

mkowalski:
/retest

Prow is choking

config/v1/types_infrastructure.go
```diff
@@ -691,7 +691,32 @@ type BareMetalPlatformLoadBalancer struct {
 
 // BareMetalPlatformSpec holds the desired state of the BareMetal infrastructure provider.
 // This only includes fields that can be modified in the cluster.
-type BareMetalPlatformSpec struct{}
+type BareMetalPlatformSpec struct {
+	// apiServerInternalIPs are the IP addresses to contact the Kubernetes API
+	// server that can be used by components inside the cluster, like kubelets
+	// using the infrastructure rather than Kubernetes networking. These are the
+	// IPs for a self-hosted load balancer in front of the API servers. In dual
+	// stack clusters this list contains two IPs otherwise only one.
```
JoelSpeed, line 699: Lets make this a little more explicit 
```suggestion
	// IPs for a self-hosted load balancer in front of the API servers.
	// In dual stack clusters this list contains two IP addresses.
	// In single stack clusters a single IP address is expected.
```

config/v1/types_infrastructure.go
```diff
@@ -691,7 +691,32 @@ type BareMetalPlatformLoadBalancer struct {
...
+	//
+	// +kubebuilder:validation:Format=ip
+	// +kubebuilder:validation:MaxItems=2
+	// +listType=set
+	APIServerInternalIPs []string `json:"apiServerInternalIPs"`
```
JoelSpeed, line 704: This needs to have an optional tag right?
What happens when this list is empty?

config/v1/types_infrastructure.go
```diff
@@ -691,7 +691,32 @@ type BareMetalPlatformLoadBalancer struct {
...
+
+	// ingressIPs are the external IPs which route to the default ingress
+	// controller. The IPs are suitable targets of a wildcard DNS record used to
+	// resolve default route host names. In dual stack clusters this list
+	// contains two IPs otherwise only one.
```
JoelSpeed, line 709: Same suggestion as above applies here

config/v1/types_infrastructure.go
```diff
@@ -691,7 +691,32 @@ type BareMetalPlatformLoadBalancer struct {
...
+	//
+	// +kubebuilder:validation:Format=ip
+	// +kubebuilder:validation:MaxItems=2
+	// +listType=set
+	IngressIPs []string `json:"ingressIPs"`
```
JoelSpeed, line 714: Same as above, optional tag and what happens when omitted?

config/v1/types_infrastructure.go
```diff
@@ -691,7 +691,32 @@ type BareMetalPlatformLoadBalancer struct {
...
+
+	// IP networks used to connect all the OpenShift cluster nodes.
```
JoelSpeed, line 716: This godoc needs an update, what is the format of the MachineNetworks string?

JoelSpeed (commented):
IIUC these fields are copying from status fields? What happens if there's a diff between spec and status, you need a controller to reconcile the difference, validate and accept the change, what is going to do that?

config/v1/types_infrastructure.go
mkowalski: Of course, modified description in Spec

config/v1/types_infrastructure.go
mkowalski: Added `optional`. As unsetting all VIPs is illegal right now but maybe in a future we allow that. Or not. In either case, I am not sure, should we explicitly write it in the description here or leave it up to our product docs?

config/v1/types_infrastructure.go
mkowalski: Updated by adding `Each network is provided in the CIDR format, e.g. "192.0.2.0/24"` in the description. I had a look at `types_network.go` and in there it's not at all handled so I think the description here is already better

```
	// allowedCIDRs is the list of allowed CIDRs.
	AllowedCIDRs []string `json:"allowedCIDRs,omitempty"`
```

I am not sure if we could do any further validation of the format given we need to support IPv4 and IPv6 (I am looking at how complex regex that would produce)

config/v1/types_infrastructure.go
JoelSpeed: You should treat these godoc as documentation. So, I would expect something along the lines of explaining how these values get used.

```
// When omitted, values from the status.<field> will be used.
// Once set, cannot be removed.
```

We should add a validation using CEL for that second one.

And then on upgrade, it can be unset, and the CNO will just ignore it, but once it's set, the user can't unset it and the CNO will copy it over

mkowalski: Fixed. Now the godoc should be self-explanatory and no additional documentation needed to understand this field

config/v1/0000_10_config-operator_01_infrastructure-CustomNoUpgrade.crd.yaml
```diff
@@ -72,7 +72,35 @@ spec:
...
+                          x-kubernetes-list-type: set
+                        ingressIPs:
+                          description: ingressIPs are the external IPs which route to the default ingress controller. The IPs are suitable targets of a wildcard DNS record used to resolve default route host names. In dual stack clusters this list contains two IP addresses. In single stack clusters a single IP address is expected. When omitted, values from the status.ingressIPs will be used. Once set, the list cannot be completely removed (but its second entry can).
+                          format: ip
+                          items:
+                            type: string
+                          maxItems: 2
+                          type: array
+                          x-kubernetes-list-type: set
+                        machineNetworks:
```
MaysaMacedo, line 92: Is `machineNetworks` required once `apiServerInternalIPs` and `ingressIPs` are set? If yes, should that be documented/verified?

mkowalski: No it is not because we need to support upgraded clusters. As the only source of this information is install-config.yaml to which Cluster Network Operator has no access, it can be only o/installer who sets this field during installation time.

So for newly installed 4.15 clusters we will populate this field and for clusters that originated in 4.14 or older this field will remain empty.

mkowalski:
Hey folks, can we get approve+lgtm on this unless there are more changes requested? That would unblock the works in o/installer and CNO

MaysaMacedo:
/lgtm
For the OpenStack part

config/v1/types_infrastructure.go
```diff
@@ -691,7 +691,44 @@ type BareMetalPlatformLoadBalancer struct {
...
+	// apiServerInternalIPs are the IP addresses to contact the Kubernetes API
+	// server that can be used by components inside the cluster, like kubelets
+	// using the infrastructure rather than Kubernetes networking. These are the
+	// IPs for a self-hosted load balancer in front of the API servers.
+	// In dual stack clusters this list contains two IP addresses.
+	// In single stack clusters a single IP address is expected.
+	// When omitted, values from the status.apiServerInternalIPs will be used.
+	// Once set, the list cannot be completely removed (but its second entry can).
+	//
+	// +kubebuilder:validation:Format=ip
```
JoelSpeed, line 706: Does this work? Can you add an integration test to prove if it works or not? 

Does this support both ipv4 and ipv6?

You mention in the godoc that in dual stack clusters we expect one ipv4 and one ipv6

config/v1/types_infrastructure.go
```diff
@@ -691,7 +691,44 @@ type BareMetalPlatformLoadBalancer struct {
...
+	// +kubebuilder:validation:MaxItems=2
+	// +listType=set
+	// +optional
+	IngressIPs []string `json:"ingressIPs"`
+
+	// IP networks used to connect all the OpenShift cluster nodes. Each network
+	// is provided in the CIDR format, e.g. "192.0.2.0/24"
```
JoelSpeed, line 727: Godoc should start with the json tag name

config/v1/types_infrastructure.go
```diff
@@ -691,7 +691,44 @@ type BareMetalPlatformLoadBalancer struct {
...
+	// +listType=set
+	// +optional
+	MachineNetworks []string `json:"machineNetworks"`
```
JoelSpeed, line 730: Can we validate that this is a CIDR?

config/v1/types_infrastructure.go
```diff
@@ -744,6 +781,11 @@ type BareMetalPlatformStatus struct {
 	// +openshift:enable:FeatureSets=CustomNoUpgrade;TechPreviewNoUpgrade
 	// +optional
 	LoadBalancer *BareMetalPlatformLoadBalancer `json:"loadBalancer,omitempty"`
+
+	// IP networks used to connect all the OpenShift cluster nodes.
```
JoelSpeed, line 785: Godoc should be updated here

config/v1/types_infrastructure.go
mkowalski: Validating `is an address IPv4 or IPv6` via regex is some hardcore piece of code, like
```
'/^(?>(?>([a-f0-9]{1,4})(?>:(?1)){7}|(?!(?:.*[a-f0-9](?>:|$)){8,})((?1)(?>:(?1)){0,6})?::(?2)?)|(?>(?>(?1)(?>:(?1)){5}:|(?!(?:.*[a-f0-9]:){6,})(?3)?::(?>((?1)(?>:(?1)){0,4}):)?)?(25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])(?>\.(?4)){3}))$/iD'
```

I thought about it, then looked how other APIs validate it (hint: they don't really) and decided not to do it directly here

config/v1/types_infrastructure.go
mkowalski: `+kubebuilder:validation:Format=ip` is a copy-pasta from the already existing PlatformStatus structs

config/v1/types_infrastructure.go
mkowalski: The best existing out there seems to be `+kubebuilder:validation:Format=cidr` so I will go with that

JoelSpeed: Can we test that format CIDR actually works?

config/v1/types_infrastructure.go
JoelSpeed: Format IP doesn't work, we have other ways to validate IP addresses that we can look to add, will check what others are doing here

config/v1/types_infrastructure.go
mkowalski: Seems like it does not work, the same way that `format=ip` does not (:

config/v1/types_infrastructure.go
mkowalski: Done

config/v1/types_infrastructure.go
mkowalski: After all used `[]CIDR` with a monster regex validation

config/v1/types_infrastructure.go
mkowalski: Done

config/v1/types_infrastructure.go
mkowalski: Refactored to use custom `[]IP` with regex validation

mkowalski:
/test e2e-aws-ovn

Not my failure

```
level=info msg=Creating infrastructure resources...
level=error
level=error msg=Error: no matching EC2 Subnet found 
```

mkowalski:
/test e2e-aws-ovn

Failure of unrelated test

config/v1/types_infrastructure.go
mkowalski: Fixed

config/v1/types_infrastructure.go
```diff
@@ -1125,6 +1207,39 @@ type VSpherePlatformSpec struct {
...
+
+	// apiServerInternalIPs are the IP addresses to contact the Kubernetes API
+	// server that can be used by components inside the cluster, like kubelets
+	// using the infrastructure rather than Kubernetes networking. These are the
+	// IPs for a self-hosted load balancer in front of the API servers.
+	// In dual stack clusters this list contains two IP addresses.
+	// In single stack clusters a single IP address is expected.
+	// When omitted, values from the status.apiServerInternalIPs will be used.
+	// Once set, the list cannot be completely removed (but its second entry can).
+	//
```
JoelSpeed, line 1219: The same comment applies here that I've put in the `ingressIPs` field

config/v1/types_infrastructure.go
```diff
@@ -1102,6 +1182,8 @@ type VSpherePlatformNodeNetworking struct {
 // VSpherePlatformSpec holds the desired state of the vSphere infrastructure provider.
 // In the future the cloud provider operator, storage operator and machine operator will
 // use these fields for configuration.
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.apiServerInternalIPs) || has(self.apiServerInternalIPs)", message="Value is required once set"
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.ingressIPs) || has(self.ingressIPs)", message="Value is required once set"
```
JoelSpeed, line 1186: Please update the error messages here to explain which field you are referring to, the error coming out won't tell the user which field failed because you're adding this to the struct

config/v1/types_infrastructure.go
```diff
@@ -1176,6 +1291,11 @@ type VSpherePlatformStatus struct {
 	// +openshift:enable:FeatureSets=CustomNoUpgrade;TechPreviewNoUpgrade
 	// +optional
 	LoadBalancer *VSpherePlatformLoadBalancer `json:"loadBalancer,omitempty"`
+
+	// machineNetworks are IP networks used to connect all the OpenShift cluster nodes.
```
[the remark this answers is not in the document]
mkowalski: There is no limit on machine networks supported today. There are no rules as for the order of IP stacks in the list of machine networks today. This one field already in install-config.yaml is a bit of wild west, mainly for the reason that it's not used anywhere for real

We have a comment explaining notation added to the Spec. In Status I am not doing this because it's read only and no one should modify it

config/v1/types_infrastructure.go
```diff
@@ -1125,6 +1207,39 @@ type VSpherePlatformSpec struct {
...
+	// When omitted, values from the status.ingressIPs will be used.
+	// Once set, the list cannot be completely removed (but its second entry can).
+	//
+	// +kubebuilder:validation:MaxItems=2
+	// +listType=set
+	// +optional
+	IngressIPs []IP `json:"ingressIPs"`
+
+	// machineNetworks are IP networks used to connect all the OpenShift cluster
+	// nodes. Each network is provided in the CIDR format, e.g. "192.0.2.0/24".
```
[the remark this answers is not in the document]
mkowalski: Changed a comment to

```
	// machineNetworks are IP networks used to connect all the OpenShift cluster
	// nodes. Each network is provided in the CIDR format and should be IPv4 or IPv6,
	// for example, "10.0.0.0/8" or "fd00::/8".
```

config/v1/types_infrastructure.go
mkowalski: Fair enough, fixed

config/v1/types_infrastructure.go
mkowalski: Fair enough, fixed

config/v1/types_infrastructure.go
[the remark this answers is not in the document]
JoelSpeed: Ok. so while I appreciate it's currently wild west, what do you think users are actually using? Do you think we could justify limiting this to 32 networks? This new API will only affect new clusters anyway right?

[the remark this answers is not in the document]
JoelSpeed: I'm gonna wager we won't get that request any time soon, but you may tell me you told me so if we do. It would be relaxing the API validation so is a compatible change given that we control the consumers of the field.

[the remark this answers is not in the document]
mkowalski: Yep, it will be a very easy change to do shall we ever need it

PR #1594, opened 2023-09-20
NO-ISSUE: Fix mismatching spaces and tabs

mkowalski:
Commit 2f259e8fcccdc7eaf21a2c17709cdd6bbaefdb6a introduced a change where spaces and tabs are mixed inside one single type definition. Because of that various code editors are reporting issues and display this struct in a badly formated way.

This PR modifies `IBMCloudPlatformStatus` definition to use consistent whitespacing.


mkowalski:
/cc @cjschaef
Author of original commit

PR #1602, opened 2023-09-27
operator/v1: Move KubeScheduler CRD to run-level 50

wking:
The outgoing level 25 is from
openshift/cluster-kube-scheduler-operator@530e585f63 (openshift/cluster-kube-scheduler-operator#118).  But while it makes sense to want a scheduler early so that later pods can get scheduled, cluster bootstrapping and other outgoing clusters already have a scheduler in place.  For example, you need a functional scheduler, Deployment controller, and ReplicaSet controller to roll out [the scheduler operator][1].  By consolidating in the highly-parallel default level 50, we allow for some time savings (especially once we shift kube-controller-manager-operator from 25 to 50 as well, which I plan on doing next).

[1]: https://github.com/openshift/cluster-kube-scheduler-operator/blob/e0690f4ec17ae9dcd5de65fb49c0a810e7d46b0b/manifests/0000_25_kube-scheduler-operator_06_deployment.yaml


deads2k:
Scheduler should install and update before any other component so that new scheduling techniques can be used.

/hold

deads2k:
other pieces need to be updated, but this looks reasonable

/approve

PR #1614, opened 2023-10-11
Direct OIDC configuration

stlaz:
This is an API to https://github.com/openshift/enhancements/pull/1493


config/v1/types_authentication.go
```diff
@@ -69,6 +69,16 @@ type AuthenticationSpec struct {
 	// This allows internal components to transition to use new service account issuer without service distruption.
 	// +optional
 	ServiceAccountIssuer string `json:"serviceAccountIssuer"`
+
+	// OIDCProviders are OIDC identity providers that can issue tokens
+	// for this cluster
+	// Can only be set if "Type" is set to "OIDC".
```
deads2k, line 75: does type also impact WebhookTokenAuthenticator ?

config/v1/types_authentication.go
```diff
@@ -159,3 +170,74 @@ const (
...
+	// ClaimMappings describes rules on how to transform information from an
+	// ID token into a cluster identity
+	ClaimMappings TokenClaimMappings `json:"claimMappings"`
+}
+
+type TokenIssuer struct {
+	// URL is the serving URL of the token issuer.
+	// Must use the https:// scheme.
+	//
+	// +kubebuilder:validation:Required
```
deads2k, line 190: min length 9.  required allows ""

config/v1/types_authentication.go
```diff
@@ -159,3 +170,74 @@ const (
...
+	// +required
+	URL string `json:"url"`
+
+	// Audiences is an array of audiences that the token was issued for.
+	// Valid tokens must include at least one of these values in their
+	// "aud" claim.
+	// Must be set to exactly one value.
+	//
+	// +listType=set
+	// +kubebuilder:validation:Required
```
deads2k, line 200: can I have one item and be "" or is that invalid?  if invalid, we'll have to work out how to enforce.

config/v1/types_authentication.go
```diff
@@ -159,3 +170,74 @@ const (
...
+	// +kubebuilder:validation:MaxItems=1
+	// +required
+	Audiences []string `json:"audiences"`
+
+	// CertificateAuthority is a reference to a config map in the
+	// configuration namespace. The .data of the configMap must contain
+	// the "ca.crt" key.
+	CertificateAuthority ConfigMapNameReference `json:"certificateAuthority,omitempty"`
```
deads2k, line 208: no omit empty.

If this is the zero-object, what happens?  Invalid or "use system certs"?  I suggest "use system certs"

config/v1/types_authentication.go
```diff
@@ -159,3 +170,74 @@ const (
...
+	// Default value: "sub"
+	Username PrefixedTokenClaimMapping `json:"username,omitempty"`
+
+	// Groups is a name of the claim that should be used to construct
+	// groups for the cluster identity.
+	// The referenced claim must use array of strings values.
+	Groups PrefixedTokenClaimMapping `json:"groups,omitempty"`
+}
+
+type TokenClaimMapping struct {
```
deads2k, line 229: prefix to come once we hit beta?  Logically, prefix and claim and mutually exclusive?

config/v1/types_authentication.go
```diff
@@ -159,3 +170,74 @@ const (
...
+	// Claim is a JWT token claim to be used
+	//
+	// +kubebuilder:validation:Required
+	// +required
+	Claim string `json:"claim"`
+}
+
+type PrefixedTokenClaimMapping struct {
+	TokenClaimMapping `json:",inline"`
```
deads2k, line 238: why nest this way?

config/v1/types_authentication.go
```diff
@@ -159,3 +170,74 @@ const (
...
+
+	// Prefix is a string to prefix the value of the token in the result of the
```
deads2k, line 240: need better doc here.  This is prefixing the kube username and has a default?  Will we prefix the issuer or pass `-`?

config/v1/types_authentication.go
```diff
@@ -159,3 +170,74 @@ const (
 	// KubeConfigKey is the key for the kube config file data in a secret
 	KubeConfigKey = "kubeConfig"
 )
+
+type OIDCProvider struct {
```
deads2k, line 174: will we support required claims?

stlaz: I did not want to add that just yet and instead in case somebody needs it

config/v1/types_authentication.go
stlaz: That would be system certs. Good point, that should be documented.

config/v1/types_authentication.go
stlaz: Which way would make more sense? The struct extends plain `TokenClaimMapping`.

config/v1/types_authentication.go
stlaz: The default for prefix is different for groups and username :neutral_face:

config/v1/types_authentication.go
stlaz: Documented at `WebhookTokenAuthenticator`

config/v1/types_authentication.go
stlaz: Added in the last patch

config/v1/types_authentication.go
```diff
@@ -159,3 +170,74 @@ const (
...
+	Issuer TokenIssuer `json:"issuer"`
+
+	// ClaimMappings describes rules on how to transform information from an
+	// ID token into a cluster identity
+	ClaimMappings TokenClaimMappings `json:"claimMappings"`
+}
+
+type TokenIssuer struct {
+	// URL is the serving URL of the token issuer.
+	// Must use the https:// scheme.
```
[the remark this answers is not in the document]
stlaz: added regex

config/v1/types_authentication.go
stlaz: expressed by a regex

config/v1/types_authentication.go
stlaz: Added `TokenAudiences` type with MinLength restriction

config/v1/types_authentication.go
```diff
@@ -159,3 +175,105 @@ const (
...
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxItems=1
+	// +required
+	Audiences []TokenAudience `json:"audiences"`
+
+	// CertificateAuthority is a reference to a config map in the
+	// configuration namespace. The .data of the configMap must contain
+	// the "ca.crt" key.
+	// If unset, system trust is used instead.
+	CertificateAuthority ConfigMapNameReference `json:"certificateAuthority"`
```
deads2k, line 229: @sjenning what does this do on hypershift?

config/v1/types_authentication.go
deads2k: > The default for prefix is different for groups and username

Two different types then.

config/v1/types_authentication.go
```diff
@@ -159,3 +175,112 @@ const (
...
+
+type PrefixedClaimMapping struct {
+	TokenClaimMapping `json:",inline"`
+
+	// Prefix is a string to prefix the value from the token in the result of the
+	// claim mapping
+	Prefix string `json:"prefix"`
+}
+
+type TokenClaimValidationRule struct {
```
deads2k, line 271: discriminated union for "simple" and expression.

config/v1/types_authentication.go
```diff
@@ -159,3 +175,130 @@ const (
 	// KubeConfigKey is the key for the kube config file data in a secret
 	KubeConfigKey = "kubeConfig"
 )
+
+type OIDCProvider struct {
```
deads2k, line 179: Is this the right scope for configuring a clientID and reference to a clientSecret for the console

Pros:
1. keeps configuration close together from a user perspective
2. will allow different clientIDs for different oidc providers for console

Cons?

config/v1/types_authentication.go
```diff
@@ -69,6 +72,17 @@ type AuthenticationSpec struct {
...
+	// OIDCProviders are OIDC identity providers that can issue tokens
+	// for this cluster
+	// Can only be set if "Type" is set to "OIDC".
+	//
+	// At most one provider can be configured.
+	//
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=1
+	OIDCProviders []OIDCProvider `json:"oidcProviders,omitempty"`
```
deads2k, line 85: Gate with techpreview: https://github.com/openshift/api/blob/master/example/v1/types_stable.go#L36-L39

config/v1/types_authentication.go
```diff
@@ -159,3 +175,130 @@ const (
...
+	// Default value: "sub"
+	Username UsernameClaimMapping `json:"username,omitempty"`
+
+	// Groups is a name of the claim that should be used to construct
+	// groups for the cluster identity.
+	// The referenced claim must use array of strings values.
+	Groups PrefixedClaimMapping `json:"groups,omitempty"`
+}
+
+type TokenClaimMapping struct {
```
deads2k, line 245: What else will go in here?

config/v1/types_authentication.go
```diff
@@ -159,3 +175,130 @@ const (
...
+	// Claim is a JWT token claim to be used in the mapping
+	//
+	// +kubebuilder:validation:Required
+	// +required
+	Claim string `json:"claim"`
+}
+
+type UsernameClaimMapping struct {
+	TokenClaimMapping `json:",inline"`
```
deads2k, line 254: prefer the inline string here.  Or if we're going to have a type for a claim, make it an alias for string unless something else goes in the struct.

config/v1/types_authentication.go
```diff
@@ -159,3 +175,130 @@ const (
...
+	// Set to "-" to disable prefixing
+	Prefix string `json:"prefix"`
+}
+
+type PrefixedClaimMapping struct {
+	TokenClaimMapping `json:",inline"`
+
+	// Prefix is a string to prefix the value from the token in the result of the
+	// claim mapping
+	Prefix string `json:"prefix"`
```
deads2k, line 268: what is the behavior if unset?

config/v1/types_authentication.go
```diff
@@ -159,3 +175,130 @@ const (
...
+	// prevent naming clashes with other plugins.
+	//
+	// Set to "-" to disable prefixing
+	Prefix string `json:"prefix"`
+}
+
+type PrefixedClaimMapping struct {
+	TokenClaimMapping `json:",inline"`
+
+	// Prefix is a string to prefix the value from the token in the result of the
```
deads2k, line 266: provide an example.  I only know what this does because I read the code.

config/v1/types_authentication.go
```diff
@@ -159,3 +175,130 @@ const (
...
+	//
+	// +kubebuilder:validation:Required
+	// +required
+	Claim string `json:"claim"`
+}
+
+type UsernameClaimMapping struct {
+	TokenClaimMapping `json:",inline"`
+
+	// By default, claims other than `email` will be prefixed with the issuer URL to
```
deads2k, line 256: provide an example.  I only know what this does because I read the code.  People must know in order to write RBAC rules.

config/v1/types_authentication.go
```diff
@@ -159,3 +175,130 @@ const (
...
+	//
+	// +listType=set
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxItems=1
+	// +required
+	Audiences []TokenAudience `json:"audiences"`
+
+	// CertificateAuthority is a reference to a config map in the
+	// configuration namespace. The .data of the configMap must contain
+	// the "ca.crt" key.
```
deads2k, line 227: we normally use ca-bundle.crt

config/v1/types_authentication.go
```diff
@@ -159,3 +175,130 @@ const (
...
+	// If unset, system trust is used instead.
+	CertificateAuthority ConfigMapNameReference `json:"certificateAuthority"`
```
deads2k, line 229:
```suggestion
	CertificateAuthority ConfigMapNameReference `json:"issuerCertificateAuthority"`
```

config/v1/types_authentication.go
```diff
@@ -159,3 +175,130 @@ const (
...
+type TokenAudience string
+
+type TokenIssuer struct {
+	// URL is the serving URL of the token issuer.
+	// Must use the https:// scheme.
+	//
+	// +kubebuilder:validation:Pattern=`^https:\/\/[^\s]`
+	// +kubebuilder:validation:Required
+	// +required
+	URL string `json:"url"`
```
deads2k, line 212:
```suggestion
	URL string `json:"issuerURL"`
```

deads2k:
a few final comments.  I think this is good to squash up and merge after

config/v1/types_authentication.go
```diff
@@ -69,6 +72,17 @@ type AuthenticationSpec struct {
...
+
+	// OIDCProviders are OIDC identity providers that can issue tokens
+	// for this cluster
+	// Can only be set if "Type" is set to "OIDC".
+	//
+	// At most one provider can be configured.
+	//
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=1
```
sjenning, line 84: i assume this is temporary if upstream adds multiple? otherwise, what is the point of a slice limited to a singe entry?

config/v1/types_authentication.go
```diff
@@ -159,3 +175,130 @@ const (
...
+	URL string `json:"url"`
+
+	// Audiences is an array of audiences that the token was issued for.
+	// Valid tokens must include at least one of these values in their
+	// "aud" claim.
+	// Must be set to exactly one value.
+	//
+	// +listType=set
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxItems=1
```
sjenning, line 221: why `MaxItems=1`?

config/v1/types_authentication.go
```diff
@@ -159,3 +175,130 @@ const (
...
+	// +required
+	Audiences []TokenAudience `json:"audiences"`
```
sjenning, line 223: to what OIDC KAS flag does this map?

config/v1/types_authentication.go
stlaz: Because the current configuration does not allow anything else. This is future proofing for https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/3331-structured-authentication-configuration

config/v1/types_authentication.go
stlaz: Currently, that's `oidc-client-id`

config/v1/types_authentication.go
stlaz: +1, I'm for keeping this just "url"

config/v1/types_authentication.go
stlaz: "ca.crt" is usually used in kube. I can change that to what we use in OpenShift if you want.

config/v1/types_authentication.go
stlaz: +1, I don't see a reason to prefix these.

config/v1/types_authentication.go
stlaz: CEL expressions for the mapping to keep it similar to upstream - https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/3331-structured-authentication-configuration#configuration-file

config/v1/types_authentication.go
stlaz: if unset, don't prefix

config/v1/types_authentication.go
stlaz: yep, that's a future preparation

config/v1/types_authentication.go
stlaz: Each client for each application should be different, or should be provided with means to request tokens with different audiences (e.g. the `resource` parameter). That means that should we choose to configure client parameters via this API, we need to make it configurable for a number of clients. That can get messy, but is still doable.

config/v1/types_authentication.go
deads2k: Accepted

config/v1/types_authentication.go
deads2k: please add to godoc

config/v1/types_authentication.go
```diff
@@ -159,3 +176,144 @@ const (
...
+	// By default, claims other than `email` will be prefixed with the issuer URL to
+	// prevent naming clashes with other plugins.
+	//
+	// Set to "-" to disable prefixing.
+	//
+	// Example:
+	//     (1) `prefix` is set to `myoidc:` and `claim` is set to "username".
+	//         If the JWT claim `username` contains value `userA`, the resulting
+	//         mapped value will be "myoidc:userA".
+	//     (2) `prefix` is set to `myoidc:` and `claim` is set to "email". If the
```
deads2k, line 266: really?  this is weird.  So you can set the value and it is ignored?

config/v1/types_authentication.go
```diff
@@ -159,3 +176,144 @@ const (
...
+
+type UsernameClaimMapping struct {
+	TokenClaimMapping `json:",inline"`
+
+	// By default, claims other than `email` will be prefixed with the issuer URL to
+	// prevent naming clashes with other plugins.
+	//
+	// Set to "-" to disable prefixing.
+	//
+	// Example:
```
deads2k, line 262: examples of defaulting.

config/v1/types_authentication.go
```diff
@@ -159,3 +176,144 @@ const (
...
+	//     (1) `prefix` is set to `myoidc:` and `claim` is set to "username".
+	//         If the JWT claim `username` contains value `userA`, the resulting
+	//         mapped value will be "myoidc:userA".
+	//     (2) `prefix` is set to `myoidc:` and `claim` is set to "email". If the
+	//         JWT `email` claim contains value "userA@myoidc.tld", the resulting
+	//         mapped value will be "userA@myoidc.tld".
+	Prefix string `json:"prefix"`
```
deads2k, line 269: playing with 


```go

type PrefixPolicy string

var(
  // NoOpinion let's the cluster assign prefixes.  If the username claim is email, there is no prefix
  // If the username claim is anything else, it is prefixed by the issuerURL
  NoOpinion PrefixPolicy = ""

  // NoPrefix means the username claim value will not have any  prefix
  NoPrefix PrefixPolicy = "NoPrefix"

  // Prefix means the prefix value must be specified.  It cannot be empty
  Prefix PrefixPolicy = "Prefix"
)

type UsernamePrefix struct{
  PrefixPolicy PrefixPolicy

  Prefix *Prefix
}

type Prefix struct{
  // required
  // minLength=1
  Prefix string
}

```

This structure would allow NoOpinion to match the kube-apiserver doc (I think).  And would make a selection here clearly indicate intent.

config/v1/types_authentication.go
stlaz: I misunderstood the docs, the value would not be ignored when specifically set. Let me fix that

config/v1/types_authentication.go
```diff
@@ -159,3 +176,175 @@ const (
...
+	//         JWT `email` claim contains value "userA@myoidc.tld", the resulting
+	//         mapped value will be "myoidc:userA@myoidc.tld".
+	//     (3) `prefix` is unset, `issuerURL` is set to `https://myoidc.tld`,
+	//         the JWT claims include "username":"userA" and "email":"userA@myoidc.tld",
+	//         and `claim` is set to:
+	//         (a) "username": the mapped value will be "https://myoidc.tld#userA"
+	//         (b) "email": the mapped value will be "userA@myoidc.tld"
+	//
+	// +kubebuilder:validation:Enum={"", "NoPrefix", "Prefix"}
+	PrefixPolicy UsernamePrefixPolicy
```
deads2k, line 278: missing json tag

config/v1/types_authentication.go
```diff
@@ -159,3 +178,177 @@ const (
...
+	//     (3) `prefix` is unset, `issuerURL` is set to `https://myoidc.tld`,
+	//         the JWT claims include "username":"userA" and "email":"userA@myoidc.tld",
+	//         and `claim` is set to:
+	//         (a) "username": the mapped value will be "https://myoidc.tld#userA"
+	//         (b) "email": the mapped value will be "userA@myoidc.tld"
+	//
+	// +kubebuilder:validation:Enum={"", "NoPrefix", "Prefix"}
+	PrefixPolicy UsernamePrefixPolicy `json:"prefixPolicy"`
+
+	Prefix *UsernamePrefix `json:"prefix"`
```
deads2k, line 283: There is still no CEL here.

deads2k: Test should be able to prove CEL is missing.

deads2k: Oh, I see my review mistake.  Realized it in the shower this morning.  Agreed.

PR #1615, opened 2023-10-11
MON-3410: Add FeatureGate for MetricsServer

slashpai:
enhancement proposal: openshift/enhancements#1426

This feature will TechPreview in OCP 4.15 hence adding a feature gate where the feature is to be enabled only on TechPreview Clusters and not in standard cluster.

Goal is to disable the feature by default in 4.15 but enable it only in TechPreviewNoUpgrade clusters

cc: @simonpasquier


slashpai:
@simonpasquier PTAL

slashpai:
rebased

slashpai:
@deads2k Could you please review?

PR #1616, opened 2023-10-11
OCPBUGS-20164: Capability Annotation for Build CRD

adambkaplan:
Add missing `capability.openshift.io/name` annotation to the Build cluster configuration CRD. This will ensure that the config object for the Build sub-system is not created when the Build capability is not enabled.

JIRA ID: OCPBUGS-20164


adambkaplan:
Related implentation PR: https://github.com/openshift/cluster-config-operator/pull/360

bparees:
/approve

i know you're still working on this but making sure you're not blocked on the api approval portion anyway.

wking:
Trying to shake Tide loose.

/skip
/retest-required

bparees:
luckily i think we can consider this PR moot since we now need the rename PR anyway.

PR #1617, opened 2023-10-12
OCPBUGS-20439: Capability Annotation for Build CRD

adambkaplan:
Add missing `capability.openshift.io/name` annotation to the Build cluster configuration CRD. This will ensure that the config object for the Build sub-system is not created when the Build capability is not enabled.

This is a cherrypick of #1616


adambkaplan:
Vendor update for cluster-config-operator: https://github.com/openshift/cluster-config-operator/pull/363

bparees:
manually labeling this as valid-bug, we're going to merge out of order (merge to 4.14 first, then follow with 4.15) due to time constraints for getting this into 4.14.0.  There is ample time to avoid any regression concerns.

bparees:
/approve
/lgtm
/hold

to be unheld upon completion of cluster-bot testing

PR #1618, opened 2023-10-12
OCPBUGS-20164: Capabilities fix for Build CRD

adambkaplan:
Marking the Build CRD as associated with the "Build" capability and  "owned" by openshift-controller-manager-
operator. This means that its CRD manifest will be included in the ocm-o image, with the right annotations to mark it as part of the "Build" capability.

Follow up on #1616


adambkaplan:
Also requires https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/306 for everything to work

adambkaplan:
Removing WIP - confirmed with the linked PRs that the Build CRD does not show up when a cluster installs with no capabilities.

```
$ oc get build
error: the server doesn't have a resource type "build"
$ oc api-resources | grep "build"
$ 
```

bparees:

job is complaining because we renamed+touched the CRD, but the complaints are not about the actual changes being made to the CRD and they are issues w/ the CRD that (at least mostly) can't/won't be fixed at this point (the CRD predates the api rules in question)

PR #1596, opened 2023-09-21
MCO-821: Introduce MachineConfigNode types

cdoern:
For Openshift 4.15, the MCO is working on a new state reporting mechanism: The MachineConfigStateController.  The main type behind this architecture is the MachineConfigNode. A MachineConfigNode tracks an upgrade progression that relates behaviors between a node, its machineconfigpool, and the MCD processes within the MCO. 

Each of these will have its own CR in the MCO which necessitates a CRD that I have also included in this PR.

where doing `oc describe machineconfignode/<name>` will of course display more in depth information of the type `metav1.Condition`. These conditions are composed of a state, phase, reason, and time. The state refers to what part of the upgrade we are in (UpdatePreparing, Completing etc), Phase is a in depth description of what part of a State we are in, and the Reason is a quick one or two word statement describing what "Stopping Point" we are at: ReconcilingConfigs, StoppingCfgDrift. The following accepted Condition types are:

```
	// MachineConfigNodeUpdatePrepared describes a machine that is preparing in the daemon to trigger an update
	MachineConfigNodeUpdatePrepared StateProgress = "UpdatePrepared"
	// MachineConfigNodeUpdateExecuted describes a machine that has executed the body of the upgrade
	MachineConfigNodeUpdateExecuted StateProgress = "UpdateExecuted"
	// MachineConfigNodeUpdatePostActionComplete describes a machine that has executed its post update action
	MachineConfigNodeUpdatePostActionComplete StateProgress = "UpdatePostActionComplete"
	// MachineConfigNodeUpdateComplete describes a machine that has completed the core parts of an upgrade.
	MachineConfigNodeUpdateComplete StateProgress = "UpdateComplete"
	// MachineConfigNodeUpdated describes a machine that has a matching desired and current config after executing an update
	MachineConfigNodeUpdated StateProgress = "Updated"
	// MachineConfigNodeUpdateResumed describes a machine that has resumed normal processes
	MachineConfigNodeResumed StateProgress = "Resumed"
	// MachineConfigNodeUpdateCompatible the part of the preparing phase where the mco decides whether it can update
	MachineConfigNodeUpdateCompatible StateProgress = "UpdateCompatible"
	// MachineConfigNodeUpdateDrained describes the part of the inprogress phase where the node drains
	MachineConfigNodeUpdateDrained StateProgress = "Drained"
	// MachineConfigNodeUpdateFilesAndOS describes the part of the inprogress phase where the nodes file and OS config change
[...]

PR #1628, opened 2023-10-18
add featuregate for MachineConfigNodes

cdoern:
This adds a featuegate for the work on openshift/api#1596 which I will then bring into the MCO. Keeping this separate as it can merge before the types get into the API.


JoelSpeed (approved):
/lgtm

PR #1639, opened 2023-10-26
Cleanup OWNERS file

bparees:


bparees:
@mfojtik leaving you in for now since you're at least still around and in the org, but if you don't think you'd ever actually approve anything then let me know and we should remove you.  We need to get more disciplined about understanding who our "active" approvers are/how many we have.

I think this file is still a bit optimistic (even with this cleanup) because there are only two @openshift/api-reviewers  (https://github.com/orgs/openshift/teams/api-reviewers) (though @JoelSpeed should probably be in that list, so 3, but i doubt derek has approved anything in recent memory, so 2).  The rest of the people who are approvers here might approve a small/simple api change or mechanical fix, but are not going to sign off on a substantial new api or change in most cases without feedback from @openshift/api-reviewers.  An unscientific survey of recent PRs to this repo largely confirms that.

/assign @deads2k

bparees:
according to some github scraping, out of the current list of approvers who are still with redhat, here's who has `/approve`d a PR in the last year:

```
deads2k
JoelSpeed
knobunc
mfojtik
sjenning
bparees
```

of course that still doesn't distinguish between "approved a typo fix" and "approved a new api contribution" but at least it means they were paying attention and responded when pinged/requested to approve something.

bparees:
@deads2k bump

deads2k:
I'll prune further and remove derek and michal.  Can be followup.

/lgtm

mfojtik:
Feel free to remove me :-)

PR #1641, opened 2023-10-27
AUTH-439: add an API for openID connect clients config

stlaz:
I chose adding the API to the global OIDC config, mainly because:
1. it saves us pain from referring to the name of an OIDC provider should it appear outside the `OIDCProvider` API
2. it positions the API next to the issuer and issuer CA information that would still be required for any client configuration

/assign @deads2k 
/cc @sjenning


config/v1/types_authentication.go
```diff
@@ -192,6 +192,14 @@ type OIDCProvider struct {
 	// +required
 	Issuer TokenIssuer `json:"issuer"`
 
+	// OIDCClients contains configuration for the platform's clients that
+	// need to request tokens from the issuer
+	//
+	// +listType=map
+	// +listMapKey=component.name
```
stlaz, line 199: @deads2k - would you know if this actually works this way?

deads2k: dont' know.  You'll need to create it on real cluster and try using `oc apply`

deads2k, line 199: name and namespace, right?  I think you just want Component?  Does that not work?

stlaz: the docs say mapKey must be of scalar type

stlaz: Unfortunately, nesting does not seem to work. I inlined it to `componentName` and `componentNamespace`

config/v1/types_authentication.go
```diff
@@ -253,6 +266,106 @@ type TokenClaimMapping struct {
...
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Required
+	// +required
+	ClientID string `json:"clientID"`
+
+	// ClientSecret refers to a secret in the `openshift-config` namespace that
+	// contains the client secret in the `clientSecret` key of the `.data` field
+	ClientSecret SecretNameReference `json:"clientSecret"`
+
+	// ExtraScopes is an optional set of scopes to request tokens with.
```
deads2k, line 297: Doc an example of when I'd want to set this.  It's not clear to me why a cluster-admin would have knowledge of scopes required that a component does not.

config/v1/types_authentication.go
```diff
@@ -192,6 +197,14 @@ type OIDCProvider struct {
 	// +required
 	Issuer TokenIssuer `json:"issuer"`
 
+	// OIDCClients contains configuration for the platform's clients that
+	// need to request tokens from the issuer
+	//
+	// +listType=map
+	// +listMapKey=componentNamespace
+	// +listMapKey=componentName
+	OIDCClients []OIDCClientConfig `json:"oidcClients"`
```
deads2k, line 206: My gut says it's possible to write a CEL expression at the `Authentication` level that requires items in the spec list to either

stlaz: my first attempt is apparently only 50.3x more complex than the rule cost allows us if we limit all OIDCClients to max 20
```
self.spec.oidcProviders.all(
  p, p.oidcClients.all(
    specC, self.status.oidcClients.exists(
      statusC, statusC.componentNamespace == specC.componentNamespace && statusC.componentName == specC.componentName)
      || oldSelf.spec.oidcProviders.exists(
            oldP, oldP.oidcClients.exists(
            oldC, oldC.componentNamespace == specC.componentNamespace && oldC.componentName == specC.componentName))))
```
I'll try to see if we can optimize it somehow but I'm not sure we'd fit within the rule cost budget once we allow multiple OIDC providers anyway :neutral_face:

stlaz: I was hoping that we could perhaps try to use the fact that the arrays are meant to be accessed as maps but as I originally suspected, that's apparently only useful for SSA.

stlaz:
@deads2k the CEL has now been added, along with tests. However, the CEL tends to test the "Stable" version of the Authentication struct which does not contain the fields that appear in the CEL.

Do we have a way to solve that with the featuregate flags?

config/v1/types_authentication.go
```diff
@@ -6,6 +6,7 @@ import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 // +genclient:nonNamespaced
 // +kubebuilder:subresource:status
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:validation:XValidation:rule="self.spec.oidcProviders.all(p, !has(p.oidcClients) || p.oidcClients.all(specC, self.status.oidcClients.exists(statusC, statusC.componentNamespace == specC.componentNamespace && statusC.componentName == specC.componentName) || (has(oldSelf.spec.oidcProviders) && oldSelf.spec.oidcProviders.exists(oldP, oldP.name == p.name && has(oldP.oidcClients) && oldP.oidcClients.exists(oldC, oldC.componentNamespace == specC.componentNamespace && oldC.componentName == specC.componentName)))))",message="all oidcClients in the oidcProviders must match their componentName and componentNamespace to either a previously configured oidcClient or they must exist in the status.oidcClients"
```
deads2k, line 9: Try `openshift:validation:FeatureSetAwareXValidation` Example here: https://github.com/openshift/api/blob/master/example/v1/types_stable.go#L32

config/v1/techpreview.authentication.testsuite.yaml
```diff
@@ -108,3 +108,166 @@ tests:
             username:
               claim: "preferred_username"
               prefixPolicy: NoPrefix
+  onUpdate:
```
deads2k, line 111: tests appear to do what I expect.

config/v1/types_authentication.go
```diff
@@ -104,8 +105,14 @@ type AuthenticationStatus struct {
...
-	// TODO if we add support for an in-cluster operator managed Keycloak instance
-	// KeycloakOAuthMetadata ConfigMapNameReference `json:"keycloakOAuthMetadata"`
+	// OIDCClients is where participating operators place the current OIDC client status
+	// for OIDC clients that can be customized by the cluster-admin.
+	//
+	// +listType=map
+	// +listMapKey=componentNamespace
+	// +listMapKey=componentName
+	// +kubebuilder:validation:MaxItems=20
+	OIDCClients []OIDCClientStatus `json:"oidcClients"`
```
deads2k, line 115:
```suggestion
	// +listType=map
	// +listMapKey=componentNamespace
	// +listMapKey=componentName
	// +kubebuilder:validation:MaxItems=20
	// +openshift:enable:FeatureSets=CustomNoUpgrade;TechPreviewNoUpgrade
	OIDCClients []OIDCClientStatus `json:"oidcClients"`
```

somehow forgot/lost this

config/v1/types_authentication.go
```diff
@@ -192,6 +199,15 @@ type OIDCProvider struct {
...
 	Issuer TokenIssuer `json:"issuer"`
 
+	// OIDCClients contains configuration for the platform's clients that
+	// need to request tokens from the issuer
+	//
+	// +listType=map
+	// +listMapKey=componentNamespace
+	// +listMapKey=componentName
+	// +kubebuilder:validation:MaxItems=20
+	OIDCClients []OIDCClientConfig `json:"oidcClients"`
```
deads2k, line 209:
```suggestion
	// +listType=map
	// +listMapKey=componentNamespace
	// +listMapKey=componentName
	// +openshift:enable:FeatureSets=CustomNoUpgrade;TechPreviewNoUpgrade
	// +kubebuilder:validation:MaxItems=20
	OIDCClients []OIDCClientConfig `json:"oidcClients"`
```

somehow forgot/lost this

stlaz: the whole struct is featuregated though

PR #1649, opened 2023-11-02
STOR-1487: IBMCloud: Add CSIDriverConfig support for IBM Cloud

cjschaef:
Add support to the CSIDriverConfig to include encyrption key for storage class volumes on IBM Cloud.

Related: https://issues.redhat.com/browse/STOR-1487


operator/v1/types_csi_cluster_driver.go
```diff
@@ -248,6 +252,15 @@ type GCPCSIDriverConfigSpec struct {
...
 }
 
+// IBMCloudCSIDriverConfigSpec defines the properties that can be configured for the IBM Cloud CSI driver.
+type IBMCloudCSIDriverConfigSpec struct {
+	// encryptionKeyCRN is the IBM Cloud CRN of the customer-managed root key to use
+	// for disk encryption of volumes for the default storage class(es).
+	// +kubebuilder:validation:Pattern:=`^crn:v[0-9]+:bluemix:(public|private):(kms|hs-crypto):[a-z-]+:a/[0-9a-f]+:[0-9a-f-]{36}:key:[0-9a-f-]{36}$`
+	// +optional
+	EncryptionKeyCRN string `json:"encryptionKeyCRN,omitempty"`
+}
```
dobsonj, line 262: Sharing an example CRN here, because I wasn't familiar with the format:
```
crn:v1:bluemix:public:kms:us-south:a/18240e5ed3f647cb96ca52dc92d9addd:d5e11420-3f0f-447a-92c0-13c978cf9096:key:c4a72f5b-9412-4ce8-80a4-2acb1987f3ba
```
The validation regex looks reasonable to me for this example, but is it valid to reference a key alias in this field? What does a key alias CRN look like, and should this field allow it?

openapi/openapi.json
```diff
@@ -26086,6 +26091,16 @@
         }
       }
     },
+    "com.github.openshift.api.operator.v1.IBMCloudCSIDriverConfigSpec": {
+      "description": "IBMCloudCSIDriverConfigSpec defines the properties that can be configured for the IBM Cloud CSI driver.",
+      "type": "object",
+      "properties": {
+        "encryptionKeyCRN": {
+          "description": "encryptionKeyCRN is the IBM Cloud CRN of the customer-managed root key to use for disk encryption of volumes for the default storage class(es).",
```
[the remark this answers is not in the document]
dobsonj: Perhaps a more general question: Will this field be used _only_ for IBM VPC? Or will it be used for the PowerVS platform as well?

dobsonj:
I left a few comments, but this will need to be reviewed by an API approver too.
/assign @JoelSpeed

dobsonj:
Just for context: this expands on the work in https://github.com/openshift/api/pull/1401 with the goal of supporting BYOK on IBM Cloud.

operator/v1/types_csi_cluster_driver.go
```diff
@@ -248,6 +252,15 @@ type GCPCSIDriverConfigSpec struct {
 	KMSKey *GCPKMSKeyReference `json:"kmsKey,omitempty"`
 }
 
+// IBMCloudCSIDriverConfigSpec defines the properties that can be configured for the IBM Cloud CSI driver.
+type IBMCloudCSIDriverConfigSpec struct {
+	// encryptionKeyCRN is the IBM Cloud CRN of the customer-managed root key to use
+	// for disk encryption of volumes for the default storage class(es).
+	// +kubebuilder:validation:Pattern:=`^crn:v[0-9]+:bluemix:(public|private):(kms|hs-crypto):[a-z-]+:a/[0-9a-f]+:[0-9a-f-]{36}:key:[0-9a-f-]{36}$`
```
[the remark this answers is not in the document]
cjschaef: Sure, I should be able to come up with a max length, or best assumption, for the string, along with a min length, to try to confirm we have a CRN and not just a subset of values.

operator/v1/types_csi_cluster_driver.go
```diff
@@ -143,6 +143,10 @@ type CSIDriverConfigSpec struct {
 	// +optional
 	GCP *GCPCSIDriverConfigSpec `json:"gcp,omitempty"`
 
+	// ibmcloud is used to configure the IBM Cloud CSI driver.
+	// +optional
+	IBMCloud *IBMCloudCSIDriverConfigSpec `json:"ibmcloud,omitempty"`
```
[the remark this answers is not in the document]
cjschaef: Added a validation to the `CSIDriverConfigSpec` for IBMCloud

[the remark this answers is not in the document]
cjschaef: Added.

openapi/openapi.json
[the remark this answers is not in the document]
cjschaef: Updated.

I am unsure ATM. I would have to discuss with PowerVS to determine their expectations and how they would approach this feature (only need CRN, supported key encryption services, etc.).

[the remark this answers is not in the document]
cjschaef: It sounds like due to limitations and time restraints, we will keep this targeted for IBM Cloud VPC (x86_64) and allow PowerVS to piggyback if they desire this support and find they have the same requirements (only a CRK CRN) to implement. Likely in a follow up release.

operator/v1/types_csi_cluster_driver.go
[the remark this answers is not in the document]
dobsonj: I see the new MinLength and MaxLength rules, but now the regex is gone. Was this intentional?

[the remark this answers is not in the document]
cjschaef: I think that was the request, otherwise I can restore the regex along with the min/max length.

operator/v1/types_csi_cluster_driver.go
```diff
@@ -109,23 +109,25 @@ type ClusterCSIDriverSpec struct {
...
+	AWSDriverType      CSIDriverType = "AWS"
+	AzureDriverType    CSIDriverType = "Azure"
+	GCPDriverType      CSIDriverType = "GCP"
+	IBMCloudDriverType CSIDriverType = "IBMCloud"
+	VSphereDriverType  CSIDriverType = "vSphere"
 )
 
 // CSIDriverConfigSpec defines configuration spec that can be
 // used to optionally configure a specific CSI Driver.
+// +kubebuilder:validation:XValidation:rule="has(self.driverType) && self.driverType == 'IBMCloud' ? (has(self.ibmcloud) && size(self.ibmcloud.encryptionKeyCRN) > 0) : !has(self.ibmcloud)",message="ibmcloud must be set if driverType is 'IBMCloud', but remain unset otherwise"
```
JoelSpeed, line 125: Why `encryptionKeyCRN > 0` in this? That implies to me that the encryption key should be a required field?

operator/v1/types_csi_cluster_driver.go
```diff
@@ -248,6 +254,16 @@ type GCPCSIDriverConfigSpec struct {
...
 }
 
+// IBMCloudCSIDriverConfigSpec defines the properties that can be configured for the IBM Cloud CSI driver.
+type IBMCloudCSIDriverConfigSpec struct {
+	// encryptionKeyCRN is the IBM Cloud CRN of the customer-managed root key to use
+	// for disk encryption of volumes for the default storage classes.
+	// +kubebuilder:validation:MaxLength:=154
+	// +kubebuilder:validation:MinLength:=144
+	// +optional
+	EncryptionKeyCRN string `json:"encryptionKeyCRN,omitempty"`
```
JoelSpeed, line 264: I'm guessing given the precise length, that there's a pattern for this CRN? What is it? What characters are allowed? Can we add a regex?

cjschaef: Yes, the IBM CRN object is defined here (sorry, I didn't see any docs with more helpful details)
https://github.com/IBM-Cloud/bluemix-go/blob/master/crn/crn.go

I can add back the regex I had previously, in additional to the min/max length. That was the best interpretation I had into the format for a CRN for an IBM Cloud Key Protect or Hyper Protect root key.

operator/v1/types_csi_cluster_driver.go
JoelSpeed: If we make it required now, we could make it optional in the future when we add a second field to the mix

cjschaef: I changed `ecnryptionKey` to required

operator/v1/types_csi_cluster_driver.go
cjschaef: Added the regex I had back and made the value required.

operator/v1/types_csi_cluster_driver.go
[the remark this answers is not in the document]
cjschaef: I added the regex back

operator/v1/types_csi_cluster_driver.go
JoelSpeed: You should remove the encryption key part from this CEL validation
```suggestion
// +kubebuilder:validation:XValidation:rule="has(self.driverType) && self.driverType == 'IBMCloud' ? has(self.ibmcloud) : !has(self.ibmcloud)",message="ibmcloud must be set if driverType is 'IBMCloud', but remain unset otherwise"
```

operator/v1/types_csi_cluster_driver.go
```diff
@@ -248,6 +254,17 @@ type GCPCSIDriverConfigSpec struct {
...
 
+// IBMCloudCSIDriverConfigSpec defines the properties that can be configured for the IBM Cloud CSI driver.
+type IBMCloudCSIDriverConfigSpec struct {
+	// encryptionKeyCRN is the IBM Cloud CRN of the customer-managed root key to use
+	// for disk encryption of volumes for the default storage classes.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength:=154
+	// +kubebuilder:validation:MinLength:=144
+	// +kubebuilder:validation:Pattern:=`^crn:v[0-9]+:bluemix:(public|private):(kms|hs-crypto):[a-z-]+:a/[0-9a-f]+:[0-9a-f-]{36}:key:[0-9a-f-]{36}$`
+	EncryptionKeyCRN string `json:"encryptionKeyCRN,omitempty"`
```
JoelSpeed, line 265: Should not be omitempty when required
```suggestion
	EncryptionKeyCRN string `json:"encryptionKeyCRN"`
```

JoelSpeed (commented):
2 nits and good to go

operator/v1/types_csi_cluster_driver.go
cjschaef: oops, yep, I'll fix.

cjschaef: Updated both items.

PR #1668, opened 2023-11-16
features: add alternative installers to TP

patrickdillon:
Adds the AWS SDK and CAPI Install path to the tech preview feature set, but includes them as Disabled by default.


JoelSpeed (commented):
/lgtm

PR #1670, opened 2023-11-17
MON-3486: add feature gate for Collection Profiles

rexagod:
Adds a custom feature gate for CMO's Collection Profiles feature. Also preserve the earlier behavior to enable it by default on TP enabled clusters.

EP: https://github.com/openshift/enhancements/pull/1298

***

- [ ] Raise a PR like so: https://github.com/openshift/cluster-config-operator/pull/377/files once this is in.


slashpai:
@rexagod can you also link enhacemement proposal here?

config/v1/feature_gates.go
```diff
@@ -341,4 +341,14 @@ var (
 		ResponsiblePerson:   "padillon",
 		OwningProduct:       ocpSpecific,
 	}
+
+	FeatureGateCollectionProfiles = FeatureGateName("CollectionProfiles")
```
simonpasquier, line 345: Maybe prefix with `metrics` to make it more explicit?
```suggestion
	FeatureGateMetricsCollectionProfiles = FeatureGateName("MetricsCollectionProfiles")
```

jan--f:
I think all comments has been addressed and to me this
/lgtm
Iiuc this is simply inactive until we merge the handling code in CMO?

rexagod:
@jan--f There are no blockers for this PR and this can be merged straightaway. Once that's done, a PR updating the `api` version in COO needs to be raised like so: https://github.com/openshift/cluster-config-operator/pull/377/files, and then pulled into https://github.com/openshift/cluster-monitoring-operator/pull/2153/files#diff-99c0290799daf9abc6240df64063e20bfaf67b371577b67ac7eec6f4725622ffR439 to make `FeatureGateMetricsCollectionProfiles` available.

Also, as a side note, https://github.com/openshift/cluster-monitoring-operator/pull/2046 is independent of any other PRs and can be merged in parallel.

jan--f:
To me this looks good. Lets rebase this and get api reviewer eyes on this.
/lgtm

jan--f:
@rexagod This will need an update according to #1815.

rexagod:
Ah, it seems all `SingleNode` manifests were removed from https://github.com/openshift/api/tree/master/payload-manifests/featuregates, however, `update-payload-featuregates` ends up recreating those files.

config/v1/feature_gates.go
```diff
@@ -559,4 +559,11 @@ var (
 					productScope(ocpSpecific).
 					enableIn(TechPreviewNoUpgrade).
 					mustRegister()
+
+	FeatureGateMetricsCollectionProfiles = newFeatureGate("MetricsCollectionProfiles").
+						reportProblemsToJiraComponent("Monitoring").
+						contactPerson("rexagod").
+						productScope(ocpSpecific).
+						enableIn(Default, TechPreviewNoUpgrade).
```
deads2k, line 567: To be accessible by default, we need to see automated testing results (99% pass over two weeks) or have a QE ack.

If you remove the Default, it's mergeable immediately.

jan--f:
@deads2k seems like the bot merged this when it said its not mergeable...

rexagod:
[From a discussion with Jan earlier today] It seems this commit was pulled into Jan's PR, so when that went it, this was marked as merged as well.

PR #1646, opened 2023-10-31
add feature gate for MCO boot images updates

djoshy:
This adds a feature gate for the implementation of [openshift/enhancements#1496](https://github.com/openshift/enhancements/pull/1496) by the MCO team.  This feature is planned to be released in 4.15 as TechPreview.

PR #1672, opened 2023-11-17
MCO-820: Add boot images update opt-in mechanism

djoshy:
In addition to the feature gate that will be added by #1646, we'd also like to add a secondary opt-in knob for the user to toggle the boot image update mechanism. We expect that some users would want to keep their boot images static and this will serve as the primary switch when this feature leaves tech preview. 

This PR introduces a new struct in [types_machineconfiguration.go](https://github.com/openshift/api/blob/master/operator/v1/types_machineconfiguration.go), `ManagedBootImages` which encloses an array of `MachineManager` objects. A `MachineManager` object contains the resource type of  the machine management object that is being opted-in, the API group of that object and a union discriminant object of the type `MachineManagerSelector`.  This object `MachineManagerSelector` encloses:
- The union discriminator, `Mode`, can be set to three values : `All`, `Partial` and `None`.
- `Partial` This is a label selector that will be used by objects to opt-in. When the `Mode` is set to `Partial` mode, all machinesets in the selector list would be considered enrolled for updates. For all other values of `Mode`, this selector does not exist.

Here is a YAML snippet of what this config could look like
```
managedBootImages:
  machineManagers:
  - resource: machinesets
    apiGroup: cluster.x-k8s.io
    selection:
      mode: Partial
      partial:
        matchLabels: {}
  - resource: machinesets
    apiGroup: machine.openshift.io
    selection:
       mode: All
```
The above YAML would provide a label selector for enrolling Cluster API MachineSets and all Machine API MachineSets. Only one machineManager per unique pair of resource/APIGroup will be accepted. This is to avoid conflicting instructions provided by multiple machineManagers.

We may also [add future configuration points](https://github.com/openshift/api/pull/1672#discussion_r1425699287) to `ManagedBootImages`. This PR also adds a new degrade condition to the `MachineConfigPool` CRD. This will be used by the new subcontroller to signal a failure in updating boot images. 

More details about this feature can be found in [enhancement #1496](https://github.com/openshift/enhancements/pull/1496). Comments welcome!


operator/v1/types_machineconfiguration.go
```diff
@@ -32,6 +32,18 @@ type MachineConfigurationSpec struct {
...
+	// will attempt to update boot images for every machineset
+	// If unset, this property defaults to disabled
+	// +optional
+	BootImageUpdateMode BootImageUpdateModeEnum `json:"bootImageUpdateMode,omitempty"`
+
+	// bootImageUpdateEnrolledMachineSets contains a list of machinesets that will be enrolled for updates
+	// +listType=atomic
+	// +nullable
+	// +optional
+	BootImageUpdateEnrolledMachineSets []string `json:"bootImageUpdateEnrolledMachineSets"`
```
JoelSpeed, line 46: Why not put these into a struct that combines the two together?

JoelSpeed, line 46: Why is this a list of names, why not use something like a selector like you would for a deployment to select pods, that would be move flexible/able to evolve over time

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +68,16 @@ type MachineConfigurationList struct {
...
+type BootImageUpdateModeEnum string
+
+const (
+	// Enabled describes an operator that is managing the boot images of all machine sets
+	Enabled BootImageUpdateModeEnum = "Enabled"
+	// Selected describes an operator that is managing the boot images of the selected machine sets, listed in the secondary field BootImageUpdateEnrolledMachineSets
+	Selected BootImageUpdateModeEnum = "Selected"
+	// Disabled describes an operator that is not managing boot images
+	Disabled BootImageUpdateModeEnum = "Disabled"
+)
```
JoelSpeed, line 81: Looking at this, I think you want a [discriminated union](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#discriminated-unions)

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +60,20 @@ type MachineConfigurationList struct {
 	// Items contains the items
 	Items []MachineConfiguration `json:"items"`
 }
+
+// ManagedBootImageConfig contains configuration data for the bootimages management feature
+
+type ManagedBootImageConfig struct {
+	// mode has three settings:
+	// enabled - MCO will manage all machinesets, and ignore machinesets that match with the labelselector
```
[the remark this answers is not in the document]
djoshy: Ah, sorry about the confusing comment. I was trying to imply that the selector list is not checked at all(not in a negations sense) as all machinesets will be enrolled. I'll remove that bit so it is less confusing.

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +60,21 @@ type MachineConfigurationList struct {
 	// Items contains the items
 	Items []MachineConfiguration `json:"items"`
 }
+
+// ManagedBootImageConfig contains configuration data for the bootimages management feature
+// +union
+type ManagedBootImageConfig struct {
```
JoelSpeed, line 66: Needs CEL expression to make sure that `MatchSelector` is only set when type is `MatchSelector`. See the `example/v1` `CELUnion` type for an example

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +60,21 @@ type MachineConfigurationList struct {
...
+	// mode has three settings:
+	// enabled - MCO will manage all machinesets
+	// matchSelector - MCO will manage machinesets that match with the label selector
+	// disabled - MCO will not manage any machinesets
+	// If unset, mode will default to disabled
+	// +unionDiscriminator
+	// +kubebuilder:validation:Enum:="Enabled";"Disabled";"MatchSelector"
+	// +kubebuilder:validation:Required
+	Mode string `json:"mode,omitempty"`
```
JoelSpeed, line 75: Please create a typed string alias with constants that are exported
```
type ManagedBootImageConfigMode string
```

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +60,21 @@ type MachineConfigurationList struct {
...
+
+	// matchSelector specifies a label selector for MachinesSets that will be managed by this mechanism
```
JoelSpeed, line 77:
```suggestion
	// matchSelector specifies a label selector for MachinesSets that will be managed by this mechanism.
	// Only permitted when mode is "MatchSelector".
```

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +60,21 @@ type MachineConfigurationList struct {
...
+	// +optional
+	MatchSelector *metav1.LabelSelector `json:"matchSelector,omitempty"`
```
JoelSpeed, line 79: Are we sure this is the member we want, rather than creating a struct? With this, we have absolutely no way to evolve the API down the line, we cannot add new ways of matching a selector, we would have to add a new `mode`. That may be ok but something to consider, do we expect that MatchSelector may ever need more fields than what exists here today?

djoshy: I thought a little bit about this, I can't think of any future customization as of now. But I would like to leave room for something we haven't planned for - so I made it a struct and enclosed the selector within that as you advised.

Also did some renaming so it makes sense with the new struct. If I'm overlooking some existing naming conventions, please let me know...I'm not a 100% satisfied with `ManagedBootImageConfigCustomization` ... it is quite a mouthful :(

djoshy:
So with this new approach, we still have have three modes:
- Enabled - this will run the feature on all machinesets
- Disabled - this will turn off the feature completely
- CustomConfig - this will run the feature on machinesets that match with the selector enclosed within the `ManagedBootImageConfigCustomization`  struct

We could then add any future configurations as new fields in the  `ManagedBootImageConfigCustomization` object. 
In the event that we do add future configurations, `Enabled` will have assume certain defaults for them. I think this is ok, as long as we document these assumptions/defaults so users will know what to expect.

JoelSpeed (commented):
Thinking about the shape of the union, I'm wondering if `CustomConfig` is one level too many. What would go alongside a MachineSetSelector? Whatever it is, it would have to work in tandem with the selector. An exclusions list perhaps?

Either way, I think Custom is too broad. We do have the option of adding more enum values to the discriminator later, so perhaps for now having the selector is a better discriminator.

Or perhaps we go for something like:

```
managedBootImages:
  managedImages: None | Partial | All
  partial:
    machineSetSelector: ...
   controlPlaneMachineSetSelector: ... <- a potential future addition?
```
If we take that route, I guess then `machineSetSelector` doesn't need to be required, but we should include a validation that `partial` contains at least one non-empty property using the `MinItems` validation
```

File: operator/v1/types_machineconfiguration.go [hunk 8]
```diff
@@ -56,3 +60,42 @@ type MachineConfigurationList struct {
...
+// ManagedBootImageConfig contains configuration data for the bootimages management feature
+// +kubebuilder:validation:XValidation:rule="has(self.mode) && self.mode == 'CustomConfig' ?  has(self.customConfig) : !has(self.customConfig)",message="customConfig is required when type is CustomConfig, and forbidden otherwise"
+// +union
+type ManagedBootImageConfig struct {
+	// mode determines which of the union members should be populated.
+	// +unionDiscriminator
+	// +kubebuilder:validation:Required
+	Mode ManagedBootImageConfigMode `json:"mode,omitempty"`
+
+	// customConfig provides granular control of the ManagedBootImages feature to the user.
```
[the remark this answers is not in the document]
djoshy: Honestly, me neither. Custom/Config was all starting to become a mouthful. I do like the `Partial` approach.

operator/v1/types_machineconfiguration.go
```diff
@@ -32,6 +32,10 @@ type MachineConfigurationSpec struct {
 	StaticPodOperatorSpec `json:",inline"`
 
 	// TODO(jkyros): This is where we put our knobs and dials
+
+	// managedBootImageConfig contains configuration data for the bootimages management feature
```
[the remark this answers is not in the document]
djoshy: Thank you!

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +60,42 @@ type MachineConfigurationList struct {
...
+	// Disabled represents an operator that will not manage any machinesets.
+	Disabled ManagedBootImageConfigMode = "Disabled"
+
+	// CustomConfig represents an operator that will manage machinesets specified by a ManagedBootImageConfigCustomization object.
+	CustomConfig ManagedBootImageConfigMode = "CustomConfig"
+)
+
+// ManagedBootImageConfigCustomization provides granular control of the ManagedBootImages feature.
+// This object is only permitted when mode is set to "CustomConfig".
+type ManagedBootImageConfigCustomization struct {
```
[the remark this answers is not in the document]
djoshy: I do envision there will be some configuration of the revert/storing boot image history mechanism, but nothing solid at the moment. And the reason I wanted to leave empty as a valid option is the this scenario: 

The user wants all boot images enrolled, but they also want to adjust another configuration field(currently, unimplemented) independently of the selector. This gives them that flexibility and not have to go through the trouble of setting up the selector when in `Partial` mode.

I suppose controlPlaneMachineSetsSelector is a possible future one, as you suggested in an earlier comment!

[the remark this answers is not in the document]
djoshy: Ah, I see you alluded to this in your main comment as well. Yes, the `MinItems` validation looks like a good approach here

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +60,42 @@ type MachineConfigurationList struct {
...
+	// machineSetSelector specifies a label selector for MachinesSets that will be managed by this feature.
+	// An empty MachineSetSelector will enroll all MachineSets.
```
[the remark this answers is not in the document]
djoshy: I think the [comment above](https://github.com/openshift/api/pull/1672#discussion_r1425600710) answers why I left this in, but I trhink we're in agreement overall using `Partial` with `MinItems`

operator/v1/types_machineconfiguration.go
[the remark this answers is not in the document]
JoelSpeed: Do you think that the configuration of the boot image history will be tied to the partial/none/all enum?

I wonder if we need a different level here and maybe to bring the MachinesSet concept further up

```
managedBootImages:
  machineSets:
    mode: None | All | Partial
    partial: <- A label selector
      matchLabels: ...
  controlPlaneMachineSets: {} <- will look the same as machineSets
  bootImageHistory:
    mode: Enabled | Disabled
```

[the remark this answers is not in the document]
djoshy: I don't think all of the configuration necessarily needs to be tied to a global switch - I guess I was trying a little too hard) to make it all work within a single union discriminant struct. I do like the idea of moving it up to a higher level, I see it as the main thing users would want to configure.

djoshy:
I've updated the struct to follow the shape mentioned in the comment here: https://github.com/openshift/api/pull/1672#discussion_r1425699287
It does clean up a lot of weird naming issues, thanks @JoelSpeed !

djoshy:
/reopen

Err, not sure what happened

operator/v1/types_machineconfiguration.go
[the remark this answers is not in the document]
djoshy: I've updated the PR to follow the design above. I opted to keep the union as a required field as the configuration would not make sense with out - I'm open to making it optional tough!

yuqi-zhang (approved):
lgtm from the MCO side, thanks David!

djoshy:
Cleaned up an inaccurate comment

djoshy:
I updated the description to better reflect our approach as it was pretty outdated.

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +63,41 @@ type MachineConfigurationList struct {
 	// Items contains the items
 	Items []MachineConfiguration `json:"items"`
 }
+
+type ManagedBootImages struct {
+	// machineSets allows granular control of the machine resources that will be enrolled to the ManagedBootImages feature.
+	// +required
```
JoelSpeed, line 69: This is not the correct tag
```suggestion
	// +kubebuilder:validation:Required
```

JoelSpeed (commented):
Currently, this API is evolvable to the point where we could add other resources like MachineDeployments and ControlPlaneMachineSets, but it makes no distinction between Machine API and Cluster API, this is probably something we want to fix. Perhaps if we can come up with a term, maybe `machineManager` that supports setting the namespace, the resource type and API group, then this could work for all types, provided the MCO understood how to handle them

```
machineManagers:
- resource: machinesets
  apiGroup: cluster.x-k8s.io
  namespace:
    type: Name
    name: openshift-cluster-api
  selection:
    mode: Partial
    partial:
      matchLabels: {}
- resource: machinesets
  apiGroup: machine.openshift.io
  namespace:
    type: Selector
    selector:
      matchLabels: {}
  selection:
     mode: All
```

We could permit only certain configurations for now, eg Machine API only understands `openshift-machine-api` as a namespace, but cluster API may support more namespaces.

We could also restrict the API group and resource to be enums with single values until we support more types.

djoshy:
Updated per latest discussion with Joel. We now have an array of `machineManagers`, within the `managedBootImages` object. Each element in `machineManagers` can be used to point a group of machine management resources, such as a MAPI/CAPI MachineSet, a CAPI MachineDeployment etc. With this approach, we can evolve our API to account for future resources that may need boot image updates. We'll support MAPI machinesets to start, so I've restricted the enums to just that.

YAML example:
```
managedBootImages:
  machineManagers:
  - resource: machinesets
    apiGroup: cluster.x-k8s.io
    selection:
      mode: Partial
      partial:
        matchLabels: {}
  - resource: machinesets
    apiGroup: machine.openshift.io
    selection:
       mode: All
```

JoelSpeed (commented):
I'd appreciate if you could review our API guidelines and guidance on writing godoc and see if you can expand on any of the godoc you've added to this API. Shape is looking better though I think and should be able to be expanded later down the line.

Where did we get to on the conversation about a `Namespace` field?

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +63,78 @@ type MachineConfigurationList struct {
...
+)
+
+// MachineManagerSelectorMode is a string enum used in the MachineManagerSelector union discriminator.
+// +kubebuilder:validation:Enum:="machinesets"
+type MachineManagerResourceType string
+
+const (
+	// machinesets represent the MachineSet resource type, which manage a group of machines.
+	// Although this could belong to a MachineAPI or a ClusterAPI, only MAPI is currently supported.
+	MachineSets MachineManagerResourceType = "machinesets"
```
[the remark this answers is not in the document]
djoshy: If we were to add MachineDeployments, this is the enum for it, hence the somewhat abstract name. 
How about `MachineManagerManagedResourceType`?
```
const (
	// machinesets represent the MachineSet resource type, which manage a group of machines.
	// Although this could belong to a MachineAPI or a ClusterAPI, only MAPI is currently supported.
	MachineSets MachineManagerManagedResourceType = "machinesets"
        // machinedeployments represent the MachineDeployment resource type, which manage a group of machinesets, and is only prsent in ClusterAPI.
	MachineDeployments MachineManagerManagedResourceType = "machinedeployments"
)
```

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +63,78 @@ type MachineConfigurationList struct {
 	// Items contains the items
 	Items []MachineConfiguration `json:"items"`
 }
+
+type ManagedBootImages struct {
+	// machineManagers is an array of machineManager objects.
```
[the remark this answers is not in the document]
djoshy: I gave it a brief description and also fleshed out go doc in other places based on the guidelines. Let me know what you think!

djoshy:
>Where did we get to on the conversation about a Namespace field?

Based on our discussion on slack, I believe we concluded it would complicate the configuration for the first pass and decided to hold off on a namespace filter for future expansion.

djoshy:
I somehow missed the enum validation on the `MachineManagerSelectorMode`, so the last push was just putting that back in.

operator/v1/types_machineconfiguration.go
```diff
@@ -32,6 +32,14 @@ type MachineConfigurationSpec struct {
 	StaticPodOperatorSpec `json:",inline"`
 
 	// TODO(jkyros): This is where we put our knobs and dials
+
+	// managedBootImages allows configuration for the management of boot images for machine
+	// resources within the cluster. This configuration allows users to select resources that should
+	// be updated to the latest boot images during cluster upgrades, ensuring that new machines
+	// always boot with the current cluster version's boot image. When omitted, the ManagedBootImages
+	// feature is essentially disabled.
```
JoelSpeed, line 40: I would like to reword this to be a little more direct
```suggestion
	// always boot with the current cluster version's boot image.
	// When omitted, no boot images will be updated.
```

operator/v1/types_machineconfiguration.go
```diff
@@ -32,6 +32,14 @@ type MachineConfigurationSpec struct {
...
+	// +optional
+	ManagedBootImages ManagedBootImages `json:"managedBootImages,omitempty"`
```
JoelSpeed, line 42: `omitempty` doesn't work with structs, it only works with pointer to struct. For the sake of observability, I think we might want to remove the omitempty, and make the fields within optional. If we have the below example, it allows users to discover the API without having any live configuration
```
managedBootImages:
  machineManagers: []
```

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +64,94 @@ type MachineConfigurationList struct {
 	// Items contains the items
 	Items []MachineConfiguration `json:"items"`
 }
+
+type ManagedBootImages struct {
+	// machineManagers is an array of machineManager objects.
+	// The MCO will watch for changes to this list and enroll/un-enroll machine management resources. An entry in this list
```
JoelSpeed, line 70: Enroll and un-enroll from? Also nit, I'm thinking maybe register and deregister, not sure I like "un-enroll"

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +64,94 @@ type MachineConfigurationList struct {
...
+	// consists of the resource type, the API group that the resource belongs to and a selection filter on the resources.
+	// All fields are required.
+	//
+	// Warning: Only one entry per machine management resource type is allowed. If multiple entries target the same type of machine management
+	// resource, all entries associated with that resource type will be ignored. For example, if one entry describes that MAPI machinesets are
+	// disabled for boot image updates and a later entry describes that MAPI machinesets are enabled for boot image updates, both entries
+	// will be ignored.
```
JoelSpeed, line 77: Hmm, I wonder if there is a way we can set a priority, so that we don't have to have this caveat 🤔 
What if I want all MachineSets in a namespace, but, not a particular machineset? I guess I can achieve that with the matchexpressions?

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +64,94 @@ type MachineConfigurationList struct {
...
+	//
+	// +kubebuilder:validation:Required
+	MachineManagers []MachineManager `json:"machineManagers"`
```
JoelSpeed, line 80: Should there be a unique `resource/apigroup` pair in this list?

Do you need more than one 
```
- resource: machinesets
  apiGroup: machine.openshift.io
  ...
```
Should the partial matchexpressions not cater for this?

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +64,94 @@ type MachineConfigurationList struct {
...
+	// selection allows granular control of the machine management resources that will be enrolled to the ManagedBootImages feature.
+	// +kubebuilder:validation:Required
+	Selection MachineManagerSelector `json:"selection"`
+}
+
+// +kubebuilder:validation:XValidation:rule="has(self.mode) && self.mode == 'Partial' ?  has(self.partial) : !has(self.partial)",message="Partial is required when type is partial, and forbidden otherwise"
+// +union
+type MachineManagerSelector struct {
+	// mode is a union discriminator for MachineManagerSelector and can have three possible values.
+	// - All: All machinesets are enrolled into the ManagedBootImages feature.
```
JoelSpeed, line 109: I would prefer not to describe the feature by the name of the feature, describe it instead by the behaviour. Right now, I have to know what ManagedBootImages feature is, you could instead talk about having automatically updated boot images during cluster upgrades, and then I wouldn't need to understand the feature name

operator/v1/types_machineconfiguration.go
JoelSpeed, line 80: We should add a maxmium length to this list, I think right now, that might be one, but for that, we need to answer the `partial`

If we were to add a namespace dimension, then the list would obviously be much longer

operator/v1/types_machineconfiguration.go
djoshy: Hmm, so to do the above, I'd have to set `machineManager` to optional, but the fields within can remain required, correct? The list in the example you have is empty, but an empty element in the list wouldn't be allowed? (see below)
```
managedBootImages:
  machineManagers: [[]]
```

operator/v1/types_machineconfiguration.go
djoshy: Yeah, I figure a match expression would work for that case. Couple of other approaches to avoid this caveat I had was:
- Only respect the last entry for a machine management resource if there are duplicates. This seems rather unfriendly to me but atleast we are respecting some of the user's needs
- Enforce a unique resource/apigroup pair. Two ways to do this:
-- Instead of a list of machineManagers with unbound size, have three specific definitions of machineManagers for MAPI MachineSets, CAPI MachineSets and CAPI MachineDeployments. We can add support for more types if we see this changing:
```
MAPIMachineSetManager MachineManager `json:"mapiMachineManager"`
MAPIMachineSetManager MachineManager `json:"mapiMachineManager"`
```
--

operator/v1/types_machineconfiguration.go
djoshy: I had the same concern when I was working this out, I wrote up some ideas about other ways[here](https://github.com/openshift/api/pull/1672#discussion_r1457860390)

djoshy: Agree, resolving the caveat will determine the limits on the list

operator/v1/types_machineconfiguration.go
JoelSpeed: Personally, I think we want to keep the machineManagers as it is, but add
```
// +listType=map
// +listMapKey=apiGroup
// +listMapKey=resource
```

With that, we restrict each item in the list, there must be only one entry for each combination of `apiGroup` and `resource`, which currently, limits us to 1. But expands as we add more values in the future to the enums.

Then, the partial selector becomes the only way to selector or deselect machinesets. You could use `matchExpressions` to be expressive about which to include or not include.

This is 2 in your suggestion.

I prefer 2 over 1 because it means we have specific fields which are hard to name, and, it prevents future expansion into having a namespace field.

FYI, if we add `namespace` in the future, that would also have to be a `listMapKey`.

operator/v1/types_machineconfiguration.go
JoelSpeed: Correct, the list itself is empty, but you wouldn't be allowed an empty element in the list
```
managedBootImages:
  machineManagers:
  - {}
```
The above would fail validation based on the required fields within a `MachineManager`

operator/v1/types_machineconfiguration.go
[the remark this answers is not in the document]
JoelSpeed: I appreciate that it's a little specific in my suggestion, however, this is just for the Go API, and the user facing API is not affected by the suggestion to have a more specific name in the Go types.

My concern is that we may want MachineSets or MachineDeployments in some other enum later, yet the consts should be distinguished as separate things

operator/v1/types_machineconfiguration.go
djoshy: This makes sense to me, let's go with your approach for (2). 

Thanks for the example - I totally missed that `listMapKey` could be used in that fashion.

operator/v1/types_machineconfiguration.go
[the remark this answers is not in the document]
djoshy: Makes sense, I'll update to `MachineManagerMachineSetsResourceType`. Thanks!

djoshy:
Fixed and added some new tests in the last push. Still not sure about the error on ci/prow/verify-crd-schema  - any pointers?

djoshy:
Alright, per conversation with @JoelSpeed  I've fixed up the missing tags in the embedded fields(which changed multiple operator CRDs as they used them too) and the only remaining errors are the following label selector ones which we have decided to override:
```
error in 0000_80_machine-config-operator_01_config.TechPreviewNoUpgrade.crd.yaml: ListsMustHaveSSATags: crd/machineconfigurations.operator.openshift.io version/v1 field/^.spec.managedBootImages.machineManagers[*].selection.partial.matchExpressions must set ListsMustHaveSSATags
error in 0000_80_machine-config-operator_01_config.TechPreviewNoUpgrade.crd.yaml: ListsMustHaveSSATags: crd/machineconfigurations.operator.openshift.io version/v1 field/^.spec.managedBootImages.machineManagers[*].selection.partial.matchExpressions[*].values must set x-kubernetes-list-type
error in 0000_80_machine-config-operator_01_config.TechPreviewNoUpgrade.crd.yaml: NoMaps: crd/machineconfigurations.operator.openshift.io version/v1 field/^.spec.managedBootImages.machineManagers[*].selection.partial.matchLabels may not be a map
```
Some background: The embedded field errors had popped up because the TechPreview version of the MCO Operator CRD is "new" to the API. `verify-crd-schema` only runs on changes and existing invalid fields are ignored. By including them in this new CRD, these older(invalid) embedded fields were re-evaluated causing test failures. As an attempt to clean this up for future work, Joel asked me to add the missing tags.

djoshy:
Rebased (:

operator/v1/types.go
```diff
@@ -115,6 +115,7 @@ type OperatorStatus struct {
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
 
 	// conditions is a list of conditions and their status
+	// +listType=atomic
```
JoelSpeed, line 118: This should behave as a map
```suggestion
	// +listType=map
	// +listMapKey=type
```

operator/v1/types.go
```diff
@@ -126,6 +127,7 @@ type OperatorStatus struct {
 	ReadyReplicas int32 `json:"readyReplicas"`
 
 	// generations are used to determine when an item needs to be reconciled or has changed in a way that needs a reaction.
+	// +listType=atomic
```
JoelSpeed, line 130: Should be a map
```suggestion
	// +listType=map
	// +listMapKey=group
	// +listMapKey=resource
	// +listMapKey=namespace
	// +listMapKey=name
```

operator/v1/types.go
```diff
@@ -208,6 +210,7 @@ type StaticPodOperatorStatus struct {
 	LatestAvailableRevisionReason string `json:"latestAvailableRevisionReason,omitempty"`
 
 	// nodeStatuses track the deployment values and errors across individual nodes
+	// +listType=atomic
```
JoelSpeed, line 213: Should be a map
```suggestion
	// +listType=map
	// +listMapKey=nodeName
```

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +65,98 @@ type MachineConfigurationList struct {
...
+	MachineManagers []MachineManager `json:"machineManagers"`
+}
+
+// MachineManager contains identifying information of a machine management resource(eg. a machineset) that will be
+// registered for boot image updates. This is likely to evolve as support for more machine management resources are added.
+type MachineManager struct {
+	// resource is the machine management resource's type.
+	//
+	// The following values are accepted:
+	// - MachineSets: The machine manager will only register resources of the type MachineSet, which may belong to MachineAPI or ClusterAPI.
```
JoelSpeed, line 91: Casing matters, this should be lowercased
```suggestion
	// - machinesets: The machine manager will only register resources of the type MachineSet, which may belong to MachineAPI or ClusterAPI.
```

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +65,98 @@ type MachineConfigurationList struct {
...
+	//
+	// +kubebuilder:validation:Required
+	Resource MachineManagerMachineSetsResourceType `json:"resource"`
+	// apiGroup is name of the APIGroup that the machine management resource belongs to.
+	//
+	// The following values are accepted:
+	// - MachineAPI: The machine manager will only register resources that belong to MachineAPI APIGroup.
```
JoelSpeed, line 98:
```suggestion
	// - machine.openshift.io: The machine manager will only register resources that belong to MachineAPI APIGroup.
```

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +65,98 @@ type MachineConfigurationList struct {
...
+	Selection MachineManagerSelector `json:"selection"`
+}
+
+// +kubebuilder:validation:XValidation:rule="has(self.mode) && self.mode == 'Partial' ?  has(self.partial) : !has(self.partial)",message="Partial is required when type is partial, and forbidden otherwise"
+// +union
+type MachineManagerSelector struct {
+	// mode is a union discriminator for MachineManagerSelector and can have three possible values.
+	// - All: All resources specified by the parent MachineManager are registered for boot image updates.
+	// - None: No resources specified by the parent MachineManager are registered for boot image updates.
+	// - Partial: resources specified by the parent MachineManager are registered for boot image updates only if they match with the label selector.
```
JoelSpeed, line 114: Have you tried an `oc explain` of these fields with the current patch? I'm not sure the formatting will carry over and it would be good to check how this looks in an `oc explain`

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +65,98 @@ type MachineConfigurationList struct {
...
+	// selection allows granular control of the machine management resources that will be registered for boot image updates.
+	//
+	// +kubebuilder:validation:Required
+	Selection MachineManagerSelector `json:"selection"`
+}
+
+// +kubebuilder:validation:XValidation:rule="has(self.mode) && self.mode == 'Partial' ?  has(self.partial) : !has(self.partial)",message="Partial is required when type is partial, and forbidden otherwise"
+// +union
+type MachineManagerSelector struct {
+	// mode is a union discriminator for MachineManagerSelector and can have three possible values.
```
JoelSpeed, line 111: You're talking here in technical terms, you should talk more in end user terminology.
```suggestion
	// mode determines how machine managers will be selected for updates.
	// Valid values are All, None and Partial.
	// All means that every resource matched by the machine manager will be updated.
	// None means that no resources matched by the machine manager will be updated.
	// Partial requires a specified selector and allows customisation of which resources matched by the machine manager will be updated.
```
This got me thinking, what's the point to None? Surely you just turn off the feature completely?

JoelSpeed (commented):
I'll double check the suggested map types for the operators with @deads2k but I think that's what we want.

I'd encourage you to review the comments you have on fields. The style we normally go for is quite a direct statement, but with a view of the end user in mind. For example, rather than talking about a union discrminator, which the end user doesn't care about, talk instead about the behaviour that setting the field enables or changes.

It will make the documentation much easier for end users to understand the behaviours of the API when they get to using this feature.

operator/v1/types_machineconfiguration.go
djoshy: I'm not opposed to taking this out, but I figured this would give them an easy way to negate a MachineSet/group of MachineSets

operator/v1/types_machineconfiguration.go
djoshy: Ahh, I accidentally put in the type names instead of the string. Will fix

operator/v1/types_machineconfiguration.go
djoshy: oof, yeah I missed that. Just tried it, it's not pretty. Will reformat.

operator/v1/types.go
JoelSpeed: We can't use namespace as a map key because it's not a required field. You'll need to not include that one here

djoshy: I did have to add some required field tags for some of the other list "maps", so I assume that's okay other than for this namespace one?

JoelSpeed: Can you give me a list of which fields you had to make required? I don't think we can actually. So we may have to move back to atomic, give me some time and I'll confirm before you go back and forth changing and changing back 😓

djoshy: Sure, the following lists asked me to add `+kubebuilder:validation:Required` tags. These internal fields did not have any tags on them:
- `GenerationStatus` - group, resource, name
- `OperatorCondition` - type
- `NodeStatus` - nodeName

JoelSpeed:
Spoke with @deads2k offline about the operator status changes. We are happy to make `OperatorCondition.Type` and `NodeStatus.NodeName` required. They will then become the `listMapKey` for their respective lists. For the `GenerationStatus`, there should actually be 4 keys, but since none of them are currently required, we don't feel confident in making that change. `Generations` should therefore be atomic.

Every condition should have a type, every nodestatus should have a nodename, if they do not, this bug should be fixed.

This is why we don't use package level `+kubebuilder:validation:Optional` 😞

deads2k:
> Spoke with @deads2k offline about the operator status changes. We are happy to make `OperatorCondition.Type` and `NodeStatus.NodeName` required. They will then become the `listMapKey` for their respective lists. For the `GenerationStatus`, there should actually be 4 keys, but since none of them are currently required, we don't feel confident in making that change. `Generations` should therefore be atomic.
> 
> Every condition should have a type, every nodestatus should have a nodename, if they do not, this bug should be fixed.
> 
> This is why we don't use package level `+kubebuilder:validation:Optional` 😞

I agree on all counts.  Sins of the past led to the schema validation we're now writing.

djoshy:
Alright, so I've broken out the OperatorStatus changes into a different commit and updated as per the ask above.  I cleaned up all the user facing comments to make it easier to read and also fixed up the formatting. Here are the `oc explain` outputs:

```
$ oc explain machineconfigurations.operator.openshift.io.spec.managedBootImages
GROUP:      operator.openshift.io
KIND:       MachineConfiguration
VERSION:    v1

FIELD: managedBootImages <Object>

DESCRIPTION:
    managedBootImages allows configuration for the management of boot images for
    machine resources within the cluster. This configuration allows users to
    select resources that should be updated to the latest boot images during
    cluster upgrades, ensuring that new machines always boot with the current
    cluster version's boot image. When omitted, no boot images will be updated.
    
FIELDS:
  machineManagers	<[]Object>
    machineManagers can be used to register machine management resources for
    boot image updates. The Machine Config Operator will watch for changes to
    this list. Only one entry is permitted per type of machine management
    resource.


$ oc explain machineconfigurations.operator.openshift.io.spec.managedBootImages.machineManagers
GROUP:      operator.openshift.io
KIND:       MachineConfiguration
VERSION:    v1

FIELD: machineManagers <[]Object>

DESCRIPTION:
    machineManagers can be used to register machine management resources for
    boot image updates. The Machine Config Operator will watch for changes to
    this list. Only one entry is permitted per type of machine management
    resource.
    MachineManager describes a target machine resource that is registered for
    boot image updates. It stores identifying information such as the resource
    type and the API Group of the resource. It also provides granular control
    via the selection field.
    
FIELDS:
  apiGroup	<string> -required-
    apiGroup is name of the APIGroup that the machine management resource
    belongs to. 
     Valid value(s):  machine.openshift.io 
     * machine.openshift.io means that the machine manager will only register
    resources that belong to MachineAPI APIGroup.

  resource	<string> -required-
    resource is the machine management resource's type. 
     Valid value(s): machinesets 
     * machinesets means that the machine manager will only register resources
    of the type MachineSet, which may belong to MachineAPI or ClusterAPI.

  selection	<Object> -required-
    selection allows granular control of the machine management resources that
    will be registered for boot image updates.

$ oc explain machineconfigurations.operator.openshift.io.spec.managedBootImages.machineManagers.selection
GROUP:      operator.openshift.io
KIND:       MachineConfiguration
VERSION:    v1

FIELD: selection <Object>

DESCRIPTION:
    selection allows granular control of the machine management resources that
    will be registered for boot image updates.
    
FIELDS:
  mode	<string> -required-
    mode determines how machine managers will be selected for updates. 
     Valid value(s): All, Partial. 
     * All means that every resource matched by the machine manager will be
    updated. 
     * Partial requires a specified selector and allows customisation of which
    resources matched by the machine manager will be updated.

  partial	<Object>
    partial provides a label selector that can be used to match machine
    management resources. Only permitted when mode is set to "Partial".

```
Let me know if that works better (:

operator/v1/types_machineconfiguration.go
```diff
@@ -32,6 +32,15 @@ type MachineConfigurationSpec struct {
 	StaticPodOperatorSpec `json:",inline"`
 
 	// TODO(jkyros): This is where we put our knobs and dials
+
+	// managedBootImages allows configuration for the management of boot images for machine
+	// resources within the cluster. This configuration allows users to select resources that should
+	// be updated to the latest boot images during cluster upgrades, ensuring that new machines
+	// always boot with the current cluster version's boot image. When omitted, no boot images
+	// will be updated.
+	// +openshift:enable:FeatureSets=TechPreviewNoUpgrade
```
JoelSpeed, line 41: Need a `CustomNoUpgrade` on this too else you'll break anyone who tries to use that featureset, see other uses of this tag

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +65,94 @@ type MachineConfigurationList struct {
...
+	Partial MachineManagerSelectorMode = "Partial"
+)
+
+// MachineManagerManagedResourceType is a string enum used in the MachineManager type to describe the resource
+// type to be registered.
+// +kubebuilder:validation:Enum:="machinesets"
+type MachineManagerMachineSetsResourceType string
+
+const (
+	// machinesets represent the MachineSet resource type, which manage a group of machines.
```
JoelSpeed, line 144: Nit, comment on this one should be capitalised
```suggestion
	// MachineSets represent the MachineSet resource type, which manage a group of machines.
```

JoelSpeed (commented):
Couple of nits, but otherwise the API here looks like it works for me

djoshy:
Fixed up the following:
- Added the new tag and CRD for CustomNoUpgrade
- Add the feature set validation rule for I had missed earlier
- fixed the capitalization in the MachineSets comment

djoshy:
Hmm, that doesn't look good

/test images

djoshy:
/test images

Holding until I figure out how I broke this :thinking:

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +66,94 @@ type MachineConfigurationList struct {
...
+}
+
+// MachineManager describes a target machine resource that is registered for boot image updates. It stores identifying information
+// such as the resource type and the API Group of the resource. It also provides granular control via the selection field.
+type MachineManager struct {
+	// resource is the machine management resource's type.
+	//
+	// Valid value(s): machinesets
+	//
+	// * machinesets means that the machine manager will only register resources of the type MachineSet, which may belong to MachineAPI or ClusterAPI.
```
JoelSpeed, line 88: This won't render nicely in `kubectl explain`, make it more of a paragraph and more like prose. Lets not mention Cluster API just yet since it's not being supported just yet
```suggestion
	// resource is the machine management resource's type.
	// The only current valid value is machinesets.
	// machinesets means that the machine manager will only register resources of the kind MachineSet.
```

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +66,94 @@ type MachineConfigurationList struct {
...
+	//
+	// +kubebuilder:validation:Required
+	Resource MachineManagerMachineSetsResourceType `json:"resource"`
+	// apiGroup is name of the APIGroup that the machine management resource belongs to.
+	//
+	// Valid value(s):  machine.openshift.io
+	//
+	// * machine.openshift.io means that the machine manager will only register resources that belong to MachineAPI APIGroup.
```
JoelSpeed, line 96:
```suggestion
	// apiGroup is name of the APIGroup that the machine management resource belongs to.
	// The only current valid value is  machine.openshift.io.
	// machine.openshift.io means that the machine manager will only register resources that belong to OpenShift machine API group.
```

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +66,94 @@ type MachineConfigurationList struct {
...
+// +kubebuilder:validation:XValidation:rule="has(self.mode) && self.mode == 'Partial' ?  has(self.partial) : !has(self.partial)",message="Partial is required when type is partial, and forbidden otherwise"
+// +union
+type MachineManagerSelector struct {
+	// mode determines how machine managers will be selected for updates.
+	//
+	// Valid value(s): All, Partial.
+	//
+	// * All means that every resource matched by the machine manager will be updated.
+	//
+	// * Partial requires a specified selector and allows customisation of which resources matched by the machine manager will be updated.
```
JoelSpeed, line 115:
```suggestion
	// mode determines how machine managers will be selected for updates.
	// Valid values are All and Partial.
	// All means that every resource matched by the machine manager will be updated.
	// Partial requires a specified selector and allows customisation of which resources matched by the machine manager will be updated.
```

JoelSpeed (commented):
Few formatting things and a question on one of the validations to work out

operator/v1/types_machineconfiguration.go
```diff
@@ -28,10 +28,20 @@ type MachineConfiguration struct {
 	Status MachineConfigurationStatus `json:"status"`
 }
 
+// +openshift:validation:FeatureSetAwareXValidation:featureSet=CustomNoUpgrade;TechPreviewNoUpgrade,rule="has(oldSelf.managedBootImages) ? has(self.managedBootImages) : true",message="managedBootImages may not be removed once set"
```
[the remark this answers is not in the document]
djoshy: Ah, I had thought this was a standard rule for the feature set gated fields based on the example. I thought the idea was to not be able to unset the field itself(not the list field within), but you're right - in this case it makes it more confusing, I will remove.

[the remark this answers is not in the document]
JoelSpeed: We have defined in the enhancement what happens if you deselect a MachineSet right?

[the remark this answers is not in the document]
djoshy: I don't think we have explicitly laid that out(I'm assuming you mean initially registered, then de-registered?). I can state that under the opt-in section..Something like "the machineset would retain what the last boot image update was". 

Tangentially, there is a bit later on under Upgrade/Downgrade strategy, in which I mention a couple of scenarios like this, but the effects of that are cluster wide, based on the feature being present at all.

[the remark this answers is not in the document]
JoelSpeed: Yeah I think it would be good to clarify what happens if I'm a user, and I opt-in initially, and then decide to opt out later

JoelSpeed:
Ok, from an API perspective I'm happy with this now, I would like to take a final pass over the enhancement before we merge as well

@yuqi-zhang I believe you've been reviewing the EP, are you happy with the shape of this API?

yuqi-zhang (approved):
Looked over the managedbootimages commit again, and the current API lgtm!

djoshy:
Updated based on discussion from the enhancement:

- removed the `MachineConfigPool` degrade type since we are switching to an alert
- enclosed the `MachineSet` label selector within a struct so we have flexibility to expand in the future

JoelSpeed:

Introduction fo techpreview/custom versions of CRDs has caused some pre-existing failures that cannot be fixed

PR #1676, opened 2023-11-20
OCPNODE-1890: Bump k8s api to v0.29.0

sairameshv:
cc: @soltysh


go.mod
```diff
@@ -1,45 +1,45 @@
 module github.com/openshift/api
 
-go 1.20
+go 1.21.3
```
soltysh, line 3: We don't want to be that precise. So only `go 1.21` is enough. But for now, I'd stick with golang 1.20, since that will allow you to test this and other PRs with the current builders available.

soltysh: When you're changing this, you'll need to update https://github.com/openshift/api/blob/master/.ci-operator.yaml to match the golang version.

soltysh (changes requested):

go.mod
sairameshv: >We don't want to be that precise. So only go 1.21 is enough. But for now, I'd stick with golang 1.20, since that will allow you to test this and other PRs with the current builders available.

`go mod tidy` has automatically updated it to `1.21.3`. However, as of now, I updated the go version to `1.20`

soltysh (approved):
/lgtm
/approve

soltysh:
The failures in e2e-aws-serial-techpreview and e2e-upgrade are not related with this PR. We go green runs for both of the jobs (https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_api/1676/pull-ci-openshift-api-master-e2e-upgrade/1732398236494204928 and https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_api/1676/pull-ci-openshift-api-master-e2e-aws-serial-techpreview/1732398236372570112), so I'm going to override to be able to unblock k8s bump.

/override ci/prow/e2e-upgrade

PR #1681, opened 2023-11-27
fix MCN CRD to use proper Drained and Cordoned condition types

cdoern:
These CRD selectors do not match the constants in types_machineconfignode.go I'd prefer to change the CRD now before 4.15 is cut. But if not, I can change the types later.


wking (approved):
/lgtm

PR #1695, opened 2023-12-05
OPNET-415: Revert the revert "Extend infra CR to store VIPs and MachineNetwork""

mkowalski:
In https://github.com/openshift/api/pull/1593 we have implemented a new API targeting OCP 4.15

In https://github.com/openshift/api/pull/1694 we have reverted this change as the feature has been deferred till OCP 4.16

Here we revert the revert with the desire to merge it after the 4.15 branching so that it lands in `master` but not in `release-4.15`


mkowalski:
/cc @JoelSpeed
/cc @cybertron 
/hold

To unhold only after branching

mkowalski:
Hey @JoelSpeed, back in 4.16, can we get this back again?

mkowalski:
/test e2e-aws-serial-techpreview

Failure in namespace/openshift-monitoring

PR #1696, opened 2023-12-05
Added Feature Gate for On Cluster Build

dkhater-redhat:


jkyros:
This is for https://github.com/openshift/enhancements/pull/1515, are we cool with that enhancement still being in Draft for now, or does it need to be "proper open" to get a feature gate?

jkyros:
Implementation would be: https://github.com/openshift/machine-config-operator/pull/4060

dkhater-redhat:
The feature we are gating was initially merged in 4.14. When we were discussing the feature then, we were ambivalent to gate it because you have to "opt-in" to the feature anyways. Furthermore, the feature in 4.14 is in dev preview so we didn't see it necessary to gate it (as it is unsupported anyways).

You are definitely right on it being late in the cycle- but we were wanting to land it in 4.15 so we can get it in before tech preview.

PR #1609, opened 2023-10-06
Add `PinnedImageSet`

jhernand:
This patch adds a new `PinnedImageSet` object that describes how to pre-load and pin container images.

Related: https://github.com/openshift/enhancements/pull/1432
Related: https://issues.redhat.com/browse/RFE-4482

PR #1713, opened 2024-01-02
MCO-838: add PinnedImageSet types and crd

hexfusion:
This PR adds `PinnedImageSet` types to the ` machineconfiguration.openshift.io` group and a CRD as a continuation of https://github.com/openshift/api/pull/1609.

ref:
 - [MCO-838](https://issues.redhat.com//browse/MCO-838)
 - updated enhancement https://github.com/openshift/enhancements/pull/1599


machineconfiguration/v1/types.go
```diff
@@ -760,3 +760,107 @@ type ContainerRuntimeConfigList struct {
...
+	// +optional
+	Status PinnedImageSetStatus `json:"status"`
+}
+
+// PinnedImageSetSpec defines the desired state of a PinnedImageSet.
+type PinnedImageSetSpec struct {
+	// nodeSelector defines the set of nodes that this configuration applies to. If not
+	// specified then the configuration applies to all the nodes of the cluster.
+	// +optional
+	NodeSelector *corev1.NodeSelector `json:"nodeSelector,omitempty"`
```
hexfusion, line 787: since the MachineConfigPoolSpec defines 
```golang
	NodeSelector *metav1.LabelSelector `json:"nodeSelector,omitempty"`
```
We will want to also use this allow us to easily validate.

machineconfiguration/v1/types.go
```diff
@@ -760,3 +760,137 @@ type ContainerRuntimeConfigList struct {
...
+// +openshift:compatibility-gen:level=1
+type PinnedImageSetList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata"`
+
+	Items []PinnedImageSet `json:"items"`
+}
+
+// NodeStatus provides information about the current state of a particular node managed by this operator.
+type NodeStatus struct {
```
sinnykumari, line 875: Curious, how are we planning to use this status in MCO? Perhaps an addition to @cdoern 's Status Reporting https://github.com/openshift/api/pull/1596.

machineconfiguration/v1/types.go
```diff
@@ -760,3 +760,137 @@ type ContainerRuntimeConfigList struct {
...
 }
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// PinnedImageSet describes a set of images that should be pinned.
+//
+// Compatibility level 1: Stable within a major release for a minimum of 12 months or 3 minor releases (whichever is longer).
+// +openshift:compatibility-gen:level=1
```
cdoern, line 771: do we want this in tech preview? I think we should probably put this under v1alpha1

machineconfiguration/v1/types.go
```diff
@@ -760,3 +760,137 @@ type ContainerRuntimeConfigList struct {
...
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+	
+	// conditions represents the latest available observations of current state.
+	// +listType=atomic
+	// +optional
+	Conditions []PinnedImageSetCondition `json:"conditions"`
+}
+
+// PinnedImageSetCondition contains condition information for a PinnedImageSet.
+type PinnedImageSetCondition struct {
```
cdoern, line 823: we typically opt to use metav1.conditions rather than bespoke condition types.  I think we can fit this into a metav1.condition

machineconfiguration/v1/types.go
cdoern, line 875: yeah, this should live in v1alpha1/machineconfignode we shouldn't have two separate places tracking node state. Lets collaborate to figure out how this should look in the MCN :)

cdoern (commented):
This is on the right track! I think we:

1. need this in tech-preview and therefore v1alpha1 version.
2. the NodeStatus needs to be baked into the MCN CRD rather than re-creating that here.
3. its best practice to use metav1.conditions rather than bespoke state reporting.

machineconfiguration/v1/types.go
hexfusion: I didnt mean to push this was playing with an idea and pushed by accident. reverting

hexfusion: see below

machineconfiguration/v1/types.go
hexfusion: I am not sure honestly we can fine-tune this when we get the API stable.

hexfusion: Yeah let's revisit once we get the PR stable. v1alpha sounds reasonable.

config/v1/types_tlssecurityprofile.go
```diff
@@ -177,7 +177,7 @@ type TLSProfileSpec struct {
 // TLSProtocolVersion is a way to specify the protocol version used for TLS connections.
 // Protocol versions are based on the following most common TLS configurations:
 //
-//   https://ssl-config.mozilla.org/
+//	https://ssl-config.mozilla.org/
 //
```
hexfusion, line 181: TODO: drop files that leaked in from make update..

machineconfiguration/v1/types.go
```diff
@@ -760,3 +760,77 @@ type ContainerRuntimeConfigList struct {
...
+	// These image references should all be by digest, tags aren't allowed.
+	// +listType=set
+	// +optional
+	PinnedImages []PinnedImageRef `json:"pinnedImages,omitempty"`
+}
+
+// PinnedImageRef is an image reference by digest.
+//
+// +kubebuilder:validation:Pattern:=`@sha256:[a-fA-F0-9]{64}$`
+type PinnedImageRef string
```
JoelSpeed, line 809: Can you apply a maximal length to this string?

machineconfiguration/v1/types.go
```diff
@@ -760,3 +760,77 @@ type ContainerRuntimeConfigList struct {
...
+	//              "quay.io/openshift-release-dev/ocp-release@sha256:...",
+	//              "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:...",
+	//              "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:...",
+	//              ...
+	//      ]
+	//
+	// These image references should all be by digest, tags aren't allowed.
+	// +listType=set
+	// +optional
+	PinnedImages []PinnedImageRef `json:"pinnedImages,omitempty"`
```
JoelSpeed, line 803: Can we set an upper limit on the length of this list? Can we set a lower limit?

machineconfiguration/v1/types.go
```diff
@@ -760,3 +760,77 @@ type ContainerRuntimeConfigList struct {
...
+	// observedGeneration represents the generation observed by the controller.
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// conditions represent the observations of the PinnedImageSet controller.
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
```
JoelSpeed, line 822: Idiomatically conditions should be the first item in the status

machineconfiguration/v1/types.go
```diff
@@ -760,3 +760,77 @@ type ContainerRuntimeConfigList struct {
...
+// PinnedImageRef is an image reference by digest.
+//
+// +kubebuilder:validation:Pattern:=`@sha256:[a-fA-F0-9]{64}$`
+type PinnedImageRef string
+
+// PinnedImageSetStatus defines the observed state of a PinnedImageSet.
+type PinnedImageSetStatus struct {
+	// observedGeneration represents the generation observed by the controller.
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
```
JoelSpeed, line 815: Can you expand on the godoc here, what does it mean that the controller has observed this generation?

machineconfiguration/v1/types.go
```diff
@@ -760,3 +760,77 @@ type ContainerRuntimeConfigList struct {
...
+	// +optional
+	Status PinnedImageSetStatus `json:"status"`
+}
+
+// PinnedImageSetSpec defines the desired state of a PinnedImageSet.
+type PinnedImageSetSpec struct {
+	// MachineConfigPoolSelector selects which pools the PinnedImageSet should apply to.
+	// A nil selector will result in no pools being selected.
+	// +optional
+	MachineConfigPoolSelector *metav1.LabelSelector `json:"machineConfigPoolSelector,omitempty"`
```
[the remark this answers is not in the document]
hexfusion: It can be more than 2 as custom pools are a feature[1] but I would imagine most clusters operator with only 2.

It would be possible to label the pools IE and thus define the elector to match `pinned-image-set-opt-in: true`
```
metadata:
  labels:
    pinned-image-set-opt-in: true
```

[1] PinnedImageSet describes a set of images that should be pinned.

machineconfiguration/v1/types.go
```diff
@@ -760,3 +760,77 @@ type ContainerRuntimeConfigList struct {
...
+	// +required
+	Spec PinnedImageSetSpec `json:"spec"`
+	// +optional
+	Status PinnedImageSetStatus `json:"status"`
+}
+
+// PinnedImageSetSpec defines the desired state of a PinnedImageSet.
+type PinnedImageSetSpec struct {
+	// MachineConfigPoolSelector selects which pools the PinnedImageSet should apply to.
+	// A nil selector will result in no pools being selected.
```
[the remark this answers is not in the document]
hexfusion: its a good point feels like it should error as to your point there is not thing to reconcile.

machineconfiguration/v1/types.go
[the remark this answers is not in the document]
hexfusion: We could explicitly define an enum something like.
```
 {Worker,Master}PinnedImageSetMachineConfigPool
 ```
 I would like to enforce one set per pool so this would work from my perspective.

File: openapi/openapi.json [hunk 12]
```diff
@@ -45238,6 +45342,59 @@
       "description": "IntOrString is a type that can hold an int32 or a string.  When used in JSON or YAML marshalling and unmarshalling, it produces or consumes the inner type.  This allows you to have, for example, a JSON field that can accept a name or number.",
       "type": "string",
       "format": "int-or-string"
+    },
+    "io.k8s.apimachinery.pkg.version.Info": {
```
hexfusion, line 45346: not sure why this is getting generated

machineconfiguration/v1/types.go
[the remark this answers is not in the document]
JoelSpeed: If there are custom pools, then I don't think an enum will work, as that will prevent users from working with the custom pools.

Enforcing a "one set per pool" will be very difficult if you have a selector, with multiple entries in the list, how would you enforce that the selectors do not overlap? And, even if they don't overlap at admission time, what's to stop the labels being modified later and then having them overlap outside of the admission control

Could this be a property on the machine config pool itself?

Or perhaps, a resource that's `.metadata.name` must match a machine pool name?

Having the `.metadata.name` match ensures your uniqueness that you desire

The alternative is to have some sort of maybe policy and binding, in theory they could be two separate CRDs, or they could be in a cluster wide singleton somewhere

```
pinnedImages:
  policies:
  - name: generic
    pinnedImages:
    - ....
    - ....
  bindings:
  - poolName: worker
    policyName: generic
```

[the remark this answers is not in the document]
JoelSpeed: Another suggestion from David, keep the image lists as CRs, and then reference the image list from the MCP, solves having multiple lists, and being able to bind specifically one of those lists to the MCP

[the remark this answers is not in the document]
hexfusion: makes sense good feedback

machineconfiguration/v1alpha1/types_pinnedimageset.go
```diff
@@ -0,0 +1,83 @@
...
+	//              "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:...",
+	//              "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:...",
+	//              ...
+	//      ]
+	//
+	// These image references should all be by digest, tags aren't allowed.
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=500
+	// +listType=set
+	// +optional
```
JoelSpeed, line 49: This feels to me like a required property, don't you expect?

machineconfiguration/v1alpha1/types_pinnedimageset.go
```diff
@@ -0,0 +1,83 @@
...
+	PinnedImages []PinnedImageRef `json:"pinnedImages,omitempty"`
+}
+
+// PinnedImageRef is an image reference by digest.
+// +kubebuilder:validation:MaxLen=MaxLength=256
```
JoelSpeed, line 54: Typo in this, where did 256 come from?
```suggestion
// +kubebuilder:validation:=MaxLength=256
```

machineconfiguration/v1alpha1/types_pinnedimageset.go
```diff
@@ -0,0 +1,83 @@
...
+	// +optional
+	Status PinnedImageSetStatus `json:"status"`
+}
+
+// PinnedImageSetSpec defines the desired state of a PinnedImageSet.
+type PinnedImageSetSpec struct {
+	// MachineConfigPoolSelector selects which pools the PinnedImageSet should apply to.
+	// A nil selector will result in no pools being selected.
+	// +optional
+	MachineConfigPoolSelector *metav1.LabelSelector `json:"machineConfigPoolSelector,omitempty"`
```
JoelSpeed, line 32: Moving to MCP selecting a set of pinned image lists?

machineconfiguration/v1alpha1/types_pinnedimageset.go
```diff
@@ -0,0 +1,83 @@
...
+	PinnedImages []PinnedImageRef `json:"pinnedImages,omitempty"`
+}
+
+// PinnedImageRef is an image reference by digest.
+// +kubebuilder:validation:MaxLen=MaxLength=256
+// +kubebuilder:validation:Pattern:=`@sha256:[a-fA-F0-9]{64}$`
+type PinnedImageRef string
+
+// PinnedImageSetStatus defines the observed state of a PinnedImageSet.
+type PinnedImageSetStatus struct {
```
JoelSpeed, line 59: I wonder if this is just a resource to collect a list of images, whether it even needs a status

machineconfiguration/v1alpha1/types_pinnedimageset.go
hexfusion: arbitrary felt logical

machineconfiguration/v1alpha1/types_pinnedimageset.go
hexfusion: correct as per conv offline/above.

machineconfiguration/v1alpha1/types_pinnedimageset.go
hexfusion: good point

hexfusion: MCP will have status I believe

machineconfiguration/v1alpha1/types_pinnedimageset.go
hexfusion: ensuring valid with crio/docker specs

machineconfiguration/v1alpha1/types_pinnedimageset.go
```diff
@@ -0,0 +1,108 @@
...
+
+// PinnedImageSetSpec defines the desired state of a PinnedImageSet.
+type PinnedImageSetSpec struct {
+	// machineConfigPools define which pools the PinnedImageSet should be
+	// applied to. If the referenced pools do not exist, the PinnedImageSet will
+	// be ignored.
+
+	// +kubebuilder:validation:Required
+	// +listType=set
+	MachineConfigPools []MachineConfigPoolRef `json:"machineConfigPools"`
```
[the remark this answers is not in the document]
hexfusion: The flow would need to be 
- define CR
- MCP syncs on new CR
- MCP adds set to spec
- MCD reconciles updating node annotation as signal for success/fail
- MCP reports status

machineconfiguration/v1alpha1/types_pinnedimageset.go
```diff
@@ -0,0 +1,108 @@
...
+// PinnedImageRef is an image reference by digest.
+// It should be in the format of <registry>/<image>@sha256:<digest>
+// The total length of the image reference (<registry>/<image>) should be at most 256 characters.
+// ref. https://github.com/distribution/distribution/blob/6a568c100fe130b083d945511f379ea8be52db26/docs/content/spec/api.md#overview
+// The total length of the digest should be 64 characters + 8 (@sha256:) for a total of at most 72 characters.
+// +kubebuilder:validation:MaxLength=328
+// +kubebuilder:validation:Pattern:=`@sha256:[a-fA-F0-9]{64}$`
+type PinnedImageRef string
+
+type PinnedImageSetRef struct {
```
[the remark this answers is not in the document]
hexfusion: used by MCP

machineconfiguration/v1alpha1/types_pinnedimageset.go
```diff
@@ -0,0 +1,108 @@
...
+	// +listType=set
+	PinnedImages []PinnedImageRef `json:"pinnedImages"`
+}
+
+type MachineConfigPoolRef struct {
+	// name is the pool name.
+	// Must adhere to RFC-1035 (https://tools.ietf.org/html/rfc1035)
+	// It may consist of only alphanumeric characters, hyphens (-) and periods (.)
+	// and must be at most 63 characters in length.
+	// +kubebuilder:validation:MaxLength=63
```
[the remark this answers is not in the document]
hexfusion: OK I was going by metadata.name updating

machineconfiguration/v1alpha1/types_pinnedimageset.go
```diff
@@ -0,0 +1,108 @@
...
+	Name string `json:"name"`
+}
+
+// PinnedImageRef is an image reference by digest.
+// It should be in the format of <registry>/<image>@sha256:<digest>
+// The total length of the image reference (<registry>/<image>) should be at most 256 characters.
+// ref. https://github.com/distribution/distribution/blob/6a568c100fe130b083d945511f379ea8be52db26/docs/content/spec/api.md#overview
+// The total length of the digest should be 64 characters + 8 (@sha256:) for a total of at most 72 characters.
+// +kubebuilder:validation:MaxLength=328
+// +kubebuilder:validation:Pattern:=`@sha256:[a-fA-F0-9]{64}$`
```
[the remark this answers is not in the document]
hexfusion: I looked at CEL I am not 100% sure I can make this more clear that way.

I did update the regex to be more complete.

`^([a-zA-Z0-9-]+\.)*[a-zA-Z0-9-]+\/([a-zA-Z0-9-_]{1,61}\/)*[a-zA-Z0-9-_.]{1,61}@sha256:[a-f0-9]{64}$`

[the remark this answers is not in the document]
hexfusion: <hostname>/<namespace>/<name>@sha256:<digest>
253 + 1 + 61 + 1 +  61 + 8 + 64 = 447

I think that is correct

hexfusion:
There has been some back and forth with regards to do we need a new CRD vs can we inline into the MCP spec. As the enhancement has a core user story as pre-load-and-pin-application-images. That inlining into the spec would limit the capabilities as the set to remain within 1MB would need to max out at ~2000 images and its almost impossible to ensure this is reasonable.

So the PR still keeps a new PIS (eek) CRD which also will require an VerificationWebHook to ensure the name of the Set is valid before passing to MCP.

With regards to status the MCP would need to be in progress until the set was completed so as a user I could be reassured that the set was complete if all the machines were ready.

```
NAME     CONFIG                                             UPDATED   UPDATING   DEGRADED   MACHINECOUNT   READYMACHINECOUNT   UPDATEDMACHINECOUNT   DEGRADEDMACHINECOUNT   AGE
master   rendered-master-c3c6acd8f064f38fcffc4bbfb5cf9f4f   True      False      False      3              3                   3                     0                      8d
```

[1] https://github.com/openshift/enhancements/blob/master/enhancements/machine-config/pin-and-pre-load-images.md#pre-load-and-pin-application-images

machineconfiguration/v1alpha1/types_pinnedimageset.go
```diff
@@ -0,0 +1,108 @@
...
+	//              "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:...",
+	//              ...
+	//      ]
+	//
+	// These image references should all be by digest, tags aren't allowed.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=500
+	// +listType=set
+	PinnedImages []PinnedImageRef `json:"pinnedImages"`
```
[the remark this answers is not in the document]
hexfusion: adjusted count to 2000

machineconfiguration/v1alpha1/types_pinnedimageset.go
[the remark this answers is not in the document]
hexfusion: made the rror message more clear by using CEL match

sinnykumari:
Originally, we wanted to also have progress of pre-fetching of images on each node displayed into PIS. Wondering where are we going to track that now? Will it be incorporated into MachineConfigNode now?

hexfusion:
> Originally, we wanted to also have progress of pre-fetching of images on each node displayed into PIS. Wondering where are we going to track that now?

The new approach would move that status to MCP. I guess this needs a bit more discussion.

machineconfiguration/v1alpha1/types_pinnedimageset.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// PinnedImageSetSpec defines the desired state of a PinnedImageSet.
+type PinnedImageSetSpec struct {
+	// machineConfigPools define which pools the PinnedImageSet should be
+	// applied to. If the referenced pools do not exist, the PinnedImageSet will
+	// be ignored.
+
+	// +kubebuilder:validation:Required
+	// +listType=set
```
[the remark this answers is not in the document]
hexfusion: For example maybe  optional release when CVO has integration?
```
name:  payload
release: ocp-release:4.15.2-x86_64
```

JoelSpeed:
> Wondering where are we going to track that now? Will it be incorporated into MachineConfigNode now?

Pre-fetching sounds like a node level thing right? So I would expect MCN to host the status for each node's pre-fetch

JoelSpeed:
> So the PR still keeps a new PIS (eek) CRD which also will require an VerificationWebHook to ensure the name of the Set is valid before passing to MCP.

This works when the name is added to the MCP, but does not work if the PIS is later deleted, as the validation will then fail for an already persisted object. If you take this approach, you may need to make it a warning rather than failure. You'll need to account for the failure in other regards perhaps.

machineconfiguration/v1alpha1/types_pinnedimageset.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// PinnedImageRef is an OCI Image referenced by digest.
+//
+// The format of the image ref is:
+// host[:port][/namespace]/name@sha256:<digest>
+// +kubebuilder:validation:Required
+// +kubebuilder:validation:MinLength=1
+// +kubebuilder:validation:MaxLength=447
+// +kubebuilder:validation:XValidation:rule=`self.matches('^([a-zA-Z0-9-]+\\.)+[a-zA-Z0-9-]+(:[0-9]{2,5})?/([a-zA-Z0-9-_]{0,61}/)?[a-zA-Z0-9-_.]*@sha256:[a-f0-9]{64}$')`,message="The image reference must be in the format host[:port][/namespace]/name@sha256:<digest> with a valid SHA256 digest"
```
JoelSpeed, line 60: If we exclude the digest (split on @), then am I right in thinking that the rest of the repository name is a valid URL just without a scheme? I wonder if we can use the URL validation in CEL to our advantage here

```
isURL('https://'+self.split('@')[0])
```
We could also then validate the second half independently
```
self.split('@')[1].matches('^sha256:[a-f0-9]{64}$')
```
We would also want to check that the list when split does have 2 entries

But these breakdowns would potentially provide more specific feedback

machineconfiguration/v1/types.go
```diff
@@ -398,6 +398,24 @@ type MachineConfigPoolSpec struct {
 	// The targeted MachineConfig object for the machine config pool.
 	// +optional
 	Configuration MachineConfigPoolStatusConfiguration `json:"configuration"`
+
+	// pinnedImageSets is a list of PinnedImageSetRef objects that should be
+	// applied to the nodes in this pool.
+	// +openshift:enable:FeatureGate=PinnedImages
+	// +optional
+	// +listType=atomic
+	PinnedImageSets []PinnedImageSetRef `json:"pinnedImageSets"`
```
sinnykumari, line 407: One naive question: `PinnedImageSetRef` contains one element with type string. Is there a reason to make PinnedImageSetRef a struct instead of directly defining as string?

JoelSpeed: It allows for potential future expansion. For example, perhaps in the future we want to `errorIfMissing` or set a priority or something else for each referenced image set

hexfusion:
> > Wondering where are we going to track that now? Will it be incorporated into MachineConfigNode now?
> 
> Pre-fetching sounds like a node level thing right? So I would expect MCN to host the status for each node's pre-fetch

The implementation will be very tightly coupled with MCP and MCD. it will not add additional node annotations instead use those already used by MCD. 

So the goal will be to pivot to MCN when it becomes v1 but not before.

machineconfiguration/v1/types.go
```diff
@@ -398,6 +398,24 @@ type MachineConfigPoolSpec struct {
...
+}
+
+type PinnedImageSetRef struct {
+	// name is a reference to the name of a PinnedImageSet.
+	// Must adhere to RFC-1123 (https://tools.ietf.org/html/rfc1123)
```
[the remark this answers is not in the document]
hexfusion: >  Made up of one of more period-separated (.) segments

updated regex to reflect this as well. basically the old regex doesn't allow for 

```
^(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$
```

machineconfiguration/v1alpha1/types_pinnedimageset.go
hexfusion: hmm seems sort of hacky for minimal gain I believe the error message is fairly clear here but I will try to improve it a bit.

```
host[:port][/namespace]/name@sha256:<digest>
```

machineconfiguration/v1/types.go
```diff
@@ -398,6 +398,24 @@ type MachineConfigPoolSpec struct {
 	// The targeted MachineConfig object for the machine config pool.
 	// +optional
 	Configuration MachineConfigPoolStatusConfiguration `json:"configuration"`
+
+	// pinnedImageSets is a list of PinnedImageSetRef objects that should be
+	// applied to the nodes in this pool.
+	// +openshift:enable:FeatureGate=PinnedImages
+	// +optional
+	// +listType=atomic
```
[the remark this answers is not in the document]
hexfusion: updated

machineconfiguration/v1/types.go
```diff
@@ -398,6 +398,24 @@ type MachineConfigPoolSpec struct {
 	// The targeted MachineConfig object for the machine config pool.
 	// +optional
 	Configuration MachineConfigPoolStatusConfiguration `json:"configuration"`
+
+	// pinnedImageSets is a list of PinnedImageSetRef objects that should be
+	// applied to the nodes in this pool.
```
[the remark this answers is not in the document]
hexfusion: I believe this is 

```
// +patchStrategy=merge
// +patchMergeKey=name
```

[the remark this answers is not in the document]
JoelSpeed: Oh I more meant from an end user perspective, what happens if I have multiple PISs specified, what happens to the image lists, is there a priority in which they are pulled, is there a merge strategy for the lists, or are they pulled in order, if there's a maximal list of images that the CRI can handle, what happens if the combination of PISs exceeds this.

I'm basically trying to think through various failure modes and what they may mean to a user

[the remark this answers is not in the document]
hexfusion: Yeah I am updating the docs but basically, the controller will do a pre check to ensure that there is enough storage to pull the sets. If not it will set the pool degraded with an error message describing the problem.

So the CRI does not impose a limit to my knowledge it is limited by the resources availabl.e.

machineconfiguration/v1/types.go
```diff
@@ -421,6 +421,38 @@ type MachineConfigPoolSpec struct {
 	// The targeted MachineConfig object for the machine config pool.
 	// +optional
 	Configuration MachineConfigPoolStatusConfiguration `json:"configuration"`
+
+	// pinnedImageSets specifies a sequence of PinnedImageSetRef objects for the
+	// pool. Nodes within this pool will preload and pin images defined in the
+	// PinnedImageSet. Before pulling images the controller will ensure the
+	// total uncompressed size of all the images does not exceed available
```
JoelSpeed, line 428: What happens if we do exceed the available disk space?

machineconfiguration/v1alpha1/types_pinnedimageset.go
```diff
@@ -0,0 +1,80 @@
...
+	//              "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:...",
+	//              "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:...",
+	//              ...
+	//      ]
+	//
+	// These image references should all be by digest, tags aren't allowed.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=2000
+    // +listType=map
```
JoelSpeed, line 49: Indentation off

machineconfiguration/v1/types.go
hexfusion: controller does a precheck calculating the storage required for all images + buffer ~20GB - allocatable, before we pull. If this fails the pool goes degraded. The node is checked for KubeletHasNoDiskPressure if that at anytime is `true` controller workers will bail and retry if that can't be reconciled eventually it will be Unreconcilable. Also the node will become NotReady as a result of disk pressure (pretty sure).

Kubelet would force a GC on images(need to verify) and potentially evict pods. On the control-plane worst case it could take the cluster down. So there is a fair amount of assumption that the kubelet reports allocatable storage correctly.

hexfusion: For this reason, we will not pin the images until all images are pulled to allow the Kubelet the ability to GC>

hexfusion:
make verify is failing in the new TechPreview crd for v1 unrelated to this PR

```
error running generator schemacheck on group machineconfiguration.openshift.io: 
        could not run schemacheck generator for group/version machineconfiguration.openshift.io/v1: 
                error in 0000_80_machineconfigpool-CustomNoUpgrade.crd.yaml: ListsMustHaveSSATags: crd/machineconfigpools.machineconfiguration.openshift.io version/v1 field/^.spec.machineConfigSelector.matchExpressions must set x-kubernetes-list-type
                error in 0000_80_machineconfigpool-CustomNoUpgrade.crd.yaml: ListsMustHaveSSATags: crd/machineconfigpools.machineconfiguration.openshift.io version/v1 field/^.spec.machineConfigSelector.matchExpressions[*].values must set x-kubernetes-list-type
`

**Comment by hexfusion** [comment]
> [0000_80_machineconfigpool-Default.crd.yaml](https://github.com/openshift/api/pull/1713/files#diff-057a453383305642c544c30a4c1dea3de8cdbdb48c3901e20856aa6d272d698a)

Do I want this? It seems recent updates now generate this file, perhaps it should replace the other stable CRD?

**Comment by hexfusion** [comment]
> In some cases you may have changed or renamed a file that already contained api violations, but you are not introducing a new violation. In such cases it is appropriate to /override the failing CI job.

:)

File: machineconfiguration/v1/custom.machineconfigpool.testsuite.yaml [hunk 30]
```diff
@@ -0,0 +1,18 @@
...
+crd: 0000_80_machineconfigpool-TechPreviewNoUpgrade.crd.yaml
+tests:
+  onCreate:
+  - name: Should be able to create a minimal MachineConfigPool
+    initial: |
+      apiVersion: machineconfiguration.openshift.io/v1
+      kind: MachineConfigPool
+      spec:
+        pinnedImageSets:
+          - name: test-pinnedimageset
```
JoelSpeed, line 12: This isn't a required field is it? Generally the `minimal` version of this test shows only required fields

machineconfiguration/v1alpha1/types_pinnedimageset.go
```diff
@@ -0,0 +1,80 @@
...
+type PinnedImageRef struct {
+	// name is an OCI Image referenced by digest.
+	//
+	// The format of the image ref is:
+	// host[:port][/namespace]/name@sha256:<digest>
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=447
+	// +kubebuilder:validation:XValidation:rule=`self.matches('^([a-zA-Z0-9-]+\\.)+[a-zA-Z0-9-]+(:[0-9]{2,5})?/([a-zA-Z0-9-_]{0,61}/)?[a-zA-Z0-9-_.]*@sha256:[a-f0-9]{64}$')`,message="The OCI Image reference must be in the format host[:port][/namespace]/name@sha256:<digest> with a valid SHA256 digest"
+	Name string `json:"name"`
```
yuqi-zhang, line 63: I'm wondering if there were alternatives to just "name" here, as this is more of a... url? digest? I guess I don't have a good alternative

yuqi-zhang (commented):
Overall makes sense to me from the MCO perspective. Do you foresee the need to add any new status types?

hexfusion:
THis PR evolved a lot over time as the design drifted away from the enhancement while this is good historical context closing and moving to #1822

PR #1716, opened 2024-01-08
AUTH-439: authentication: use CRD validation for the 'type' field

stlaz:
/assign @deads2k

### DIFF
--- a/config/v1/0000_10_config-operator_01_authentication.crd.yaml
+++ b/config/v1/0000_10_config-operator_01_authentication.crd.yaml
@@ -52,6 +52,10 @@ spec:
                 type:
                   description: type identifies the cluster managed, user facing authentication mode in use. Specifically, it manages the component that responds to login attempts. The default is IntegratedOAuth.
                   type: string
+                  enum:
+                    - ""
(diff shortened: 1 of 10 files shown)


config/v1/0000_10_config-operator_01_authentication.crd.yaml
```diff
@@ -52,6 +52,10 @@ spec:
                 type:
                   description: type identifies the cluster managed, user facing authentication mode in use. Specifically, it manages the component that responds to login attempts. The default is IntegratedOAuth.
                   type: string
+                  enum:
```
deads2k, line 55: What happened in 4.14 if this string value had a different value for this field?

stlaz: we had a separate validation in our o/k fork that did not allow this - https://github.com/openshift/kubernetes/pull/1770/

deads2k:
This was previously restricted in kube-apiserver: https://github.com/openshift/kubernetes/pull/1770, therefore this doesn't change compatibility

/lgtm
/approve

PR #1718, opened 2024-01-08
OCPBUGS-25821: add MCO management of api-int CA

cdoern:
the MCO needs to add some support for long living certs (10 years) that get populated in /etc/kubernetes/kubeconfig. These certs come from the openshift-config-managed/kube-apiserver-server-ca configmap and currently are only applied on bootstrap. So, if a cluster rotates often or exists for a long period of time, a user will eventually run into the "certificate signed by unknown authority" error when these certs expire.

What I did here is 
- add `KubeAPIServerClientCA` to represent what was previously being stored in `KubeAPIServerServingCAData`. We were storing the client-ca data in the serverca entity. 
- made `KubeAPIServerServingCAData` store the data for the api-int CA


machineconfiguration/v1/types.go
```diff
@@ -100,6 +100,11 @@ type ControllerConfigSpec struct {
 	// +nullable
 	InternalRegistryPullSecret []byte `json:"internalRegistryPullSecret"`
 
+	// InternalAPICert is used to generate and update /etc/kubelet/kubeconfig
```
wking, line 103: I don't think it's clear from the Godocs/`description` how this differs from `kubeAPIServerServingCAData`.  In fact, I'd initially thought (from its Godocs/`description`) this was what `kubeAPIServerServingCABytes` was for, but it turns out that's currently [populated from consumes `kube-apiserver-client-ca`](https://github.com/openshift/machine-config-operator/blob/e0dda8272f54efbf738ef43ea4b0e7b21450b7bf/pkg/operator/sync.go#L458) and [written](https://github.com/openshift/machine-config-operator/pull/3575/files#diff-504415817512e4a226a6b024a2defbba204a1ff821e1a10c6b797ca3219d58e4R105-R107) to [`/etc/kubernetes/kubelet-ca.crt`](https://github.com/yuqi-zhang/machine-config-operator/blob/4fef55681565469c588ae888e3e244a11f995c34/pkg/daemon/daemon.go#L185). Perhaps this unwinding can add a new `kubeAPIServerClientCA` or some such to hold that `kube-apiserver-client-ca` to `/etc/kubernetes/kubelet-ca.crt` pipe? And we can pivot to using `kubeAPIServerServingCAData` for the `kube-apiserver-server-ca` to `/var/lib/kubelet/kubeconfig` pipe?

machineconfiguration/v1/types.go
```diff
@@ -100,6 +100,11 @@ type ControllerConfigSpec struct {
...
+	// +kubebuilder:validation:Required
+	// +nullable
+	InternalAPICert []byte `json:"internalAPICert"`
```
deads2k, line 106: the CA bundle to use is https://github.com/openshift/api/tree/master/tls/docs/kube-apiserver%20Serving%20Certificates#kube-apiserver-total-serving-ca located at `-n openshift-config-managed 	configmap/kube-apiserver-server-ca`

yuqi-zhang:
/close

We're going to close this and not do this management with an API process for now so its no longer attached to the bug. We'll revisit in newer versions

PR #1722, opened 2024-01-10
OCPBUGS-26498: Add UnservableInFutureVersions route status condition type

gcs278:
Add UnservableInFutureVersions as a route status condition type. This change allows the router to indicate that a route is using an unsupported configuration and will be unservable in the future. The ingress operator may use this condition to block upgrades.

More details in forum-ocp-updates [slack thread](https://redhat-internal.slack.com/archives/CEGKQ43CP/p1704894930666179) and latest updates discussion in forum-api-review [slack thread](https://redhat-internal.slack.com/archives/CE4L0F143/p1706626476105289)

### DIFF
--- a/route/v1/types.go
+++ b/route/v1/types.go
@@ -369,14 +369,16 @@ type RouteIngressConditionType string
 const (
 	// RouteAdmitted means the route is able to service requests for the provided Host
 	RouteAdmitted RouteIngressConditionType = "Admitted"
-	// TODO: add other route condition types
+	// RouteUnservableInFutureVersions indicates that the route is using an unsupported
+	// configuration that may be incompatible with a future version of OpenShift.
+	RouteUnservableInFutureVersions RouteIngressConditionType = "UnservableInFutureVersions"
 )
 
 // RouteIngressCondition contains details for the current condition of this route on a particular
 // router.
 type RouteIngressCondition struct {
 	// Type is the type of the condition.
-	// Currently only Admitted.
+	// Currently only Admitted or UnservableInFutureVersions.
 	Type RouteIngressConditionType `json:"type" protobuf:"bytes,1,opt,name=type,casttype=RouteIngressConditionType"`
 	// Status is the status of the condition.
 	// Can be True, False, Unknown.
--- a/openapi/generated_openapi/zz_generated.openapi.go
+++ b/openapi/generated_openapi/zz_generated.openapi.go
@@ -56730,7 +56730,7 @@ func schema_openshift_api_route_v1_RouteIngressCondition(ref common.ReferenceCal
 				Properties: map[string]spec.Schema{
 					"type": {
 						SchemaProps: spec.SchemaProps{
-							Description: "Type is the type of the condition. Currently only Admitted.",
+							Description: "Type is the type of the condition. Currently only Admitted or UnservableInFutureVersions.",
 							Default:     "",
 							Type:        []string{"string"},
 							Format:      "",
--- a/openapi/openapi.json
+++ b/openapi/openapi.json
@@ -33386,7 +33386,7 @@
           "default": ""
         },
         "type": {
-          "description": "Type is the type of the condition. Currently only Admitted.",
+          "description": "Type is the type of the condition. Currently only Admitted or UnservableInFutureVersions.",
           "type": "string",
           "default": ""
         }
--- a/route/v1/generated.proto
+++ b/route/v1/generated.proto
@@ -213,7 +213,7 @@ message RouteIngress {
 // router.
 message RouteIngressCondition {
   // Type is the type of the condition.
-  // Currently only Admitted.
+  // Currently only Admitted or UnservableInFutureVersions.
   optional string type = 1;
 
   // Status is the status of the condition.
--- a/route/v1/route-CustomNoUpgrade.crd.yaml
+++ b/route/v1/route-CustomNoUpgrade.crd.yaml
@@ -581,7 +581,7 @@ spec:
                             type: string
                           type:
                             description: Type is the type of the condition. Currently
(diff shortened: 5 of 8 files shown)


route/v1/types.go
```diff
@@ -369,14 +369,15 @@ type RouteIngressConditionType string
 const (
 	// RouteAdmitted means the route is able to service requests for the provided Host
 	RouteAdmitted RouteIngressConditionType = "Admitted"
-	// TODO: add other route condition types
+	// RouteUpgradable indicates the route is upgradeable to the next version of OpenShift
```
deads2k, line 372: please describe what happens if this is false.  Is the upgrade stopped?  That's a lot of power to give a namespace editor compared a cluster-admin assessing a CVE for instance.

If the upgrade isn't stopped, how do we expect a cluster-admin to avoid accidentally causing this situation?

cc @sdodson 

for discussion of how to handle upgradeability.

gcs278: > please describe what happens if this is false. Is the upgrade stopped? That's a lot of power to give a namespace editor compared a cluster-admin assessing a CVE for instance.

I will update code comment soon - the cluster ingress operator will read the route's Upgradeable status and set the cluster operator ingress Upgradeable status which will prevent upgrades. So yes, the route owner can stop an upgrade, agreed.

See https://redhat-internal.slack.com/archives/CEGKQ43CP/p1704894930666179 discussion.

gcs278:
@deads2k I'd like to push forward with this change. @Miciah and I discussed the API changes.

We currently have settled on a `Upgradeable` condition on the route after exploring using `Deprecated` because `Deprecated` doesn't exactly imply the route won't be upgradeable (and we want to imply that). We liked the flexibility of `Deprecated`, in that we could backport further to 4.14 and warn users even earlier. But this would require the ingress operator to parse the `condition.reason` field, to determine ingress upgradeable status, which seems kludgy.

The other option would be using annotation. That seemed quite kludgy too.

Please advise if you have any concerns or questions.

gcs278:
/hold
Actually, NE discussed this again. We discussed in favor of using `Deprecated` due to its greater flexibility compared to `Upgradeable`. My assumption that we would need to parse the `Deprecated` message is not necessarily true (at least, for the initial implementation). I was designing a more generic solution in which we'd have "permanent" code (4.15+) in the ingress operator that would interpret `Deprecated` conditions and take upgradeable action based on their reason or message.

The other solution is to just have the ingress operator upgradeable code temporary, so it only exists in the OCP version N-1 where N is the option (e.g. SHA1) is unsupported. The code would be make an assumption that any `Deprecated` route status in 4.15 is going to block upgrades via an admin ack.

It becomes more complex if we intend to reuse the deprecated status in the future to determine the co's ingress `upgradeable=false` because a user could have ack'd an SHA1 route and it could still have the deprecated condition. In that case, we have to start parsing the deprecated condition's reason/message to ensure we don't pick up the old SHA1 deprecated condition, which would cause an admin to have to re-ack it. Not the biggest concern, but it's something we can handle later down the road if this scenario ever arises again. CC: @Miciah

route/v1/types.go
```diff
@@ -369,14 +369,17 @@ type RouteIngressConditionType string
 const (
 	// RouteAdmitted means the route is able to service requests for the provided Host
 	RouteAdmitted RouteIngressConditionType = "Admitted"
-	// TODO: add other route condition types
+	// RouteDeprecated indicates that the route is using a deprecated configuration that may
+	// be incompatible with a future version of OpenShift. If this set to true on any route,
+	// the ingress operator may block upgrades.
```
wking, line 374: micro-nit, feel free to ignore.

> If this set to true on any route, the ingress operator may block upgrades.

This seems like an internal implementation detail.  In cases where the ingress operator decides to use `Deprecated` on `Routes` to drive `Upgradeable=False`, that `Upgradeable` ClusterOperator condition is the API talking to the CVO, and the CVO will in turn set `Upgradeable=False` on ClusterVersion, and that ClusterVersion condition is the API talking to the cluster-admin or other OCP-managing actor.  I don't think we need to hint at that possible chain in this Route condition's docs.  Instead, we can keep the first sentence explaining the deprecation aspect, which lets Route-admins know that they should be thinking about a config update.  And then leave it to the cluster-admin-facing ClusterVersion API to talk about whether we're concerned about anything (including deprecated Route config) to talk to cluster-admins about that aspect.

But also, "may block upgrades" is wiggly enough that I doubt anyone will feel like we are surprising them regardless of whether the ingress operator decides to take any action based on this new Route condition.  So no worries if you want to leave this sentence in as a semi-informative bread-crumb.

gcs278: thanks for the info - I remove it as I agree it's an implementation detail.

route/v1/types.go
```diff
@@ -369,14 +369,17 @@ type RouteIngressConditionType string
...
+	RouteDeprecated RouteIngressConditionType = "Deprecated"
```
deads2k, line 375: The condition should be more assertive of the problem.  Deprecated is generally read in a "well someday I'll fix this" and we're actually highlight a security concern that we know will prevent this route from functioning in two versions.

`UnservableInFutureVersions` could carry a reason and message that we choose for each case where you need this case.  The reason and message should emphasize the insecure aspect of the route.

route/v1/types.go
gcs278: Whoops, I forgot to remove this.

route/v1/types.go
gcs278: Done.

gcs278:
There's enough discussion and agreement that the hold is no longer needed
/hold cancel
@deads2k @frobware @Miciah ready for review

PR #1739, opened 2024-01-24
WIP: Enable UnauthenticatedHTTP2DOSMitigation by default

dgrisonnet:
This should help getting rid of the carry patch that we previously added to enforce the `UnauthenticatedHTTP2DOSMitigation` feature-gate


dgrisonnet:
Closing as the feature gate is enabled by default since Kubernetes 1.29.

PR #1752, opened 2024-02-05
OCPBUGS-18991: Add BuildInterrupted type

cdoern:
BuildInterrupted indicates that the build has been stopped by some sort of user action or outside force. It is not necessarily a failure but causes the build to be thrown away.


JoelSpeed (approved):
/lgtm

PR #1764, opened 2024-02-09
MCO-1010: Add node disruption policies to MachineConfiguration CRD

yuqi-zhang:
Draft API of https://github.com/openshift/enhancements/pull/1525

This extension of MachineConfiguration object allows uses to specify how their MachineConfig object changes affect node disruption, allowing for non-drain and non-reboot updates to some config files. The MachineConfigController and MachineConfigDaemon will ultimately be implementing and executing on this object.

Also currently based on https://github.com/openshift/api/pull/1672 by @djoshy since that should probably merge first

Major questions:
1. re lists, we currently want users to specify policies such as follows:
```
userPolicies:
  - type: file
    value: "/etc/my-file"
    actions:
      - type: reload
        reload: my-service
      - type: daemon-reload
  - type: file
    value: "/etc/my-other-file"
    actions:
      - type: none
  - type: sshkey
    actions:
      - type: none
```

such that the list doesn't have unique keys (it would be type + value, but value isn't required for all types so it cannot be made into a listMapKey). Maybe it would be better to make it more MachineConfig like and have it instead be:

```
userPolicies:
  files:
    - path: "/etc/my-file"
      actions:
        - type: reload
          reload: my-service
        - type: daemon-reload
    - path: "/etc/my-other-file"
      actions:
        - type: none
  sshkey:
    actions:
      - type: none
```

for the actions, we have the same issue, as we'd like to allow users to specify multiple actions with the same key
```
  - type: file
    value: "/etc/my-other-file"
    actions:
      - type: reload
        reload: service-1
      - type: reload
        reload: service-2
      - type: daemon-reload
```

So there is not a unique key here either.

Currently the proposed set up is, actions are union discriminators (so if you reload, you have to specify the services to reload), type/value pairs are also validated, and otherwise there is no verification on the lists. The MachineConfigController will do the rest of the validation, so a split of responsibilities.

Alternatively, we could either:
1. have no validation in the API, and have the MCC do all the validation
2. have no validation in the MCC and have the API logic do all the heavy lifting

controller validation and status:

Currently we list out clusterDefaultPolicies in the spec, which will be populated by the MCController every time it updates. Users should not modify that sub-spec, but since the daemon will only apply what's in status (so validated version) the current approach has no additional gating and just has the controller overwrite any changes before applying to status.

Is there a way we would be able to let just the MCController (and not say, someone using the MCC service account or other admin privs) write that sub-object? It should be mutate-able, just not by anything other than the MCO/MCC


operator/v1/types_machineconfiguration.go
```diff
@@ -156,3 +169,116 @@ const (
...
+
+	// clusterDefaultPolicies is managed by the Machine Config Operator, and reflects the latest cluster defaults
+	// +optional
+	ClusterDefaultPolicies []NodeDisruptionPolicyConfig `json:"clusterDefaultPolicies"`
+}
+
+type NodeDisruptionPolicyStatus struct {
+	// clusterPolicies is a merge of cluster default and user provided node disruption policies.
+	// +optional
+	ClusterPolicies []NodeDisruptionPolicyConfig `json:"clusterPolicies"`
```
deads2k, line 186: status should use a different type because you're likely to grow different fields

operator/v1/types_machineconfiguration.go
```diff
@@ -156,3 +169,116 @@ const (
...
+type NodeDisruptionPolicyAction struct {
+	// type represents the commands that will be carried out if this NodeDisruptionPolicyActionType is executed
+	// Valid value(s): reboot, drain, reload, daemon-reload, none, special
+	// reload requires a corresponding service target specified in the reload field. All other fields cannot have a reload field
+	// +unionDiscriminator
+	// +kubebuilder:validation:Required
+	Type NodeDisruptionPolicyActionType `json:"type"`
+	// reload specifies the service to reload, only valid if type is reload
+	// +optional
+	Reload *string `json:"reload,omitempty"`
```
deads2k, line 258: probably want a struct

operator/v1/types_machineconfiguration.go
```diff
@@ -156,3 +169,116 @@ const (
...
+	// fields (files, services, kernel arguments, and sshkeys)
+	// +kubebuilder:validation:Required
+	Type NodeDisruptionPolicyType `json:"type"`
+	// value represents the canoncial name of the object that this policy is in effect for.
+	// This could be a file path, a service name, the specific kernel argument, etc.
+	// +optional
+	Value string `json:"value"`
+	// actions represents the series of commands to be executed on changes to the corresponding type and value
+	// +kubebuilder:validation:Required
+	Actions []NodeDisruptionPolicyAction `json:"actions"`
```
deads2k, line 211: this should probably be atomic

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +78,207 @@ type MachineConfigurationList struct {
...
+type NodeDisruptionPolicyType string
+
+const (
+	// File represents a NodeDisruption policy that is in effect for changes to a file.
+	File NodeDisruptionPolicyType = "file"
+
+	// Service represents a NodeDisruption policy that is in effect for changes to a service.
+	Service NodeDisruptionPolicyType = "service"
+
+	// File represents a NodeDisruption policy that is in effect for changes to a kernel argument.
```
sinnykumari, line 239:
```suggestion
	// kernelArgument represents a NodeDisruption policy that is in effect for changes to a kernel argument.
```

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +78,207 @@ type MachineConfigurationList struct {
...
+	KernelArgument NodeDisruptionPolicyType = "kernelArgument"
```
sinnykumari, line 240: MCO doesn't have a way today to apply kargs without reboot. Are we thinking of adding a way to apply these kargs live? If not, perhaps we may want to skip letting user skip drain/reboot in the initial implementation.

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +78,207 @@ type MachineConfigurationList struct {
...
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'reload' ? has(self.reload) : !has(self.reload)",message="Reload is required when type is reload, and forbidden otherwise"
+// +union
+type NodeDisruptionPolicyAction struct {
+	// type represents the commands that will be carried out if this NodeDisruptionPolicyActionType is executed
+	// Valid value(s): reboot, drain, reload, daemon-reload, none, special
+	// reload requires a corresponding service target specified in the reload field. All other fields cannot have a reload field
+	// +unionDiscriminator
+	// +kubebuilder:validation:Required
+	Type NodeDisruptionPolicyActionType `json:"type"`
+	// reload specifies the service to reload, only valid if type is reload
```
sinnykumari, line 256: Admin maybe benefited with  `service restart ` as well as I beleive sometime just reloading service is not enough to apply the config changes.

operator/v1/types_machineconfiguration.go
sinnykumari: should this be a list so that user can specify multiple servicename together that needs to be reloaded for a NodeDisruptionPolicyType ?

operator/v1/types_machineconfiguration.go
yuqi-zhang: done

operator/v1/types_machineconfiguration.go
yuqi-zhang: done. I opted them to be separate so a user can theoretically order them as they wish to apply in sequence with other actions

operator/v1/types_machineconfiguration.go
```diff
@@ -41,12 +41,25 @@ type MachineConfigurationSpec struct {
 	// +openshift:enable:FeatureSets=TechPreviewNoUpgrade
 	// +optional
 	ManagedBootImages ManagedBootImages `json:"managedBootImages"`
+
+	// nodeDisruptionPolicySpec allows an admin to set granular node disruption actions for
+	// MachineConfig-based updates, such as drains, service reloads, etc. Specifying this will allow
+	// for less downtime when doing small configuration updates to the cluster.
```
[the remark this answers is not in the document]
yuqi-zhang: added a line

operator/v1/types_machineconfiguration.go
```diff
@@ -41,12 +41,25 @@ type MachineConfigurationSpec struct {
...
+	// +openshift:enable:FeatureSets=TechPreviewNoUpgrade
```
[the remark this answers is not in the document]
yuqi-zhang: done

operator/v1/types_machineconfiguration.go
```diff
@@ -41,12 +41,25 @@ type MachineConfigurationSpec struct {
...
 }
 
 type MachineConfigurationStatus struct {
 	StaticPodOperatorStatus `json:",inline"`
 
 	// TODO(jkyros): This is where we can put additional bespoke status fields
+
+	// nodeDisruptionPolicyStatus status reflects what the latest cluster-validated policies are,
+	// and will be used by the Machine Config Daemon during future node updates.
+	// +openshift:enable:FeatureSets=TechPreviewNoUpgrade
```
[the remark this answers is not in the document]
yuqi-zhang: done

operator/v1/types_machineconfiguration.go
yuqi-zhang: removed kargs as per review on the enhacement

operator/v1/types_machineconfiguration.go
```diff
@@ -56,3 +78,207 @@ type MachineConfigurationList struct {
...
+	Reload NodeDisruptionPolicyActionType = "reload"
+
+	// DaemonReload represents an action that TBD
+	DaemonReload NodeDisruptionPolicyActionType = "daemon-reload"
+
+	// None represents an action that no handling is required by the MCO.
+	None NodeDisruptionPolicyActionType = "none"
+
+	// Special represents an action that is internal to the MCO, and is not allowed in user defined NodeDisruption policies.
+	Special NodeDisruptionPolicyActionType = "special"
```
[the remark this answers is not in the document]
yuqi-zhang: I'm open to discussion on this, I don't intend this to be user specify-able

operator/v1/types_machineconfiguration.go
yuqi-zhang: added, thanks!

operator/v1/types_machineconfiguration.go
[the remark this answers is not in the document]
sinnykumari: Since removing API later is hard, maybe we can add this or similar field later on depending upon usecase?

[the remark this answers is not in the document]
yuqi-zhang: The use case right now is that we would like to display the current cluster defaults, so the user can see (and optionally override) the image registry logic. The initial set we decided on is to have this keyword and description (that you can see via `oc describe`, etc.) to explain what the default is.

Happy to change this if there's a better route though! Mostly just wanted to be able to show the current cluster settings.

This right now is also tech preview only so we should be able to change it before GA too if needed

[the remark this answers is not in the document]
sinnykumari: Thanks for the context. Let's keep it then and we can update it if needed while API will be in TechPreview. Maybe MCOInternal, MCODefault etc would provide more clarity.

operator/v1/types_machineconfiguration.go
```diff
@@ -153,3 +167,181 @@ const (
...
+type NodeDisruptionPolicySSHKey struct {
+	// actions represents the series of commands to be executed on changes to the corresponding type and value
+	// +kubebuilder:validation:Required
+	// +listType=atomic
+	Actions []NodeDisruptionPolicyAction `json:"actions"`
+}
+
+// Validation is a string enum used in a NodeDisruptionPolicyStatus object. They describe whether the controller
+// was able to validate the latest spec
+// +kubebuilder:validation:Enum:="success";"failed"
```
JoelSpeed, line 261: Enums like this should be PascalCase, use `Success` and `Failed`

operator/v1/types_machineconfiguration.go
```diff
@@ -153,3 +167,181 @@ const (
...
+	// success represents a valid cluster node disruption policy
+	Success Validation = "success"
+
+	// failed represents an invalid cluster node disruption policy
+	Failed Validation = "failed"
+)
+
+// NodeDisruptionPolicyType is a string enum used in a NodeDisruptionPolicyConfig object. They describe the type of
+// the object that this policy will be in effect for.
+// +kubebuilder:validation:Enum:="file";"service";"sshKey"
```
JoelSpeed, line 274: Use `File`, `Service`, `SSHKey`

operator/v1/types_machineconfiguration.go
```diff
@@ -153,3 +167,181 @@ const (
...
+	File NodeDisruptionPolicyType = "file"
+
+	// Service represents a NodeDisruption policy that is in effect for changes to a service.
+	Service NodeDisruptionPolicyType = "service"
+
+	// Service represents a NodeDisruption policy that is in effect for changes to a sshKey
+	SSHKey NodeDisruptionPolicyType = "sshKey"
+)
+
+// TODO maybe do this in the MCC validation and skip here?
```
JoelSpeed, line 288: Prefer to keep as much API validation at admission time as possible rather than later, either this should be done here or in a webhook, CEL is preferable in this case IMO

operator/v1/types_machineconfiguration.go
```diff
@@ -153,3 +167,181 @@ const (
...
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'reload' ? has(self.reload) : !has(self.reload)",message="Reload is required when type is reload, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'restart' ? has(self.restart) : !has(self.restart)",message="Restart is required when type is restart, and forbidden otherwise"
+// +union
+type NodeDisruptionPolicyAction struct {
+	// type represents the commands that will be carried out if this NodeDisruptionPolicyActionType is executed
+	// Valid value(s): reboot, drain, reload, restart, daemon-reload, none, special
```
JoelSpeed, line 294: Values of this enum should be pascal case

operator/v1/types_machineconfiguration.go
```diff
@@ -153,3 +167,181 @@ const (
...
+	// reload/restart requires a corresponding service target specified in the reload/restart field. All other fields cannot have a these fields
```
JoelSpeed, line 295: Maybe `Other values require no further configuration`?

operator/v1/types_machineconfiguration.go
```diff
@@ -153,3 +167,181 @@ const (
...
+	// +optional
+	Reload *ReloadService `json:"reload,omitempty"`
+	// restart specifies the service to restart, only valid if type is restart
+	// +optional
+	Restart *RestartService `json:"restart,omitempty"`
+}
+
+// ReloadService allows the user to specify the services to be reloaded
+type ReloadService struct {
+	// ServiceName is the full name (e.g. crio.service) of the service to be reloaded
```
JoelSpeed, line 309: Should be lower case at the start of the godoc. Similar questions above about service and the godoc here

operator/v1/types_machineconfiguration.go
```diff
@@ -153,3 +167,181 @@ const (
...
+	// +kubebuilder:validation:Required
+	ServiceName string `json:"serviceName,omitempty"`
+}
+
+// RestartService allows the user to specify the services to be restarted
+type RestartService struct {
+	// ServiceName is the full name (e.g. crio.service) of the service to be restarted
```
JoelSpeed, line 316: Ditto for reload

operator/v1/types_machineconfiguration.go
```diff
@@ -153,3 +167,181 @@ const (
...
+	// +kubebuilder:validation:Required
+	ServiceName string `json:"serviceName,omitempty"`
+}
+
+// NodeDisruptionPolicyActionType is a string enum used in a NodeDisruptionPolicyAction object. They describe an action to be performed.
+// +kubebuilder:validation:Enum:="reboot";"drain";"reload";"restart";"daemon-reload";"none";"special"
```
JoelSpeed, line 322: PascalCase these values please

operator/v1/types_machineconfiguration.go
```diff
@@ -153,3 +167,181 @@ const (
...
+	Drain NodeDisruptionPolicyActionType = "drain"
+
+	// Reload represents an action that will cause nodes to reload the service described by the Target field.
+	Reload NodeDisruptionPolicyActionType = "reload"
+
+	// Restart represents an action that will cause nodes to restart the service described by the Target field.
+	Restart NodeDisruptionPolicyActionType = "restart"
+
+	// DaemonReload represents an action that TBD
+	DaemonReload NodeDisruptionPolicyActionType = "daemon-reload"
```
JoelSpeed, line 340: Should be `DaemonReload`

operator/v1/types_machineconfiguration.go
```diff
@@ -153,3 +167,181 @@ const (
...
+
+	// None represents an action that no handling is required by the MCO.
+	None NodeDisruptionPolicyActionType = "none"
+
+	// Special represents an action that is internal to the MCO, and is not allowed in user defined NodeDisruption policies.
```
JoelSpeed, line 345: Need to make sure the spec for user defined does not allow this

operator/v1/types_machineconfiguration.go
```diff
@@ -41,12 +41,26 @@ type MachineConfigurationSpec struct {
 	// +openshift:enable:FeatureSets=CustomNoUpgrade;TechPreviewNoUpgrade
 	// +optional
 	ManagedBootImages ManagedBootImages `json:"managedBootImages"`
+
+	// nodeDisruptionPolicySpec allows an admin to set granular node disruption actions for
+	// MachineConfig-based updates, such as drains, service reloads, etc. Specifying this will allow
+	// for less downtime when doing small configuration updates to the cluster. This is NOT intended
```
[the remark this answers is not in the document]
yuqi-zhang: Will change, a cluster upgrade in practice almost always has an associated OS update, even if it's just a minor package bump. In all of OCP 4 I think I've seen one z-stream bump not come with an associated update, so the overlap is 0

operator/v1/types_machineconfiguration.go
```diff
@@ -153,3 +167,181 @@ const (
...
 )
+
+type NodeDisruptionPolicySpec struct {
+	// userPolicies define user-provided node disruption policies
+	// +optional
+	UserPolicies NodeDisruptionPolicyConfig `json:"userPolicies"`
+
+	// clusterDefaultPolicies is managed by the Machine Config Operator, and reflects the latest cluster defaults
+	// +optional
+	ClusterDefaultPolicies NodeDisruptionPolicyConfig `json:"clusterDefaultPolicies"`
```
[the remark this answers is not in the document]
yuqi-zhang: The current design has it such that there is:
1. a user spec, for the user to set
2. a cluster spec, for the MCO to set, which may change depending on version
3. a status which is the validated merge of the two

The expectation being that it is easier for the user to see what the cluster defaults are.

If we remove this from the spec, then the status will hold the cluster defaults + anything overriden by the user, which... I guess still mostly achieves our goal?

operator/v1/types_machineconfiguration.go
```diff
@@ -153,3 +167,181 @@ const (
 	// This feature only supports MAPI machinesets at this time.
 	MachineAPI MachineManagerMachineSetsAPIGroupType = "machine.openshift.io"
 )
+
+type NodeDisruptionPolicySpec struct {
+	// userPolicies define user-provided node disruption policies
+	// +optional
+	UserPolicies NodeDisruptionPolicyConfig `json:"userPolicies"`
```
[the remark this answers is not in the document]
yuqi-zhang: Off the top of my head any potential future extensions would be with the new image based workflow. I think an example could be I could specify some additional non-machineconfig content (image based) and specify something here to indicate how it can be applied.

But maybe for now we can just collapse this to the higher level policy and add a new field when that comes into play?

operator/v1/types_machineconfiguration.go
```diff
@@ -153,3 +167,181 @@ const (
...
+	// clusterPolicies is a merge of cluster default and user provided node disruption policies.
+	// +optional
+	ClusterPolicies NodeDisruptionPolicyClusterStatus `json:"clusterPolicies"`
+
+	// conditions represent the current validation status of the NodeDisruptionPolicy in spec
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
```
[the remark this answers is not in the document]
yuqi-zhang: Related to the discussion we had in the call, I think this makes more sense living in the `MachineConfigurationStatus` object directly, which currently has `StaticPodOperatorStatus` that we were thinking of deprecating. Could you help me point to a resource on how that should be done?

operator/v1/types_machineconfiguration.go
```diff
@@ -153,3 +167,181 @@ const (
...
+	// +optional
+	// +listType=map
+	// +listMapKey=path
+	Files []NodeDisruptionPolicyFile `json:"files,omitempty"`
+	// units is a list MachineConfig unit definitions and actions to take on changes to those services
+	// +optional
+	// +listType=map
+	// +listMapKey=name
+	Units []NodeDisruptionPolicyUnit `json:"units,omitempty"`
+	// sshkey is the overall sshkey MachineConfig definition
```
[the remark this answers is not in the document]
yuqi-zhang: Added a comment. Essentially the sshkeys are a separate field in MachineConfigs, and today we already have cluster defaults to do a no-op.

So most of the time this won't need any additional input from the user, but we'd like to expose it for the status

operator/v1/types_machineconfiguration.go
```diff
@@ -153,3 +167,181 @@ const (
...
+	// +optional
+	SSHKey NodeDisruptionPolicySSHKey `json:"sshkey,omitempty"`
+}
+
+// NodeDisruptionPolicyFile is a file entry and corresponding actions to take
+type NodeDisruptionPolicyFile struct {
+	// path represents the filepath
+	// +kubebuilder:validation:Required
+	Path string `json:"path"`
+	// actions represents the series of commands to be executed on changes to the corresponding type and value
```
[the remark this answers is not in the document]
yuqi-zhang: Added a lengthier explanation, let me know if that makes sense

operator/v1/types_machineconfiguration.go
```diff
@@ -153,3 +167,181 @@ const (
...
+	// +kubebuilder:validation:Required
+	// +listType=atomic
+	Actions []NodeDisruptionPolicyAction `json:"actions"`
+}
+
+// NodeDisruptionPolicyUnit is a systemd unit name and corresponding actions to take
+type NodeDisruptionPolicyUnit struct {
+	// name represents the service name
```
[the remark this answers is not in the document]
yuqi-zhang: Ignition does apply a valid name check it looks like: https://github.com/coreos/ignition/blob/d84b20fadef7cedfd3577f27272c89b5ddb0f4d2/config/v3_4/types/unit.go#L48-L55

Should we do that here? I was thinking of having the MCO validation take care of that

operator/v1/types_machineconfiguration.go
yuqi-zhang: Removed in favour of having a more proper condition, instead of a "success/fail" flag

yuqi-zhang:
Fixed based on comments and rebased on master

operator/v1/types_machineconfiguration.go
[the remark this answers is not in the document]
JoelSpeed: Ok, you should use separate structs for spec and status, that way you have the option to expose in status but not make available in spec

operator/v1/types_machineconfiguration.go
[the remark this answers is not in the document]
JoelSpeed: Ideally, we apply as much validation as possible at admission time, so that we never persist an invalid object. If it's possible to validate this name here (which it looks like it is), then we should aim to do so.

We prefer openapi validations (eg CEL and kubebuilder patterns), but validating webhooks can also be used for more complex validations. But the ideal world is that all validations happen at admission, and are not "Check by controller, set status bad", the latter pattern meaning a user can apply the object, get a success and walk away, thinking they've achieved something that really they have not.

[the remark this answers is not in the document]
yuqi-zhang: Ack, will try to construct a validator, thanks!

operator/v1/types_machineconfiguration.go
```diff
@@ -153,3 +197,183 @@ const (
...
+// NodeDisruptionPolicyFile is a file entry and corresponding actions to take
+type NodeDisruptionPolicyFile struct {
+	// path is the file path to a file on disk managed through a MachineConfig.
+	// Actions specified will be applied when changes to the file at the path
+	// configured in this field.
+	// +kubebuilder:validation:Required
+	Path string `json:"path"`
+	// actions represents the series of commands to be executed on changes to the file at
+	// corresponding file path. This is an atomic list, which will be validated by
+	// the MachineConfigOperator, with any conflicts reflecting as an error in the
```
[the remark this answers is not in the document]
JoelSpeed: Discussed in slack, we can use CEL validation to enforce that reboot or none are singletons, something along the lines of the ternary below

```
+kubebuilder:validation:XValidation:rule="self.exists(x, x == 'Reboot') ? size(self) == 1 : true"
```

operator/v1/types_machineconfiguration.go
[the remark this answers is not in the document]
yuqi-zhang: I think we would still like to expose it in spec potentially, in case the user wants to take some action on their sshkey changes

yuqi-zhang:
Pushed some changes based on the discussion above, now only the user spec should exist in spec, and status will reflect cluster defaults.

Attempted to add some more CEL validation, but:
1. trying to verify the suffixes of service names (12 valid options) errors with exceeding the budget by 2.5x
2. trying to verify the list items 	`// +kubebuilder:validation:XValidation:rule="self.exists(x, x.type=='Reboot') ? size(self) == 1 : true", message="Reboot action can only be specified standalone, as it will override any other actions" ` exceeds budget by 7.9x

So neither seem to work at the moment. Any advice appreciated!

yuqi-zhang:
Rebased to use `openshift:enable:FeatureGate=NodeDisruptionPolicy` instead of FeatureSet. A few remaining questions:

1. this PR removes StaticOperatorPodStatus, without any additional handling, so will likely not pass tests. Spun that off as a separate effort via https://github.com/openshift/api/pull/1789
2. the newly added CEL (for example, parsing service names) works based on the tests, but is rather lengthy as it just parses for various `endsWith()`. Any ideas on improving it is appreciated
3. the spec and status currently uses the same object typing, which mostly works except for the `Special` keyword in the actions list. we would like to prevent users from being able to set `Special` via spec and make it status-only for the MCO to apply, but I'm not sure if there's an easy way to do this other than making every intermediary object slightly different (so, a lot of duplication on the objects). Is there another way we can prevent this just for spec? Or should we split out the objects anyways for better compatibility in the future?

operator/v1/types_machineconfiguration.go
```diff
@@ -158,3 +179,170 @@ const (
...
+	// +kubebuilder:validation:XValidation:rule="self.exists(x, x.type=='None') ? size(self) == 1 : true", message="None action can only be specified standalone, as it will override any other actions"
+	Actions []NodeDisruptionPolicyAction `json:"actions"`
+}
+
+// NodeDisruptionPolicyUnit is a systemd unit name and corresponding actions to take
+type NodeDisruptionPolicyUnit struct {
+	// name represents the service name of a systemd service managed through a MachineConfig
+	// Actions specified will be applied for changes to the named service
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self.endsWith('.service') || self.endsWith('.socket') || self.endsWith('.device') || self.endsWith('.mount') || self.endsWith('.automount') || self.endsWith('.swap') || self.endsWith('.target') || self.endsWith('.path') || self.endsWith('.timer') || self.endsWith('.snapshot') || self.endsWith('.slice') || self.endsWith('.scope')",message="Invalid service name schema, needs to be in the format of ${NAME}.${SERVICETYPE}"
```
deads2k, line 255: if name should be more than zero chars, need another CEL rule in addition to this one.

operator/v1/types_machineconfiguration.go
```diff
@@ -158,3 +179,170 @@ const (
...
+	// +optional
+	Reload *ReloadService `json:"reload,omitempty"`
+	// restart specifies the service to restart, only valid if type is restart
+	// +optional
+	Restart *RestartService `json:"restart,omitempty"`
+}
+
+// ReloadService allows the user to specify the services to be reloaded
+type ReloadService struct {
+	// serviceName is the full name (e.g. crio.service) of the service to be reloaded
```
JoelSpeed, line 308: Since we have this same sort of field multiple times, the godoc should probably be similar, as above, would like to see examples and enumerate the valid suffixes ideally

operator/v1/types_machineconfiguration.go
```diff
@@ -46,12 +46,33 @@ type MachineConfigurationSpec struct {
...
+	// +openshift:enable:FeatureGate=NodeDisruptionPolicy
+	// +optional
+	NodeDisruptionPolicy NodeDisruptionPolicyConfig `json:"nodeDisruptionPolicy"`
 }
 
 type MachineConfigurationStatus struct {
-	StaticPodOperatorStatus `json:",inline"`
+	// observedGeneration is the last generation change you've dealt with
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
```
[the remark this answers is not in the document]
djoshy: Not sure we have any plans for this yet, I think Jerry mentioned perhaps in an error condition, but I'll let him weigh in

operator/v1/types_machineconfiguration.go
```diff
@@ -46,12 +46,33 @@ type MachineConfigurationSpec struct {
...
+	// MachineConfig-based updates, such as drains, service reloads, etc. Specifying this will allow
+	// for less downtime when doing small configuration updates to the cluster. This configuration
+	// has no effect on cluster upgrades which will still incur node disruption where required.
+	// +openshift:enable:FeatureGate=NodeDisruptionPolicy
+	// +optional
+	NodeDisruptionPolicy NodeDisruptionPolicyConfig `json:"nodeDisruptionPolicy"`
 }
 
 type MachineConfigurationStatus struct {
-	StaticPodOperatorStatus `json:",inline"`
```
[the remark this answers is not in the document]
djoshy: Yeah, I believe Jerry opened a PR for this https://github.com/openshift/api/pull/1789; I do not think this was based off of that commit though

operator/v1/types_machineconfiguration.go
```diff
@@ -158,3 +179,170 @@ const (
...
+	// units is a list MachineConfig unit definitions and actions to take on changes to those services
+	// +optional
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=50
+	Units []NodeDisruptionPolicyUnit `json:"units"`
+	// sshkey maps to the ignition.sshkeys field in the MachineConfig object, definition an action for this
+	// will apply to all sshkey changes in the cluster
+	// +optional
+	SSHKey NodeDisruptionPolicySSHKey `json:"sshkey"`
```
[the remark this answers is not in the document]
djoshy: I think maybe you are thinking about the "Special" action type? We don't want the user to set that one. The only way out we could think of was having a different action type for Status and Spec, and making it such that Special is only settable on the Status side. Unfortunately, this would result in a lot of duplication, as this type is encapsulated through `NodeDisruptionPolicyAction` -> `NodeDisruptionPolicySSHKey`, `NodeDisruptionPolicyUnit`, `NodeDisruptionPolicyFile`  -> `NodeDisruptionPolicyClusterStatus` or `NodeDisruptionPolicyConfig`. I'm planning to do a follow-up commit for that, but wanted to run it by you in case there is another path?

djoshy:
Changes from this pass:

1. Switched out service name validation to regex
2. Added language for size limits 
3. Other missing go doc comments/clarifications

operator/v1/types_machineconfiguration.go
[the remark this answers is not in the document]
JoelSpeed: Generally we do advise having separate types for the spec and status to account for actions such as this, even though there would be duplication

The alternative is to use a CEL validation to prevent the special action, but that doesn't give you much freedom for future customisation nor does it make it easy to document the behaviour.

I'd recommend spec and status versions of the items

operator/v1/types_machineconfiguration.go
```diff
@@ -158,3 +179,205 @@ const (
...
+	// ".snapshot", ".slice" or ".scope".
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength=30
+	// +kubebuilder:validation:Pattern:=`^[a-zA-Z0-9_-]+\.(service|socket|device|mount|automount|swap|target|path|timer|snapshot|slice|scope)$`
+	Name string `json:"name"`
+
+	// Name regex breakdown:
+	// [a-zA-Z0-9_-]+ : Matches one or more characters consisting of uppercase letters, lowercase letters, digits, underscores, or hyphens.
+	// \. : Matches a period.
+	// (service|socket|device|mount|automount|swap|target|path|timer|snapshot|slice|scope) : Matches any of the specified service types.
```
JoelSpeed, line 269: Maybe do this before the field, rather than after. Also please check that none of the generated docs or openapi include this text

operator/v1/types_machineconfiguration.go
```diff
@@ -158,3 +179,205 @@ const (
...
+	// ".snapshot", ".slice" or ".scope".
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength=30
+	// +kubebuilder:validation:Pattern:=`^[a-zA-Z0-9_-]+\.(service|socket|device|mount|automount|swap|target|path|timer|snapshot|slice|scope)$`
+	ServiceName string `json:"serviceName"`
+
+	// Name regex breakdown:
+	// [a-zA-Z0-9_-]+ : Matches one or more characters consisting of uppercase letters, lowercase letters, digits, underscores, or hyphens.
+	// \. : Matches a period.
+	// (service|socket|device|mount|automount|swap|target|path|timer|snapshot|slice|scope) : Matches any of the specified service types.
```
JoelSpeed, line 354: Do we need to duplicate this?

djoshy:
Updated with the following:
- added two CEL based regex validations for the service name, with different errors
- added new character limit of 255 for service name
- added/updated some tests to match the new validations
- added Status type objects per discussion from [this thread](https://github.com/openshift/api/pull/1764#discussion_r1525529613)

operator/v1/types_machineconfiguration.go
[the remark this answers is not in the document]
yuqi-zhang: I was hoping to work on that separately as we had discussed. Would it be acceptable to have this PR merge without tombstones?

yuqi-zhang:
@JoelSpeed I believe @djoshy has covered the latest round of reviews (thanks!). There's just the open question about tombstone-ing the static operator status. Could this PR merge without that or would that have to go in first?

FWIW I did another brief check and nothing is currently depending on this (the CRD itself was added to the MCO in 4.15 but no objects exist and no other repo is referring to it outside of the API and MCO)

operator/v1/zz_generated.crd-manifests/0000_80_machine-config_01_machineconfigurations-Default.crd.yaml
```diff
@@ -120,137 +120,80 @@ spec:
...
+                        (see .node.status.conditions), the ability to deconflict is
+                        important. The regex it matches is (dns1123SubdomainFmt/)?(qualifiedNameFmt)
+                      maxLength: 316
+                      pattern: ^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])$
                       type: string
                   required:
+                  - lastTransitionTime
+                  - message
+                  - reason
+                  - status
```
JoelSpeed, line 185: Change to required fields, has this CRD been included in payload yet?

operator/v1/types_machineconfiguration.go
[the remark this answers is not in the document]
JoelSpeed: Should we revert this for now and handle separately? We can't merge this before the handling of removing the operator status in this hybrid state

operator/v1/types_machineconfiguration.go
```diff
@@ -158,3 +179,170 @@ const (
...
+	Units []NodeDisruptionPolicyUnit `json:"units,omitempty"`
+	// sshkey is the overall sshkey MachineConfig definition
+	// +optional
+	SSHKey NodeDisruptionPolicySSHKey `json:"sshkey,omitempty"`
+}
+
+// NodeDisruptionPolicyFile is a file entry and corresponding actions to take
+type NodeDisruptionPolicyFile struct {
+	// path is the file path to a file on disk managed through a MachineConfig.
+	// Actions specified will be applied when changes to the file at the path
```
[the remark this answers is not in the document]
JoelSpeed: Not resolved

operator/v1/types_machineconfiguration.go
```diff
@@ -158,3 +179,304 @@ const (
...
+	// ${NAME} must be atleast 1 character long and can only consist of alphabets, digits, ":", "-", "_", ".", and "\".
+	// ${SERVICETYPE} must be one of ".service", ".socket", ".device", ".mount", ".automount", ".swap", ".target", ".path", ".timer", ".snapshot", ".slice" or ".scope".
+	// +kubebuilder:validation:XValidation:rule=`self.matches('\\.(service|socket|device|mount|automount|swap|target|path|timer|snapshot|slice|scope)$')`, message="Invalid ${SERVICETYPE} in service name. Expected format is ${NAME}${SERVICETYPE}, where ${SERVICETYPE} must be one of \".service\", \".socket\", \".device\", \".mount\", \".automount\", \".swap\", \".target\", \".path\", \".timer\",\".snapshot\", \".slice\" or \".scope\"."
+	// +kubebuilder:validation:XValidation:rule=`self.matches('^[a-zA-Z0-9:._\\\\-]+\\..')`, message="Invalid ${NAME} in service name. Expected format is ${NAME}${SERVICETYPE}, where {NAME} must be atleast 1 character long and can only consist of alphabets, digits, \":\", \"-\", \"_\", \".\", and \"\\\""
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength=255
+	ServiceName string `json:"serviceName"`
```
JoelSpeed, line 426: Since this is re-used several times, you could create a type alias for this and assign the validations to that instead, not blocking though. Similar to how you have done the action types

operator/v1/techpreview.machineconfiguration.testsuite.yaml
```diff
@@ -115,3 +115,212 @@ tests:
...
+      apiVersion: operator.openshift.io/v1
+      kind: MachineConfiguration
+      spec:
+        nodeDisruptionPolicy:
+          files:
+            - path: "/etc/my-file"
+              actions:
+                - type: DaemonReload
+                - type: Reload
+    expectedError: "Reload is required when type is reload, and forbidden otherwise"
```
JoelSpeed, line 233: This is the wrong way around, think about it from an end user perspective, they see the field as lower r and the type as upper r
```suggestion
    expectedError: "reload is required when type is Reload, and forbidden otherwise"
```

operator/v1/techpreview.machineconfiguration.testsuite.yaml
```diff
@@ -115,3 +115,212 @@ tests:
...
+                - type: DaemonReload
+                - type: Reload
+                  reload:
+                    serviceName: a:b_c-d\.e.swap
+            - path: "/etc/my-other-file"
+              actions:
+                - type: Drain
+                - type: Restart
+                  restart:
+                    serviceName: a.b.c.d.e.snapshot
```
JoelSpeed, line 326: Nit, should include new line char as final character in every file

operator/v1/zz_generated.crd-manifests/0000_80_machine-config_01_machineconfigurations-Default.crd.yaml
yuqi-zhang: The only place this is currently referenced is by the MCO code, where I think we technically bring in the CRD. (So I think it does exist in clusters 4.15? and up, but no objects)

What is making this generated field required now and is that an issue?

operator/v1/types_machineconfiguration.go
yuqi-zhang: Ack, moved to a new type

operator/v1/types_machineconfiguration.go
```diff
@@ -158,3 +171,301 @@ const (
...
+	// ${NAME} must be atleast 1 character long and can only consist of alphabets, digits, ":", "-", "_", ".", and "\".
+	// ${SERVICETYPE} must be one of ".service", ".socket", ".device", ".mount", ".automount", ".swap", ".target", ".path", ".timer", ".snapshot", ".slice" or ".scope".
+	// +kubebuilder:validation:XValidation:rule=`self.matches('\\.(service|socket|device|mount|automount|swap|target|path|timer|snapshot|slice|scope)$')`, message="Invalid ${SERVICETYPE} in service name. Expected format is ${NAME}${SERVICETYPE}, where ${SERVICETYPE} must be one of \".service\", \".socket\", \".device\", \".mount\", \".automount\", \".swap\", \".target\", \".path\", \".timer\",\".snapshot\", \".slice\" or \".scope\"."
+	// +kubebuilder:validation:XValidation:rule=`self.matches('^[a-zA-Z0-9:._\\\\-]+\\..')`, message="Invalid ${NAME} in service name. Expected format is ${NAME}${SERVICETYPE}, where {NAME} must be atleast 1 character long and can only consist of alphabets, digits, \":\", \"-\", \"_\", \".\", and \"\\\""
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength=255
+	Name string `json:"name"`
```
JoelSpeed, line 304: NodeDisruptionPolicyServiceName?

operator/v1/types_machineconfiguration.go
```diff
@@ -158,3 +171,301 @@ const (
...
```
JoelSpeed, line 276: NodeDisruptionPolicyServiceName?

operator/v1/types_machineconfiguration.go
```diff
@@ -158,3 +171,301 @@ const (
...
+	// +kubebuilder:validation:Required
+	// +listType=atomic
+	// +kubebuilder:validation:MaxItems=10
+	// +kubebuilder:validation:XValidation:rule="self.exists(x, x.type=='Reboot') ? size(self) == 1 : true", message="Reboot action can only be specified standalone, as it will override any other actions"
+	// +kubebuilder:validation:XValidation:rule="self.exists(x, x.type=='None') ? size(self) == 1 : true", message="None action can only be specified standalone, as it will override any other actions"
+	Actions []NodeDisruptionPolicyStatusAction `json:"actions"`
+}
+
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'Reload' ? has(self.reload) : !has(self.reload)",message="reload is required when type is Reload, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'Restart' ? has(self.restart) : !has(self.restart)",message="Restart is required when type is restart, and forbidden otherwise"
```
JoelSpeed, line 356:
```suggestion
// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'Restart' ? has(self.restart) : !has(self.restart)",message="restart is required when type is Restart, and forbidden otherwise"
```

JoelSpeed (commented):
/lgtm

Will follow up separately on the static pod status removal, build out on this first and transition to metav1.Condition before we ship.

yuqi-zhang:
Whoops, was failing verify since it wasn't rebased on master

JoelSpeed (commented):
/lgtm

PR #1768, opened 2024-02-12
image registry auth token type

sanchezl:
- add imageRegistryAuthTokenType field
- make update
- add openshift.io/image-registry-pull-secrets controller name


sanchezl:
Going another way with this.

PR #1771, opened 2024-02-14
OCPBUGS-29435: allow multiple audiences to be configured for kube-apiserver

soltysh:


config/v1/types_authentication.go
```diff
@@ -240,7 +240,8 @@ type TokenIssuer struct {
 	//
 	// +listType=set
 	// +kubebuilder:validation:Required
-	// +kubebuilder:validation:MaxItems=1
+	// +kubebuilder:validation:MinItems=1
```
deads2k, line 243: interesting catch.

PR #1783, opened 2024-03-06
STOR-1803: add vsphere snapshot configuration fields to ClusterCSIDriver

RomanBednar:
cc @openshift/storage

Enhancement: https://github.com/openshift/enhancements/pull/1563

### DIFF
--- a/config/v1/feature_gates.go
+++ b/config/v1/feature_gates.go
@@ -595,4 +595,11 @@ var (
 						productScope(ocpSpecific).
 						enableIn(TechPreviewNoUpgrade).
 						mustRegister()
+
+	FeatureGateVSphereDriverConfiguration = newFeatureGate("VSphereDriverConfiguration").
+						reportProblemsToJiraComponent("Storage / Kubernetes External Components").
+						contactPerson("rbednar").
+						productScope(ocpSpecific).
+						enableIn(TechPreviewNoUpgrade).
+						mustRegister()
 )
--- a/openapi/openapi.json
+++ b/openapi/openapi.json
@@ -32146,6 +32146,21 @@
       "description": "VSphereCSIDriverConfigSpec defines properties that can be configured for vsphere CSI driver.",
       "type": "object",
       "properties": {
+        "globalMaxSnapshotsPerBlockVolume": {
+          "description": "globalMaxSnapshotsPerBlockVolume is a global configuration parameter that applies to volumes on all kinds of datastores. If omitted, the platform chooses a default, which is subject to change over time, currently that default is 3. Snapshots can not be disabled using this parameter. Increasing number of snapshots above 3 can have negative impact on performance, for more details see: https://kb.vmware.com/s/article/1025279 Volume snapshot documentation: https://docs.vmware.com/en/VMware-vSphere-Container-Storage-Plug-in/3.0/vmware-vsphere-csp-getting-started/GUID-E0B41C69-7EEB-450F-A73D-5FD2FF39E891.html",
+          "type": "integer",
+          "format": "int64"
+        },
+        "granularMaxSnapshotsPerBlockVolumeInVSAN": {
+          "description": "granularMaxSnapshotsPerBlockVolumeInVSAN is a granular configuration parameter on vSAN datastore only. It overrides GlobalMaxSnapshotsPerBlockVolume if set, while it falls back to the global constraint if unset. Snapshots for VSAN can not be disabled using this parameter.",
+          "type": "integer",
+          "format": "int64"
+        },
+        "granularMaxSnapshotsPerBlockVolumeInVVOL": {
+          "description": "granularMaxSnapshotsPerBlockVolumeInVVOL is a granular configuration parameter on Virtual Volumes datastore only. It overrides GlobalMaxSnapshotsPerBlockVolume if set, while it falls back to the global constraint if unset. Snapshots for VVOL can not be disabled using this parameter.",
+          "type": "integer",
+          "format": "int64"
+        },
         "topologyCategories": {
           "description": "topologyCategories indicates tag categories with which vcenter resources such as hostcluster or datacenter were tagged with. If cluster Infrastructure object has a topology, values specified in Infrastructure object will be used and modifications to topologyCategories will be rejected.",
           "type": "array",
--- a/operator/v1/types_csi_cluster_driver.go
+++ b/operator/v1/types_csi_cluster_driver.go
@@ -281,6 +281,35 @@ type VSphereCSIDriverConfigSpec struct {
 	// will be rejected.
 	// +optional
 	TopologyCategories []string `json:"topologyCategories,omitempty"`
+
+	// globalMaxSnapshotsPerBlockVolume is a global configuration parameter that applies to volumes on all kinds of
+	// datastores. If omitted, the platform chooses a default, which is subject to change over time, currently that default is 3.
+	// Snapshots can not be disabled using this parameter.
+	// Increasing number of snapshots above 3 can have negative impact on performance, for more details see: https://kb.vmware.com/s/article/1025279
+	// Volume snapshot documentation: https://docs.vmware.com/en/VMware-vSphere-Container-Storage-Plug-in/3.0/vmware-vsphere-csp-getting-started/GUID-E0B41C69-7EEB-450F-A73D-5FD2FF39E891.html
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=32
+	// +openshift:enable:FeatureGate=VSphereDriverConfiguration
+	// +optional
+	GlobalMaxSnapshotsPerBlockVolume *uint32 `json:"globalMaxSnapshotsPerBlockVolume,omitempty"`
+
+	// granularMaxSnapshotsPerBlockVolumeInVSAN is a granular configuration parameter on vSAN datastore only. It
+	// overrides GlobalMaxSnapshotsPerBlockVolume if set, while it falls back to the global constraint if unset.
+	// Snapshots for VSAN can not be disabled using this parameter.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=32
+	// +openshift:enable:FeatureGate=VSphereDriverConfiguration
+	// +optional
+	GranularMaxSnapshotsPerBlockVolumeInVSAN *uint32 `json:"granularMaxSnapshotsPerBlockVolumeInVSAN,omitempty"`
+
+	// granularMaxSnapshotsPerBlockVolumeInVVOL is a granular configuration parameter on Virtual Volumes datastore only.
+	// It overrides GlobalMaxSnapshotsPerBlockVolume if set, while it falls back to the global constraint if unset.
+	// Snapshots for VVOL can not be disabled using this parameter.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=32
+	// +openshift:enable:FeatureGate=VSphereDriverConfiguration
+	// +optional
+	GranularMaxSnapshotsPerBlockVolumeInVVOL *uint32 `json:"granularMaxSnapshotsPerBlockVolumeInVVOL,omitempty"`
 }
 
 // ClusterCSIDriverStatus is the observed status of CSI driver operator
--- a/features.md
+++ b/features.md
@@ -38,6 +38,7 @@
 | SigstoreImageVerification| | | <span style="background-color: #519450">Enabled</span> | <span style="background-color: #519450">Enabled</span>  |
 | TranslateStreamCloseWebsocketRequests| | | <span style="background-color: #519450">Enabled</span> | <span style="background-color: #519450">Enabled</span>  |
 | UpgradeStatus| | | <span style="background-color: #519450">Enabled</span> | <span style="background-color: #519450">Enabled</span>  |
+| VSphereDriverConfiguration| | | <span style="background-color: #519450">Enabled</span> | <span style="background-color: #519450">Enabled</span>  |
 | ValidatingAdmissionPolicy| | | <span style="background-color: #519450">Enabled</span> | <span style="background-color: #519450">Enabled</span>  |
 | VolumeGroupSnapshot| | | <span style="background-color: #519450">Enabled</span> | <span style="background-color: #519450">Enabled</span>  |
 | ExternalOIDC| <span style="background-color: #519450">Enabled</span> | | <span style="background-color: #519450">Enabled</span> | <span style="background-color: #519450">Enabled</span>  |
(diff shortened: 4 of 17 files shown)


RomanBednar:
verify-crd-schema error is probably not related to this change as I have not changed `topologyCategories` field: `spec.driverConfig.vSphere.topologyCategories must set x-kubernetes-list-type` - should it be overriden?

operator/v1/types_csi_cluster_driver.go
```diff
@@ -281,6 +281,27 @@ type VSphereCSIDriverConfigSpec struct {
 	// will be rejected.
 	// +optional
 	TopologyCategories []string `json:"topologyCategories,omitempty"`
+
+	// globalMaxSnapshotsPerBlockVolume is a global configuration parameter that applies to volumes on all kinds of
+	// datastores. By default, it is set to three.
+	// +kubebuilder:validation:Optional
+	// +openshift:enable:FeatureSets=TechPreviewNoUpgrade
+	// +optional
+	GlobalMaxSnapshotsPerBlockVolume *uint32 `json:"globalMaxSnapshotsPerBlockVolume,omitempty"`
```
jsafrane, line 290: For all new fields:

* There should be some validation, for example the value should not be negative. Or semantics of negative values should be described.
* Are there any max values?
* Value `0` should have some description - is it a valid value? Does it turn off snapshot support for VSAN / VVOL / globally? 
* What are the impacts of setting the values high? I would expect higher values mean less performance, otherwise we can set the values to MAXINT by default and we don't need any configuration.
* It could help if there was link to vSphere docs, the answers could be answered there.

operator/v1/types_csi_cluster_driver.go
```diff
@@ -281,6 +281,27 @@ type VSphereCSIDriverConfigSpec struct {
...
+
+	// granularMaxSnapshotsPerBlockVolumeInVSAN is a granular configuration parameter on vSAN datastore only. It
+	// overrides the global constraint if set, while it falls back to the global constraint if unset.
+	// +kubebuilder:validation:Optional
```
jsafrane, line 294: IMO it would be better to use `overrides globalMaxSnapshotsPerBlockVolume` instead of `the global constraint`.

operator/v1/types_csi_cluster_driver.go
RomanBednar: > There should be some validation, for example the value should not be negative. Or semantics of negative values should be described.

- Correct - currently the operator would fail to unmarshal as the field is uint32 type
- Adding validation for min/max volumes.

```
W0312 17:09:26.526662    8935 reflector.go:539] k8s.io/client-go@v0.29.1/tools/cache/reflector.go:229: failed to list *v1.ClusterCSIDriver: json: cannot unmarshal number -1 into Go struct field VSphereCSIDriverConfigSpec.items.spec.driverConfig.vSphere.globalMaxSnapshotsPerBlockVolume of type uint32
E0312 17:09:26.526720    8935 reflector.go:147] k8s.io/client-go@v0.29.1/tools/cache/reflector.go:229: Failed to watch *v1.ClusterCSIDriver: failed to list *v1.ClusterCSIDriver: json: cannot unmarshal number -1 into Go struct field VSphereCSIDriverConfigSpec.items.spec.driverConfig.vSphere.globalMaxSnapshotsPerBlockVolume of type uint32
```

> Are there any max values?

- From VMware documentation: `Maximum of 32 snapshots are **supported** in a chain. However, for a better performance use only 2 to 3 snapshots.`
- Adding validation for max value of 32.
- Setting any higher value in the config is not prevented currently:
```
[Snapshot]
global-max-snapshots-per-block-volume = 33
```

> Value `0` should have some description - is it a valid value? Does it turn off snapshot support for VSAN / VVOL / globally?

- Not documented at all, after testing it seems like `0` value results in disabling the option and default is used.
- Snapshots are not disabled when `0` is set
- Setting min value to 1 should work fine as there is no reason to set `0`

```
$ oc -n openshift-cluster-csi-drivers get cm/vsphere-csi-config -o yaml | grep global-max-snapshots-per-block-volume
    global-max-snapshots-per-block-volume = 0
```

```
message: 'Failed to create snapshot: failed to take snapshot of the volume c0e2dc41-33de-4566-9323-9aa422741493: "rpc error: code = FailedPrecondition desc = the number of snapshots on the source volume c0e2dc41-33de-4566-9323-9aa422741493 reaches the configured maximum (3)"'
```

> What are the impacts of setting the values high? I would expect higher values mean less performance, otherwise we can set the values to MAXINT by default and we don't need any configuration.

- This is documented under "Best practices for snapshots": https://kb.vmware.com/s/article/1025279
- 2-3 snapshots are recommended for better performance so we can expect a performance drop when using more than 3 snapshots

 > It could help if there was link to vSphere docs, the questions could be answered there.

- Do we actually want to link to external documentation from comments here?
 - Adding the link to relevant doc: https://docs.vmware.com/en/VMware-vSphere-Container-Storage-Plug-in/3.0/vmware-vsphere-csp-getting-started/GUID-E0B41C69-7EEB-450F-A73D-5FD2FF39E891.html

jsafrane: Thanks for the update. I think linking vSphere docs will help users to actually understand what the options are good for and they don't need to guess.

jsafrane:
The API looks good to me, you probably need to fix the tests somehow.

RomanBednar:
Yeah, so there are some changes in progress to change the process so new API fields are not enabled by feature set but feature gate, documentation should arrive soon. From current observations it seems there should be 3 separate files now for clustercsidriver CRD, that means we have to [change](https://github.com/openshift/cluster-storage-operator/pull/464/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R10-R12) CSO image a bit.

Then verification requires a test file for each of those so I've added that, with a small test case for TechPreviewNoUpgrade with new fields - `ci/prow/verify` seems happy now.

The `verify-crd-schema` job seems to fail due to `topologyCategories` which is not in scope of this change and probably should be overriden:

``` error running generator schemacheck on group operator.openshift.io: 
	could not run schemacheck generator for group/version operator.openshift.io/v1: 
		error in 0000_90_cluster_csi_driver_01_config-CustomNoUpgrade.crd.yaml: ListsMustHaveSSATags: crd/clustercsidrivers.operator.openshift.io version/v1 field/^.spec.driverConfig.vSphere.topologyCategories must set x-kubernetes-list-type
		error in 0000_90_cluster_csi_driver_01_config-Default.crd.yaml: ListsMustHaveSSATags: crd/clustercsidrivers.operator.openshift.io version/v1 field/^.spec.driverConfig.vSphere.topologyCategories must set x-kubernetes-list-type
		error in 0000_90_cluster_csi_driver_01_config-TechPreviewNoUpgrade.crd.yaml: ListsMustHaveSSATags: crd/clustercsidrivers.operator.openshift.io version/v1 field/^.spec.driverConfig.vSphere.topologyCategories must set x-kubernetes-list-type
+ echo This verifier checks all files that have changed. In some cases you may have changed or renamed a file that already contained api violations, but you are not introducing a new violation. In such cases it is appropriate to /override the failing CI job. 
```

RomanBednar:
> > The verify-crd-schema job seems to fail due to topologyCategories which is not in scope of this change and probably should be overriden:
> 
> No. The test works in other PRs without any issues.

I think that's because codegen checks only changed files (diff to master), since I renamed the files the error shows up only here. I've fixed that by [adding](https://github.com/openshift/api/pull/1783/commits/142414dc280863ff9d3000e68ac6a685b1ca1aa8#diff-1890d43482146b467ff1905bfe70773c54f1bb865f348b2b93fd8f914a91bf09R282) `listType` marker to `topologyCategories` field.

RomanBednar:
@deads2k Given the simplicity and potential cadence of similar PRs we want to release this without TechPreview. With the new process which is based on feature gates, can this be achieved by simply including the feature gate in `Default` cluster profile? https://github.com/openshift/api/pull/1783/files#diff-503cbb11f85749eb9ced1d2fcebdd1474b8df9ef32be2d845673838de8b7ff0cR596

operator/v1/types_csi_cluster_driver.go
```diff
@@ -279,8 +279,35 @@ type VSphereCSIDriverConfigSpec struct {
 	// If cluster Infrastructure object has a topology, values specified in
 	// Infrastructure object will be used and modifications to topologyCategories
 	// will be rejected.
+	// +listType=set
 	// +optional
 	TopologyCategories []string `json:"topologyCategories,omitempty"`
+
+	// globalMaxSnapshotsPerBlockVolume is a global configuration parameter that applies to volumes on all kinds of
+	// datastores. If unset it defaults to 3.
```
deads2k, line 287: Can we do something like

```
If omitted, the platform chooses a default, which is subject to change over time, currently that default is 3.
```

This matches other fields.  Rather than promising a value.

operator/v1/types_csi_cluster_driver.go
```diff
@@ -279,8 +279,35 @@ type VSphereCSIDriverConfigSpec struct {
...
+	// Increasing number of snapshots above 3 can have negative impact on performance, for more details see: https://kb.vmware.com/s/article/1025279
+	// Volume snapshot documentation: https://docs.vmware.com/en/VMware-vSphere-Container-Storage-Plug-in/3.0/vmware-vsphere-csp-getting-started/GUID-E0B41C69-7EEB-450F-A73D-5FD2FF39E891.html
+	// +kubebuilder:validation:Minimum=1
```
deads2k, line 290: how do I choose: "no snapshots"?

config/v1/feature_gates.go
```diff
@@ -595,4 +595,11 @@ var (
 						productScope(ocpSpecific).
 						enableIn(TechPreviewNoUpgrade).
 						mustRegister()
+
+	FeatureGateVSphereDriverConfiguration = newFeatureGate("VSphereDriverConfiguration").
+						reportProblemsToJiraComponent("Storage / Kubernetes External Components").
+						contactPerson("rbednar").
+						productScope(ocpSpecific).
+						enableIn(Default, TechPreviewNoUpgrade).
```
deads2k, line 603: TechPreview first please. Demonstrate stability (either via automation or QE sign off) and we can promote in 4.16.

operator/v1/types_csi_cluster_driver.go
```diff
@@ -279,8 +279,35 @@ type VSphereCSIDriverConfigSpec struct {
 	// If cluster Infrastructure object has a topology, values specified in
 	// Infrastructure object will be used and modifications to topologyCategories
 	// will be rejected.
+	// +listType=set
```
deads2k, line 282: this should have failed our CRD schema checker. Why would this be a safe change to make for clients?

operator/v1/types_csi_cluster_driver.go
RomanBednar: Sounds good, changed.

operator/v1/types_csi_cluster_driver.go
RomanBednar: Already discussed here: https://github.com/openshift/api/pull/1783#discussion_r1523271016

vmWare does not document this value but testing showed that it does not disable snapshots, but default is applied instead. I can add a clarifying statement:

`Setting this value to 0 does not disable volume snapshots, but results in default value being used.`

operator/v1/types_csi_cluster_driver.go
RomanBednar: Previously the checker was failing without listType, I've retried it again now and with the recent changes verification is passing without it - dropping this change.

RomanBednar:
@deads2k PTAL again

openapi/openapi.json
```diff
@@ -32146,6 +32146,21 @@
       "description": "VSphereCSIDriverConfigSpec defines properties that can be configured for vsphere CSI driver.",
       "type": "object",
       "properties": {
+        "globalMaxSnapshotsPerBlockVolume": {
+          "description": "globalMaxSnapshotsPerBlockVolume is a global configuration parameter that applies to volumes on all kinds of datastores. If omitted, the platform chooses a default, which is subject to change over time, currently that default is 3. Setting this value to 0 does not disable volume snapshots, but results in default value being used. Increasing number of snapshots above 3 can have negative impact on performance, for more details see: https://kb.vmware.com/s/article/1025279 Volume snapshot documentation: https://docs.vmware.com/en/VMware-vSphere-Container-Storage-Plug-in/3.0/vmware-vsphere-csp-getting-started/GUID-E0B41C69-7EEB-450F-A73D-5FD2FF39E891.html",
```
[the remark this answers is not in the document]
RomanBednar: @Phaow The comment about `0` value is meant to be rather informative but not needed, and just explains why `0` does not make sense. I can write it better or remove it. The current validation (`minimum: 1`)  is rejecting `0` which is correct, or am I missing something?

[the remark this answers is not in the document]
Phaow: @RomanBednar Yeah, current validation (minimum: 1) is rejecting 0 looks good to me, just a bit strange that we already reject `0`, just the explain if set to `0` which will still use the default maybe make a bit confused.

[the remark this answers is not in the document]
RomanBednar: @Phaow ok, removing

config/v1/feature_gates.go
deads2k: Before becoming accessible-by-default, we must have evidence of completeness and reliability.  The preferred mechanism is to run automated CI tests in our existing TechPreview jobs and see 95+% pass rates over at least 14 runs.  The alternative mechanism is to have QE sign off on a promotion PR.  That can all happen in a single release.  For instance,

1. in 4.16
2. merge as techpreview
3. add functionality and tests
4. after there are 14 runs, open PR promoting to GA
5. CI automatically checks for the associated tests (see example failure [here](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_api/1851/pull-ci-openshift-api-master-verify/1778771838210412544))
6. PR is merged on green OR QE signs off on the PR and we override the automated stability check (we track data on this now in 4.16)
7. we ship 4.16


So this doesn't delay availability in 4.16 assuming the feature is tested and is as stable as expected

operator/v1/types_csi_cluster_driver.go
deads2k: > Already discussed here: [#1783 (comment)](https://github.com/openshift/api/pull/1783#discussion_r1523271016)

Please add a sentence in the documentation indicating that snapshots cannot be disabled.  On this item and the others if they cannot be disabled either.

RomanBednar: Added, for all 3 options as it's applicable to all of them.

config/v1/feature_gates.go
RomanBednar: @deads2k Removed Default cluster profile for this feature, can we merge this now as TechPreview?

PR #1787, opened 2024-03-08
OPNET-477: PlatformLoadBalancer becomes GA for Nutanix

mkowalski:
This PR promotes the PlatformLoadBalancer to GA for Nutanix.

The feature has been in TP for long enough so that we are ready to promote it.


mkowalski:
/close

Covered by https://github.com/openshift/api/pull/1757

PR #1819, opened 2024-03-19
OCPBUGS-31058: config/ingress: Make Hostname godoc user-friendly

Miciah:
Add the necessary kubebuilder syntax to the godoc for the ingress config API's Hostname type in order to hide developer notes from the generated CRD's description for the `status.componentRoutes.currentHostnames` field.  Also, add a user-oriented description of the field with a reference to the relevant RFC.

Before this change, the `oc explain` output for the API field described it in terms of its Go type, kubebuilder syntax, and regular-expression syntax for validation.  After this change, the `oc explain` output for the API field describes it in terms that are relevant to the end-user.

The commit that added the validation did not mention *which* RFC it was enforcing.  The closest RFC seems to be [RFC-1123](https://www.rfc-editor.org/rfc/rfc1123) in that the validation allows the host name to start with a decimal digit, although the validation seems a little looser than RFC-1123 in that allows arbitrary Unicode letters and symbols with `\p{L}` and `\p{S}`.  The intent seems to be that the end-user should use RFC-1123 compliant host names, and validation is looser only as necessary for backwards-compatibility.

Follow-up to #1120.

### DIFF
--- a/config/v1/types_ingress.go
+++ b/config/v1/types_ingress.go
@@ -169,20 +169,20 @@ const (
 // +kubebuilder:validation:MaxLength=512
 type ConsumingUser string
 
-// Hostname is an alias for hostname string validation.
-//
-// The left operand of the | is the original kubebuilder hostname validation format, which is incorrect because it
-// allows upper case letters, disallows hyphen or number in the TLD, and allows labels to start/end in non-alphanumeric
-// characters.  See https://bugzilla.redhat.com/show_bug.cgi?id=2039256.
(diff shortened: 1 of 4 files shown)


Miciah:
Regenerated `config/v1/zz_generated.crd-manifests/0000_10_config-operator_01_ingresses.crd.yaml` in https://github.com/openshift/api/compare/91589b24811a47a9149ff6ab9fddbb362e031999..1923c36769f6039a92af195635195dcc409221b6.

PR #1824, opened 2024-03-22
MCO-838: machineconfiguration/v1alpha1: add PinnedImageSet to MCN

hexfusion:
This PR adds type support for reporting `Node` level status of  `PinnedImageSet` through `MachineConfigNode` Spec and Status.


machineconfiguration/v1alpha1/types_machineconfignode.go
```diff
@@ -187,4 +187,8 @@ const (
 	MachineConfigNodeUpdateRebooted StateProgress = "RebootedNode"
 	// MachineConfigNodeUpdateReloaded describes the part of the post action phase where the node reloads its CRIO service
 	MachineConfigNodeUpdateReloaded StateProgress = "ReloadedCRIO"
+	// MachineConfigNodePrefetchInProgress describes a machine currently in the process of pulling the images of a pinned image set
```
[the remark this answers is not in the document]
hexfusion: Yeah I think let's start here my thought is I am in one of 3 states. If I can prove I need more we can add for 4.17? But I am open to ideas.

- InProgress (working)
- Complete (done)
- Degraded (error)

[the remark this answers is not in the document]
hexfusion: Also I don't love the naming was between this and `ImagePrefetch  or `PinnedImageSet` not sure if you have any feeling on that.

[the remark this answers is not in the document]
hexfusion: I guess even though it would be log and annoying PinnedImageSet is clear

cdoern (approved):
Looks good from an MCO side of things. Adding these into MCN status as an optional field works great.

machineconfiguration/v1alpha1/types_machineconfignode.go
```diff
@@ -107,6 +112,36 @@ type MachineConfigNodeStatus struct {
...
+	// name is the name of the pinned image set.
+	// Must be a lowercase RFC-1123 hostname (https://tools.ietf.org/html/rfc1123)
+	// It may consist of only alphanumeric characters, hyphens (-) and periods (.)
+	// and must be at most 253 characters in length.
+	// +kubebuilder:validation:MaxLength:=253
+	// +kubebuilder:validation:Pattern=`^([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])(\.([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9]))*$`
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+	// currentGeneration is the generation of the pinned image set that has most recently been successfully pulled and pinned on this node.
+	// +kubebuilder:validation:Required
```
deads2k, line 134: when first create, you may not have a currentgeneration.

machineconfiguration/v1alpha1/types_machineconfignode.go
```diff
@@ -107,6 +112,36 @@ type MachineConfigNodeStatus struct {
...
+
+type MachineConfigNodeStatusPinnedImageSetVersion struct {
+	// name is the name of the pinned image set.
+	// Must be a lowercase RFC-1123 hostname (https://tools.ietf.org/html/rfc1123)
+	// It may consist of only alphanumeric characters, hyphens (-) and periods (.)
+	// and must be at most 253 characters in length.
+	// +kubebuilder:validation:MaxLength:=253
+	// +kubebuilder:validation:Pattern=`^([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])(\.([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9]))*$`
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
```
deads2k, line 132: how does a pinned image set get chosen on a node?

deads2k: nm

machineconfiguration/v1alpha1/types_machineconfignode.go
```diff
@@ -155,6 +190,20 @@ type MachineConfigNodeSpecMachineConfigVersion struct {
...
+	// desiredGeneration is the generation of the pinned image set that should be pinned and pulled on this node.
+	// +kubebuilder:validation:Required
+	DesiredGeneration int32 `json:"desiredGeneration"`
```
deads2k, line 204: This construct is a little strange to me. I understand trying to set which pinnedimageset to use (maybe), but I wouldn't expect to see the generation codified since anything other than "current" is impossible to read from clusterstate.  On status this makes sense so you can see "I have a diff", but I don't understand how a desiredgeneration less than the current generation makes sense on spec.

machineconfiguration/v1alpha1/types_machineconfignode.go
```diff
@@ -107,6 +112,36 @@ type MachineConfigNodeStatus struct {
 	// This desired machine config has been compared to the current machine config and has been validated by the machine config operator as one that is valid and that exists.
 	// +kubebuilder:validation:Required
 	ConfigVersion MachineConfigNodeStatusMachineConfigVersion `json:"configVersion"`
+	// pinnedImageSetVersion describes the current and desired pinned image sets for this node.
+	// The current version is the generation of the pinned image set that has most recently been successfully pulled and pinned on this node.
+	// The desired version is the generation of the pinned image set that is targeted to be pulled and pinned on this node.
+	// +listType=map
+	// +listMapKey=name
+	// +optional
+	PinnedImageSetVersion []MachineConfigNodeStatusPinnedImageSetVersion `json:"pinnedImageSetVersion,omitempty"`
```
deads2k, line 121: When you make use of this API, I suggest limiting updates by doing something like

on each node
1. read the set of pinnedimagesets that should be on the node
2. try to make actual state match desired state for up to day two minutes (API consumers will see the mismatch by looking at current state)
3. if actual state matches desired state, update once with conditions and state
4. if actual state does not match desired state, update once with conditions and state.

This gets the "normal" case down to a single update, keeps the normal case mismatch observable via the API, and keeps the latency of "why isn't working tolerable without having multiple updates for every node in the cluster at about the same time.

hexfusion: great feedback :+1:

machineconfiguration/v1alpha1/types_machineconfignode.go
hexfusion: yeah I dropped it and it allows for MCP spec to reflect 1:1

machineconfiguration/v1alpha1/types_machineconfignode.go
```diff
@@ -87,6 +87,10 @@ type MachineConfigNodeSpec struct {
 	// the new machine config against the current machine config.
 	// +kubebuilder:validation:Required
 	ConfigVersion MachineConfigNodeSpecMachineConfigVersion `json:"configVersion"`
+
+	// pinnedImageSetVersion holds the desired pinned image sets that this node should pin and pull.
+	// +optional
+	PinnedImageSetVersion []MachineConfigNodeSpecPinnedImageSetVersion `json:"pinnedImageSetVersion,omitempty"`
```
deads2k, line 93: with the change, this is no longer a version, right?

deads2k, line 93: pinnedImageSets?

machineconfiguration/v1alpha1/types_machineconfignode.go
```diff
@@ -107,6 +111,36 @@ type MachineConfigNodeStatus struct {
...
+	CurrentGeneration int32 `json:"currentGeneration"`
+	// desiredGeneration version is the generation of the pinned image set that is targeted to be pulled and pinned on this node.
+	// +optional
+	DesiredGeneration int32 `json:"desiredGeneration,omitempty"`
+	// lastFailedGeneration is the generation of the most recent pinned image set that failed to be pulled and pinned on this node.
+	// +optional
+	LastFailedGeneration int32 `json:"lastFailedGeneration,omitempty"`
+	// lastFailedGenerationErrors is a list of errors why the lastFailed generation failed to be pulled and pinned.
+	// +optional
+	LastFailedGenerationErrors []string `json:"lastFailedGenerationErrors,omitempty"`
```
deads2k, line 143: limit to 10 errors

machineconfiguration/v1alpha1/types_machineconfignode.go
```diff
@@ -107,6 +111,36 @@ type MachineConfigNodeStatus struct {
...
+	// Must be a lowercase RFC-1123 hostname (https://tools.ietf.org/html/rfc1123)
+	// It may consist of only alphanumeric characters, hyphens (-) and periods (.)
+	// and must be at most 253 characters in length.
+	// +kubebuilder:validation:MaxLength:=253
+	// +kubebuilder:validation:Pattern=`^([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])(\.([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9]))*$`
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+	// currentGeneration is the generation of the pinned image set that has most recently been successfully pulled and pinned on this node.
+	// +optional
+	CurrentGeneration int32 `json:"currentGeneration"`
```
deads2k, line 134: 0 or positive

machineconfiguration/v1alpha1/types_machineconfignode.go
```diff
@@ -107,6 +111,36 @@ type MachineConfigNodeStatus struct {
...
+	// desiredGeneration version is the generation of the pinned image set that is targeted to be pulled and pinned on this node.
+	// +optional
+	DesiredGeneration int32 `json:"desiredGeneration,omitempty"`
```
deads2k, line 137: positive only.

deads2k: This should never be smaller than currentgeneration or lastFailedGeneration,  right?

deads2k, line 137: gets bigger only?

machineconfiguration/v1alpha1/types_machineconfignode.go
```diff
@@ -107,6 +111,36 @@ type MachineConfigNodeStatus struct {
...
+	// lastFailedGeneration is the generation of the most recent pinned image set that failed to be pulled and pinned on this node.
+	// +optional
+	LastFailedGeneration int32 `json:"lastFailedGeneration,omitempty"`
```
deads2k, line 140: should never be greater than desired, right?

hexfusion: adding cel

machineconfiguration/v1alpha1/types_machineconfignode.go
```diff
@@ -88,9 +88,12 @@ type MachineConfigNodeSpec struct {
...
 	ConfigVersion MachineConfigNodeSpecMachineConfigVersion `json:"configVersion"`
 
-	// pinnedImageSetVersion holds the desired pinned image sets that this node should pin and pull.
+	// pinnedImageSets holds the desired pinned image sets that this node should pin and pull.
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=100
 	// +optional
-	PinnedImageSetVersion []MachineConfigNodeSpecPinnedImageSetVersion `json:"pinnedImageSetVersion,omitempty"`
+	PinnedImageSets []MachineConfigNodeSpecPinnedImageSet `json:"pinnedImageSetVersion,omitempty"`
```
hexfusion, line 96: nit

machineconfiguration/v1alpha1/types_machineconfignode.go
```diff
@@ -111,16 +114,20 @@ type MachineConfigNodeStatus struct {
...
+	// +kubebuilder:validation:MaxItems=100
 	// +optional
-	PinnedImageSetVersion []MachineConfigNodeStatusPinnedImageSetVersion `json:"pinnedImageSetVersion,omitempty"`
+	PinnedImageSets []MachineConfigNodeStatusPinnedImageSet `json:"pinnedImageSetVersion,omitempty"`
 }
 
-type MachineConfigNodeStatusPinnedImageSetVersion struct {
+// +kubebuilder:validation:XValidation:rule="self.desiredGeneration > self.currentGeneration",message="desired generation must be greater than current generation"
+// +kubebuilder:validation:XValidation:rule="self.desiredGeneration > self.lastFailedGeneration",message="desired generation must be greater than last failed generation"
+// +kubebuilder:validation:XValidation:rule="self.currentGeneration > 0 || self.currentGeneration == null",message="current generation must be greater than 0"
```
hexfusion, line 129: nit these all need || null checks I think

machineconfiguration/v1alpha1/types_machineconfignode.go
```diff
@@ -107,6 +114,41 @@ type MachineConfigNodeStatus struct {
...
+	// The current version is the generation of the pinned image set that has most recently been successfully pulled and pinned on this node.
+	// The desired version is the generation of the pinned image set that is targeted to be pulled and pinned on this node.
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=100
+	// +optional
+	PinnedImageSets []MachineConfigNodeStatusPinnedImageSet `json:"pinnedImageSets,omitempty"`
+}
+
+// +kubebuilder:validation:XValidation:rule="has(self.desiredGeneration) ? self.desiredGeneration > self.currentGeneration : true",message="desired generation must be greater than current generation"
```
deads2k, line 127: can also be equal

machineconfiguration/v1alpha1/types_machineconfignode.go
```diff
@@ -107,6 +114,41 @@ type MachineConfigNodeStatus struct {
...
+// +kubebuilder:validation:XValidation:rule="has(self.lastFailedGeneration) ? self.desiredGeneration > self.lastFailedGeneration : true",message="desired generation must be greater than last failed generation"
```
deads2k, line 128: both desired and lastFailed are optional, right?  So do you have to check both?  also >= because when the desiredGeneration has failed, they'll be equal, right?

machineconfiguration/v1alpha1/types_machineconfignode.go
```diff
@@ -107,6 +114,41 @@ type MachineConfigNodeStatus struct {
...
+// +kubebuilder:validation:XValidation:rule="has(self.currentGeneration) ? self.currentGeneration > 0 : true",message="current generation must be greater than 0"
```
deads2k, line 129: this one (and other "must be positive") can be `kubebuilder:validation:Minimum`, right?

machineconfiguration/v1alpha1/types_machineconfignode.go
deads2k: > can also be equal when you're updating the currentGenreation, right?

Probably worth adding the tests to the integration suite so we can be sure.

hexfusion: added  status update tests to cover

machineconfiguration/v1alpha1/types_machineconfignode.go
hexfusion: > try to make actual state match desired state for up to day two minutes
just to be clear do work for at most 2 minutes then report status and retry if not complete.

PR #1828, opened 2024-03-26
Add models-schema to make clean

soltysh:
/assign @ingvagabund

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -153,7 +153,8 @@ write-available-featuresets:
 
 .PHONY: clean
 clean:
-	rm render write-available-featuresets
+	rm -f render write-available-featuresets models-schema
+	rm -rf tools/_output
 
 VERSION     ?= $(shell git describe --always --abbrev=7)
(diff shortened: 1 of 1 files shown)


ingvagabund:
How does one generate `models-schema` directory?

soltysh:
> How does one generate `models-schema` file?

it's part of `make update` target

ingvagabund:
Worth removing models-schema from https://github.com/openshift/api/blob/master/.gitignore#L19 as well

ingvagabund:
@JoelSpeed is there any significant benefit in keeping `models-schema` directory around?

soltysh:
> @JoelSpeed is there any significant benefit in keeping `models-schema` directory around?

that's not a directory, that's one of the build binaries used during generation



> Worth removing models-schema from https://github.com/openshift/api/blob/master/.gitignore#L19 as well?

Why? It's there to ensure we don't accidentally commit that binary.

dinhxuanvu (commented):
/lgtm

PR #1829, opened 2024-03-26
OCPBUGS-28673: Ingress - add connect timeout to tuning options

alebedev87:
This PR updates the Ingress Controller API with the connect timeout tuning option.


operator/v1/types_ingress.go
```diff
@@ -1649,6 +1649,15 @@ type IngressControllerTuningOptions struct {
 	// +optional
 	TunnelTimeout *metav1.Duration `json:"tunnelTimeout,omitempty"`
 
+	// ConnectTimeout defines the maximum time to wait for
+	// a connection attempt to a server/backend to succeed.
+	//
+	// If unset, the default timeout is 5s.
```
deads2k, line 1655: why isn't this the standard
>	// When omitted, this means the user has no opinion and the platform is left
>	// to choose reasonable defaults. These defaults are subject to change over time.
>	// The current default is 5s.

operator/v1/types_ingress.go
```diff
@@ -1649,6 +1649,15 @@ type IngressControllerTuningOptions struct {
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Format=duration
+	// +optional
+	ConnectTimeout *metav1.Duration `json:"connectTimeout,omitempty"`
```
deads2k, line 1659: missing format test for the new field?

operator/v1/types_ingress.go
alebedev87: Because the other timeouts in the tuning options use this phrasing, wanted to keep it consistent like if it was implemented this way from the very beginning (hence the bug).

operator/v1/types_ingress.go
alebedev87: I'm not sure which format test you are referring to.

operator/v1/types_ingress.go
deads2k: > Because the other timeouts in the tuning options use this phrasing, wanted to keep it consistent like if it was implemented this way from the very beginning (hence the bug).

Past mistakes rarely justify continuing those mistakes.  Let's update this to match the standard.

operator/v1/types_ingress.go
deads2k: > I'm not sure which format test you are referring to.

Yeah, it's missing. Please create a test showing that an incorrect duration format is rejected.

alebedev87: A validation pattern is added. The regexp is taken from other duration fields of the API.

operator/v1/types_ingress.go
alebedev87: Done.

operator/v1/types_ingress.go
deads2k: Please add a test for the validation on this field to https://github.com/openshift/api/blob/master/operator/v1/tests/ingresscontrollers.operator.openshift.io/AAA_ungated.yaml

alebedev87: Done.

operator/v1/types_ingress.go
```diff
@@ -1649,6 +1649,23 @@ type IngressControllerTuningOptions struct {
 	// +optional
 	TunnelTimeout *metav1.Duration `json:"tunnelTimeout,omitempty"`
 
+	// ConnectTimeout defines the maximum time to wait for
```
Miciah, line 1652: Whoops, we missed this—[the godoc should use the same capitalization as the json field name](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#use-json-field-names-in-godoc): "connectTimeout".  
```suggestion
	// connectTimeout defines the maximum time to wait for
```

PR #1830, opened 2024-03-26
machineconfiguration/v1alpha1: add PinnedImageSetStatus

hexfusion:
This PR adds a basic status to PinnedImageSet. The reason for the status which would be unused would be to ensure a Generation is populated if a set is mutated.


machineconfiguration/v1alpha1/types_pinnedimageset.go
```diff
@@ -26,6 +26,20 @@ type PinnedImageSet struct {
 	// spec describes the configuration of this pinned image set.
 	// +kubebuilder:validation:Required
 	Spec PinnedImageSetSpec `json:"spec"`
+
+	// status describes the last observed state of this pinned image set.
+	// +optional
+	Status PinnedImageSetStatus `json:"status"`
```
deads2k, line 32: missing status subresource in `// +` tags above?

deads2k:
neat that this didn't fail on e2e tests, but there's an e2e test that ensures schemas with status have status subresources.

machineconfiguration/v1alpha1/types_pinnedimageset.go
hexfusion: gotcha kubebuilder adding

```
// +kubebuilder:subresource:status
```

PR #1832, opened 2024-03-26
bump(golang.org/x/tools) to match k8s 1.28

soltysh:
We're using old version of `golang.org/x/tools` (0.8.0), which is causing problems when we bump go to 1.22, which is required when bringing in k8s 1.30. Also k8s 1.28 which is what `tools/go.mod` is using is on v0.16.1 (see [here](https://github.com/kubernetes/kubernetes/blob/release-1.28/go.mod#L88)). 

/assign @deads2k


dinhxuanvu (commented):
/lgtm

PR #1837, opened 2024-03-27
OCPBUGS-16418: Show more clear info about the DeploymentConfig is deprecated

sanchezl:
* Updated deprecated from OCP 4.14 to the corresponding k8s **1.27**
* Because this API is currently never going away, set removed=**0.0**

With these settings the warning header should change from this: 

```
apps.openshift.io/v1 DeploymentConfig is deprecated in v4.14+, unavailable in v4.10
```

to 

```
apps.openshift.io/v1 DeploymentConfig is deprecated in v1.27+
```

Open question:

Given that this type is never actually going to be removed, should we just remove the deprecation warning?
Marking it as deprecated, usually starts the clock towards removal.


sanchezl:
Proof PR: https://github.com/openshift/api/pull/1837

apps/v1/types.go
```diff
@@ -14,9 +14,9 @@ import (
 // +genclient:method=GetScale,verb=get,subresource=scale,result=k8s.io/api/extensions/v1beta1.Scale
 // +genclient:method=UpdateScale,verb=update,subresource=scale,input=k8s.io/api/extensions/v1beta1.Scale,result=k8s.io/api/extensions/v1beta1.Scale
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
-// +k8s:prerelease-lifecycle-gen:introduced=3.0
-// +k8s:prerelease-lifecycle-gen:deprecated=4.14
-// +k8s:prerelease-lifecycle-gen:removed=4.10000
+// +k8s:prerelease-lifecycle-gen:introduced=1.1
+// +k8s:prerelease-lifecycle-gen:deprecated=1.27
```
ingvagabund, line 18: Note: the versioning was originally intentionally set to follow the openshift versioning

apps/v1/types.go
```diff
@@ -14,9 +14,9 @@ import (
...
+// +k8s:prerelease-lifecycle-gen:removed=0.0
```
ingvagabund, line 19: Oh, if this removes the removed part in the message => ++

ingvagabund:
> Given that this type is never actually going to be removed, should we just remove the deprecation warning?

The warning message is definitely annoying. +1 for removing it when listing all resources. I'd still prefer to keep it when CRUD'ing the DC resource.

apps/v1/types.go
sanchezl: Understood, but we are making use of upstream functionality that only truly understands kubernetes versions.

sanchezl:
> > Given that this type is never actually going to be removed, should we just remove the deprecation warning?
> 
> The warning message is definitely annoying. +1 for removing it when listing all resources. I'd still prefer to keep it when CRUD'ing the DC resource.

@ingvagabund I've opened a companion PR to remove it from `oc status` and `oc get all`.

PR #1840, opened 2024-03-29
SDN-4596: Promote `AdminNetworkPolicy` to GA

tssurya:
This PR:

1. Moves ANP featuregate to also be part of the default set so that the feature is enabled by default (gate was being used from CNO)
2. Work related to https://issues.redhat.com/browse/SDN-4157 4.16 epic ANP
3. Aim is to have this feature ON by default on all OCP/HCP clusters

Unsure if I need to do PRs in other repos like CCC? I tried to look through https://docs.google.com/document/d/1zOL38_KDKwAvsx-LMHfyDdX4-jq3HQU1YWBjfuHaM0Q/edit but not sure what the new process is since the API code around FG has changed quiet a bit. After this PR I wonder if I need to do something else, pointers in right direction will be much appreciated!

E2E Test Info: This feature's tests live in: https://github.com/kubernetes-sigs/network-policy-api/tree/main/conformance and they are run as presubmits in: https://github.com/ovn-org/ovn-kubernetes/blob/master/test/conformance/network_policy_v2_test.go#L58 repo

Feature has been TechPreview for 2 releases, time to GA this.

### DIFF
--- a/config/v1/feature_gates.go
+++ b/config/v1/feature_gates.go
@@ -356,7 +356,7 @@ var (
 					reportProblemsToJiraComponent("Networking/ovn-kubernetes").
 					contactPerson("tssurya").
 					productScope(ocpSpecific).
-					enableIn(TechPreviewNoUpgrade).
+					enableIn(Default, TechPreviewNoUpgrade).
 					mustRegister()
 
 	FeatureGateNetworkLiveMigration = newFeatureGate("NetworkLiveMigration").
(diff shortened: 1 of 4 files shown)


deads2k:
Can you either
1. link to the automated tests results in periodic jobs running on a schedule to ensure regression protection both from the owning component and other components via side-effect (cluster-node-tuning-operator just benefited from such protection when an MCO PR broke them)
2. A direct QE sign off that this feature is ready.

See deeper examples of how to be promoted here: https://github.com/openshift/enhancements/blob/master/dev-guide/featuresets.md#id-like-to-declare-a-feature-accessible-by-default--what-is-the-process

tssurya:
thanks @deads2k ! that makes totally sense.
@asood-rh: could you please give your blessing for this: https://github.com/openshift/enhancements/blob/master/dev-guide/featuresets.md#obtaining-qe-sign-off IIRC you already did the necessary testing for 4.14/4.15 to ensure this feature is isolated enough to not break anything else?

I also plan to do some e2e's on origin i.e add https://github.com/ovn-org/ovn-kubernetes/blob/master/test/conformance/network_policy_v2_test.go#L58 also in OCP which is tracked via https://issues.redhat.com/browse/SDN-3807 but that needs the test infra to be more stable, I want to spend some time soaking the new tests upstream in OVNK before bringing them into OCP.

PR #1849, opened 2024-04-09
MCO-838: operator/machineconfigution: add NodeSynchronizers

hexfusion:
This PR adds  the NodeSynchronizer types to the MCO spec and status to serve as an aggregate reporting mechanism for configuration synchronizers such as PinnedImageSets which reconcile a node level configuration outside of `MachineConfig`.  For this reason it is not appropriate to aggregate this status via the MCP's existing types as it would cause confusion.

In general this is a future direction that the MCO will be heading which will allow for node level vs pool level granular configuration.


operator/v1/types_machineconfiguration.go
```diff
@@ -65,8 +74,58 @@ type MachineConfigurationStatus struct {
...
+	// +kubebuilder:validation:Minimum=0
+	MachineCount int64 `json:"machineCount"`
+	// updatedMachineCount is the number of machines that have been updated by the node synchronizer.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Minimum=0
+	UpdatedMachineCount int64 `json:"updatedMachineCount"`
+	// readyMachineCount is the number of machines managed by the node synchronizer that are in a ready state.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Minimum=0
+	ReadyMachineCount int64 `json:"readyMachineCount"`
```
deads2k, line 105: ready must be less than or equal to Available.

operator/v1/types_machineconfiguration.go
```diff
@@ -65,8 +74,58 @@ type MachineConfigurationStatus struct {
...
+	// +kubebuilder:validation:Minimum=0
+	AvailableMachineCount int64 `json:"availableMachineCount"`
+	// unavailableMachineCount is the number of machines managed by the node synchronizer but are unavailable.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Minimum=0
+	UnavailableMachineCount int64 `json:"unavailableMachineCount"`
+	// observedGeneration is the last generation change that has been applied.
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
```
deads2k, line 117: observedgeneration can move zero to anything positive.  anythign positive to zero.  anything positive to something greater than current.

operator/v1/types_machineconfiguration.go
```diff
@@ -65,8 +74,58 @@ type MachineConfigurationStatus struct {
...
+}
+
+// NodeSynchronizerType is a string enum used in a NodeSynchronizers object. They describe the type of node synchronizer.
+// +kubebuilder:validation:Enum:="PinnedImageSets"
+type NodeSynchronizerType string
+
+const (
+	// PinnedImageSets represents a node synchronizer for pinned image sets.
+	PinnedImageSets NodeSynchronizerType = "PinnedImageSets"
```
deads2k, line 126: I'm flexible on this aspect.  Named or free-form.  One option would be to have an instance in the slice for each pinnedimageset for example.  I'm happy to start here and potentially relax before going accessible-by-default.

operator/v1/types_machineconfiguration.go
```diff
@@ -54,6 +54,15 @@ type MachineConfigurationSpec struct {
...
 	NodeDisruptionPolicy NodeDisruptionPolicyConfig `json:"nodeDisruptionPolicy"`
+
+	// nodeSynchronizers is a list of node synchronizer that will be registered
+	// by this operator.  A node synchronizer manages a set of machines and
+	// ensures that they are updated to the latest configuration. The aggregate
+	// status of the machines managed by the node synchronizer will be reported
+	// in the Machine Configuration status.
+	// +openshift:enable:FeatureGate=PinnedImages
+	// +optional
+	NodeSynchronizers []NodeSynchronizerType `json:"nodeSynchronizers"`
```
deads2k, line 65: the spec field is new.  why a spec instead of just status?

deads2k: Still wondering about why we need the spec field.  The pinnedimageset thing is unconditional, right?

operator/v1/types_machineconfiguration.go
```diff
@@ -65,8 +65,71 @@ type MachineConfigurationStatus struct {
...
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])(\.([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9]))*$`
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+	// nodeSynchronizerType is the name of the node synchronizer type.
+	// +kubebuilder:validation:Required
+	NodeSynchronizerType NodeSynchronizerType `json:"nodeSynchronizerType"`
+	// nodeSelector specifies a label selector for Machines that match this synchronizer.
+	// +optional
+	NodeSelector *metav1.LabelSelector `json:"nodeSelector,omitempty"`
```
hexfusion, line 96: since the existing synchronizers are pool based I considered having a pool name in status but the node selector fells my flexible and future proof. So for example if a synchronizer is defined on the pool level then that label should be defined here.

operator/v1/types_machineconfiguration.go
```diff
@@ -65,8 +65,71 @@ type MachineConfigurationStatus struct {
...
+	// and must be at most 253 characters in length.
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])(\.([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9]))*$`
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+	// nodeSynchronizerType is the name of the node synchronizer type.
+	// +kubebuilder:validation:Required
+	NodeSynchronizerType NodeSynchronizerType `json:"nodeSynchronizerType"`
+	// nodeSelector specifies a label selector for Machines that match this synchronizer.
+	// +optional
```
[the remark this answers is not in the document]
hexfusion: this might make sense grouped with the type

operator/v1/types_machineconfiguration.go
hexfusion:
```
          nodeSynchronizersStatus:
            - name: master-pool
              nodeSynchronizerType: PinnedImageSets
              nodeSelector:
                matchLabels:
                  machineconfiguration.openshift.io/role: "master"
              machineCount: 3
              updatedMachineCount: 3
              readyMachineCount: 3
              availableMachineCount: 3
              unavailableMachineCount: 0
              observedGeneration: 1
  ```

PR #1852, opened 2024-04-11
MCO-838: machineconfig/v1: add PoolSynchronizers to MCP status

hexfusion:
This PR adds the PoolSynchronizer types to the MCP status to serve as an aggregate reporting mechanism for configuration synchronizers such as PinnedImageSets which reconcile a node-level configuration outside of MachineConfig.


sinnykumari:
Current status syncs looks good. Thanks Sam!

machineconfiguration/v1/types.go
```diff
@@ -503,8 +503,73 @@ type MachineConfigPoolStatus struct {
...
+	// The total length of the name must not exceed 253 characters.
+	// +openshift:enable:FeatureGate=PinnedImages
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])(\.([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9]))*$`
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+	// poolSynchronizerType describes the type of the pool synchronizer.
+	// +kubebuilder:validation:Required
+	PoolSynchronizerType PoolSynchronizerType `json:"poolSynchronizerType"`
```
deads2k, line 535: For a PinnedImageSet give me an example of the name and poolSynchronizerType please.  If they're the same, why have both?

machineconfiguration/v1/types.go
```diff
@@ -503,8 +503,73 @@ type MachineConfigPoolStatus struct {
...
+	// +kubebuilder:validation:Minimum=0
+	AvailableMachineCount int64 `json:"availableMachineCount"`
+	// unavailableMachineCount is the number of machines managed by the node synchronizer but are unavailable.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Minimum=0
+	UnavailableMachineCount int64 `json:"unavailableMachineCount"`
+	// observedGeneration is the last generation change that has been applied.
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
```
deads2k, line 559: this should only allow
1. 0 to positive
2. positive to zero
3. positive to bigger positive

It should reject
1. positive to smaller positive

right?

machineconfiguration/v1/types.go
```diff
@@ -503,8 +503,73 @@ type MachineConfigPoolStatus struct {
...
+	// +listMapKey=name
+	// +optional
+	PoolSynchronizersStatus []PoolSynchronizerStatus `json:"poolSynchronizersStatus"`
+}
+
+// +kubebuilder:validation:XValidation:rule="self.machineCount >= self.updatedMachineCount", message="machineCount must be greater than or equal to updatedMachineCount"
+// +kubebuilder:validation:XValidation:rule="self.machineCount >= self.availableMachineCount", message="machineCount must be greater than or equal to availableMachineCount"
+// +kubebuilder:validation:XValidation:rule="self.machineCount >= self.unavailableMachineCount", message="machineCount must be greater than or equal to unavailableMachineCount"
+// +kubebuilder:validation:XValidation:rule="self.machineCount >= self.readyMachineCount", message="machineCount must be greater than or equal to readyMachineCount"
+// +kubebuilder:validation:XValidation:rule="self.availableMachineCount >= self.readyMachineCount", message="availableMachineCount must be greater than or equal to readyMachineCount"
```
deads2k, line 519: add tests please.  These appear backwards at first glance.

machineconfiguration/v1/types.go
hexfusion: I see a possibility of using the name as a ref and the type as explicit relationship. Or another usecase is maybe it makes sense to group the sets into a group which has a name.

name: worker-set <-report on single pinnedImageSet
type: PinnedImageSet

name: rendered-worker-set <-report on a set that is a combination of more than one set
type: PinnedImageSet

hexfusion: without a name we would be bound to a single synchronizer per type which may or may not be desired depending on the type or future type.

deads2k: so the mapkey is actually "name,poolSynchronizerType" in the status field?

machineconfiguration/v1/types.go
```diff
@@ -503,8 +503,73 @@ type MachineConfigPoolStatus struct {
 	// +listType=atomic
 	// +optional
 	CertExpirys []CertExpiry `json:"certExpirys"`
+
+	// poolSynchronizersStatus is the status of the machines managed by the pool synchronizers.
+	// +openshift:enable:FeatureGate=PinnedImages
+	// +listType=map
+	// +listMapKey=name
```
deads2k, line 510: did you mean for this to be name,poolSynchronizerType

hexfusion: good point, yes

machineconfiguration/v1/types.go
hexfusion: updating cel to check oldSelf plus test

hexfusion: done

machineconfiguration/v1/types.go
hexfusion: correct it should be

machineconfiguration/v1/types.go
```diff
@@ -575,6 +642,14 @@ const (
 	// MachineConfigPoolRenderDegraded means the rendered configuration for the pool cannot be generated because of an error
 	MachineConfigPoolRenderDegraded MachineConfigPoolConditionType = "RenderDegraded"
 
+	// MachineConfigPoolPinnedImageSetsDegraded means the pinned image sets for the pool cannot be populated because of an error
+	// +openshift:enable:FeatureGate=PinnedImages
+	MachineConfigPoolPinnedImageSetsDegraded MachineConfigPoolConditionType = "PinnedImageSetsDegraded"
+
+	// MachineConfigPoolSynchronizerDegraded means the pool synchronizer can not be updated because of an error
+	// +openshift:enable:FeatureGate=PinnedImages
+	MachineConfigPoolSynchronizerDegraded MachineConfigPoolConditionType = "PoolSynchronizerDegraded"
```
yuqi-zhang, line 651: Just to understand this degrade, for example, for pinnedimagesets, how would this differ from `MachineConfigPoolPinnedImageSetsDegraded`?

For example, is `MachineConfigPoolPinnedImageSetsDegraded` -> a pull failed because not enough space on disk,
and `MachineConfigPoolSynchronizerDegraded` -> there's unavailable machines for that type (so more generic)?

yuqi-zhang (commented):
Mostly lgtm, one question inline

machineconfiguration/v1/types.go
hexfusion: well in this context the degraded types would be for the controllers that populate the data for MCP. So for example the pinnedImageSet controller would just populate the MCP spec. If the controller was degraded we would say why. THis is the same pattern used by for example the render controller.

So I would expect a small controller that watches on the synchronizer type to populate the status. If more types are created that controller would watch another resource. 

So these are status for controllers maintaining MCP vs status for the synchronization actions if that makes sense.

machineconfiguration/v1/types.go
```diff
@@ -503,8 +503,75 @@ type MachineConfigPoolStatus struct {
...
+	// Made up of one of more period-separated (.) segments, where each segment
+	// consists of alphanumeric characters and hyphens (-), must begin and end
+	// with an alphanumeric character, and is at most 63 characters in length.
+	// The total length of the name must not exceed 253 characters.
+	// +openshift:enable:FeatureGate=PinnedImages
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])(\.([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9]))*$`
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
```
hexfusion, line 533: this is an open question to the team do you want to support the option for multiple synchronizers per type? Currently, the only type is PinnedImageSets. We could instead populate a list of references that make up the synchronizer.

```
// MCOObjectReference holds information about an object the MCO either owns
// or modifies in some way
type MCOObjectReference struct {
	// name is the object name.
	// Must be a lowercase RFC-1123 hostname (https://tools.ietf.org/html/rfc1123)
	// It may consist of only alphanumeric characters, hyphens (-) and periods (.)
	// and must be at most 253 characters in length.
	// +kubebuilder:validation:MaxLength:=253
	// +kubebuilder:validation:Pattern=`^([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])(\.([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9]))*$`
	Name string `json:"name"`
}
```

sinnykumari: It makes sense to me to refer multiple PinnedImageSets by a synchronizer type.
If I understand correctly, this would mean that we would be able to express PinnedImageSet related information in MCP something like this:
```
        kind: MachineConfigPool
        spec:
          pinnedImageSets:
            - name: test-pinnedimageset
            - name: test1-pinnedimageset
            - name: test2-pinnedimageset
        status:
          poolSynchronizersStatus:
            - source: 
              - name: test-pinnedimageset
              - name: test1-pinnedimageset
              - name: test2-pinnedimageset
              poolSynchronizerType: PinnedImageSets
              machineCount: 3
              updatedMachineCount: 3
              readyMachineCount: 3
              availableMachineCount: 3
              unavailableMachineCount: 0
              observedGeneration: 3
```

sinnykumari: thanks, I find this as a useful change since this provides PinnedImageSet related status more user friendly to read.

hexfusion: The possible downside is a pool can declare up to 100 pinnedImageSets plus to be complete we would need to report on the current generation as well?

sinnykumari: I thought that's what we are anyway doing by allowing users to create and specify multiple pinnedImageSets targeting a pool. Are we limiting number of pinnedImageSets user can specify in spec? I may have missed if any changes were made.

hexfusion: second proposal

```
        kind: MachineConfigPool
        spec:
          pinnedImageSets:
            - name: test-pinnedimageset
            - name: test1-pinnedimageset
            - name: test2-pinnedimageset
        status:
          poolSynchronizersStatus:
            - source: 
              - name: test-pinnedimageset
                desiredGeneration: 3
                currentGeneration: 2
              - name: test1-pinnedimageset
                desiredGeneration: 3
                currentGeneration: 2
              - name: test2-pinnedimageset
                desiredGeneration: 3
                currentGeneration: 2
              poolSynchronizerType: PinnedImageSets
              machineCount: 3
              updatedMachineCount: 3
              readyMachineCount: 3
              availableMachineCount: 3
              unavailableMachineCount: 0
              observedGeneration: 3
  ```

machineconfiguration/v1/types.go
hexfusion: After talking with MCO team we dropped the name field because the feeling is that a single synchronizer per type is reasonable.

yuqi-zhang (approved):
/lgtm

PR #1861, opened 2024-04-18
NO-JIRA: Change PreviousFlakePercentage value type from int to float

dinhxuanvu:
PreviousFlakePercentage field in SippyTestInfo struct should be float instead of int given percentage value can be decimal. The data from Sippy does return previous_flake_percentage as decimal.


dinhxuanvu:
/cc @deads2k @soltysh
PTAL

PR #1864, opened 2024-04-22
Expose chunksize variable for S3 storage driver

deepsm007:
https://github.com/openshift/cluster-image-registry-operator/pull/1026

/hold


imageregistry/v1/types.go
```diff
@@ -200,6 +200,11 @@ type ImageRegistryConfigStorageS3 struct {
 	// Optional, defaults based on the Region that is provided.
 	// +optional
 	RegionEndpoint string `json:"regionEndpoint,omitempty"`
+	// The S3 API requires multipart upload chunks to be at least 5MB.
+	// This value should be a number that is larger than 5 * 1024 * 1024.
+	// The default is 10 MB.
+	// +optional
+	Chunksize int `json:"chunksize,omitempty"`
```
JoelSpeed, line 207: Why `Chunksize` and not `ChunkSize`?

imageregistry/v1/types.go
```diff
@@ -200,6 +200,11 @@ type ImageRegistryConfigStorageS3 struct {
 	// Optional, defaults based on the Region that is provided.
 	// +optional
 	RegionEndpoint string `json:"regionEndpoint,omitempty"`
+	// The S3 API requires multipart upload chunks to be at least 5MB.
```
JoelSpeed, line 203: Godoc should start with the json tag name, see other fields nearby for example

flavianmissi:
left a comment. this looks good for a registry perspective otherwise.

imageregistry/v1/types.go
JoelSpeed, line 207: Should add kubebuilder tags for the minimum and maximum values

deepsm007: consistency with https://docker-docs.uclv.cu/registry/storage-drivers/s3/

imageregistry/v1/types.go
```diff
@@ -200,6 +200,11 @@ type ImageRegistryConfigStorageS3 struct {
...
+	// This value should be a number that is larger than 5 * 1024 * 1024.
+	// The default is 10 MB.
```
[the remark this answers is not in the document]
JoelSpeed: Is there a default applied by the S3 API? We should mention the "no opinion", so typically we have notes like below
```
// The current default value is 10 MiB.
```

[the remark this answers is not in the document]
flavianmissi: We can probably link to the distribution docs here? 
related: https://distribution.github.io/distribution/storage-drivers/s3/ and https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html#sigv4-chunked-upload-sig-calculation-chunk0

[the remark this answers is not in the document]
flavianmissi: This is a data transmission setting, I'm not sure `*resource.Quantity` covers that?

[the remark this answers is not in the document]
deepsm007: Based on limits defined by amazone https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html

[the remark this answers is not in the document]
JoelSpeed: So you have a couple of options, you could make this `ChunkSizeMiB` and set the min to 5 and max to 5120 and keep as an `Int32`. Or use resource quanitity and restrict that to `Mi` and `Gi` suffixes, and then users case supply `5Mi` or `5Gi` and you can convert that internally to whatever you need to pass to the upstream API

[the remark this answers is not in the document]
deepsm007: Can I keep it in bytes consistent with [this](https://github.com/distribution/distribution/blob/df983747649396a7377a4b428ad3a1aebf7a9001/registry/storage/driver/s3-aws/s3.go#L47-L53) ?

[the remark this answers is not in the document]
JoelSpeed: It makes the API awkward to use from an end user perspective, how easy is it for me to decipher `chunkSize: 26214400` vs `chunkSizeMiB: 25`. It's better to create an API that is easier to handle for our end users, and do some internal conversion on our side.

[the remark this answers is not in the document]
deepsm007: Ack, I’ll update it. Thanks 😊

imageregistry/v1/types.go
```diff
@@ -200,6 +200,12 @@ type ImageRegistryConfigStorageS3 struct {
 	// Optional, defaults based on the Region that is provided.
 	// +optional
 	RegionEndpoint string `json:"regionEndpoint,omitempty"`
+	// chunkSizeMiB defines the size of the multipart upload chunks of the S3 API.
+	// The S3 API requires multipart upload chunks to be at least 5MB.
```
JoelSpeed, line 204: I left a comment before about what we should express when the value is unset, that should be added here

imageregistry/v1/types.go
```diff
@@ -200,6 +200,12 @@ type ImageRegistryConfigStorageS3 struct {
...
+	// +kubebuilder:validation:Minimum=5
+	// +kubebuilder:validation:Maximum=5120
+	// +optional
+	ChunkSizeMiB int64 `json:"chunkSizeMiB,omitempty"`
```
JoelSpeed, line 208: Make this an int32, with a maximum of 5120 you don't need a 64

imageregistry/v1/types.go
JoelSpeed, line 204: `5MiB` to make sure it's obvious it's binary bytes

Also, need to express in prose that the value has a maximum, and that the value is an integer number of MiB

flavianmissi:
> Given the proximity to the feature freeze date, what is your plan for getting this feature in? Are you intending to have it in 4.17 onwards?

@deepsm007 and I discussed this and given that we don't have any trackers for this yet (on the works) and the proximity of feature freeze we are currently expecting to land on 4.17.
@deepsm007 let's put it on hold to communicate that intention?

imageregistry/v1/types.go
```diff
@@ -200,6 +200,14 @@ type ImageRegistryConfigStorageS3 struct {
 	// Optional, defaults based on the Region that is provided.
 	// +optional
 	RegionEndpoint string `json:"regionEndpoint,omitempty"`
+	// chunkSizeMiB defines the size of the multipart upload chunks of the S3 API.
```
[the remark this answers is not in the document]
JoelSpeed: You have 2 choices that are valid, one is to include the units, which can be helpful when the range is small, or you can use a quantity. In this case we chose to include the units, but you're right, a quantity type could have been used, in combination with some CEL or regex to limit the range of valid values

imageregistry/v1/types.go
```diff
@@ -200,6 +200,16 @@ type ImageRegistryConfigStorageS3 struct {
...
+	// The S3 API requires multipart upload chunks to be at least 5MiB.
+	// When omitted, this means no opinion and the platform is left to choose a reasonable default, which is subject to change over time.
+	// The current default value is 10 MiB.
+	// The value is an integer number of MiB
+	// +kubebuilder:validation:Minimum=5
+	// +kubebuilder:validation:Maximum=5120
+	// +openshift:enable:FeatureGate=ChunkSizeMiB
```
JoelSpeed, line 210: Does this feature gate actually exist? The verify job suggests it does not

imageregistry/v1/types.go
```diff
@@ -200,6 +200,16 @@ type ImageRegistryConfigStorageS3 struct {
 	// Optional, defaults based on the Region that is provided.
 	// +optional
 	RegionEndpoint string `json:"regionEndpoint,omitempty"`
+	// chunkSizeMiB defines the size of the multipart upload chunks of the S3 API.
+	// The S3 API requires multipart upload chunks to be at least 5MiB.
+	// When omitted, this means no opinion and the platform is left to choose a reasonable default, which is subject to change over time.
+	// The current default value is 10 MiB.
+	// The value is an integer number of MiB
```
JoelSpeed, line 207: Nit
```suggestion
	// The value is an integer number of MiB.
```

imageregistry/v1/types.go
```diff
@@ -200,6 +200,16 @@ type ImageRegistryConfigStorageS3 struct {
...
+	// +kubebuilder:validation:Minimum=5
+	// +kubebuilder:validation:Maximum=5120
```
JoelSpeed, line 209: This needs to be explained in the godoc, please add.
```
// The minimum value is 5 and the maximum value is 5120 (5 GiB).
```

imageregistry/v1/tests/configs.imageregistry.operator.openshift.io/ChunkSizeMiB.yaml
```diff
@@ -0,0 +1,99 @@
...
+      initial: |
+        apiVersion: imageregistry.operator.openshift.io/v1
+        kind: Config
+        spec:
+          replicas: 1
+          storage:
+            s3:
+              bucket: bucket-1
+              region: region-1
+              ChunkSizeMiB: 10
```
JoelSpeed, line 30: Casing is incorrect here

imageregistry/v1/tests/configs.imageregistry.operator.openshift.io/ChunkSizeMiB.yaml
```diff
@@ -0,0 +1,99 @@
...
+        kind: Config
+        spec:
+          logLevel: Normal
+          operatorLogLevel: Normal
+          replicas: 1
+          storage:
+            s3:
+              bucket: bucket-1
+              region: region-1
+              ChunkSizeMiB: 10485760
```
JoelSpeed, line 42: Why is this expected?

imageregistry/v1/tests/configs.imageregistry.operator.openshift.io/ChunkSizeMiB.yaml
```diff
@@ -0,0 +1,99 @@
...
+    - name: Should not be able to configure ChunkSizeMiB for storage s3 without details
```
JoelSpeed, line 43: Should not implies an error, but the test case has no error?

imageregistry/v1/tests/configs.imageregistry.operator.openshift.io/ChunkSizeMiB.yaml
```diff
@@ -0,0 +1,99 @@
...
+        apiVersion: imageregistry.operator.openshift.io/v1
+        kind: Config
+        spec:
+          replicas: 1
+          storage:
+            s3:
+              bucket: bucket-2
+              region: region-2
+              ChunkSizeMiB: 2
+      expectedError: "spec.storage.s3.chunksizemib in body should match '^(5\\d{0,2}|[1-4]\\d{3}|5120)$'"
```
JoelSpeed, line 75: The casing on `chunksisemib` is not as expected, and the regex looks off, looks like the tests need a good update?

JoelSpeed (commented):
Your tests seem to be rather out of date, please resolve the comments on the API itself (that's pretty close) and fix the missing feature gate and the test cases please

features/features.go
```diff
@@ -508,4 +508,11 @@ var (
 						contactPerson("rvanderp3").
 						productScope(ocpSpecific).
 						mustRegister()
+
+	FeatureGateChunkSizeMiB = newFeatureGate("ChunkSizeMiB").
+				reportProblemsToJiraComponent("Image Registry").
+				contactPerson("flavianmissi").
+				productScope(ocpSpecific).
+				enableIn(DevPreviewNoUpgrade, TechPreviewNoUpgrade).
```
JoelSpeed, line 516: This isn't right and causes build failures
```suggestion
				enableIn(configv1.DevPreviewNoUpgrade, configv1.TechPreviewNoUpgrade).
```

JoelSpeed:
Verify failures are because we cannot detect that you are duplicating the CRD that already exists, however, we do want you to fix `ListsMustHaveSSATags`. Can you raise a separate PR that fixes all of the SSA tags on the image registry type? This will be needed to move forward

The tests are genuinely broken, you are not using the correct casing for you field which means it's being dropped, please fix the casing

flavianmissi:
PR needs rebase and the verify-crd-schema job is still broken.
The code otherwise lgtm from an image registry perspective.

JoelSpeed:
/lgtm

Failures in the CRD schema check are pre-existing and from corev1 types that cannot be fixed

deads2k:
Failures in the CRD schema check are pre-existing and from corev1 types that cannot be fixed

PR #1882, opened 2024-05-07
add MachineConfiguration conditions for boot image updates

djoshy:
This PR adds a few new `MachineConfiguration` conditions, which will be used by the MSBIC to inform the user of the results of the boot image reconciliation loop. These conditions will be behind the ManagedBootImages feature gate.

More info:

- [Failure conditions for boot image updates](https://github.com/openshift/enhancements/blob/master/enhancements/machine-config/manage-boot-images.md#error--alert-mechanism)

### DIFF
--- a/operator/v1/types_machineconfiguration.go
+++ b/operator/v1/types_machineconfiguration.go
@@ -494,3 +494,16 @@ const (
 	// Special represents an action that is internal to the MCO, and is not allowed in user defined NodeDisruption policies.
 	SpecialStatusAction NodeDisruptionPolicyStatusActionType = "Special"
 )
+
+// These strings will be used for MachineConfiguration Status conditions.
+const (
+	// MachineConfigurationBootImageUpdateDegraded means that the MCO ran into an error while reconciling boot images. This
+	// will cause the clusteroperators.config.openshift.io/machine-config to degrade. This  condition will indicate the cause
+	// of the degrade, the progress of the update and the generation of the boot images configmap that it degraded on.
+	MachineConfigurationBootImageUpdateDegraded string = "BootImageUpdateDegraded"
+
+	// MachineConfigurationBootImageUpdateProgressing means that the MCO is in the process of reconciling boot images. This
+	// will cause the clusteroperators.config.openshift.io/machine-config to be in a Progressing state. This condition will
+	// indicate the progress of the update and the generation of the boot images configmap that triggered this update.
+	MachineConfigurationBootImageUpdateProgressing string = "BootImageUpdateProgressing"
+)


djoshy:
I do have an open question. If boot image updates are disabled(which is possible via the opt-in mechanism), none of the conditions are exactly applicable. Should I make a specific condition for the "Off" state, and set it to true?

operator/v1/types_machineconfiguration.go
```diff
@@ -494,3 +494,21 @@ const (
 	// Special represents an action that is internal to the MCO, and is not allowed in user defined NodeDisruption policies.
 	SpecialStatusAction NodeDisruptionPolicyStatusActionType = "Special"
 )
+
+// MachineConfigurationCondition are used to indicate success and failures for MachineConfiguration knobs.
+// Please note: These conditions are subject to change. Both additions and deletions may be made.
+type MachineConfigurationConditionType string
```
deads2k, line 500: I don't think you need the type.  These could just be strings.

operator/v1/types_machineconfiguration.go
```diff
@@ -494,3 +494,19 @@ const (
 	// Special represents an action that is internal to the MCO, and is not allowed in user defined NodeDisruption policies.
 	SpecialStatusAction NodeDisruptionPolicyStatusActionType = "Special"
 )
+
+// These strings will be used to indicate success and failures for MachineConfiguration knobs via conditions.
+// Please note: These are subject to change. Both additions and deletions may be made.
+const (
+	// MachineConfigurationBootImageUpdateFailed means that the MCO ran into an error while reconciling boot images.
+	// +openshift:enable:FeatureGate=ManagedBootImages
```
deads2k, line 502: this tag has no effect on this.  The way conditions work, we dont' restrict the condition types.

operator/v1/types_machineconfiguration.go
```diff
@@ -494,3 +494,19 @@ const (
...
+	MachineConfigurationBootImageUpdateFailed string = "BootImageUpdateFailed"
+
+	// MachineConfigurationBootImageUpdateSuccess means that the MCO was able to successfully update all boot images.
+	// +openshift:enable:FeatureGate=ManagedBootImages
+	MachineConfigurationBootImageUpdateSuccess string = "BootImageUpdateSuccess"
```
deads2k, line 507: can both success and failure be true at the same time?  Can they both be false?

operator/v1/types_machineconfiguration.go
```diff
@@ -494,3 +494,19 @@ const (
 	// Special represents an action that is internal to the MCO, and is not allowed in user defined NodeDisruption policies.
 	SpecialStatusAction NodeDisruptionPolicyStatusActionType = "Special"
 )
+
+// These strings will be used to indicate success and failures for MachineConfiguration knobs via conditions.
+// Please note: These are subject to change. Both additions and deletions may be made.
```
deads2k, line 499: The cost of having these in openshift/api is going to be, "will not be removed, semantically changed, or abandoned".  I won't say it's completely impossible to remove them, but the impact to user scripts is pretty extreme.

operator/v1/types_machineconfiguration.go
```diff
@@ -494,3 +494,19 @@ const (
...
+const (
+	// MachineConfigurationBootImageUpdateFailed means that the MCO ran into an error while reconciling boot images.
+	// +openshift:enable:FeatureGate=ManagedBootImages
+	MachineConfigurationBootImageUpdateFailed string = "BootImageUpdateFailed"
+
+	// MachineConfigurationBootImageUpdateSuccess means that the MCO was able to successfully update all boot images.
```
deads2k, line 505: is there a generation involved here?  If a new bootimage is created, this immediately becomes false?

djoshy:
Chatted with @deads2k on slack, and we boiled this down to two conditions:
- `BootImageUpdateDegraded`. This can be a failure mode in the controller loop or a user mis-configuration. This will flip co/machine-config to degrade. 
- `BootImageUpdateProgressing` - A boot image update is ongoing. This will flip co/machine-config to progressing.

All of them should include progress counts of the machinesets, and the generation info of the source configmap. If neither of these conditions are true, this means that the controller is in a happy/done state.

operator/v1/types_machineconfiguration.go
djoshy: Took this bit off, I don't envision us removing this in the future...I shouldn't have stolen this boilerplate from elsewhere :P

operator/v1/types_machineconfiguration.go
djoshy: I know we've removed the "Success" condition as it is seemed redundant, but for the 2 remainder states the generation of the boot images configmap will be included in every condition update as we discussed

djoshy:
> I do have an open question. If boot image updates are disabled(which is possible via the opt-in mechanism), none of the conditions are exactly applicable. Should I make a specific condition for the "Off" state, and set it to true?

Boiling it down to two conditions has helped with this too; as there we are no longer tracking a success state, the other two conditions being false feels valid when the feature is "off" as well.

Also to make a note: The MCO team will consider implementing something like [PoolSynchronizerStatus](https://github.com/openshift/api/blob/051406bdc6f5e1a15aa88d77e3ecfd7cd59e1857/machineconfiguration/v1/types.go#L520) for this in the future

PR #1883, opened 2024-05-07
Remove struct SSA tag markers

deepsm007:
Partial undo of https://github.com/openshift/api/pull/1877


deads2k:
/hold

this won't fix the referenced problem.

deads2k:
/close

confusion over which type was causing a compilation error.

PR #1887, opened 2024-05-10
OCPBUGS-33787: add console and download URLs to console operator config

alebedev87:
This PR adds new fields to the console operator configuration: `consoleURL`, `clientDownloadsURL`. They allow the user to set an alternative console and client download base addresses (routes being the default). This API aims at helping to [tolerate the absence of ingress capability on HyperShift managed clusters](https://github.com/openshift/console-operator/pull/886).


operator/v1/types_console.go
```diff
@@ -146,6 +146,13 @@ type ConsoleCustomization struct {
 	// +listMapKey=id
 	// +optional
 	Perspectives []Perspective `json:"perspectives"`
+	// consoleURL is an URL to be used as the base console address.
+	// If not specified, the console route hostname will be used.
+	// This is required when using console on clusters where the ingress capability is disabled.
```
jhadvig, line 151: nit:
I would add additional warning which would make it clear to add a resolvable URL.
```
// Invalid value will prevent a console rollout.
```

jhadvig (commented):
The change looks good 👍 Adding a small nit.
@spadgett thoughts ?

operator/v1/types_console.go
alebedev87: Done.

alebedev87:
All the latest commits are to fix `verify` and `integration` test jobs.

operator/v1/types_console.go
```diff
@@ -146,6 +146,14 @@ type ConsoleCustomization struct {
...
```
spadgett, line 151: What happens when you try to use this value when ingress is not disabled?

operator/v1/types_console.go
```diff
@@ -146,6 +146,14 @@ type ConsoleCustomization struct {
...
+	// Invalid value will prevent a console rollout.
+	// +kubebuilder:validation:Optional
+	// +optional
+	// +kubebuilder:validation:Pattern=`^https?:\/\/[^\s]`
```
spadgett, line 155: I'd think we'd need to require https

operator/v1/types_console.go
jhadvig: That part should be programatically conditioned by the console-operator

spadgett: OK, but what does the operator do? Is the value ignored?

alebedev87: The only use case for the moment is when the ingress is disabled. The comment needs to be updated to highlight this. I think the implementation may be expanded in the future as use cases come up.

operator/v1/types_console.go
alebedev87: Done.

operator/v1/types_console.go
alebedev87: Field comment updated.

alebedev87:
Field for the download ingress is added.

operator/v1/types_console.go
```diff
@@ -146,6 +146,22 @@ type ConsoleCustomization struct {
...
+	// +kubebuilder:validation:Pattern=`^https:\/\/[^\s]`
+	ConsoleURL string `json:"consoleURL,omitempty"`
+	// downloadURL is an URL to be used as the address to download client binaries.
+	// This field is intended for clusters without ingress capability,
+	// where access to routes is not possible.
+	// Invalid value will prevent a console rollout.
+	// +kubebuilder:validation:Optional
+	// +optional
+	// +kubebuilder:validation:Pattern=`^https:\/\/[^\s]`
+	DownloadURL string `json:"downloadURL,omitempty"`
```
jhadvig, line 164:
```suggestion
	DownloadsURL string `json:"downloadsURL,omitempty"`
```

operator/v1/types_console.go
```diff
@@ -146,6 +146,22 @@ type ConsoleCustomization struct {
...
 	Perspectives []Perspective `json:"perspectives"`
+	// consoleURL is an URL to be used as the base console address.
+	// This field is intended for clusters without ingress capability,
+	// where access to routes is not possible.
+	// Invalid value will prevent a console rollout.
+	// +kubebuilder:validation:Optional
+	// +optional
+	// +kubebuilder:validation:Pattern=`^https:\/\/[^\s]`
+	ConsoleURL string `json:"consoleURL,omitempty"`
+	// downloadURL is an URL to be used as the address to download client binaries.
```
jhadvig, line 157:
```suggestion
	// downloadsURL is an URL to be used as the address to download client binaries.
```

jhadvig (commented):
Thanks @alebedev87 adding couple of nits.
Also in ideally we should have both fields unified under a common field, something like `alternativeBaseAddress.consoleURL` && `alternativeBaseAddress.downloadsURL`, although Im not really sure about the `alternativeBaseAddress` name, so Im open for suggestions.

operator/v1/types_console.go
alebedev87: Done.

operator/v1/types_console.go
alebedev87: Done.

operator/v1/types_console.go
```diff
@@ -146,6 +146,22 @@ type ConsoleCustomization struct {
 	// +listMapKey=id
 	// +optional
 	Perspectives []Perspective `json:"perspectives"`
+	// consoleURL is an URL to be used as the base console address.
+	// This field is intended for clusters without ingress capability,
+	// where access to routes is not possible.
+	// Invalid value will prevent a console rollout.
+	// +kubebuilder:validation:Optional
+	// +optional
+	// +kubebuilder:validation:Pattern=`^https:\/\/[^\s]`
```
hexfusion, line 155: In general we have been trying to move away from regex validation instead using CEL (kubebuilder:validation:XValidation:rule). perhaps `startsWith()` and a clear message indicating the failure reason?

alebedev87: Replaced regexp with CEL validation.

operator/v1/types_console.go
```diff
@@ -146,6 +146,22 @@ type ConsoleCustomization struct {
...
+	// +kubebuilder:validation:XValidation:rule="self.startsWith('https://')",message="console url scheme must be https"
+	ConsoleURL string `json:"consoleURL,omitempty"`
+	// downloadsURL is an URL to be used as the address to download client binaries.
+	// This field is intended for clusters without ingress capability,
+	// where access to routes is not possible.
+	// Invalid value will prevent a console rollout.
+	// +kubebuilder:validation:Optional
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self.startsWith('https://')",message="downloads url scheme must be https"
+	DownloadsURL string `json:"downloadsURL,omitempty"`
```
hexfusion, line 164: I am curious why download was made plural.  You are downloading a binary with a downloadURL or a downloadsURL?

operator/v1/types_console.go
```diff
@@ -146,6 +146,22 @@ type ConsoleCustomization struct {
...
 	// +optional
 	Perspectives []Perspective `json:"perspectives"`
+	// consoleURL is an URL to be used as the base console address.
+	// This field is intended for clusters without ingress capability,
+	// where access to routes is not possible.
+	// Invalid value will prevent a console rollout.
+	// +kubebuilder:validation:Optional
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self.startsWith('https://')",message="console url scheme must be https"
+	ConsoleURL string `json:"consoleURL,omitempty"`
```
hexfusion, line 156: We probably want to put these new fields on this v1 type behind a featuregate? defer to @JoelSpeed

operator/v1/types_console.go
alebedev87: There are multiple binary artifacts which are exposed on this address. So historically the route and the corresponding DNS name were plural. [Jakub asked me](https://github.com/openshift/api/pull/1887#discussion_r1600123520) to keep the naming consistent.
cc @jhadvig.

operator/v1/types_console.go
alebedev87: The only 2 cents I can add to this is that the fields are intended to be used on HyperShift managed clusters only. Where disabling of the ingress capability is possible. Other OpenShift form factors won't be impacted. The field descriptions highlight this.

JoelSpeed: I think this should be gated in the first instance. One thing to note is that HyperShift is now able to have fields that do not exist in regular OCP clusters. There is a way to enable a feature gated field only for the HyperShift profile.

I know this was done to enable temporary deltas beteween OCP and HCP, but, I don't know if the plan is to allow those to become permanent or not.

What is the target release here?

operator/v1/types_console.go
JoelSpeed: What is this expecting to download, is it OC? You description says binaries so I guess it must be more than that?

operator/v1/types_console.go
```diff
@@ -146,6 +146,22 @@ type ConsoleCustomization struct {
...
+	// downloadsURL is an URL to be used as the address to download client binaries.
+	// This field is intended for clusters without ingress capability,
+	// where access to routes is not possible.
+	// Invalid value will prevent a console rollout.
+	// +kubebuilder:validation:Optional
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self.startsWith('https://')",message="downloads url scheme must be https"
```
JoelSpeed, line 163: Use the URL library to validate this

operator/v1/types_console.go
```diff
@@ -146,6 +146,22 @@ type ConsoleCustomization struct {
...
+	// +kubebuilder:validation:Optional
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self.startsWith('https://')",message="console url scheme must be https"
+	ConsoleURL string `json:"consoleURL,omitempty"`
+	// downloadsURL is an URL to be used as the address to download client binaries.
+	// This field is intended for clusters without ingress capability,
+	// where access to routes is not possible.
+	// Invalid value will prevent a console rollout.
+	// +kubebuilder:validation:Optional
+	// +optional
```
JoelSpeed, line 162: You only need one optional tag, drop the kubebuilder version

operator/v1/types_console.go
```diff
@@ -146,6 +146,22 @@ type ConsoleCustomization struct {
...
 	Perspectives []Perspective `json:"perspectives"`
+	// consoleURL is an URL to be used as the base console address.
+	// This field is intended for clusters without ingress capability,
+	// where access to routes is not possible.
+	// Invalid value will prevent a console rollout.
+	// +kubebuilder:validation:Optional
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self.startsWith('https://')",message="console url scheme must be https"
+	ConsoleURL string `json:"consoleURL,omitempty"`
+	// downloadsURL is an URL to be used as the address to download client binaries.
```
JoelSpeed, line 157: Nit, `a URL` not `an URL`

operator/v1/types_console.go
```diff
@@ -146,6 +146,22 @@ type ConsoleCustomization struct {
 	// +listMapKey=id
 	// +optional
 	Perspectives []Perspective `json:"perspectives"`
+	// consoleURL is an URL to be used as the base console address.
+	// This field is intended for clusters without ingress capability,
+	// where access to routes is not possible.
+	// Invalid value will prevent a console rollout.
+	// +kubebuilder:validation:Optional
+	// +optional
```
JoelSpeed, line 154: You only need one optional tag, drop the kubebuilder version

operator/v1/types_console.go
```diff
@@ -146,6 +146,22 @@ type ConsoleCustomization struct {
...
+	// +kubebuilder:validation:XValidation:rule="self.startsWith('https://')",message="console url scheme must be https"
```
JoelSpeed, line 155: Use the URL library to validate this

operator/v1/types_console.go
alebedev87: Yes, it's `oc` binary for different architectures.

operator/v1/types_console.go
alebedev87: > I don't know if the plan is to allow those to become permanent or not.

Currently, it's needed on HCP only. There is no use case for the usage of these fields on other form factors.

> What is the target release here?

Well, it was 4.16 but I'm beginning to doubt this :)

JoelSpeed: >Currently, it's needed on HCP only. There is no use case for the usage of these fields on other form factors.

David is out this week so I can't confirm with him til next week.

We are now at the `acknowledge-critical-fixes` only stage of 4.16 so I think it's probably not going to make 4.16 unless you want to claim this is critical (which I guess it could be? That's your call 👀 )

operator/v1/types_console.go
JoelSpeed: Considered calling it `clientDownloadURL`? Might be a bit more obvious if it had client in the name

operator/v1/types_console.go
alebedev87: Done. The `startWith("https://")` validation is kept to make sure the URL is HTTPS.

operator/v1/types_console.go
alebedev87: Done.

operator/v1/types_console.go
alebedev87: Done.

operator/v1/types_console.go
alebedev87: Done.

operator/v1/types_console.go
alebedev87: Done. The `startWith("https://")` validation is kept to make sure the URL is HTTPS.

operator/v1/types_console.go
JoelSpeed: You could also do `url(self).getScheme() == 'HTTPS'` I believe, though I'm not sure it makes much difference

operator/v1/types_console.go
```diff
@@ -146,6 +146,22 @@ type ConsoleCustomization struct {
...
+	// where access to routes is not possible.
+	// Invalid value will prevent a console rollout.
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self.isURL(self)",message="console url must be a valid absolute URL"
+	// +kubebuilder:validation:XValidation:rule="self.startsWith('https://')",message="console url scheme must be https"
+	ConsoleURL string `json:"consoleURL,omitempty"`
+	// downloadsURL is a URL to be used as the address to download client binaries.
+	// This field is intended for clusters without ingress capability,
+	// where access to routes is not possible.
+	// Invalid value will prevent a console rollout.
```
JoelSpeed, line 160: As above, include that it must be HTTPS

operator/v1/types_console.go
```diff
@@ -146,6 +146,22 @@ type ConsoleCustomization struct {
...
 	// +optional
 	Perspectives []Perspective `json:"perspectives"`
+	// consoleURL is a URL to be used as the base console address.
+	// This field is intended for clusters without ingress capability,
+	// where access to routes is not possible.
+	// Invalid value will prevent a console rollout.
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self.isURL(self)",message="console url must be a valid absolute URL"
+	// +kubebuilder:validation:XValidation:rule="self.startsWith('https://')",message="console url scheme must be https"
+	ConsoleURL string `json:"consoleURL,omitempty"`
```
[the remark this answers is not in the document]
alebedev87: Not really but we can define some sane max. Let's say `1024`.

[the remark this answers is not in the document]
JoelSpeed: SGTM

operator/v1/types_console.go
alebedev87: @jhadvig: What do you think about `clientDownloadURL`? Or maybe in the plural form: `clientDownloadsURL`?

jhadvig: @alebedev87 I would stick to `clientDownloadsURL`.
Also thinking if we could nest it:
```go
clients:
  consoleURL: https://url-for-console.com
  downloadsURL: https://url-for-downloads.com
```
although I know that we are under time pressure 

@JoelSpeed @hexfusion just for clarification, all the downloads related resources(service, route, pod, deployment) use the plural name.

JoelSpeed: Why does the consoleURL come under this clients categorisation? I'm not sure I'm following what these are actually for yet

operator/v1/types_console.go
```diff
@@ -146,6 +146,22 @@ type ConsoleCustomization struct {
 	// +listMapKey=id
 	// +optional
 	Perspectives []Perspective `json:"perspectives"`
+	// consoleURL is a URL to be used as the base console address.
+	// This field is intended for clusters without ingress capability,
+	// where access to routes is not possible.
+	// Invalid value will prevent a console rollout.
```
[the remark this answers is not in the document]
alebedev87: Done.

operator/v1/types_console.go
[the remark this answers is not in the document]
alebedev87: `MaxLength=1024` added.

operator/v1/types_console.go
alebedev87: `url(self).getScheme() == 'https'` validation replaced the regexp.

operator/v1/types_console.go
jhadvig: @JoelSpeed just a though on how to make both URLs nested under one field.

alebedev87: > I'm not sure I'm following what these are actually for yet

The console and downloads workloads are exposed via `Route` objects of the same names. The OpenShift router resides at the edge of the cluster to handle all the traffic for all the `Route` objects. The ingress capability has been introduced in OCP 4.16. The ingress capability includes the OpenShift router. Therefore on the clusters where the ingress capability can be disabled (HyperShift ones) the console and downloads routes are left unserved.
The console operator which manages the console and downloads workloads relies on the presence of the routes: to do some health check pings and to manage authentication flows. With this new fields we enable user to provide an alternative (to route host) ingress URL. We can imagine a user which disabled the ingress capability on a ROSA cluster and used ALBO to provision AWS ALBs for its console and downloads deployments. The hostnames of these ALBs can be provided to the console operator to keep doing its duty: health checks, authentication management.

operator/v1/types_console.go
alebedev87: url(self).getScheme() == 'https' validation replaced the regexp.

operator/v1/types_console.go
JoelSpeed: > Therefore on the clusters where the ingress capability can be disabled (HyperShift ones) the console and downloads routes are left unserved.

What prevents me disabling the capability on a regular OCP cluster?

Ok, that context is helpful. Makes me thing this is something like `clientEndpoints` or `routeEndpoints`?

What would happen with these fields if they were set by the user and the Route objects existed as normal, does this take precedence?

alebedev87: > just a though on how to make both URLs nested under one field

I can think of this:
```yaml
spec:
  customization:
     ingress:
       consoleURL: https://console-alb.com
       clientDownloadsURL: https://downloads-alb.com
```

JoelSpeed: I like that suggestion, happy to go with that

alebedev87: > What prevents me disabling the capability on a regular OCP cluster?

[The CVO doesn't allow this](https://github.com/openshift/cluster-version-operator/pull/946).

> Makes me thing this is something like clientEndpoints or routeEndpoints?

Yes. Not sure whether `endpoint` is the right word though. I would say it's like `ingress` status in `Service` object.

> What would happen with these fields if they were set by the user and the Route objects existed as normal, does this take precedence?

No. The fields will only be used when the ingress is disabled (== routes are not admitted but they may exist).

alebedev87: Implemented the grouping of the new fields under `ingress` as described [here](https://github.com/openshift/api/pull/1887#discussion_r1601522967).

operator/v1/types_console.go
alebedev87: Finally [the url library was used](https://github.com/openshift/api/pull/1887#discussion_r1601337866) to do the check for https scheme.

operator/v1/types_console.go
```diff
@@ -146,6 +146,11 @@ type ConsoleCustomization struct {
 	// +listMapKey=id
 	// +optional
 	Perspectives []Perspective `json:"perspectives"`
+	// ingress allows to configure the alternative ingress for the console.
+	// This field is intended for clusters without ingress capability,
+	// where access to routes is not possible.
+	// +optional
+	Ingress IngressCustomization `json:"ingress,omitempty"`
```
jhadvig, line 153: Sorry I've missed this one.
The `Ingress` field should go directly under the `ConsoleSpec`, not under `Customization`. In `Customization` we keep the UI related customizations.

alebedev87: Done.

Setup of the alternative ingress URLs for the console looks like this now:
```
spec:
  ingress:
    consoleURL: https://console-alb.com
    clientDownloadsURL: https://downloads-alb.com
```

operator/v1/types_console.go
```diff
@@ -375,6 +380,28 @@ const (
...
 
+// IngressCustomization allows cluster admin to configure alternative ingress for the console.
+type IngressCustomization struct {
+	// consoleURL is a URL to be used as the base console address.
+	// This field is intended for clusters without ingress capability,
+	// where access to routes is not possible.
+	// Must use the HTTPS scheme.
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="isURL(self)",message="console url must be a valid absolute URL"
+	// +kubebuilder:validation:XValidation:rule="url(self).getScheme() == 'https'",message="console url scheme must be https"
```
JoelSpeed, line 391: I'm not sure how this would error in the case that it isn't a URL, need to check, do we run the rules in order or not.

A fix to remove ambiguity would be to `isURL(self) && url(self).getScheme() == 'https'`

Your tests indicate they run in order so probably fine as is

operator/v1/types_console.go
```diff
@@ -375,6 +380,28 @@ const (
 	BrandROSA Brand = "ROSA"
 )
 
+// IngressCustomization allows cluster admin to configure alternative ingress for the console.
+type IngressCustomization struct {
+	// consoleURL is a URL to be used as the base console address.
+	// This field is intended for clusters without ingress capability,
+	// where access to routes is not possible.
```
JoelSpeed, line 387: You need to explain the behaviour when these are set on clusters that have ingress, I believe you said the Ingress overrides this?

Is there any use case where having this set would be preferable over the Route's that exist? Should we instead allow these to be overrides so users could add secondary/preferred DNS not via the Routes?

JoelSpeed (commented):
Are we going to look at featuregates for this?

operator/v1/types_console.go
alebedev87: @jhadvig: what do you think about the feature gate? If you think it's needed I'll postpone the change to 4.17 as I won't have time to think this thoroughly.

operator/v1/types_console.go
alebedev87: > I believe you said the Ingress overrides this?

No, it won't have any effect on a cluster with ingress. There was no use case for this so far.

> Is there any use case where having this set would be preferable over the Route's that exist?

No, not yet.

> Should we instead allow these to be overrides so users could add secondary/preferred DNS not via the Routes?

May be, one day when such a request would come.

operator/v1/types_console.go
alebedev87: I wanted more detailed error messages so I decided not to combine them. I added a test case to validate that the order of the validations is respected (when the url is invalid and has no scheme).

jhadvig (approved):
Looks good to me 👍 
Thanks @alebedev87 for this change. Leaving the approval to api review team.

operator/v1/types_console.go
alebedev87: @spadgett : so you would prefer the new fields to be respected in any context, not only when the ingress is disabled?

alebedev87: Rephrased from
> This field is intended for clusters without ingress capability

to:
> This field can be useful in clusters without ingress capability

To open the possibility for broader interpretation.

JoelSpeed: Posted in slack, copying here. I think we do want them to take precedence, and will need to document that within the godoc, something along the lines of
```
This will take precedence over any existing routes configured into the cluster.
Please make sure that appropriate ingress is configured at this URL before configuring the value.
Failure to do so may result in the degradation of several operators.
```

operator/v1/types_console.go
```diff
@@ -375,6 +380,34 @@ const (
...
+	// This field is required for clusters without ingress capability,
+	// where access to routes is not possible.
+	// Make sure that appropriate ingress is set up at this URL before assigning the value,
+	// otherwise the console operator may go degraded.
+	// Must use the HTTPS scheme.
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="isURL(self)",message="console url must be a valid absolute URL"
+	// +kubebuilder:validation:XValidation:rule="url(self).getScheme() == 'https'",message="console url scheme must be https"
+	// +kubebuilder:validation:MaxLength=1024
+	ConsoleURL string `json:"consoleURL,omitempty"`
```
JoelSpeed, line 396: Since these two URLs (and the object itself) are part of a config API, we tend to prefer to get rid of the omitempty.

If you were to do that, you would have to prefix the CEL to avoid running the validation when the string has length 0.

Can you give this a go?

The benefit here is that there's improved visibility and end users will see the fields in the rendered config as empty values, so they can see that the option exists and maybe then want to populate it

alebedev87: > The benefit here is that there's improved visibility and end users will see the fields in the rendered config as empty values, so they can see that the option exists and maybe then want to populate it

What's wrong with the good old `oc explain`?

operator/v1/types_console.go
```diff
@@ -375,6 +380,34 @@ const (
...
 )
 
+// Ingress allows cluster admin to configure alternative ingress for the console.
+type Ingress struct {
+	// consoleURL is a URL to be used as the base console address.
+	// If not specified, the console route hostname will be used.
+	// This field is required for clusters without ingress capability,
+	// where access to routes is not possible.
+	// Make sure that appropriate ingress is set up at this URL before assigning the value,
+	// otherwise the console operator may go degraded.
```
wking, line 390: Feels like there's tension between being required for clusters without `Ingress` (with no timing notes) and needing a functional target.  I understand that the console operator wants a functional `consoleURL`.  And that when `Ingress` is enabled, it can provide that default console URL itself.  But if you're in a cluster without `Ingress`, you probably want to set `consoleURL` to point at where you expect the console to be (as soon as you know where that is), without worrying about delaying until you know it's working, right?  That way the console operator can check whether it's working for you, and let you know if you have more work to do.  Can we reword this (and the similar `clientDownloadsURL` wording) to something like:

> .... This field is required for clusters without Ingress capability, where access to Routes is not possible.  The console operator will monitor the target and may set its ClusterOperator Degraded=True if the target is unreachable for sufficiently long to represent a lower quality of service.

which lifts some "lower quality of service" wording from [the `Degraded` Godocs][1].  Alternatively, just leave it implicit that the console operator is on top of this, and stop after "where access to routes is not possible."?

[1]: https://github.com/openshift/api/blob/0314f31e77748a7bdfa7c02286193760dee85195/config/v1/types_cluster_operator.go#L175-L178

wking: I see some discussion of this wording around [here][1] (sorry, I hadn't gone through `Outdated` but still unresolved threads before posting 😅), so feel free to ignore if I'm just missing context that's already been hashed out.

[1]: https://github.com/openshift/api/pull/1887#discussion_r1608003797

alebedev87: > But if you're in a cluster without Ingress, you probably want to set consoleURL to point at where you expect the console to be (as soon as you know where that is), without worrying about delaying until you know it's working, right? That way the console operator can check whether it's working for you, and let you know if you have more work to do

Good point. Indeed, the admin cannot be sure the URL is working until it gives the custom URL to the console operator. Since the operator has to update the internal and public configs before the full flow can work. Let me think of how I can reflect this in the description.

alebedev87: Rephrased to:
```
    // consoleURL is a URL to be used as the base console address.
    // If not specified, the console route hostname will be used.
    // This field is required for clusters without ingress capability,
    // where access to routes is not possible.
    // Make sure that appropriate ingress is set up at this URL.
    // The console operator will monitor the URL and may go degraded
    // if it's unreachable for an extended period.                                                                                    
    // Must use the HTTPS scheme.
```

operator/v1/types_console.go
alebedev87: Rephrased again to reflect [Trevor's point](https://github.com/openshift/api/pull/1887#discussion_r1609449647):
```
    // consoleURL is a URL to be used as the base console address.
    // If not specified, the console route hostname will be used.
    // This field is required for clusters without ingress capability,
    // where access to routes is not possible.
    // Make sure that appropriate ingress is set up at this URL.
    // The console operator will monitor the URL and may go degraded
    // if it's unreachable for an extended period.                                                                                    
    // Must use the HTTPS scheme.
```

operator/v1/types_console.go
JoelSpeed: Sometimes a user may have to go through several levels of oc explain to understand what options are available. Printing the whole struct out in configuration APIs makes it clear that there are levels below what they might see that they had no idea they might even want to `oc explain`

alebedev87: Doesn't this look a little misleading? When the CEL validation wants the value to be an URL but the user sees an empty string because they omitted the field.

alebedev87:
Rebased from `master`. Should this fix `verify-client-go` job?

operator/v1/types_console.go
JoelSpeed: I would argue that users seeing an empty URL as allowed would indicate that this is an empty value that means nothing is going to happen no? It's only required to be a URL when it's length is non-zero I think is a reasonable validation no?

jhadvig:
@alebedev87 what golang version on your machine ?

JoelSpeed:
Working on client-go issue separately, it's not a PR issue

operator/v1/types_console.go
alebedev87: Done, including `spec.ingress`.

alebedev87:
`omitempty` removed from all the new fields. `clientDownloadsURL` description updated to match `consoleURL`.

operator/v1/types_console.go
wking: I don't have the ability to mark this thread resolved myself, but the c0622aa4e679d Godocs look good to me :bow:

jhadvig (approved):
/lgtm

JoelSpeed (commented):
/lgtm

PR #1912, opened 2024-05-30
OCPBUGS-26466: Disable PSA in non-previews 4.16

ibihim:
## What

Disable PSA in non-previews. Keep enabled in TechPreviewNoUpgrade and DevPreviewNoUpgrade

## Why

We didn't meet the threshold to enable it yet.


features/features.go
```diff
@@ -73,7 +73,7 @@ var (
 
 	FeatureGateOpenShiftPodSecurityAdmission = newFeatureGate("OpenShiftPodSecurityAdmission").
 							reportProblemsToJiraComponent("auth").
-							contactPerson("stlaz").
+							contactPerson("ibihim").
 							productScope(ocpSpecific).
 							enableIn(configv1.Default, configv1.DevPreviewNoUpgrade, configv1.TechPreviewNoUpgrade).
```
deads2k, line 78: needs to be changed here and regenerated.

ibihim:
PTAL @deads2k

deads2k:
to pass verify, you must modify https://github.com/openshift/api/blob/master/features/features.go#L78 and regenerate.

PR #1918, opened 2024-06-06
Insights types - unify duration validation

tremes:
This is to unify the duration's validation pattern used in the Insights API. I noticed this problem when the operator status failed to update with having duration value as `2m0s`


insights/v1alpha1/types_insights.go
```diff
@@ -171,7 +171,7 @@ type GathererStatus struct {
 	// lastGatherDuration represents the time spent gathering.
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:Type=string
-	// +kubebuilder:validation:Pattern="^([1-9][0-9]*(\\.[0-9]+)?(ns|us|µs|ms|s|m|h))+$"
+	// +kubebuilder:validation:Pattern="^(0|([0-9]+(\\.[0-9]+)?(ns|us|µs|μs|ms|s|m|h))+)$s"
```
JoelSpeed, line 174: s after the dollar does not look correct, and why do we now allow leading zeroes? Is that preferable?

tremes: ahh good catch with the `s` character. 

wrt leading zeroes...hmm I am not sure. Hopefully it's not needed. I was surprised by the `2m0s` value so not sure if there can be something like `0m46s`. Probably not right

tremes: It seems valid in Go to have duration like `0h54m` - https://play.golang.com/p/ahB1cojf5Rl

tremes:
@JoelSpeed would you please find some time for this one?

JoelSpeed:
If my regex reading skills are up to scratch, all existing values should still match this validation, and we are allowing more values, that we weren't before.

@tremes, did this regex come from somewhere? How confident are we that it matches the go duration validation? Perhaps we can add some tests?

tremes:
I've added few tests for this one and updated the regex a little bit. 

Please check @JoelSpeed, Thank you

operator/v1/types_insights.go
```diff
@@ -58,7 +58,7 @@ type GatherStatus struct {
 	// lastGatherDuration is the total time taken to process
 	// all gatherers during the last gather event.
 	// +optional
-	// +kubebuilder:validation:Pattern="^0|([1-9][0-9]*(\\.[0-9]+)?(ns|us|µs|ms|s|m|h))+$"
+	// +kubebuilder:validation:Pattern="^(0|([0-9]+(\\.[0-9]+)?(ns|us|µs|μs|ms|s|m|h))+)$"
```
tremes, line 61: Ohh I forgot to update this one.

insights/v1alpha1/types_insights.go
```diff
@@ -171,7 +171,7 @@ type GathererStatus struct {
 	// lastGatherDuration represents the time spent gathering.
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:Type=string
-	// +kubebuilder:validation:Pattern="^([1-9][0-9]*(\\.[0-9]+)?(ns|us|µs|ms|s|m|h))+$"
+	// +kubebuilder:validation:Pattern="^(([0-9]+(?:\\.[0-9]+)?(ns|us|µs|μs|ms|s|m|h))+)$"
```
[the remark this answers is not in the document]
tremes: Sorry I don't understand why. `0m59s` is valid Go duration so why wuold you limit it?

insights/v1alpha1/tests/datagathers.insights.openshift.io/InsightsOnDemandDataGather.yaml
```diff
@@ -257,4 +257,145 @@ tests:
...
+            lastGatherDuration: -0m5s
+            conditions:
+            - type: Gathered
+              status: "True"
+              reason: OK
+              message: "test message"
+              lastTransitionTime: "2024-12-01T08:04:21Z"
+      expectedStatusError: "Invalid value: \"-0m5s\": gatherers[0].lastGatherDuration in body should match '^(([0-9]+(?:\\.[0-9]+)?(ns|us|µs|μs|ms|s|m|h))+)$'"
+
+
```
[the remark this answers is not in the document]
tremes: Removed

insights/v1alpha1/types_insights.go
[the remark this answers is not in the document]
JoelSpeed: We don't presently allow it, and the bug is because of secondary units not allowing zeros right? 

Just a thought, doesn't necessarily need to be fixed, but if it were easy I'd have preferred not to allow leading zeros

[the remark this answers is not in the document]
tremes: OK I see. Honestly I am not that proficient in regular expressions. I have experimented a little bit and perhaps something like the following could work:
```
^(([1-9]\d*(\.\d+)?(h|m|s|ms|ns|us|µs))|(0(\.\d+)?(s|ms|ns|us|µs)))+$
```
The tricky thing is for example `0.2m` which is also a valid Go duration. So I think I would lean more towards the existing expression.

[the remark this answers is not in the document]
JoelSpeed: Ack, lets keep as it is then, thanks for checking

operator/v1/tests/insightsoperators.operator.openshift.io/AAA_ungated.yaml
```diff
@@ -14,3 +14,225 @@ tests:
...
+                message: "test message"
+                lastTransitionTime: "2024-12-01T08:04:21Z"
+      expected: |
+        apiVersion: operator.openshift.io/v1
+        kind: InsightsOperator
+        metadata:
+         name: cluster
+        spec:
+          logLevel: Normal
+          operatorLogLevel: Normal
```
JoelSpeed, line 237: I don't think you want this, given you have a status error expected 🤔

operator/v1/tests/insightsoperators.operator.openshift.io/AAA_ungated.yaml
```diff
@@ -14,3 +14,225 @@ tests:
...
+        status:
+          gatherStatus:
+            lastGatherDuration: 0h2m
+            gatherers:
+            - name: testGatherer
+              lastGatherDuration: 0m5s
```
JoelSpeed, line 109: This test almost seems to contradict a later test, can you please double check?

I think the inclusion of expected and expectedError is confusing the suite

tremes: Ahh good catch. Thank you. Those yamls....

tremes: Should be fixed & squashed.

insights/v1alpha1/types_insights.go
everettraven, line 174: Looks like there is a repeated instance of the `μs` units:
```suggestion
	// +kubebuilder:validation:Pattern="^(([0-9]+(?:\\.[0-9]+)?(ns|us|µs|ms|s|m|h))+)$"
```

tremes:
@JoelSpeed it's not. Fixed. Thanks @everettraven

insights/v1alpha1/types_insights.go
```diff
@@ -171,7 +171,7 @@ type GathererStatus struct {
 	// lastGatherDuration represents the time spent gathering.
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:Type=string
-	// +kubebuilder:validation:Pattern="^([1-9][0-9]*(\\.[0-9]+)?(ns|us|µs|ms|s|m|h))+$"
+	// +kubebuilder:validation:Pattern="^(([0-9]+(?:\\.[0-9]+)?(ns|us|μs|ms|s|m|h))+)$"
```
JoelSpeed, line 174: `μs` is highlighting as a change, are we sure there isn't a difference in unicodes here?

PR #1921, opened 2024-06-10
OCPNODE-2357: machineconfiguration: make crun as the default container runtime

sohankunkerkar:
Related to https://issues.redhat.com/browse/OCPNODE-2217

This commit sets crun as the default container runtime in OCP for 4.17


yuqi-zhang (approved):
Approving from the MCO perspective.

I guess this PR by itself doesn't affect anything, and the MCO PR will be where we do the testing?

PR #1927, opened 2024-06-11
OCPNODE-2358: Update (cluster)imagepolicy doc no restriction on release repo scopes

QiWang19:
https://github.com/openshift/machine-config-operator/pull/4401 implementation removes the restriction that `scopes` field can not be set for OCP release payload repo to support https://github.com/openshift/enhancements/pull/1633.
Remove the godoc comment from (Cluster)imagepolicy docs to stay compatible with the implementation of this change.


QiWang19:
@JoelSpeed @deads2k  could you review? The documented restriction was removed from the implementation to support https://github.com/openshift/enhancements/pull/1633. This PR will keep the doc up to date.

QiWang19:
@mtrmac PTAL

QiWang19:
@mtrmac Added the doc about the longest path match and warning information about the openshift repository. 
Could you review it?

config/v1alpha1/types_cluster_image_policy.go
```diff
@@ -38,8 +38,8 @@ type ClusterImagePolicySpec struct {
...
 	// namespace, or a registry host (by only specifying the host name and possibly a port number) or a wildcard expression starting with `*.`, for matching all subdomains (not including a port number).
 	// Wildcards are only supported for subdomain matching, and may not be used in the middle of the host, i.e.  *.example.com is a valid case, but example*.*.com is not.
-	// Please be aware that the scopes should not be nested under the repositories of OpenShift Container Platform images.
-	// If configured, the policies for OpenShift Container Platform repositories will not be in effect.
+	// If multiple scopes match a given image, only the policy requirements for the most specific scope apply. The policy requirements for more general scopes are ignored.
+	// Set this field to OpenShift image repositories quay.io/openshift-release-dev/ocp-release, quay.io/openshift-release-dev/ocp-v4.0-art-dev, or a super scope of these repositories like quay.io or quay.io/openshift-release-dev. This would lead to pod failure with SignatureValidationFailed if the policy requirements are not met.
```
wking, line 42: nit: "Set this field..." might be read as excluding other scopes, but I don't think that's your intention.  Can we rephrase like:

> Setting this field to OpenShift release image repositories quay.io/openshift-release-dev/ocp-release, quay.io/openshift-release-dev/ocp-v4.0-art-dev, or a super scope of these repositories like quay.io or quay.io/openshift-release-dev, will lead to SignatureValidationFailed pod failures for core system pods if the policy requirements are not met.

or something?

mtrmac: To play a bit of a devil’s advocate, I’m not even 100% sure that we need an OpenShift repo-specific warning at all; this boils down to “users who set a too restrictive policy will see deployments fail”, which is true for all repos.

Especially the `SignatureValidationFailed` detail seems a bit superfluous — or is it intentionally included to allow full text searches of that error to find this warning? That might make sense.

---

(In the future, after https://github.com/openshift/enhancements/pull/1633 fully lands, there will be a policy set on (some of) the repos we care about, so users setting a policy on a more general scope (e.g. all of `quay.io`) will not break anything any more. But we are not quite there yet.)

---

Maybe this should say something like
> In addition to setting a policy appropriate for your own deployed applications, make sure that a policy on the OpenShift image repositories quay.io/openshift-release-dev/ocp-release, quay.io/openshift-release-dev/ocp-v4.0-art-dev (or on a more general scope scope) allows deployment of the OpenShift images required for cluster operation.

mtrmac (commented):
The note about matching the most specific scope is definitely valuable.

config/v1alpha1/types_cluster_image_policy.go
QiWang19: @wking @mtrmac Thanks, updated.

wking (approved):
/lgtm

PR #1929, opened 2024-06-12
MON-3902: add initial Monitoring CRD api

marioferh:
Initial Implementation of Configuration CRD for the Cluster Monitoring Operator

Related: Enhancements Proposal [[link]](https://github.com/openshift/enhancements/pull/1627)

As we have a FeatureGate, the idea is to implement component by component to verify the correct behavior and facilitate the reviews.

The API comes from https://github.com/openshift/cluster-monitoring-operator/blob/7015893ece5ed0a6f888fffe09b6d10a807d6901/pkg/manifests/config.go#L51


config/v1/types_cmo.go
```diff
@@ -0,0 +1,71 @@
...
+	Items           []ClusterMonitoring `json:"items"`
+}
+
+// ClusterMonitoringOperatorSpec defines the desired state of ClusterMonitoringOperator
+type ClusterMonitoringSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// `UserWorkloadEnabled` is a Boolean flag that enables monitoring for user-defined projects.
+	UserWorkloadEnabled *bool `json:"enableUserWorkload,omitempty"`
```
marioferh, line 69: verify is failing due to:
```
F0612 18:28:37.679340 3391636 root.go:64] Error running codegen:                                                                                                
                error in config/v1/zz_generated.crd-manifests/clustermonitoring.crd.yaml: NoBools: crd/clustermonitoring.config.openshift.io version/v1 field/^.
spec.enableUserWorkload may not be a boolean                                                                                                                    
                error in config/v1/zz_generated.featuregated-crd-manifests/clustermonitoring.config.openshift.io/AAA_ungated.yaml: NoBools: crd/clustermonitorin
g.config.openshift.io version/v1 field/^.spec.enableUserWorkload may not be a boolean                                                                           
+ echo This verifier checks all files that have changed. In some cases you may have changed or renamed a file that already contained api violations, but you are
 not introducing a new violation. In such cases it is appropriate to /override the failing CI job.                                                              
This verifier checks all files that have changed. In some cases you may have changed or renamed a file that already contained api violations, but you are not in
troducing a new violation. In such cases it is appropriate to /override the failing CI job.                                                                     
make: *** [Makefile:65: verify-crd-schema] Error 255        
```                               

https://github.com/openshift/api/blob/5c2b46e4709a950e98cda1f3733f9a46cd9737cd/tools/vendor/github.com/openshift/crd-schema-checker/pkg/manifestcomparators/comp_no_bools.go#L20

how can we fix it?

marioferh: nevermind, add a enum as in https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#do-not-use-boolean-fields

config/v1/types_cmo.go
```diff
@@ -0,0 +1,86 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=clustermonitoring,scope=Cluster
+// +kubebuilder:subresource:status
+// +openshift:api-approved.openshift.io=https://github.com/openshift/api/pull/1929
+// Compatibility level 4: No compatibility is provided, the API can change at any point for any reason. These capabilities should not be used by applications needing long term support.
+// +openshift:compatibility-gen:internal
+// ClusterMonitoring is the Schema for the clustermonitoringoperators API
+type ClusterMonitoring struct {
```
marioferh, line 34: @simonpasquier we can go with this CRD name?

config/v1/types_monitoring.go
```diff
@@ -0,0 +1,86 @@
...
+	// +required
+	Spec MonitoringSpec `json:"spec,omitempty"`
+	// status holds observed values from the cluster. They may not be overridden.
+	// +optional
+	Status MonitoringStatus `json:"status,omitempty"`
+}
+
+// MonitoringOperatorStatus defines the observed state of MonitoringOperator
+type MonitoringStatus struct {
+	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
```
simonpasquier, line 49: should we remove this?

config/v1/types_monitoring.go
```diff
@@ -0,0 +1,86 @@
...
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// `UserWorkload` enables monitoring for user-defined projects.
+	// +kubebuilder:validation:Enum:="Enabled";"Disabled";
+	// +kubebuilder:default:="Disabled"
+	// +kubebuilder:validation:Optional
+	// +default="Disabled"
+	// +optional
+
+	UserWorkload UserWorkload `json:"enableUserWorkload,omitempty"`
```
simonpasquier, line 75: We could have a top-level field to group settings pertaining to the user-defined monitoring stack which are controlled by the cluster admins. We don't have right now but might in the future (for instance for cross-namespace rules).

````
userWorkload:
  type: Enabled|Disabled
  foo: ...
````

File: config/v1/types_monitoring.go [hunk 5]
```diff
@@ -0,0 +1,86 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=monitoring,scope=Cluster
+// +kubebuilder:subresource:status
+// +openshift:api-approved.openshift.io=https://github.com/openshift/api/pull/1929
+// Compatibility level 4: No compatibility is provided, the API can change at any point for any reason. These capabilities should not be used by applications needing long term support.
+// +openshift:compatibility-gen:internal
+// Monitoring is the Schema for the clustermonitoringoperators API
+type Monitoring struct {
```
simonpasquier, line 34: The first version had ClusterMonitoring?

marioferh: Yep sorry, fixing it

config/v1/types_monitoring.go
jan--f: The status field seems like it could be useful. Was Simon maybe referring to the comment?

config/v1/types_monitoring.go
jan--f: +1 for this. We have used this pattern in the configmap already.

simonpasquier: @jan--f I'm not very happy with `type` though. Would you have a better idea?

config/v1/types_monitoring.go
simonpasquier: I was indeed referring to the comment lines only. A status subresource might be useful eventually but not sure what it would carry...

config/v1/types_monitoring.go
simonpasquier: Discussed offline with @marioferh, the status subresource could be used to expose the different route URLs (e.g. Thanos Q, Prometheus, Alertmanager, ...).

config/v1/types_monitoring.go
marioferh: done

config/v1/types_monitoring.go
jan--f: Since we want the option to specify things beyond `enabled|disabled` maybe we call the field `Config`. Initially this can have values `Disabled|SingleNamespace`. This would leave the option to add additional "modes' of UWM later. Like so maybe?
```
...
    UserWorkload UserWorkloadConfig
...

type UserWorkloadConfig struct {
    Config UWMMode
}
```
Certainly not ideal naming either but maybe a bit better then `type`?

marioferh: Something like this better?
`
// MonitoringOperatorSpec defines the desired state of MonitoringOperator
type ClusterMonitoringSpec struct {
	// `UserWorkload` enables monitoring for user-defined projects.
	// +kubebuilder:validation:Enum:="Enabled";"Disabled";
	// +kubebuilder:default:="Disabled"
	// +default="Disabled"
	// +optional

	UserWorkload UserWorkload `json:"enableUserWorkload,omitempty"`
}

// UserWorkloadConfig is an enumeration that enable or disable monitoring of user-defined projects.
type UserWorkload struct {
	Config UserWorkloadMode
	// To add more UWM config
}

type UserWorkloadMode string

var (
	// UserWorkloadEnabled enables monitoring for user-defined projects.
	UserWorkloadEnabled UserWorkloadMode = "Enabled"
	// UserWorkloadDisabled disables monitoring for user-defined projects.
	UserWorkloadDisabled UserWorkloadMode = "Disabled"
)
`

danielmellado: Hmm I'd go with Jan's suggestion here and avoid having a type for enabled|disabled. That would also allow us to have a field for future enhancing of UWM modes. I think that'd work Mario. Just my .02 is to leave this really well defined in either enhancement or comments or the future us would totally forget.

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+type UserDefinedMode string
+
+var (
+	// UserWorkloadEnabled enables monitoring for user-defined projects.
+	UserDefinedDisabled UserDefinedMode = "Disabled"
+	// UserWorkloadDisabled disables monitoring for user-defined projects.
+	UserDefinedNamespace UserDefinedMode = "SingleNamespace"
+	// To add more UWM config
```
jan--f, line 84:
```suggestion
```

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,85 @@
...
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []ClusterMonitoring `json:"items"`
+}
+
+// MonitoringOperatorSpec defines the desired state of MonitoringOperator
+type ClusterMonitoringSpec struct {
+	// `UserWorkload` enables monitoring for user-defined projects.
+	// +kubebuilder:validation:Optional
+	// +optional
+
```
jan--f, line 68:
```suggestion
```

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,85 @@
...
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=clustermonitoring,scope=Cluster
+// +kubebuilder:subresource:status
+// +kubebuilder:metadata:annotations=release.openshift.io/bootstrap-required=true
+// ClusterMonitoring is the Schema for the clustermonitoringoperators API
+type ClusterMonitoring struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// spec holds user settable values for configuration
```
jan--f, line 40:
```suggestion
	// spec holds user configuration for the Cluster Monitoring Operator
```
Maybe a bit more explicit wording?

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,85 @@
...
+
+// Compatibility level 4: No compatibility is provided, the API can change at any point for any reason. These capabilities should not be used by applications needing long term support.
+// +openshift:compatibility-gen:internal
+// +openshift:api-approved.openshift.io=https://github.com/openshift/api/pull/1929
+// +openshift:file-pattern=cvoRunLevel=0000_10,operatorName=config-operator,operatorOrdering=01
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=clustermonitoring,scope=Cluster
+// +kubebuilder:subresource:status
+// +kubebuilder:metadata:annotations=release.openshift.io/bootstrap-required=true
+// ClusterMonitoring is the Schema for the clustermonitoringoperators API
```
jan--f, line 35:
```suggestion
// ClusterMonitoring is the Schema for the Cluster Monitoring Operators API
```

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,85 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// Compatibility level 4: No compatibility is provided, the API can change at any point for any reason. These capabilities should not be used by applications needing long term support.
+// +openshift:compatibility-gen:internal
+// +openshift:api-approved.openshift.io=https://github.com/openshift/api/pull/1929
+// +openshift:file-pattern=cvoRunLevel=0000_10,operatorName=config-operator,operatorOrdering=01
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=clustermonitoring,scope=Cluster
+// +kubebuilder:subresource:status
+// +kubebuilder:metadata:annotations=release.openshift.io/bootstrap-required=true
```
jan--f, line 34:
```suggestion
// +kubebuilder:metadata:annotations=release.openshift.io/bootstrap-required=true
// +kubebuilder:metadata:annotations="description=Cluster Monitoring Operators configuration API"
```

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,85 @@
...
+
+// Compatibility level 4: No compatibility is provided, the API can change at any point for any reason. These capabilities should not be used by applications needing long term support.
+// +openshift:compatibility-gen:internal
+type ClusterMonitoringList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []ClusterMonitoring `json:"items"`
+}
+
+// MonitoringOperatorSpec defines the desired state of MonitoringOperator
```
jan--f, line 63:
```suggestion
// MonitoringOperatorSpec defines the desired state of Cluster Monitoring Operator
```

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,85 @@
...
+type ClusterMonitoringSpec struct {
+	// `UserWorkload` enables monitoring for user-defined projects.
```
jan--f, line 65: This doesn't seem accurate anymore. After we discussed to use this to allow for future UWM modes, I think we need to rewrite this. Maybe like so?
```suggestion
	// `UserWorkload` set the deployment mode for user-defined monitoring.
```

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,85 @@
...
+
+// UserDefinedMonitoring config for user-defined projects.
+type UserDefinedMonitoring struct {
+	Mode UserDefinedMode `json:"mode,omitempty"`
+}
+
+type UserDefinedMode string
+
+var (
+	// UserWorkloadEnabled enables monitoring for user-defined projects.
```
jan--f, line 80:
```suggestion
	// UserDefinedDisabled disables monitoring for user-defined projects.
```

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,85 @@
...
+	UserDefinedDisabled UserDefinedMode = "Disabled"
+	// UserWorkloadDisabled disables monitoring for user-defined projects.
```
jan--f, line 82: This needs changing...how about something along the lines of:
```suggestion
	// UserDefinedNamespace enables monitoring for user-defined projects with namespace scoped tenancy.
```

jan--f (changes requested):
Some changes are needed here. I added a couple of suggestions as well. While we will go over this when things move to `v1` once more in detail, it's worth getting details right from the get go too I think.

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,84 @@
...
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []ClusterMonitoring `json:"items"`
+}
+
+// MonitoringOperatorSpec defines the desired state of Cluster Monitoring Operator
+type ClusterMonitoringSpec struct {
+	// `UserWorkload` set the deployment mode for user-defined monitoring.
+	// +kubebuilder:validation:Optional
+	// +optional
+	UserDefined UserDefinedMonitoring `json:"userDefinedMonitoring,omitempty"`
```
jan--f, line 69: Apologies for missing this before: With `UserDefined` being optional we currently essentially have three scenarios: `UserDefined` is absent, set to `Disabled` or set to `SingleNamespace`.
Unless I'm missing something, the first two are equivalent, we should probably have one way to disable UWM.
I seem to remember we went with a `Mode` field since we want to give users the option to disable UWM without deleting their (as of yet nonexistent) UWM config beyond mode. With that line of thinking we should maybe make this required and default to `{Mode: "Disabled"}`?

marioferh: Yep I think best idea is maybe do it `required`  and default `Disabled`, just in configmap is "optional" and "optional" means "disabled", for that reason I marked it as optional. After that we have extended the options...

```
	// `UserWorkloadEnabled` is a Boolean flag that enables monitoring for user-defined projects.
	UserWorkloadEnabled *bool `json:"enableUserWorkload,omitempty"`
```

marioferh: https://github.com/kubernetes-sigs/controller-tools/issues/893#issuecomment-1991256368

jan--f: Right, I think ran into this before. I guess `optional` plus a default disabled it is then? :) 
Mostly I want to avoid having two ways of disabling UWM (either omit the config _or_ set `mode: disabled`).

marioferh: Yep I am on it but facing some issues with tests

marioferh: When added required [config/v1/tests/clustermonitoring.config.openshift.io/AAA_ungated.yaml](url) has to be changed manually... that was the issue

jan--f:
/lgtm
only `verify` is now unhappy. Guess needs some generate step again or so?

marioferh:
> /lgtm only `verify` is now unhappy. Guess needs some generate step again or so?

something related with CustomNoUpgrade:
```
Found untracked file 0000_10_config-operator_01_clustermonitoring-CustomNoUpgrade.crd.yaml in payload CRD manifests.  Please add the file to crd_globs in hack/update-payload-crds.sh.
make: *** [Makefile:52: verify-scripts] Error 1

```

marioferh:
@JoelSpeed

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,88 @@
...
+// +openshift:compatibility-gen:internal
+type ClusterMonitoringList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []ClusterMonitoring `json:"items"`
+}
+
+// MonitoringOperatorSpec defines the desired state of Cluster Monitoring Operator
+type ClusterMonitoringSpec struct {
+	// `UserWorkload` set the deployment mode for user-defined monitoring.
```
JoelSpeed, line 67: What is this actually doing? What is user defined monitoring?

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,88 @@
...
+	// +optional
+	// +kubebuilder:default={"mode": "Disabled"}
```
JoelSpeed, line 69: We don't typically use defaults in configuration APIs unless we are certain we will never change the defaults, I would probably default this controller side rather than here.

We typically have a common pattern of saying "When omitted, this means no opinion and the platform is left to choose a reasonable default, which is subject to change over time. The current default is XXX"

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,88 @@
...
+	UserDefined UserDefinedMonitoring `json:"userDefined,omitempty"`
+}
+
+// UserDefinedMonitoring config for user-defined projects.
+type UserDefinedMonitoring struct {
+	// +kubebuilder:default="Disabled"
+	// +default:="Disabled"
+	// +kubebuilder:validation:Enum:="Disabled";"SingleNamespace"
+	Mode UserDefinedMode `json:"mode,omitempty"`
```
JoelSpeed, line 78: No godoc on this field? Please explain what the field is, what it does, what valid values are etc

JoelSpeed (commented):
Please review https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#write-user-readable-documentation-in-godoc and update the godocs based on the guidance there

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,88 @@
...
+// +openshift:api-approved.openshift.io=https://github.com/openshift/api/pull/1929
+// +openshift:file-pattern=cvoRunLevel=0000_10,operatorName=config-operator,operatorOrdering=01
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=clustermonitoring,scope=Cluster
+// +kubebuilder:subresource:status
+// +kubebuilder:metadata:annotations=release.openshift.io/bootstrap-required=true
+// +kubebuilder:metadata:annotations="description=Cluster Monitoring Operators configuration API"
+// +openshift:enable:FeatureGate=ClusterMonitoringConfig
+// ClusterMonitoring is the Schema for the Cluster Monitoring Operators API
+type ClusterMonitoring struct {
```
JoelSpeed, line 38: Is it expected that there will be exactly one of these in the cluster with a particular name?

config/v1/types_cluster_monitoring.go
danielmellado: Hi, User Workload Monitoring is monitoring for user-defined Projects. This is also mentioned on the [OpenShift documentation](https://docs.openshift.com/container-platform/4.16/observability/monitoring/enabling-monitoring-for-user-defined-projects.html) but IMHO it'd be a bit overkill to link that within the api comments.

config/v1/types_cluster_monitoring.go
danielmellado: Agree to that as changing this within the controller side will be done more easily @marioferh

config/v1/types_cluster_monitoring.go
JoelSpeed: Yeah but we need some sort of description, at the moment someone doing an `oc explain` on this would have no cluster what this field controls

danielmellado: oh, I see, then we could use something along the lines of 

> Extension to the existing cluster monitoring stack to enable observability of user namespaces

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,88 @@
...
+	// spec holds user configuration for the Cluster Monitoring Operator
+	// +kubebuilder:validation:Required
+	// +required
+	Spec ClusterMonitoringSpec `json:"spec,omitempty"`
+	// status holds observed values from the cluster. They may not be overridden.
+	// +optional
+	Status ClusterMonitoringStatus `json:"status,omitempty"`
+}
+
+// MonitoringOperatorStatus defines the observed state of MonitoringOperator
```
danielmellado, line 51: Here the comment speaks about "MonitoringOperatorStatus" where we're speaking about "ClusterMonitoringStatus", and also applies to a few other types. Could you make this match so it's correct. I'd also document the fields if possible as Joel suggests.

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,88 @@
...
+	// +kubebuilder:validation:Enum:="Disabled";"SingleNamespace"
+	Mode UserDefinedMode `json:"mode,omitempty"`
+}
+
+type UserDefinedMode string
+
+const (
+	// UserDefinedDisabled disables monitoring for user-defined projects.
+	UserDefinedDisabled UserDefinedMode = "Disabled"
+	// UserDefinedNamespace enables monitoring for user-defined projects with namespace scoped tenancy.
```
simonpasquier, line 86: "namespace scoped tenancy" is too vague for users. We'd need to go in greater details (not sure if the reference docs are doing a better job).

config/v1/types_cluster_monitoring.go
jan--f: yes indeed, same as it works currently where CMO only looks at a configmap with a predefined name

config/v1/types_cluster_monitoring.go
jan--f: In a recent discussion you mentioned `isolation` in the context of this. I thought that might lead to a better name. Perhaps `UserDefinedNamespaceIsolation UserDefinedMode = "NamespaceIsolation"`

config/v1/types_cluster_monitoring.go
marioferh: done

config/v1/types_cluster_monitoring.go
marioferh: done, I will implement in the controller side. Thanks

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,88 @@
...
+// ClusterMonitoring is the Custom Resource object which holds the current status of Cluster Monitoring Operator. CMO is a central component of the monitoring stack.
+//
+// Compatibility level 4: No compatibility is provided, the API can change at any point for any reason. These capabilities should not be used by applications needing long term support.
+// +openshift:compatibility-gen:internal
+// +openshift:api-approved.openshift.io=https://github.com/openshift/api/pull/1929
+// +openshift:file-pattern=cvoRunLevel=0000_10,operatorName=config-operator,operatorOrdering=01
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=clustermonitoring,scope=Cluster
+// +kubebuilder:subresource:status
+// +kubebuilder:metadata:annotations=release.openshift.io/bootstrap-required=true
```
JoelSpeed, line 36: Is this required for the bootstrap process? Bootstrap required implies that the cluster bootstrap node can't complete without this, I suspect this is not the case for this API

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,88 @@
...
+// +kubebuilder:metadata:annotations="description=Cluster Monitoring Operators configuration API"
+// +openshift:enable:FeatureGate=ClusterMonitoringConfig
+// ClusterMonitoring is the Schema for the Cluster Monitoring Operators API
+type ClusterMonitoring struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// spec holds user configuration for the Cluster Monitoring Operator
+	// +kubebuilder:validation:Required
+	// +required
```
JoelSpeed, line 46: You only need one of these, please use the kubebuilder one for now
```suggestion
	// +kubebuilder:validation:Required
```

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,88 @@
...
+// +openshift:compatibility-gen:internal
+type ClusterMonitoringList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []ClusterMonitoring `json:"items"`
+}
+
+// ClusterMonitoringSpec defines the desired state of Cluster Monitoring Operator
+type ClusterMonitoringSpec struct {
+	// `UserWorkload` set the deployment mode for user-defined monitoring in addition to the default platform monitoring.
```
JoelSpeed, line 69: This should be fleshed out a bit more about the kinds of configuration that can be in here.

Also, please start godocs with the json tag (serialised) version of the name

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,88 @@
...
+	// +optional
+	UserDefined UserDefinedMonitoring `json:"userDefined,omitempty"`
+}
+
+// UserDefinedMonitoring config for user-defined projects.
+type UserDefinedMonitoring struct {
+	// mode defines the different configurations of UserDefinedMontiring
+	// +kubebuilder:validation:Enum:="Disabled";"SingleNamespace"
+	Mode UserDefinedMode `json:"mode,omitempty"`
```
JoelSpeed, line 78: You need to expand the godoc to explain what the different available options are, and what they achieve

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,88 @@
...
+type ClusterMonitoringSpec struct {
+	// `UserWorkload` set the deployment mode for user-defined monitoring in addition to the default platform monitoring.
+	// +optional
+	UserDefined UserDefinedMonitoring `json:"userDefined,omitempty"`
+}
+
+// UserDefinedMonitoring config for user-defined projects.
+type UserDefinedMonitoring struct {
+	// mode defines the different configurations of UserDefinedMontiring
+	// +kubebuilder:validation:Enum:="Disabled";"SingleNamespace"
```
JoelSpeed, line 77: This enum doesn't match the values you created in the consts

config/v1/types_cluster_monitoring.go
JoelSpeed: Please flesh out and update the godoc here, need to explain what this actually does

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,86 @@
...
+	// +kubebuilder:validation:Enum:="Disabled";"NamespaceIsolation"
+	Mode UserDefinedMode `json:"mode,omitempty"`
+}
+
+type UserDefinedMode string
+
+const (
+	// UserDefinedDisabled disables monitoring for user-defined projects. Monitoring is limited to platform namespaces only, using the default monitoring stack installed in the openshift-monitoring project, preventing any custom monitoring configurations or resources from being applied to user-defined namespaces.
+	UserDefinedDisabled UserDefinedMode = "Disabled"
+	// UserDefinedNamespaceIsolation enables monitoring for user-defined projects with namespace-scoped tenancy, ensuring that metrics, alerts, and monitoring data are segregated at namespace level.
```
eromanova97, line 84:
```suggestion
	// UserDefinedNamespaceIsolation enables monitoring for user-defined projects with namespace-scoped tenancy. This ensures that metrics, alerts, and monitoring data are isolated at the namespace level.
```

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,86 @@
...
+// +openshift:compatibility-gen:internal
+type ClusterMonitoringList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []ClusterMonitoring `json:"items"`
+}
+
+// ClusterMonitoringSpec defines the desired state of Cluster Monitoring Operator
+type ClusterMonitoringSpec struct {
+	// userDefined set the deployment mode for user-defined monitoring in addition to the default platform monitoring.
```
JoelSpeed, line 67: If this is not present, what does it mean?

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,86 @@
...
+	// +optional
+	UserDefined UserDefinedMonitoring `json:"userDefined,omitempty"`
+}
+
+// UserDefinedMonitoring config for user-defined projects.
+type UserDefinedMonitoring struct {
+	// mode defines the different configurations of UserDefinedMonitoring
```
JoelSpeed, line 74: You need to expand this godoc to explain valid values and what they mean

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,86 @@
...
+	// +kubebuilder:validation:Enum:="Disabled";"NamespaceIsolation"
+	Mode UserDefinedMode `json:"mode,omitempty"`
```
JoelSpeed, line 76: If you remove `omitempty`, than a program marshalling this will render the field names and the empty string. This will aid in discover of the configuration field.
```suggestion
	// +kubebuilder:validation:Enum:="Disabled";"NamespaceIsolation";""
	Mode UserDefinedMode `json:"mode"`
```

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+type UserDefinedMode string
+
+const (
+	// UserDefinedDisabled disables monitoring for user-defined projects. This restrics the default monitoring stack, installed in the openshift-monitoring project, to monitor only platform namespaces, which prevents any custom monitoring configurations or resources from being applied to user-defined namespaces.
+	UserDefinedDisabled UserDefinedMode = "Disabled"
```
JoelSpeed, line 83: Does this supersede the presence of the user monitoring CRs that you plan to add? Could we validate somehow that disabling this disallows the creation of those CRs? And this cannot be disabled if user monitoring CRs are present?

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,86 @@
...
+// +kubebuilder:metadata:annotations="description=Cluster Monitoring Operators configuration API"
+// +openshift:enable:FeatureGate=ClusterMonitoringConfig
+// ClusterMonitoring is the Schema for the Cluster Monitoring Operators API
+type ClusterMonitoring struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// spec holds user configuration for the Cluster Monitoring Operator
+	// +kubebuilder:validation:Required
+	Spec ClusterMonitoringSpec `json:"spec,omitempty"`
```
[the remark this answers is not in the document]
marioferh: done

config/v1/types_cluster_monitoring.go
marioferh: at the begging was optional and default disabled, but there were some issues with mandatoryf ields,  at this point I think we can move it to compulsory and select between disabled and NamespaceIsolation, but in ConfigMap default is Disabled.

config/v1/types_cluster_monitoring.go
marioferh: I don't understand this, every field is defined in the const right?
I can see in other apis something like: 
`// Type sets the type of...` https://github.com/openshift/api/blob/232984653385c77e0d6bb8c3d4ec6007de618ace/apps/v1/types.go#L120
Should I define twice?

config/v1/types_cluster_monitoring.go
marioferh: If this is disabled the user-defined projects are not enabled. The CR could be there just its not taking the config to create user-defined projects. As in the current ConfigMap we can switch the mode and activate user-defined projects.

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,86 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ClusterMonitoring is the Custom Resource object which holds the current status of Cluster Monitoring Operator. CMO is a central component of the monitoring stack.
+//
+// Compatibility level 4: No compatibility is provided, the API can change at any point for any reason. These capabilities should not be used by applications needing long term support.
+// +openshift:compatibility-gen:internal
+// +openshift:api-approved.openshift.io=https://github.com/openshift/api/pull/1929
+// +openshift:file-pattern=cvoRunLevel=0000_10,operatorName=config-operator,operatorOrdering=01
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=clustermonitoring,scope=Cluster
```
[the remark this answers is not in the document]
marioferh: It is used in HyperShift

config/v1/types_cluster_monitoring.go
JoelSpeed: This documentation (and not the documentation on the constants) ends up in the API schema.

So you typically need to describe values and validations in the godoc so that users have a chance of understanding the API by using `oc explain`.

I'm expecting something like:
```
mode defines the configuration of the UserDefinedMonitoring.
Valid values are A and B.
A means that we do one thing.
B means that we do another thing.
```

marioferh: done thanks

config/v1/types_cluster_monitoring.go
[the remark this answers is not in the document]
JoelSpeed: Do you know on which side it is used? Is it on the management or guest cluster side?

config/v1/types_cluster_monitoring.go
JoelSpeed: Up to you, you could default it if you prefer, though we tend to avoid defaulting configuration APIs. What we typically do if there is a default is allow the value to be omitted and write something like
```
// Valid values are Disabled, NamespaceIsolation and omitted.
// When omitted this means no opinion and the platform is left to choose a reasonable default, which is subject to change over time.
// The current default is Disabled.
```

config/v1/types_cluster_monitoring.go
JoelSpeed: > The CR could be there just its not taking the config to create user-defined projects.

Would this be surprising to a user who has configured their monitoring, how do they understand that the monitoring is not currently enabled, is something going to tell them?

config/v1/types_cluster_monitoring.go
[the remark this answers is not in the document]
marioferh: Not sure about it.  @simonpasquier do you have this info?

config/v1/types_cluster_monitoring.go
marioferh: We declared it as mandatory, so the user has to add the field.

config/v1/types_cluster_monitoring.go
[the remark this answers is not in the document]
jan--f: Iiuc, this should be used on the management cluster side, as this will eventually contain the settings for cluster admins. The guest clusters would rather use the `UserDefinedMonitoring` struct.

jan--f:
@JoelSpeed Just fyi I'll take this over in @marioferh's absence.

config/v1/types_cluster_monitoring.go
jan--f: When we add the user defined monitoring CRD, we'll communicate in its status if user defined monitoring is enabled or not.
I tihnk this is better then preventing its creation. The prevention path would raise the question of what to do with existing user defined objects when user defined is disabled. We wouldn't want to delete them (disbaling them might be temporary) so preventing addition (and editing) seems like not the best solution.

jan--f:
@JoelSpeed PTAL?

config/v1/types_cluster_monitoring.go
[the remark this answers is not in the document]
JoelSpeed: I think we need to have a discussion with the HyperShift folks about this API then.

If this is cluster scoped, then it makes sense in the OCP world.

If it is namespace scoped, the it makes sense in HyperShift.

The converse of both of my previous statements are not true.

To be correct with our APIs, it would be preferable to have a `Hosted<CRD>` that is namespaced, and exists only in the management cluster, and then have no semblance of the feature exist within the guest cluster. 

This pattern is coming up a lot at the moment with new designs for CRDs, will need to come up with a common path forward

CC @enxebre

config/v1/types_cluster_monitoring.go
JoelSpeed: What is the use case for being able to disable?

Could you prevent disabling, unless there were no user defined CRDs?

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,88 @@
...
+// ClusterMonitoringSpec defines the desired state of Cluster Monitoring Operator
+type ClusterMonitoringSpec struct {
+	// userDefined set the deployment mode for user-defined monitoring in addition to the default platform monitoring.
+	UserDefined UserDefinedMonitoring `json:"userDefined,omitempty"`
+}
+
+// UserDefinedMonitoring config for user-defined projects.
+type UserDefinedMonitoring struct {
+	// mode defines the different configurations of UserDefinedMonitoring
+	// Valid values are UserDefinedDisabled and UserDefinedNamespaceIsolation
```
JoelSpeed, line 74: The valid values should be the yaml, not Go versions
```suggestion
	// Valid values are Disabled and NamespaceIsolation.
```

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,88 @@
...
+	Mode UserDefinedMode `json:"mode,omitempty"`
+}
+
+type UserDefinedMode string
+
+const (
+	// UserDefinedDisabled disables monitoring for user-defined projects. This restricts the default monitoring stack, installed in the openshift-monitoring project, to monitor only platform namespaces, which prevents any custom monitoring configurations or resources from being applied to user-defined namespaces.
+	UserDefinedDisabled UserDefinedMode = "Disabled"
+	// UserDefinedNamespaceIsolation enables monitoring for user-defined projects with namespace-scoped tenancy. This ensures that metrics, alerts, and monitoring data are isolated at the namespace level.
+	UserDefinedNamespaceIsolation UserDefinedMode = "NamespaceIsolation"
```
JoelSpeed, line 87: Should this really be `NamespaceIsolated`?

jan--f: I don't feel strongly about this but I'd argue the current `NamespaceIsolation` is better. Its not the namespace that is isloated but the metrics access that is isolated by namespace values.

JoelSpeed: > Its not the namespace that is isloated but the metrics access that is isolated by namespace values.

You've used the word `isolated` to describe how the metrics are accessed here, and I think to you're point, if it were `IsolatedNamespace`, I'd agree that this would imply the namespace is isolated, but my suggestions was `NamespaceIsolated`

I'm fine with either, but my thought was that `NamespaceIsolated` felt more declarative, and inline with `Disabled`. Isolation is a process, Isolated is a state, and it's the state that declarative APIs are supposed to represent

jan--f: Yeah fair point, I'm convinced. We also use `isolated at the namespace level` in the doc comments.

danielmellado: yeah, it'll probably make more sense to be coherent all along the doc comments and code.

config/v1/types_cluster_monitoring.go
jan--f: The use case is that the decision whether to enable it lies with the cluster admin. The actual configuration can be done by a non-admin user. We think its useful for admins to be able to (perhaps temporarily) disable user defined monitoring without having to remove the config owned by a separate user group.

config/v1/types_cluster_monitoring.go
marioferh: done

config/v1/types_cluster_monitoring.go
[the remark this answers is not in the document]
marioferh: any news about it?

config/v1/types_cluster_monitoring.go
JoelSpeed: If the cluster admin has disabled these resources, what signal will end users get to show that these are disabled? How will they be told that the configuration they have applied is now inert?

config/v1/types_cluster_monitoring.go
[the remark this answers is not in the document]
JoelSpeed: Having similar discussions in https://github.com/openshift/enhancements/pull/1571 about using a Hosted version of a CRD to prevent the issues we have where we have discrepancies between standalone and hosted

There's also one about CVO log levels that has taken a different approach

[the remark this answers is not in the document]
JoelSpeed: Did you have any discussion with the HyperShift folks about how to handle this? I believe someone was setting up a topic for their office hours?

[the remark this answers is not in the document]
marioferh: Yep https://issues.redhat.com/browse/HOSTEDCP-2194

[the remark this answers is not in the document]
enxebre: hcp exposes selected v1config knobs in the [HostedCluster API](https://github.com/openshift/hypershift/blob/main/api/hypershift/v1beta1/hostedcluster_types.go#L537) It would seem reasonable to expose intent for this there as well if we need to.  API Consumers would need to assess coverage and intersection of existing scenarios, https://github.com/openshift/hypershift-dataplane-metrics-forwarder/blob/main/hack/templates/00-hypershift-dataplane-metrics-forwarder.SelectorSyncSet.yaml.tmpl#L107

Now, re how this input propagates down (namespace vs cluster scope): while making it namespaced will give more flexibility for future pivots I don't think there's a current direction or effort to move the cluster monitoring into the management side. cc @csrwng @typeid

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,88 @@
...
+// +openshift:compatibility-gen:internal
+// +openshift:api-approved.openshift.io=https://github.com/openshift/api/pull/1929
+// +openshift:file-pattern=cvoRunLevel=0000_10,operatorName=config-operator,operatorOrdering=01
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=clustermonitoring,scope=Cluster
+// +kubebuilder:subresource:status
+// +kubebuilder:metadata:annotations="description=Cluster Monitoring Operators configuration API"
+// +openshift:enable:FeatureGate=ClusterMonitoringConfig
+// ClusterMonitoring is the Schema for the Cluster Monitoring Operators API
+type ClusterMonitoring struct {
```
enxebre, line 39: does this API have any intersections with https://github.com/rhobs/observability-operator/blob/main/docs/api.md?

config/v1/types_cluster_monitoring.go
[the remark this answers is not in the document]
typeid: For ROSA HCP, our ultimate goal is to fully monitor the platform purely from the management cluster side via metrics exposed by the hypershift/control-plane components.

We have [XCMSTRAT-976](https://issues.redhat.com/browse/XCMSTRAT-976) to hand off CMO completely to the user, [up to the point that they can end up enabling/disabling it as a whole](https://issues.redhat.com/browse/MON-3152), therefore I think it makes sense to keep it as is (CMO configurable from the hosted cluster side) - feel free to reach out in that XCMSTRAT in case more clarification is needed.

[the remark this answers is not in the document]
JoelSpeed: Ok, so if I'm reading the last few comments correctly, we are happy to proceed with this design as is, and will create a new abstraction at the HostedCluster level for HCP relevant parts of monitoring the management side

config/v1/types_cluster_monitoring.go
JoelSpeed: Declaring it as mandatory means marking it as required, I don't see it marked as required?

Documentation still needs to be updated about the available values and behaviour of each of those

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// UserDefinedMonitoring config for user-defined projects.
+type UserDefinedMonitoring struct {
+	// mode defines the different configurations of UserDefinedMonitoring
+	// Valid values are Disabled and NamespaceIsolated
+	// Disabled disables monitoring for user-defined projects. This restricts the default monitoring stack, installed in the openshift-monitoring project, to monitor only platform namespaces, which prevents any custom monitoring configurations or resources from being applied to user-defined namespaces.
+	// NamespaceIsolated enables monitoring for user-defined projects with namespace-scoped tenancy. This ensures that metrics, alerts, and monitoring data are isolated at the namespace level.
+	// +kubebuilder:validation:Enum:="Disabled";"NamespaceIsolated"
+	Mode UserDefinedMode `json:"mode,omitempty"`
```
JoelSpeed, line 78: Is this optional or required?

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+type UserDefinedMode string
```
JoelSpeed, line 81: Needs some sort of godoc for what this enum is used for.

Needs a `// +enum` marker added

JoelSpeed (commented):
Please review the linter issues and resolve those, in particular, lots of fields aren't marked either as optional or required, they will need to be updated.

Try `make lint-fix` to help where some issues can be auto-fixed

config/v1/types_cluster_monitoring.go
marioferh: no.

config/v1/types_cluster_monitoring.go
marioferh: We had a looong discussion about it and I made some changes, I don't remember right now what we concluded but I though this way was correct.

config/v1/types_cluster_monitoring.go
marioferh: done

config/v1/types_cluster_monitoring.go
marioferh: done

JoelSpeed:
From the linter, these ones should be fixed. The first four WRT empty json tag is a bug that I'll fix this afternoon
```
config/v1/types_cluster_monitoring.go:63:2: optionalorrequired: field Items must be marked as optional or required (kal)
	Items []ClusterMonitoring `json:"items"`
	^
config/v1/types_cluster_monitoring.go:[45](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_api/1929/pull-ci-openshift-api-master-lint/1868744969225768960#1:build-log.txt%3A45):2: requiredfields: field Spec is marked as required, but has the omitempty tag (kal)
	Spec ClusterMonitoringSpec `json:"spec,omitempty"`
	^
config/v1/types_cluster_monitoring.go:70:2: requiredfields: field UserDefined is marked as required, but has the omitempty tag (kal)
	UserDefined UserDefinedMonitoring `json:"userDefined,omitempty"`
	^
config/v1/types_cluster_monitoring.go:81:2: requiredfields: field Mode is marked as required, but has the omitempty tag (kal)
	Mode UserDefinedMode `json:"mode,omitempty"`
	^
```

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,93 @@
...
+type ClusterMonitoringStatus struct {
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// Compatibility level 4: No compatibility is provided, the API can change at any point for any reason. These capabilities should not be used by applications needing long term support.
+// +openshift:compatibility-gen:internal
+type ClusterMonitoringList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
```
JoelSpeed, line 61:
```suggestion

	// metadata is the standard list metadata.
	metav1.ListMeta `json:"metadata,omitempty"`
```

config/v1/types_cluster_monitoring.go
```diff
@@ -0,0 +1,93 @@
...
+// +openshift:file-pattern=cvoRunLevel=0000_10,operatorName=config-operator,operatorOrdering=01
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=clustermonitoring,scope=Cluster
+// +kubebuilder:subresource:status
+// +kubebuilder:metadata:annotations="description=Cluster Monitoring Operators configuration API"
+// +openshift:enable:FeatureGate=ClusterMonitoringConfig
+// ClusterMonitoring is the Schema for the Cluster Monitoring Operators API
+type ClusterMonitoring struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
```
JoelSpeed, line 41:
```suggestion
	
	// metadata is the standard object metadata.
	metav1.ObjectMeta `json:"metadata,omitempty"`
```

JoelSpeed (commented):
Couple of nits, plus the linter fixes and I think this is good to go

config/v1/types_cluster_monitoring.go
marioferh: thx

jan--f (approved):
/lgtm

JoelSpeed:

This only affects the payload in adding a new techpreview CRD. @marioferh is continuing to work on this feature over shutdown, and not having this merged is a blocker for him. I believe this is sufficiently low enough risk to merge even under the critical fixes label

PR #1934, opened 2024-06-18
NO-JIRA: remove beta enabled features from feature gates

ingvagabund:
Follow up for https://github.com/openshift/api/pull/1831. Both ServiceAccountTokenNodeBindingValidation and ServiceAccountTokenPodNodeInfo got promoted beta and enabled by default in 1.30/4.17.


ingvagabund:
Failing with `[sig-network][Feature:Router][apigroup:route.openshift.io] The HAProxy router converges when multiple routers are writing conflicting status [Suite:openshift/conformance/parallel]`

/retest-required

PR #1940, opened 2024-06-28
OCPNODE-2388: add UserNamespacesSupport to TechPreview feature gates

haircommander:


haircommander:
I think we should just carry this in https://github.com/openshift/api/pull/1939

PR #1952, opened 2024-07-04
OCPBUGS-35906: a rule to check if the featureSet is one of the known set of features

p0lyn0mial:
In the past, there was an admission plugin for checking if the value of the featureSet field matched a known one. 

With https://github.com/openshift/kubernetes/pull/1944, the validation was moved to CEL and the check was dropped. 
This PR adds the missing check. 

NOTE that there was an additional check that allowed updates to FG even with an invalid featureSet if the field wasn't modified. I haven't found a way to move this to CEL for kube < 1.30. 

The validation was moved in 4.16, and setting an invalid value breaks an operator and prevents upgrades. This PR allows transitioning from an invalid value to unblock upgrades.

For full backward compatibility we would have to bring back the admission plugin for 4.16 and use for 4.17 use `optionalOldSelf`.

### DIFF
--- a/config/v1/tests/featuregates.config.openshift.io/AAA_ungated.yaml
+++ b/config/v1/tests/featuregates.config.openshift.io/AAA_ungated.yaml
@@ -23,6 +23,13 @@ tests:
         kind: FeatureGate
         spec:
           featureSet: TechPreviewNoUpgrade
+    - name: Must reject unsupported value of a FeatureGate
+      initial: |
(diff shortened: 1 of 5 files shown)


deads2k:
add testing

p0lyn0mial:
@deads2k added the missing test cases, PTAL.

deads2k:
Setting different values caused failures, so this is ok.

/lgtm
/approve

PR #1955, opened 2024-07-10
OCPBUGS-36809: Add missing auth config fields

jonesbr17:
The `uid` and `extra` fields were added to Kubernetes in 1.29.
These fields should be included in OpenShift as well.


csrwng:
cc @openshift/api-reviewers

csrwng:
cc @ibihim

rtheis (approved):
/lgtm

liouk:
@jonesbr17 could you please adapt/add tests in `config/v1/tests/authentications.config.openshift.io/ExternalOIDC.yaml`?

rtheis (approved):
/lgtm

config/v1/tests/authentications.config.openshift.io/ExternalOIDC.yaml
```diff
@@ -120,6 +120,35 @@ tests:
               username:
                 claim: "preferred_username"
                 prefixPolicy: NoPrefix
+    - name: Can set OIDC providers with extra mappings
```
liouk, line 123: How about testing setting the `uid` field?

liouk (changes requested):

config/v1/tests/authentications.config.openshift.io/ExternalOIDC.yaml
```diff
@@ -120,6 +120,39 @@ tests:
...
+        spec:
+          type: OIDC
+          oidcProviders:
+          - name: myoidc
+            issuer:
+              issuerURL: https://meh.tld
+              audiences: ['openshift-aud']
+            claimMappings:
+              uid:
+                claim: claim
```
liouk, line 136: According to the definition, `claim` is mutually exclusive with `expression` so we should add a separate test case for that, too.

liouk (changes requested):

rtheis (approved):
/lgtm

rtheis:
@jonesbr17 ptal

jonesbr17:
Superseded by https://github.com/openshift/api/pull/2234

PR #1977, opened 2024-07-30
API-1689: features: add ShortCertRotation

vrutkovs:
Enhancement proposal: https://github.com/openshift/enhancements/pull/1670

### DIFF
--- a/features/features.go
+++ b/features/features.go
@@ -809,4 +809,12 @@ var (
 		enhancementPR("https://github.com/openshift/enhancements/pull/1756").
 		enableIn(configv1.DevPreviewNoUpgrade, configv1.TechPreviewNoUpgrade).
 		mustRegister()
+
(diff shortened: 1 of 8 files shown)


features/features.go
```diff
@@ -610,4 +610,11 @@ var (
 					productScope(kubernetes).
 					enableIn(configv1.DevPreviewNoUpgrade, configv1.TechPreviewNoUpgrade).
 					mustRegister()
+
+	FeatureShortCertRotation = newFeatureGate("ShortCertRotation").
```
tkashem, line 614: If it is meant to be a CI only feature, then can we rename it to make it obvious, something like `CIOnlyShortCertRotation`

vrutkovs: This feature will be tested in CI, but it may be used in manual testing. I don't mind renaming it though

PR #1980, opened 2024-07-31
AUTH-527: security/v1 - add constant for enforcing psa annotation

ibihim:
## What

Add an annotation constant that reveals what the label syncer would enforce.

## Why

It is unclear what the label syncer would set, if alert labels are set or the syncer is turned off.

### DIFF
--- a/security/v1/consts.go
+++ b/security/v1/consts.go
@@ -10,4 +10,7 @@ const (
 	// This annotation pins required SCCs for core OpenShift workloads to prevent preemption of custom SCCs.
 	// It is being used in the SCC admission plugin.
 	RequiredSCCAnnotation = "openshift.io/required-scc"
+
+	// MinimallySufficientPodSecurityStandard indicates the PodSecurityStandard that matched the SCCs available to the users of the namespace.
+	MinimallySufficientPodSecurityStandard = "security.openshift.io/psa"
 )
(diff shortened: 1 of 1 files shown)


security/v1/consts.go
```diff
@@ -10,4 +10,7 @@ const (
 	// This annotation pins required SCCs for core OpenShift workloads to prevent preemption of custom SCCs.
 	// It is being used in the SCC admission plugin.
 	RequiredSCCAnnotation = "openshift.io/required-scc"
+
+	// MinimallySufficientPodSecurityStandard indicates the PodSecurityStandard that matched the SCCs available to the users of the namespace.
+	MinimallySufficientPodSecurityStandard = "security.openshift.io/psa"
```
deads2k, line 15:
```suggestion
	MinimallySufficientPodSecurityStandard = "security.openshift.io/MinimallySufficientPodSecurityStandard"
```

liouk: Maybe a nit: overall we seem to be using both kebab-case and camelCase for annotations/labels, so not sure which one is preferable -- but probably PascalCase should be avoided?

i.e. suggesting
```
security.openshift.io/minimallySufficientPodSecurityStandard
```
or
```
security.openshift.io/minimally-sufficient-pod-security-standard
```
instead of
```
security.openshift.io/MinimallySufficientPodSecurityStandard
```

PR #1981, opened 2024-07-31
clean up openshift-sdn references in the API

danwinship:
- `config.openshift.io`:
  - Belatedly update a few places to now say that the default plugin is `OVNKubernetes`, not `OpenShiftSDN`.
  - Clarify the docs on `NetworkMigration` to make it clear that it is used for status, not spec.
- `operator.openshift.io`:
  - Remove references to OpenShift SDN being a supported network type.
  - Mark `OpenShiftSDNConfig` as deprecated.
  - Belatedly remove the comment claiming that `OVNKubernetes` is not supported. :grimacing: 
  - Clarify that `NetworkType` migration is no longer supported by current versions of OpenShift, and so setting the associated fields will result in CNO rejecting the configuration.
- `network.openshift.io`
  - Deprecate all the (OpenShift-SDN-only) types.

Note that all of the `network.openshift.io` types are now neither read from nor written to by any OCP component. _In theory_, users might be using them for their own purposes for unknown reasons not related to openshift-sdn, but other than that there's no reason to not just remove those types completely. (Though this PR does not do that.)


operator/v1/types_network.go
```diff
@@ -91,8 +91,8 @@ type NetworkSpec struct {
 	// deployKubeProxy specifies whether or not a standalone kube-proxy should
 	// be deployed by the operator. Some network providers include kube-proxy
 	// or similar functionality. If unset, the plugin will attempt to select
-	// the correct value, which is false when OpenShift SDN and ovn-kubernetes are
-	// used and true otherwise.
+	// the correct value, which is false when ovn-kubernetes is used and true
```
kyrtapz, line 94: While we are here should we change `ovn-kubernetes` to `OVNKubernetes`?

operator/v1/types_network.go
```diff
@@ -136,23 +136,26 @@ const (
...
 	// +optional
 	MTU *MTUMigration `json:"mtu,omitempty"`
 
+	// networkType is the target type of network migration. Set this to the
+	// target network type to allow changing the default network. If unset, the
+	// operation of changing cluster default network plugin will be rejected.
+	// Current versions of OpenShift do not support networkType migration and
+	// will reject any configuration where this is set.
+	// +optional
+	NetworkType string `json:"networkType,omitempty"`
```
kyrtapz, line 151: Could you help me understand why are we not deprecating networkType migration?

operator/v1/types_network.go
danwinship: `OVNKubernetes` with no hyphen and no space is generally only used as a value in the API, whereas we're talking about the plugin itself here really, so `ovn-kubernetes` (or `OVN-Kubernetes` or `OVN Kubernetes` seems correct)

operator/v1/types_network.go
danwinship: Hm... I guess you're right. We have no plans of ever supporting that again.

operator/v1/types_network.go
abhat: I think the name BU prefers in all documentation is `OVN Kubernetes`. cc @mcurry-rh.

danwinship: The existing API documentation is inconsistent and this does not change it, so that can be fixed later.

danwinship:
@kyrtapz can you re-review?

PR #2024, opened 2024-09-10
MULTIARCH-4559: config/v1/types_cluster_version.go: Add 'architecture' to the release structure

wking:
As [requested in the enhancement][1].  [Motivation for the new property from the enhancement][2]:

> There were a few options discussed in lieu of introducing a new ClusterVersion status field and the potential risks for doing so. The alternatives are highlighted with reasoning given for why they were not pursued:
> * Default ImportMode to PreserveOriginal everywhere: single-arch-release users maybe concerned about import size and the lack of metadata like `dockerImageLayers` and `dockerImageMetadata` for manifestlisted imagestream tags.
> * Clusters with homogeneous nodes running the multi payload who do not want to import manifestlists: The clusters can either migrate to single arch payloads or manually toggle the importMode through the image config CRD.
> * CVO provides architecural knowledge to the cluster-image-registry-operator through a configmap or the image config CRD: To limit the risk of many external consumers using CVO's status field to determine that their cluster is multi-arch ready, the idea was to expose this information to the specific controller. This solution is not necessary as we let other controller implementers decide if the CVO's new status field is the best fit for their use case.

[1]: https://github.com/openshift/enhancements/blob/36ef3f51e5b7bfb2d4af18d415ba762e0a0431ac/enhancements/multi-arch/dynamic-imagestream-importmode-setting.md#api-extensions
[2]: https://github.com/openshift/enhancements/blob/36ef3f51e5b7bfb2d4af18d415ba762e0a0431ac/enhancements/multi-arch/dynamic-imagestream-importmode-setting.md#motivations-for-a-new-clusterversion-status-property

### DIFF
--- a/config/v1/types_cluster_version.go
+++ b/config/v1/types_cluster_version.go
@@ -746,6 +746,16 @@ type Update struct {
 // Release represents an OpenShift release image and associated metadata.
 // +k8s:deepcopy-gen=true
 type Release struct {
+	// architecture is an optional field that indicates the
+	// value of the cluster architecture. In this context cluster
+	// architecture means either a single architecture or a multi
+	// architecture.
+	// Valid values are 'Multi' and empty.
+	//
+	// +openshift:enable:FeatureGate=ImageStreamImportMode
+	// +optional
+	Architecture ClusterVersionArchitecture `json:"architecture,omitempty"`
+
 	// version is a semantic version identifying the update version. When this
 	// field is part of spec, version is optional if image is specified.
 	// +required
--- a/config/v1/tests/clusterversions.config.openshift.io/ImageStreamImportMode.yaml
+++ b/config/v1/tests/clusterversions.config.openshift.io/ImageStreamImportMode.yaml
@@ -1,0 +1,441 @@
+apiVersion: apiextensions.k8s.io/v1 # Hack because controller-gen complains if we don't have this
+name: "ClusterVersion"
+crdName: clusterversions.config.openshift.io
+featureGate: ImageStreamImportMode
+tests:
+  onCreate:
+    - name: Should be able to create a minimal ClusterVersion
+      initial: |
+        apiVersion: config.openshift.io/v1
+        kind: ClusterVersion
+        spec:
+          clusterID: foo
+      expected: |
+        apiVersion: config.openshift.io/v1
+        kind: ClusterVersion
+        spec:
+          clusterID: foo
+    - name: Should allow image to be set
+      initial: |
+        apiVersion: config.openshift.io/v1
+        kind: ClusterVersion
+        spec:
+          clusterID: foo
+          desiredUpdate:
+            image: bar
+      expected: |
+        apiVersion: config.openshift.io/v1
+        kind: ClusterVersion
+        spec:
+          clusterID: foo
+          desiredUpdate:
+            image: bar
+    - name: Should allow version to be set
+      initial: |
+        apiVersion: config.openshift.io/v1
+        kind: ClusterVersion
+        spec:
+          clusterID: foo
+          desiredUpdate:
+            version: 4.11.1
+      expected: |
+        apiVersion: config.openshift.io/v1
+        kind: ClusterVersion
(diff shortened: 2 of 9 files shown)


config/v1/types_cluster_version.go
```diff
@@ -691,6 +691,15 @@ type Update struct {
 // Release represents an OpenShift release image and associated metadata.
 // +k8s:deepcopy-gen=true
 type Release struct {
+	// architecture is an optional field that indicates the
+	// value of the cluster architecture. In this context cluster
```
Prashanth684, line 695: can we also clarify in the comment that this is the architecture the cluster is trying to reconcile to and not necessarily the current payload architecture

wking: Not at this level, because `Release` shows up in other places too:

```console
$ git grep Release config/ | grep json
config/v1/types_cluster_version.go:     Desired Release `json:"desired"`
config/v1/types_cluster_version.go:     AvailableUpdates []Release `json:"availableUpdates"`
config/v1/types_cluster_version.go:     Release Release `json:"release"`
```

But up one level above this, at the `status.desired` consumer you're thinking of, we [already have][1]:

> desired is the version that the cluster is reconciling towards

which addresses your concern, right?  Or did you want additional wording there to underline "but 'reconciling towards' does not mean 'completely delivered'"?

[1]: https://github.com/openshift/api/blob/3e5de946111c67cb1f2b09ff7da569012d15931d/config/v1/types_cluster_version.go#L138

Prashanth684: I think we are fine as is. As you said, given that the wording is there in the `desired` field, we are fine.

config/v1/types_cluster_version.go
```diff
@@ -691,6 +691,15 @@ type Update struct {
...
+	// architecture means either a single architecture or a multi
+	// architecture.
+	// Valid values are 'Multi' and empty.
+	//
+	// +optional
+	Architecture ClusterVersionArchitecture `json:"architecture,omitempty"`
```
deads2k, line 701: Needs featuregate.

config/v1/types_cluster_version.go
```diff
@@ -691,6 +691,15 @@ type Update struct {
 // Release represents an OpenShift release image and associated metadata.
 // +k8s:deepcopy-gen=true
 type Release struct {
+	// architecture is an optional field that indicates the
+	// value of the cluster architecture. In this context cluster
+	// architecture means either a single architecture or a multi
+	// architecture.
+	// Valid values are 'Multi' and empty.
```
deads2k, line 698: This may very well end up extended with more detail in the near future, but starting here is acceptable.

config/v1/types_cluster_version.go
wking: Which feature gate would you like?  [The enhancement is pretty thin in that section][1].

[1]: https://github.com/openshift/enhancements/blob/36ef3f51e5b7bfb2d4af18d415ba762e0a0431ac/enhancements/multi-arch/dynamic-imagestream-importmode-setting.md#graduation-criteria

Prashanth684: the `ImageStreamImportMode` featuregate implemented here: https://github.com/openshift/api/pull/1928

Prashanth684:
@wking any progress on this?

config/v1/types_cluster_version.go
wking: Feature gate added as of 7f3382e2f1b4d6d03.  [The verify-crd-schema][1] failures are all existing properties in the new-with-this-pull `clusterversions.config.openshift.io/ImageStreamImportMode.yaml` file, and they'll need the `/override ci/prow/verify-crd-schema` mentioned in the failure message.

[1]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_api/2024/pull-ci-openshift-api-master-verify-crd-schema/1857445385325252608#1:build-log.txt%3A541-549

PR #2032, opened 2024-09-17
NO-JIRA: move storage CRDs to openshift/api

bertinatto:
This is a follow-up to https://github.com/openshift/api/pull/2030.

This makes the promotion of featuregates more atomic, as the CRD schema changes in the same PR that changes the feature.

This will prevent changes like [this](https://github.com/openshift/cluster-storage-operator/commit/076e178df7aa8bd9753a2b5de9aa14542412a757) when adding new TechPreview fields to ClusterCSIDriver CRDs.

CC @openshift/storage


deads2k:
/lgtm
/hold

Due to some out of sync-ness, the order needs to be

1. update https://github.com/openshift/cluster-storage-operator with latest openshift/api and merge
2. then re-test this PR for green and merge (notice verify is red now)
3. then, shortly after merge https://github.com/openshift/cluster-storage-operator/pull/505/files.


But the change here lgtm.

PR #2035, opened 2024-09-20
API-1843: FeatureGate(d) KMS encryption

swghosh:
Adds API fields for KMS based encryption enablement via `configv1.APIServer` through it's `spec.encryptionType`,
AWS KMS config added as the initial KMS encryption provider.

This feature is primarily targeted for **SelfManagedHA OpenShift** and is considered for TechPreview.

- [x] Feature Gate: openshift/api#2071
- [ ] Enhancement: https://github.com/openshift/enhancements/pull/1682


config/v1/types_apiserver.go
```diff
@@ -191,9 +194,23 @@ type APIServerEncryption struct {
...
+	// externally configured KMS instance.
+	//
+	// The Key Management Service (KMS) instance provides symmetric encryption and is responsible for
+	// managing the lifecyle of the encryption keys outside of the control plane.
+	// This allows integration with an external provider to manage the data encryption keys securely.
+	//
+	// +openshift:enable:FeatureGate=KMSv2
+	// +unionMember
+	// +optional
+	KMS *KMSConfig `json:"kms,omitempty"`
```
JoelSpeed, line 209: You'll need to add CEL validation to the parent struct to ensure this is only set when type is KMS. Look for CELUnion in the example v1 API group stable_types.go for an example of how to do this

config/v1/types_apiserver.go
```diff
@@ -191,9 +194,23 @@ type APIServerEncryption struct {
...
 }
 
-// +kubebuilder:validation:Enum="";identity;aescbc;aesgcm
+// +openshift:validation:FeatureSetAwareEnum:featureSet=Default,enum="";identity;aescbc;aesgcm
+// +openshift:validation:FeatureSetAwareEnum:featureSet=TechPreviewNoUpgrade,enum="";identity;aescbc;aesgcm;kms
```
JoelSpeed, line 213: Use featuregate aware, not featureset aware.

Out of interest, where did you find the example for featureset aware?

config/v1/types_apiserver.go
```diff
@@ -208,6 +225,13 @@ const (
 	// aesgcm refers to a type where AES-GCM with random nonce and a 32-byte key
 	// is used to perform encryption at the datastore layer.
 	EncryptionTypeAESGCM EncryptionType = "aesgcm"
+
+	// kms refers to a type of encryption where the encryption keys are managed
+	// outside the control plane in a Key Management Service instance,
+	// encryption is still performed at the datastore layer.
+	//
+	// +openshift:enable:FeatureGate=KMSv2
+	EncryptionTypeKMS EncryptionType = "kms"
```
JoelSpeed, line 234: Enum values must be PascalCase, in this example, that means
```suggestion
	EncryptionTypeKMS EncryptionType = "KMS"
```

config/v1/types_apiserver.go
```diff
@@ -208,6 +225,13 @@ const (
 	// aesgcm refers to a type where AES-GCM with random nonce and a 32-byte key
 	// is used to perform encryption at the datastore layer.
 	EncryptionTypeAESGCM EncryptionType = "aesgcm"
+
+	// kms refers to a type of encryption where the encryption keys are managed
+	// outside the control plane in a Key Management Service instance,
+	// encryption is still performed at the datastore layer.
+	//
+	// +openshift:enable:FeatureGate=KMSv2
```
JoelSpeed, line 233: You don't need to gate the value constant here

config/v1/types_kmsencryption.go
```diff
@@ -0,0 +1,28 @@
+package v1
+
+// KMSConfig defines the configuration for the KMS instance
+// that will be used with KMSv2 encryption
+//
+// +openshift:enable:FeatureGate=KMSv2
```
JoelSpeed, line 6: This struct does not need to be gated

config/v1/types_kmsencryption.go
```diff
@@ -0,0 +1,28 @@
...
+type KMSConfig struct {
```
JoelSpeed, line 7: What expansion do you expect in the future? Does this need to be a union with a platform type as the discriminator?

config/v1/types_kmsencryption.go
```diff
@@ -0,0 +1,28 @@
...
+	// for the encryption. The AWS KMS instance is managed
+	// by the user outside the purview of the control plane.
+	//
+	// +optional
+	AWS *AWSKMSConfig `json:"aws,omitempty"`
+}
+
+// AWSKMSConfig defines the KMS config specific to AWS KMS provider
+//
+// +openshift:enable:FeatureGate=KMSv2
```
JoelSpeed, line 18: You don't need to gate the struct

config/v1/types_kmsencryption.go
```diff
@@ -0,0 +1,28 @@
...
+type AWSKMSConfig struct {
+	// keyARN is the AWS ARN for the symmetric encryption KMS key
```
JoelSpeed, line 20: You need to explain in the godoc, what the format of the value is, and validations we apply should be explained

config/v1/types_kmsencryption.go
```diff
@@ -0,0 +1,28 @@
...
+	//
+	// +kubebuilder:validation:Pattern=`^arn:aws:kms:[a-z0-9-]+:[0-9]{12}:key/[a-f0-9-]+$`
+	KeyARN string `json:"keyARN"`
```
JoelSpeed, line 23: Mark as required explicitly 
```suggestion
	// +kubebuilder:validation:Required
	KeyARN string `json:"keyARN"`
```

config/v1/types_kmsencryption.go
```diff
@@ -0,0 +1,28 @@
...
+	AWS *AWSKMSConfig `json:"aws,omitempty"`
+}
+
+// AWSKMSConfig defines the KMS config specific to AWS KMS provider
+//
+// +openshift:enable:FeatureGate=KMSv2
+type AWSKMSConfig struct {
+	// keyARN is the AWS ARN for the symmetric encryption KMS key
+	//
+	// +kubebuilder:validation:Pattern=`^arn:aws:kms:[a-z0-9-]+:[0-9]{12}:key/[a-f0-9-]+$`
```
JoelSpeed, line 22: Use CEL for this, it allows you to craft a custom error message. Break down the rule and explain it to users in concise prose
```suggestion
	// +kubebuilder:validation:XValidation:rule="self.matches('^arn:aws:kms:[a-z0-9-]+:[0-9]{12}:key/[a-f0-9-]+$')",message="keyARN must start with `arn:aws:kms:` followed by ..."
```

config/v1/types_kmsencryption.go
```diff
@@ -0,0 +1,28 @@
...
+	KeyARN string `json:"keyARN"`
+	// region is the AWS region where the KMS instance exists
+	//
+	// +kubebuilder:validation:Pattern=`^[a-z]{2}-[a-z]+-\d{1,2}$`
+	Region string `json:"region"`
```
JoelSpeed, line 27: All the comments for keyARN apply here too

config/v1/zz_generated.crd-manifests/0000_10_config-operator_01_apiservers-Default.crd.yaml
```diff
@@ -171,11 +172,6 @@ spec:
                         3. routes.route.openshift.io
                         4. oauthaccesstokens.oauth.openshift.io
                         5. oauthauthorizetokens.oauth.openshift.io
-                    enum:
-                    - ""
-                    - identity
-                    - aescbc
-                    - aesgcm
```
JoelSpeed, line 178: Make sure this goes away when you fix the featureset to featuregate enum tag

config/v1/types_apiserver.go
JoelSpeed: Ok, both of these examples are out of date, you must use featureGate now, not featureSet

config/v1/types_apiserver.go
swghosh: yeah, I've added validation to enforce KMSconfig only when type is KMS.

config/v1/types_apiserver.go
swghosh: Using `FeatureGateAwareEnum` now.

config/v1/types_kmsencryption.go
swghosh: Yes, so I ended up adding a provider type with similar unionDiscriminator + related CEL validation.

In theory, we could have reused PlatformType which is used in Infrastructure status but I do not want to couple it immediately with all cloud providers (as 1. we want to support AWS KMS only rn, 2. not all platform providers may have KMSes).

config/v1/types_kmsencryption.go
swghosh: Added required.

config/v1/types_kmsencryption.go
swghosh: Added CEL validation.

config/v1/types_kmsencryption.go
swghosh: Added CEL validation.

config/v1/zz_generated.crd-manifests/0000_10_config-operator_01_apiservers-Default.crd.yaml
swghosh: This is fixed now that I'm using FeatureGateAwareEnum.

JoelSpeed:
The EP for this feature doesn't yet have any review, lets get some initial review going on the EP first before we get too much into perfecting the API (else it may change massively based on EP feedback from other stakeholders)

config/v1/types_apiserver.go
```diff
@@ -173,6 +173,9 @@ type APIServerNamedServingCert struct {
 	ServingCertificate SecretNameReference `json:"servingCertificate"`
 }
 
+// APIServerEncryption is used to encrypt sensitive resources on the cluster.
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'KMS' ?  has(self.kms) : !has(self.kms)",message="kms config is required when encryption type is KMS, and forbidden otherwise"
```
JoelSpeed, line 177: This has to be the feature gate aware XValidation else it will be added to the default version

config/v1/types_kmsencryption.go
JoelSpeed: I agree with not re-using the constant, and keeping the enum restricted, that's good. I wonder though if we should re-use the name here, so from a user perspective, it's still `platformType` that they are setting, WDYT?

config/v1/types_kmsencryption.go
```diff
@@ -0,0 +1,39 @@
...
+	// +unionMember
+	// +optional
+	AWS *AWSKMSConfig `json:"aws,omitempty"`
+}
+
+// AWSKMSConfig defines the KMS config specific to AWS KMS provider
+type AWSKMSConfig struct {
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self.matches('^arn:aws:kms:[a-z0-9-]+:[0-9]{12}:key/[a-f0-9-]+$')",message="keyARN must start with `arn:aws:kms:` followed by region, account and key id"
+	KeyARN string `json:"keyARN"`
```
JoelSpeed, line 27: Needs godoc on this field

config/v1/types_kmsencryption.go
```diff
@@ -0,0 +1,39 @@
...
+	//
+	// +unionMember
+	// +optional
+	AWS *AWSKMSConfig `json:"aws,omitempty"`
+}
+
+// AWSKMSConfig defines the KMS config specific to AWS KMS provider
+type AWSKMSConfig struct {
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self.matches('^arn:aws:kms:[a-z0-9-]+:[0-9]{12}:key/[a-f0-9-]+$')",message="keyARN must start with `arn:aws:kms:` followed by region, account and key id"
```
JoelSpeed, line 26:
```suggestion
	// +kubebuilder:validation:XValidation:rule="self.matches('^arn:aws:kms:[a-z0-9-]+:[0-9]{12}:key/[a-f0-9-]+$')",message="keyARN must follow the format `arn:aws:kms:<region>:<account_id>:key/<key_id>`. The account ID must be a 12 digit number and the region and key ID should consist only of lowercase hexadecimal characters and hyphens (-)."
```

config/v1/types_kmsencryption.go
```diff
@@ -0,0 +1,49 @@
+package v1
+
+// KMSConfig defines the configuration for the KMS instance
+// that will be used with KMSEncryptionProvider encryption
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'AWS' ?  has(self.aws) : !has(self.aws)",message="aws config is required when kms provider type is AWS, and forbidden otherwise"
+// +union
+type KMSConfig struct {
+	// type defines the kind of platform for the KMS provider
```
JoelSpeed, line 8:
```suggestion
	// type defines the kind of platform for the KMS provider.
	// Available platform types are AWS only.
```

config/v1/types_kmsencryption.go
```diff
@@ -0,0 +1,49 @@
...
+type AWSKMSConfig struct {
+	// keyARN specifies the Amazon Resource Name (ARN) of the AWS KMS key used for encryption.
+	// The value must adhere to the format `arn:aws:kms:<region>:<account_id>:key/<key_id>`, where:
+	// - `<region>` is the AWS region consisting of lowercase letters and hyphens followed by a number.
+	// - `<account_id>` is a 12-digit numeric identifier for the AWS account.
+	// - `<key_id>` is a unique identifier for the KMS key, consisting of lowercase hexadecimal characters and hyphens.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self.matches('^arn:aws:kms:[a-z0-9-]+:[0-9]{12}:key/[a-f0-9-]+$')",message="keyARN must follow the format `arn:aws:kms:<region>:<account_id>:key/<key_id>`. The account ID must be a 12 digit number and the region and key ID should consist only of lowercase hexadecimal characters and hyphens (-)."
+	KeyARN string `json:"keyARN"`
```
JoelSpeed, line 33: What is the maximum possible length of this string?

config/v1/types_kmsencryption.go
```diff
@@ -0,0 +1,49 @@
...
+	// region specifies the AWS region where the KMS intance exists, and follows the format
+	// `<region-prefix>-<region-name>-<number>`, e.g.: `us-east-1`.
+	// Only lowercase letters and hyphens followed by numbers are allowed.
+	//
+	// +kubebuilder:validation:XValidation:rule="self.matches('^[a-z]{2}-[a-z]+-[0-9]+$')",message="region must be a valid AWS region"
```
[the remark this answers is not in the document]
swghosh: It was inspired from: https://github.com/aws/aws-sdk-java-v2/blob/f3e9dc58e3da45c28b5283ddb3251e4bf9a9cd2e/codegen/src/main/resources/software/amazon/awssdk/codegen/rules/partitions.json.resource#L12 - which is used in the AWS SDK for Java, the aws client-go does not perform region regex validation IMHO.

Also, I wanted to avoid the `(us|eu|ap|sa|ca|me|af|il|mx)` prefixes in the hope that AWS might expand it in the future and then we have to update our validation instead used `[a-z]{2}`.

config/v1/types_kmsencryption.go
swghosh: Ti's an example ARN for KMS: `arn:aws:kms:ap-south-1:101010101010:key/9a512e29-0d9c-4cf5-8174-fc1a5b22cd6a` it has 76 characters. Guess we can use 80/85 characters? (so that we can account for regions with larger names too).

config/v1/types_kmsencryption.go
swghosh: Yeah, but in the future we might have requests to add KMS provider(s) which are not coherent with our PlatformType offerings eg. Vault or an XYZ HSM, etc. that led me to not use the `platformType` name from an UX perspective.

config/v1/types_kmsencryption.go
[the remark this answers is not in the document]
JoelSpeed: To determine a maximum length, we should first work out what the longest AWS region is so far. Looking at [this list](https://www.aws-services.info/regions.html), they all seem pretty short, so we could cap this to say 64 to give plenty of headroom for long custom region names in the future

config/v1/types_kmsencryption.go
JoelSpeed: Could just choose 128 as a round power of 2 limit, big enough that it allows some expansion, without crazy amounts of data allowed to be stored

config/v1/types_kmsencryption.go
JoelSpeed: Then perhaps we should have the enum value determine the type of the KMS, and not just "AWS"? 

Would it be possible to use the AWS implementation on an Azure cluster?

swghosh: > Would it be possible to use the AWS implementation on an Azure cluster?

Generally, the KMS instance resides outside of the control-plane so that could be a use case (however, I don't think many users may ask for it - so we can decide ourselves later how we want to support that).

On the contrary, to answer that still think that the answer is a YES, if we flip the scenario say: we can have a new provider type in the future eg. Vault that is supported on both AWS, Azure or any cloud provider for that matter.

swghosh: > Then perhaps we should have the enum value determine the type of the KMS, and not just "AWS"?

Din't quite get you on this one. Here's what is currently being used as the enum:
```go
// KMSProviderType is a specific supported KMS provider
// +kubebuilder:validation:Enum="";AWS
type KMSProviderType string
```

Are you suggesting to do it differently?

config/v1/types_kmsencryption.go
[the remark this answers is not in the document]
swghosh: Added, `&& self.size() <= 64` to the CEL validation rule.

config/v1/types_kmsencryption.go
swghosh: Added `&& self.size() <= 128` to the CEL validation rule.

config/v1/types_kmsencryption.go
```diff
@@ -0,0 +1,50 @@
...
+// that will be used with KMSEncryptionProvider encryption
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'AWS' ?  has(self.aws) : !has(self.aws)",message="aws config is required when kms provider type is AWS, and forbidden otherwise"
+// +union
+type KMSConfig struct {
+	// type defines the kind of platform for the KMS provider.
+	// Available provider types are AWS only.
+	//
+	// +unionDiscriminator
+	// +kubebuilder:validation:Required
+	Type KMSProviderType `json:"type"`
```
dgrisonnet, line 13: Do we really want/need to have this enum in the KMS config? It seems redundant to have both this and KMS specific configs. We can already tell which KMS the user want to configure by checking which config isn't nil.

config/v1/types_kmsencryption.go
```diff
@@ -0,0 +1,50 @@
+package v1
+
+// KMSConfig defines the configuration for the KMS instance
+// that will be used with KMSEncryptionProvider encryption
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'AWS' ?  has(self.aws) : !has(self.aws)",message="aws config is required when kms provider type is AWS, and forbidden otherwise"
```
dgrisonnet, line 5: Since we only want one configuration to be provided, maybe we could add something like:
```
[self.aws(,self.azure,self.gcp,...)].exists_one(c, c != "null")
```

dgrisonnet: Disregard since we want to keep KMSProviderType around

tkashem:
this looks good to me for techpreview

/lgtm

config/v1/types_kmsencryption.go
```diff
@@ -0,0 +1,50 @@
...
+	// keyARN specifies the Amazon Resource Name (ARN) of the AWS KMS key used for encryption.
+	// The value must adhere to the format `arn:aws:kms:<region>:<account_id>:key/<key_id>`, where:
+	// - `<region>` is the AWS region consisting of lowercase letters and hyphens followed by a number.
+	// - `<account_id>` is a 12-digit numeric identifier for the AWS account.
+	// - `<key_id>` is a unique identifier for the KMS key, consisting of lowercase hexadecimal characters and hyphens.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self.matches('^arn:aws:kms:[a-z0-9-]+:[0-9]{12}:key/[a-f0-9-]+$') && self.size() <= 128",message="keyARN must follow the format `arn:aws:kms:<region>:<account_id>:key/<key_id>`. The account ID must be a 12 digit number and the region and key ID should consist only of lowercase hexadecimal characters and hyphens (-)."
+	KeyARN string `json:"keyARN"`
+	// region specifies the AWS region where the KMS intance exists, and follows the format
```
JoelSpeed, line 35:
```suggestion
	// region specifies the AWS region where the KMS instance exists, and follows the format
```

config/v1/tests/apiservers.config.openshift.io/KMSEncryptionProvider.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: apiextensions.k8s.io/v1 # Hack because controller-gen complains if we don't have this
+name: "APIServer"
+crdName: apiservers.config.openshift.io
+featureGate: KMSEncryptionProvider
+tests:
+  onCreate:
```
JoelSpeed, line 6: Can you please add some cases that show the various validations returning errors so that we know that they are working please

config/v1/types_kmsencryption.go
```diff
@@ -0,0 +1,50 @@
...
+	// `<region-prefix>-<region-name>-<number>`, e.g.: `us-east-1`.
+	// Only lowercase letters and hyphens followed by numbers are allowed.
+	//
+	// +kubebuilder:validation:XValidation:rule="self.matches('^[a-z]{2}-[a-z]+-[0-9]+$') && self.size() <= 64",message="region must be a valid AWS region"
+	Region string `json:"region"`
+}
+
+// KMSProviderType is a specific supported KMS provider
+// +kubebuilder:validation:Enum="";AWS
```
JoelSpeed, line 44: Why is `""` a valid value for this enum?

config/v1/types_kmsencryption.go
```diff
@@ -0,0 +1,50 @@
...
+	// - `<key_id>` is a unique identifier for the KMS key, consisting of lowercase hexadecimal characters and hyphens.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self.matches('^arn:aws:kms:[a-z0-9-]+:[0-9]{12}:key/[a-f0-9-]+$') && self.size() <= 128",message="keyARN must follow the format `arn:aws:kms:<region>:<account_id>:key/<key_id>`. The account ID must be a 12 digit number and the region and key ID should consist only of lowercase hexadecimal characters and hyphens (-)."
+	KeyARN string `json:"keyARN"`
+	// region specifies the AWS region where the KMS intance exists, and follows the format
+	// `<region-prefix>-<region-name>-<number>`, e.g.: `us-east-1`.
+	// Only lowercase letters and hyphens followed by numbers are allowed.
+	//
+	// +kubebuilder:validation:XValidation:rule="self.matches('^[a-z]{2}-[a-z]+-[0-9]+$') && self.size() <= 64",message="region must be a valid AWS region"
```
JoelSpeed, line 39: Use `+kubebuilder:validation:MaxLength`, not checking the size in the CEL rule

config/v1/types_kmsencryption.go
```diff
@@ -0,0 +1,50 @@
...
+// AWSKMSConfig defines the KMS config specific to AWS KMS provider
+type AWSKMSConfig struct {
+	// keyARN specifies the Amazon Resource Name (ARN) of the AWS KMS key used for encryption.
+	// The value must adhere to the format `arn:aws:kms:<region>:<account_id>:key/<key_id>`, where:
+	// - `<region>` is the AWS region consisting of lowercase letters and hyphens followed by a number.
+	// - `<account_id>` is a 12-digit numeric identifier for the AWS account.
+	// - `<key_id>` is a unique identifier for the KMS key, consisting of lowercase hexadecimal characters and hyphens.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self.matches('^arn:aws:kms:[a-z0-9-]+:[0-9]{12}:key/[a-f0-9-]+$') && self.size() <= 128",message="keyARN must follow the format `arn:aws:kms:<region>:<account_id>:key/<key_id>`. The account ID must be a 12 digit number and the region and key ID should consist only of lowercase hexadecimal characters and hyphens (-)."
```
JoelSpeed, line 33: Use `+kubebuilder:validation:MaxLength`, not checking the size in the CEL rule

config/v1/types_kmsencryption.go
benluddy: Are you sure you want to be validating region names with a regular expression? I don't know that this naming scheme is anything other than a convention. What stops AWS from introducing a new region with a name that does not match your validation?

The AWS API doesn't even appear to validate this string beyond its length: https://docs.aws.amazon.com/accounts/latest/reference/API_Region.html.

swghosh: the regex expression was sourced from AWS Java client https://github.com/openshift/api/pull/2035#discussion_r1829865653
do you think we should still remove it?

config/v1/types_kmsencryption.go
swghosh: I'll remove that "" dint mean to keep an empty value lingering.

swghosh: we don't want a default type today, because there isn't one without configuring KMS

config/v1/types_kmsencryption.go
swghosh: Removed the regex validation.

config/v1/types_kmsencryption.go
swghosh: Used that instead.

config/v1/tests/apiservers.config.openshift.io/KMSEncryptionProvider.yaml
swghosh: Added more tests to check all the validations.

config/v1/types_apiserver.go
```diff
@@ -52,6 +52,7 @@ type APIServerSpec struct {
 	// server from JavaScript applications.
 	// The values are regular expressions that correspond to the Golang regular expression language.
 	// +optional
+	// +listType=atomic
```
swghosh, line 55: These `+listType=atomic` are required to make the `verify-crd-schema` pass, which fails the test saying lack of SSA tags in the generated CRD. Alternative to adding these is to override the specific test but this retains the API violations as-is today.

config/v1/types_kmsencryption.go
swghosh: updated

config/v1/tests/apiservers.config.openshift.io/KMSEncryptionProvider.yaml
```diff
@@ -0,0 +1,88 @@
...
+                region: us-east-1
+    - name: Should fail to create with an empty KMS config
+      initial: |
+        apiVersion: config.openshift.io/v1
+        kind: APIServer
+        spec:
+          encryption:
+            type: KMS
+            kms: {}
+      expectedError: "Invalid value: \"null\": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation"
```
JoelSpeed, line 40: This is not the part of the error I would expect to see here, I would expect to see something about type being required?

config/v1/tests/apiservers.config.openshift.io/KMSEncryptionProvider.yaml
```diff
@@ -0,0 +1,88 @@
...
+        apiVersion: config.openshift.io/v1
+        kind: APIServer
+        spec:
+          encryption:
+            type: KMS
+            kms:
+              type: AWS
+              aws:
+                region: us-east-1
+      expectedError: "Invalid value: \"null\": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation"
```
JoelSpeed, line 62: Again, not the part of the error we want to compare to

config/v1/tests/apiservers.config.openshift.io/KMSEncryptionProvider.yaml
```diff
@@ -0,0 +1,88 @@
...
+        kind: APIServer
+        spec:
+          encryption:
+            type: KMS
+            kms:
+              type: AWS
+              aws:
+                keyARN: arn:aws:kms:us-east-1:101010101010:key/9a512e29-0d9c-4cf5-8174-fc1a5b22cd6a
+                region: ""
+      expectedError: "Invalid value: \"string\": AWS region cannot be empty"
```
JoelSpeed, line 88: Nit, missing new line at the end of this file

config/v1/types_kmsencryption.go
```diff
@@ -0,0 +1,52 @@
...
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength=128
+	// +kubebuilder:validation:XValidation:rule="self.matches('^arn:aws:kms:[a-z0-9-]+:[0-9]{12}:key/[a-f0-9-]+$')",message="keyARN must follow the format `arn:aws:kms:<region>:<account_id>:key/<key_id>`. The account ID must be a 12 digit number and the region and key ID should consist only of lowercase hexadecimal characters and hyphens (-)."
+	KeyARN string `json:"keyARN"`
+	// region specifies the AWS region where the KMS instance exists, and follows the format
+	// `<region-prefix>-<region-name>-<number>`, e.g.: `us-east-1`.
+	// Only lowercase letters and hyphens followed by numbers are allowed.
+	//
+	// +kubebuilder:validation:MaxLength=64
+	// +kubebuilder:validation:XValidation:rule="self.size() > 0",message="AWS region cannot be empty"
```
JoelSpeed, line 41:
```suggestion
	// +kubebuilder:validation:MinLength=1
```

config/v1/types_kmsencryption.go
```diff
@@ -0,0 +1,52 @@
...
+// AWSKMSConfig defines the KMS config specific to AWS KMS provider
+type AWSKMSConfig struct {
+	// keyARN specifies the Amazon Resource Name (ARN) of the AWS KMS key used for encryption.
+	// The value must adhere to the format `arn:aws:kms:<region>:<account_id>:key/<key_id>`, where:
+	// - `<region>` is the AWS region consisting of lowercase letters and hyphens followed by a number.
+	// - `<account_id>` is a 12-digit numeric identifier for the AWS account.
+	// - `<key_id>` is a unique identifier for the KMS key, consisting of lowercase hexadecimal characters and hyphens.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength=128
```
JoelSpeed, line 33:
```suggestion
	// +kubebuilder:validation:MaxLength=128
	// +kubebuilder:validation:MinLength=1
```

config/v1/tests/apiservers.config.openshift.io/KMSEncryptionProvider.yaml
swghosh: yeah, updated it to use `"spec.encryption.kms.type: Required value"`

config/v1/tests/apiservers.config.openshift.io/KMSEncryptionProvider.yaml
swghosh: `"spec.encryption.kms.aws.keyARN: Required value"` now

config/v1/tests/apiservers.config.openshift.io/KMSEncryptionProvider.yaml
swghosh: ✅

config/v1/types_kmsencryption.go
swghosh: added `MinLength` validations

config/v1/types_kmsencryption.go
swghosh: ✅

swghosh:
updated the integration test suite to be more verbose and specific about the error checks.

config/v1/types_kmsencryption.go
```diff
@@ -0,0 +1,53 @@
...
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:XValidation:rule="self.matches('^arn:aws:kms:[a-z0-9-]+:[0-9]{12}:key/[a-f0-9-]+$')",message="keyARN must follow the format `arn:aws:kms:<region>:<account_id>:key/<key_id>`. The account ID must be a 12 digit number and the region and key ID should consist only of lowercase hexadecimal characters and hyphens (-)."
+	KeyARN string `json:"keyARN"`
+	// region specifies the AWS region where the KMS instance exists, and follows the format
+	// `<region-prefix>-<region-name>-<number>`, e.g.: `us-east-1`.
+	// Only lowercase letters and hyphens followed by numbers are allowed.
+	//
+	// +kubebuilder:validation:MaxLength=64
+	// +kubebuilder:validation:MinLength=1
+	Region string `json:"region"`
```
JoelSpeed, line 43: We used to have some validation on the valid characters here, where did it go?

swghosh: Had removed it based on suggestions from @benluddy 
https://github.com/openshift/api/pull/2035#discussion_r1852589708
Let's re-check if we want to keep that validation rule.

JoelSpeed: I would at least check the character set is lowercase alphanumeric segments separate by hyphens, even if it is not as strict as what you had. We know they use the region as part of the domain name, so it must be a valid domain segment, so below should be accurate. 
```
[a-z0-9]+(-[a-z0-9]+)*
```
And the maxLength can be 50, per AWS' own docs, based on Ben's link.

swghosh: Added

```diff
+ // +kubebuilder:validation:XValidation:rule="self.matches('^[a-z0-9]+(-[a-z0-9]+)*$')",message="region must be a valid AWS region, consisting of lowercase characters, digits and hyphens (-) only."

```

Also, added an integration test case to look for that error message when it is violated.

swghosh: btw, on that AWS region on a similar context
found from elsewhere that "aws-global" seems to also be a somewhat valid region for AWS Go SDK v2 trying to connect over AWS STS credentials with some legacy behaviour
xref: https://github.com/cert-manager/cert-manager/blob/537e71ee639a41887e93b0fd151bf063c4730536/pkg/issuer/acme/dns/route53/route53.go#L103
however the said regex, covers this case too so nvm!

JoelSpeed:
API is looking good, need to get the gate merged and go from there

Are the team happy with the shape of this API? How has feedback on the EP been?

config/v1/types_kmsencryption.go
benluddy: xref https://github.com/openshift/api/pull/2124

Here's an example of our API validation rejecting a value that would have been accepted by an AWS API. Just sharing to support my opinion that it's worth erring on the side of being lax with these. I would be a little surprised if these values were validated consistently across all services _within AWS_.

patrickdillon: > Just sharing to support my opinion that it's worth erring on the side of being lax with these. I would be a little surprised if these values were validated consistently across all services within AWS.

+1. I'm all for including documented regexes or, perhaps, relatively simple ones, but this is at least the second time I can recall that we (installer team) have had to revise regexes we attempted to reverse engineer the requirements. In [the first case](https://github.com/openshift/installer/pull/6513) we inadvertently excluded disk encryption keys with capital letters, which are allowed in Azure.

Where possible, I prefer validating against the cloud api rather than regexes, but that may just be a luxury we have in the installer.

config/v1/tests/apiservers.config.openshift.io/KMSEncryptionProvider.yaml
```diff
@@ -0,0 +1,137 @@
...
+          audit:
+            profile: Default
+          encryption:
+            type: KMS
+            kms:
+              type: AWS
+              aws:
+                keyARN: arn:aws:kms:us-east-1:101010101010:key/9a512e29-0d9c-4cf5-8174-fc1a5b22cd6a
+                region: us-east-1
+    - name: Should be able to create encrypt with KMS for AWS without region
```
dgrisonnet, line 33: Should we really allow that? Looking at the plugin's code, if the region is not provided, the plugin will try to get the region from the ec2 metadata API, but it can be disabled and I don't know if we can guarantee that the plugin will always have the permissions to call it. Maybe it will be simpler for us to just require the region

swghosh: Got it, updated to make the region field required.

swghosh:
This is a low risk change and API fields are well gated behind TechPreviw. It should not affect our OpenShift payload.

swghosh:
/hold
e2e-aws-ovn-hypershift have been generally failing across
https://sippy.dptools.openshift.org/sippy-ng/jobs/4.19/analysis?filters=%7B%22items%22%3A%5B%7B%22columnField%22%3A%22name%22%2C%22operatorValue%22%3A%22equals%22%2C%22value%22%3A%22periodic-ci-openshift-hypershift-release-4.19-periodics-e2e-aws-ovn%22%7D%5D%7D

swghosh:
/remove-hold
permafailing e2e-aws-ovn-hypershift job was supposedly fixed in https://github.com/openshift/hypershift/pull/5881

PR #2069, opened 2024-10-17
API-1858: Add ModernTLSConfiguration feature gate

sanchezl:
Introduce `ModernTLSConfiguration` as an _inaccessible-by-default_ feature gate.


sanchezl:
> Warning: The referenced jira issue has an invalid target version for the target branch this PR targets: expected the sub-task to target the "4.18.0" version, but no target version was set.
The list of targets available in JIRA seems to be stale, and 4.18.0 is not allowed as a valid value.

deads2k:
Is there an enhancement describing what this will do and how we'll know it's done?  I think I've see a lot around this one in the past and I know it involves at least two teams.

JoelSpeed:
@sanchezl Is this PR still needed?

JoelSpeed:
/close

Please reopen if you feel we still need this

PR #2070, opened 2024-10-17
Add omitempty tag for imageStreamImportMode

enxebre:
This seems to match the original proposal https://github.com/openshift/enhancements/pull/1605/files#diff-c3558e7ce5705e060c3f6ccbbf7430be6fe4dac2b8bad05220360bd977f81e70R214, recomendations and consistency with the rest of this API. The reason I came across this is that hypershift seralizes this API and use that as input to generate a hash that now transparently changes. While that fundamental problem is solved this will happen to buy hypershift some time.


JoelSpeed:
@enxebre Did you want to progress with this PR?

JoelSpeed:
/close

Please reopen if you intend to continue with this

PR #2089, opened 2024-11-07
MCO-1284: MachineOSConfig: set ImageSecretObjectReference to optional

yuqi-zhang:
Small update to the TP field to optional, so the MCO can use cluster pullsecrets by default.


machineconfiguration/v1alpha1/types_machineosconfig.go
```diff
@@ -104,9 +104,10 @@ type BuildInputs struct {
 	// +optional
 	BaseOSImagePullspec string `json:"baseOSImagePullspec,omitempty"`
 	// baseImagePullSecret is the secret used to pull the base image.
-	// must live in the openshift-machine-config-operator namespace
-	// +kubebuilder:validation:Required
-	BaseImagePullSecret ImageSecretObjectReference `json:"baseImagePullSecret"`
+	// must live in the openshift-machine-config-operator namespace if provided.
+	// defaults to using the cluster-wide pull secret if not specified.
+	// +optional
+	BaseImagePullSecret ImageSecretObjectReference `json:"baseImagePullSecret,omitempty"`
```
enxebre, line 110: You are setting this as optional but the ImageSecretObjectReference.Name is required. Which would still let you provide the optional UX since imageSecretObjectReference.Name is not validating a  min length.

If we actually want to it to be optional and omitempty I think this should better be a pointer, and then ImageSecretObjectReference.Name should have a min length. @JoelSpeed thoughts?

PR #2111, opened 2024-11-28
OCPBUGS-44693: Disable ResilientWatchCacheInitialization

p0lyn0mial:
`ResilientWatchCacheInitialization` should be disabled for all API servers. At the moment [we carry a patch ](https://github.com/openshift/kubernetes/blob/master/pkg/features/kube_features.go#L1287-L1288)for kas (which will be replaced by this PR)

The FG will be disabled temporarily until cause of storage error triggering cache reinitialization during bootstrap is fixed in https://issues.redhat.com/browse/OCPBUGS-45126.

There's a separate pre-existing issue causing storage layer errors and watch cache re-initialization during cluster bootstrap. 

With ResilientWatchCacheInitialization enabled, clients (reflectors in particular) are turned away with 429 responses while the watch cache is being repopulated and retry repeatedly. Without this feature, requests hang (consuming priority and fairness "seats") until the watch cache is initialized or they time out. We fail tests when the total number of watch requests during a job exceeds a threshold based on recent historical totals. The systemic 429/retry behavior causes this threshold to be breached. We are temporarily disabling it to reduce noise as it's a symptom and not the cause of the underlying storage errors.

### DIFF
--- a/payload-manifests/featuregates/featureGate-SelfManagedHA-Default.yaml
+++ b/payload-manifests/featuregates/featureGate-SelfManagedHA-Default.yaml
@@ -137,6 +137,9 @@
                         "name": "ProcMountType"
                     },
                     {
+                        "name": "ResilientWatchCacheInitialization"
+                    },
+                    {
                         "name": "RouteAdvertisements"
                     },
                     {
--- a/features.md
+++ b/features.md
@@ -6,6 +6,7 @@
 | MachineAPIMigration| | | | | |  |
 | MachineAPIOperatorDisableMachineHealthCheckController| | | | | |  |
 | MultiArchInstallAzure| | | | | |  |
+| ResilientWatchCacheInitialization| | | | | |  |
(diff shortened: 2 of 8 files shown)


payload-manifests/featuregates/featureGate-SelfManagedHA-Default.yaml
```diff
@@ -136,6 +136,9 @@
                     {
                         "name": "ProcMountType"
                     },
+                    {
+                        "name": "ResilientWatchCacheInitialization"
```
p0lyn0mial, line 140: @JoelSpeed will this FG be available after this PR is included in a promoted image?

p0lyn0mial: yes it was, i have just checked the feature crd in a must-gather and `ResilientWatchCacheInitialization` was disabled. Thanks!

PR #2115, opened 2024-12-04
cpumanager: add featuregate for cpumanager alpha policy options

ffromani:
add feature gate to enable selected users to consume cpumanager policy options of alfpha maturtiy.
Needs to be merged alongside https://github.com/openshift/kubernetes/pull/2136 which enables per-option granularity

for more details: https://github.com/openshift/enhancements/pull/1724


ffromani:
no longer needed per discussion in https://github.com/openshift/enhancements/pull/1724

PR #2123, opened 2024-12-11
OCPBUGS-46380: StaticPodOperatorStatus validation should reject downgrades and concurrent node rollouts

benluddy:


operator/v1/types.go
```diff
@@ -252,6 +252,7 @@ type StaticPodOperatorStatus struct {
 	// +listType=map
 	// +listMapKey=nodeName
 	// +optional
+	// +kubebuilder:validation:XValidation:rule="size(self.filter(status, status.?targetRevision.orValue(0) != 0)) <= 1",message="no more than 1 node status may have a nonzero targetRevision"
```
deads2k, line 255: techpreview.  Let's enforce in an origin test first.

deads2k, line 255: missing integration test

benluddy:
/payload-job periodic-ci-openshift-release-master-ci-4.19-e2e-gcp-ovn

benluddy:
/payload-aggregated periodic-ci-openshift-release-master-ci-4.19-e2e-gcp-ovn 3

benluddy:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.19-e2e-gcp-ovn 3

benluddy:
I caught the static pod installer controller trying to decrease currentRevision and failing validation [here](https://grafana-loki.ci.openshift.org/explore?orgId=1&left=%7B%22datasource%22:%22PCEB727DF2F34084E%22,%22queries%22:%5B%7B%22expr%22:%22%7Binvoker%3D%5C%22openshift-internal-ci%2Fopenshift-api-2123-ci-4.19-e2e-gcp-ovn%2F1867227820942430208%5C%22%7D%20%7C%20unpack%20%7C%3D%20%5C%22currentRevision%5C%22%22,%22refId%22:%22A%22,%22editorMode%22:%22code%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%221734010277578%22,%22to%22:%221734046277581%22%7D%7D) in https://prow.ci.openshift.org/view/gs/test-platform-results/logs/openshift-api-2123-ci-4.19-e2e-gcp-ovn/1867227820942430208.

operator/v1/types.go
```diff
@@ -252,16 +252,19 @@ type StaticPodOperatorStatus struct {
...
 type NodeStatus struct {
 	// nodeName is the name of the node
 	// +required
 	NodeName string `json:"nodeName"`
 
 	// currentRevision is the generation of the most recently successful deployment
+	// +kubebuilder:validation:XValidation:rule="self >= oldSelf",message="must only increase"
 	CurrentRevision int32 `json:"currentRevision"`
 	// targetRevision is the generation of the deployment we're trying to apply
 	TargetRevision int32 `json:"targetRevision,omitempty"`
```
deads2k, line 270: In the future we'll make this only increase too

operator/v1/types.go
```diff
@@ -252,16 +252,19 @@ type StaticPodOperatorStatus struct {
...
 
 // NodeStatus provides information about the current state of a particular node managed by this operator.
+// +kubebuilder:validation:XValidation:rule="has(self.currentRevision) || !has(oldSelf.currentRevision)",message="cannot be unset once set",fieldPath=".currentRevision"
 type NodeStatus struct {
 	// nodeName is the name of the node
 	// +required
 	NodeName string `json:"nodeName"`
 
 	// currentRevision is the generation of the most recently successful deployment
+	// +kubebuilder:validation:XValidation:rule="self >= oldSelf",message="must only increase"
```
p0lyn0mial, line 267: Oh, this is not true. There is a fallback logic in SNO that might revert the CurrentRevision if the new revision fails to install. I think we should revert it; otherwise, it might break an SNO cluster.

wangke19: Yes, this reminds me of an earlier bug https://bugzilla.redhat.com/show_bug.cgi?id=1985997.

benluddy: I'm not really here, but would it be compatible to instead validate that self.currentRevision == oldSelf.targetRevision?

wangke19: @p0lyn0mial A clear logic for fallback we can see https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/startup-monitor.md, CurrentRevision won't decrease, when detecting problems with the new revision, the startup-monitor will copy the pod-manifest of the /etc/kubernetes/static-pods/last-known-good link (or the previous revision if the link does not exist, or don't do anything if there is no previous revision as in bootstrapping) into /etc/kubernetes.

PR #2146, opened 2025-01-14
[release-4.18] OCPBUGS-48383: Add IdleCloseOnResponse field to IngressControllerSpec

frobware:
This is an manual cherry-pick of https://github.com/openshift/api/pull/2102. 

Set `IngressControllerConnectionTerminationPolicy` default to `Deferred` for OCP <= 4.18

In OCP 4.19 (see #2102), the default for `IngressControllerConnectionTerminationPolicy` is `Immediate`. For OCP versions <= 4.18, set the default to `Deferred` to match the behaviour of older releases, where the HAProxy `idle-close-on-response` option is added unconditionally—where `idle-close-on-response` means `Deferred`.

Quoting from #2102 

> Introduce a new knob, IdleConnectionTerminationPolicy, in the IngressController configuration to control how idle connections are handled during router reloads.

> Context
> In [OCPBUGS-32044](https://issues.redhat.com/browse/OCPBUGS-32044), the idle-close-on-response option was unconditionally added to the HAProxy confuguration to address issues with incoming HTTP requests failing during router reloads. This issue primarily affected Apache HTTPClient versions prior to 5.0, which do not gracefully handle connection resets. Adding the option ensured that idle connections were left open to handle one final request before being closed.

> Historically, HAProxy 2.2 maintained idle connections during router reloads by default, allowing requests on those connections to complete even when routing configuration changes were applied. Starting with HAProxy 2.4, the default behaviour changed to close idle connections immediately during soft reloads.

> To accommodate existing clients dependent on the HAProxy 2.2 behaviour, the unconditional addition of idle-close-on-response restored the previous OpenShift status quo, where customers upgrading their OpenShift clusters experienced a behaviour change due to the jump from HAProxy 2.2 to 2.6, which altered the default handling of idle connections during router reloads.

> However, unconditionally enabling idle-close-on-response has now led to issues ([OCPBUGS-43745](https://issues.redhat.com/browse/OCPBUGS-43745)) with Route backend switching. When a Route switches its service backend, requests on persistent connections could continue being routed to the previously active backend due to HAProxy handling these connections in the old process. This behaviour occurs until the connection is closed, either by a new request, the expiration of the client keep-alive, or the expiration of the HAProxy timeout http-keep-alive 300s. While this behaviour is desirable in some cases (e.g., for clients sensitive to connection resets), it can lead to temporary inconsistencies and unexpected routing behaviour during backend switching.

> This PR addresses these regressions by making the behaviour configurable through a new knob.


alebedev87:
https://github.com/openshift/api/pull/2178 takes over from this PR.

frobware:
https://github.com/openshift/api/pull/2178 takes over from this PR.

PR #2151, opened 2025-01-15
MachineOSBuild: make ObjectReference fields optional

yuqi-zhang:
These seem to have been made required during alpha API as it was copied from an existing reference. These do not need to be required and it helps with testing to set them as optional. This also makes it closer to the corev1 ObjectReference type.

This should be also safe from a compatibility perspective, and even if not, this API has not yet been used, so there isn't any dependency on it.


yuqi-zhang:
Some additional context via https://github.com/openshift/machine-config-operator/pull/4756#discussion_r1894310171 . Note that this is not strictly required

yuqi-zhang:
/close

Upon further discussion I think this wouldn't really work since we'd like to key off the resource for the list of ObjectReferences. Also we can always construct the resource and group fields, and the MCO should always ensure those are correct.

PR #2154, opened 2025-01-16
SPLAT-2000: Add configuration for vSphere multi disk thinProvisioned

vr4manta:
SPLAT-2000

### Changes
- Added thinProvisioned to VSphereDisk
- Fixed max size for data disk


machine/v1beta1/types_vsphereprovider.go
```diff
@@ -193,13 +193,34 @@ type VSphereDisk struct {
 	// +required
 	Name string `json:"name"`
 	// sizeGiB is the size of the disk in GiB.
-	// The maximum supported size is 57742 GiB.
+	// The maximum supported size is 16384 GiB.
 	// +kubebuilder:validation:Minimum=1
-	// +kubebuilder:validation:Maximum=57742
+	// +kubebuilder:validation:Maximum=16384
```
everettraven, line 198: Why is this being decreased? Decreasing a maximum value is considered a breaking change as there may be instances of this resource that already specify values greater than 16384.

If you are looking to do ratcheting validation on this field, it is important to understand:
- How many instances of this resource that we keep track of have a value larger than 16384?
- Testing should be added to ensure the ratcheting validation works as expected
- The controller must continue to have logic that supports the original maximum to ensure you do not break existing users.

machine/v1beta1/types_vsphereprovider.go
```diff
@@ -193,13 +193,34 @@ type VSphereDisk struct {
...
 	// +required
 	SizeGiB int32 `json:"sizeGiB"`
+	// provisioningType specifies the provisioning type to be used by default for VirtualMachine volumes exclusively
+	// owned by this VirtualMachine. This does not apply to PersistentVolumeClaim volumes that are created and managed externally.
+	// If not set, the setting will be provided by the default storage policy.
+	// +optional
+	ProvisioningType ProvisioningType `json:"provisioningType,omitempty"`
```
everettraven, line 205: `ProvisioningType` is structured like a field with enum constraints. Is this field supposed to be constrained to specific set of values?

machine/v1beta1/types_vsphereprovider.go
vr4manta: This was accidentally put in here.  It was moved to a separate PR.

vr4manta:
/unhold
Removed hold.  Upstream changes in CAPV are now merged and making MAPI match.

vr4manta:
/assign @JoelSpeed 
Ready for review.  Thanks!

vr4manta:
/assign @everettraven 
Ready for review.  Thanks!

machine/v1beta1/types_vsphereprovider.go
```diff
@@ -198,8 +198,30 @@ type VSphereDisk struct {
 	// +kubebuilder:validation:Maximum=16384
 	// +required
 	SizeGiB int32 `json:"sizeGiB"`
+	// provisioningMode specifies the provisioning type to be used by this vSphere data disk.
+	// If not set, the setting will be provided by the default storage policy.
```
everettraven, line 202: For an enum field, the Godoc generally should follow a format of:
```
provisioningMode specifies the provisioning type to be used by this vSphere data disk.

Allowed values are ...

When set to {Value} ... (repeated for allowed values)
```

machine/v1beta1/types_vsphereprovider.go
```diff
@@ -198,8 +198,30 @@ type VSphereDisk struct {
...
+	// +optional
```
everettraven, line 203: Include an explicit mention in the Godoc that the field is optional. Users using tools like `oc explain ...` are not able to see the markers we use so an explicit mention is helpful for users to know which fields are optional or required

machine/v1beta1/types_vsphereprovider.go
```diff
@@ -198,8 +198,30 @@ type VSphereDisk struct {
...
+	ProvisioningMode ProvisioningMode `json:"provisioningMode,omitempty"`
 }
 
+// provisioningMode represents the various provisioning types available to a VMs disk.
+// +kubebuilder:validation:Enum=Thin;Thick;EagerlyZeroed
+type ProvisioningMode string
+
+var (
```
everettraven, line 211:
```suggestion
const (
```

machine/v1beta1/types_vsphereprovider.go
```diff
@@ -198,8 +198,30 @@ type VSphereDisk struct {
...
+	// ThinProvisioningMode creates the disk using thin provisioning. This means a sparse (allocate on demand)
+	// format with additional space optimizations.
+	ThinProvisioningMode ProvisioningMode = "Thin"
```
everettraven, line 214: The most common pattern I've seen for naming of constants for a type alias is `{TypeAlias}{ConstantName}`. In this case it would be:
```suggestion
	ProvisioningModeThin ProvisioningMode = "Thin"
```

I would encourage consistency with how this is commonly done throughout the various OpenShift repos.

machine/v1beta1/types_vsphereprovider.go
vr4manta: @everettraven / @JoelSpeed , I can make these changes.  I did not see any existing enum in API doing this.  Do we have this requirement documented anywhere for future enhancers?  I'll be sure to add this requirement to our team's documentation on working with API to reduce missed points.

machine/v1beta1/types_vsphereprovider.go
vr4manta: I'll update this.

machine/v1beta1/types_vsphereprovider.go
vr4manta: Good catch.  I'll fix that.

machine/v1beta1/types_vsphereprovider.go
vr4manta: I can make this change, but will deviate from upstream CAPV.  The actual values will still match, just slight var difference.  (cc: @JoelSpeed).

machine/v1beta1/types_vsphereprovider.go
everettraven: The general guidance for writing user readable godoc is here: https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#write-user-readable-documentation-in-godoc

You don't have to have it match exactly what I put, but that formatting is a good baseline. As long as the Godoc is reasonably structured and answers the key questions in the api conventions doc, it should be OK.

machine/v1beta1/types_vsphereprovider.go
JoelSpeed: I agree with @everettraven on this one to be consistent with OpenShift APIs, when it comes to converting to upstream, this won't really make a difference to us

machine/v1beta1/types_vsphereprovider.go
JoelSpeed: >  I did not see any existing enum in API doing this

There are definitely many around that I've API reviewed, if you search for `Allowed values are` in this repo there are many hits

> Do we have this requirement documented anywhere for future enhancers?

Somewhat covered in the general user readable docs section that @everettraven linked, but I think we could likely add more

vr4manta: Thank, i'll use that link in my doc as well.

machine/v1beta1/types_vsphereprovider.go
vr4manta: Sounds good.  I'll keep this in mind.

machine/v1beta1/types_vsphereprovider.go
vr4manta: @everettraven how do these changed look?

machine/v1beta1/types_vsphereprovider.go
```diff
@@ -198,8 +198,34 @@ type VSphereDisk struct {
 	// +kubebuilder:validation:Maximum=16384
 	// +required
 	SizeGiB int32 `json:"sizeGiB"`
+	// provisioningMode is an optional field that specifies the provisioning type to be used by this vSphere data disk.
+	// Allowed values are Thin, Thick, and EagerlyZeroed.
+	// When set to Thin, the disk will be made using thin provisioning allocating the bare minimum space.
+	// When set to Thick, the full disk size will be allocated when disk is created.
+	// When set to EagerlyZeroed, the full disk size will be allocated and all bytes will be zeroed.
+	// If not set, the setting will be provided by the default storage policy.
```
everettraven, line 206: What is the "default storage policy"? How does a user find out what this will default to if they don't set it? Can this default change over time?

vr4manta: its based on vsphere itself.  Another layer of complexity is if you create a template, you can also set a clone level policy that will override the vsphere default and will trickle to these data disks.

everettraven: This feels like something that would good to clearly state for the users that may be reading this description. I'm not sure on what would be the best wording, but maybe something along the lines of:
> If not set, vSphere itself will be responsible for choosing the storage policy based on its defaults or platform configurations in place that may override these defaults.

WDYT?

vr4manta: I made a small adjustment.  I'm talking w/ team to see if they would change more to it.

machine/v1beta1/types_vsphereprovider.go
```diff
@@ -198,8 +198,34 @@ type VSphereDisk struct {
...
+	// ProvisioningModeThin creates the disk using thin provisioning. This means a sparse (allocate on demand)
+	// format with additional space optimizations.
+	ProvisioningModeThin ProvisioningMode = "Thin"
+
+	// ProvisioningModeThick creates the disk with all space allocated.
+	ProvisioningModeThick ProvisioningMode = "Thick"
+
+	// ProvisioningModeEagerlyZeroed creates the disk using eager zero provisioning. An eager zeroed thick disk
+	// has all space allocated and wiped clean of any previous contents on the physical media at
+	// creation time. Such disks may take longer time during creation compared to other disk formats.
```
everettraven, line 225: This description of the `EagerlyZeroed` provisioning mode is much more descriptive than the one that would currently be shown to users. I would recommend taking this description and using it in the actual field description where you say "When set to EagerlyZeroed ...".

Particularly the part about this provisioning mode taking longer during creation compared to the other disk formats would be helpful information for users to know when exploring this API.

everettraven (commented):
I've only got one more comment, but overall this looks pretty good.

I would like to see both my outstanding comments be addressed prior to getting my stamp of approval and tagging in Joel for a pass

machine/v1beta1/types_vsphereprovider.go
vr4manta: Ok, i can update that.

vr4manta:
@JoelSpeed / @everettraven , ready for next round of review.  Thanks!

everettraven (commented):
Other than a concern on some terminology I think this looks good.

I'll go ahead and pull @JoelSpeed in for a pass. If Joel thinks it looks good and doesn't share the same concerns I do for the terminology I won't block this.

machine/v1beta1/types_vsphereprovider.go
```diff
@@ -198,8 +198,34 @@ type VSphereDisk struct {
 	// +kubebuilder:validation:Maximum=16384
 	// +required
 	SizeGiB int32 `json:"sizeGiB"`
+	// provisioningMode is an optional field that specifies the provisioning type to be used by this vSphere data disk.
+	// Allowed values are "Thin", "Thick", "EagerlyZeroed", and omitted.
+	// When set to Thin, the disk will be made using thin provisioning allocating the bare minimum space.
+	// When set to Thick, the full disk size will be allocated when disk is created.
+	// When set to EagerlyZeroed, the disk will be created using eager zero provisioning. An eager zeroed thick disk has all space allocated and wiped clean of any previous contents on the physical media at creation time. Such disks may take longer time during creation compared to other disk formats.
+	// When omitted, no setting will be applied to the data disk and the provisioning mode for the disk will be provided by the default storage policy.
```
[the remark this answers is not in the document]
everettraven: If you are confident that users will understand what you mean here, I'm fine with it. I would probably put something like "the default storage policy configured in vSphere" to point a user to vSphere itself as the place to look for this default.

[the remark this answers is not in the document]
vr4manta: I can add that additional context.

everettraven (approved):
Changes LGTM. Officially tagging in @JoelSpeed for a pass.

JoelSpeed:
@vr4manta Does this field already exist in the upstream CAPI provider? And do you have a PR to MAO to implement the appropriate validation of the schema?

vr4manta:
> @vr4manta Does this field already exist in the upstream CAPI provider? And do you have a PR to MAO to implement the appropriate validation of the schema?

Yes, I've made the changes upstream first.  The link to that PR is in the top comment along w/ the dependent PRs for implementing these changes.  I have reconciler, but I'll look more into the webhook to make sure validation is done.

PR #2167, opened 2025-01-23
DNM: test rbr bootstrap-required=false

everettraven:


everettraven:
Successfully launched a 4.19 cluster with clusterbot with this PR. The `RoleBindingRestriction` CRD is likely not required as part of bootstrapping, but it might still be worth making some updates to cluster-authentication-operator such that it is still included in the bootstrapping process since the kube-apiserver has an admission plugin that uses this CRD.

PR #2171, opened 2025-01-27
(Work in progess) MCO-1518: Promote MCN to GA

isabella-janssen:


machineconfiguration/v1/types_machineconfignode.go
```diff
@@ -0,0 +1,266 @@
...
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	// +optional
+	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+	// observedGeneration represents the generation observed by the controller.
+	// This field is updated when the controller observes a change to the desiredConfig in the configVersion of the machine config node spec.
+	// +required
+	ObservedGeneration int64 `json:"observedGeneration"`
```
yuqi-zhang, line 122: I think we generally want to validate that this is >0 and incremental only. Probably make optional as well to match elsewhere

machineconfiguration/v1/types_machineconfignode.go
```diff
@@ -0,0 +1,266 @@
...
+	// pool contains a reference to the machine config pool that this machine config node's
+	// referenced node belongs to.
+	// +required
+	Pool MCOObjectReference `json:"pool"`
+
+	// configVersion holds the desired config version for the node targeted by this machine config node resource.
+	// The desired version represents the machine config the node will attempt to update to. This gets set before the machine config operator validates
+	// the new machine config against the current machine config.
+	// +required
+	ConfigVersion MachineConfigNodeSpecMachineConfigVersion `json:"configVersion"`
```
yuqi-zhang, line 87: Hmm, I'm not sure I really understand why we had this originally. It looks like the status field has current and desired (rendered machineconfigs), and then this spec also has a desired field, and the difference is... the desired here is set before the node validates, and the desired in status is post-validation? Then what's the intent of the "current" in status?

Maybe we should consider consolidating these.

machineconfiguration/v1/types_machineconfignode.go
```diff
@@ -0,0 +1,266 @@
...
+	// +optional
+	DesiredGeneration int32 `json:"desiredGeneration,omitempty"`
+	// lastFailedGeneration is the generation of the most recent pinned image set that failed to be pulled and pinned on this node.
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	LastFailedGeneration int32 `json:"lastFailedGeneration,omitempty"`
+	// lastFailedGenerationErrors is a list of errors why the lastFailed generation failed to be pulled and pinned.
+	// +kubebuilder:validation:MaxItems=10
+	// +optional
+	LastFailedGenerationErrors []string `json:"lastFailedGenerationErrors,omitempty"`
```
yuqi-zhang, line 166: This seems a somewhat weird way to keep track of a history of failures? I wonder if this should just be a singular error condition here

machineconfiguration/v1/types_machineconfignode.go
```diff
@@ -0,0 +1,266 @@
...
+	// +optional
+	CurrentGeneration int32 `json:"currentGeneration,omitempty"`
+	// desiredGeneration is the generation of the pinned image set that is targeted to be pulled and pinned on this node.
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	DesiredGeneration int32 `json:"desiredGeneration,omitempty"`
+	// lastFailedGeneration is the generation of the most recent pinned image set that failed to be pulled and pinned on this node.
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	LastFailedGeneration int32 `json:"lastFailedGeneration,omitempty"`
```
yuqi-zhang, line 162: Let's say we're on generation 2, and trying to update to 3, and we're failing, I would expect this to be
current: 2
desired: 3
lastfailed: 3

But then let's say we have a generation 4 now. Do we just block on this error? Or do we try 4? If 4 is successful, do we have:
current: 4
desired: 4
lastfailed: 3
to keep a history of errors? Or do we just clear the failure?

Alternatively, if we don't want to move ahead with newer gens when failing, this would always equate to desired, which seems a bit redundant. Maybe we should reconsider this field's usage

machineconfiguration/v1/types_machineconfignode.go
```diff
@@ -0,0 +1,266 @@
...
+	// with an alphanumeric character, and be at most 253 characters in length.
+	// +kubebuilder:validation:MaxLength:=253
+	// +kubebuilder:validation:XValidation:rule="!format.dns1123Subdomain().validate(self).hasValue()",message="a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character."
+	// +required
+	Name string `json:"name"`
+}
+
+// StateProgress is each possible state for each possible MachineConfigNodeType
+// UpgradeProgression Kind will only use the "MachinConfigPoolUpdate..." types for example
+type StateProgress string
```
yuqi-zhang, line 233: hmm, I think this should be a `// +enum` ?

machineconfiguration/v1/types_machineconfignode.go
```diff
@@ -0,0 +1,266 @@
...
+	// +required
+	ObservedGeneration int64 `json:"observedGeneration"`
+	// configVersion describes the current and desired machine config for this node.
+	// The current version represents the current machine config for the node and is updated after a successful update.
+	// The desired version represents the machine config the node will attempt to update to.
+	// This desired machine config has been compared to the current machine config and has been validated by the machine config operator as one that is valid and that exists.
+	// +required
+	ConfigVersion MachineConfigNodeStatusMachineConfigVersion `json:"configVersion"`
+	// pinnedImageSets describes the current and desired pinned image sets for this node.
+	// The current version is the generation of the pinned image set that has most recently been successfully pulled and pinned on this node.
```
yuqi-zhang, line 130: These two lines of comments are a bit confusing. I think they are trying to describe some of the subfield's details. We should probably be a bit more general in describing what the status is holding

machineconfiguration/v1/types_machineconfignode.go
```diff
@@ -0,0 +1,266 @@
...
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=100
+	// +optional
+	PinnedImageSets []MachineConfigNodeSpecPinnedImageSet `json:"pinnedImageSets,omitempty"`
+}
+
+// MCOObjectReference holds information about an object the MCO either owns
+// or modifies in some way
+type MCOObjectReference struct {
+	// name is the object name.
```
yuqi-zhang, line 100: Maybe beef up description a bit here

machineconfiguration/v1/types_machineconfignode.go
```diff
@@ -0,0 +1,266 @@
...
+	// +required
+	Pool MCOObjectReference `json:"pool"`
+
+	// configVersion holds the desired config version for the node targeted by this machine config node resource.
+	// The desired version represents the machine config the node will attempt to update to. This gets set before the machine config operator validates
+	// the new machine config against the current machine config.
+	// +required
+	ConfigVersion MachineConfigNodeSpecMachineConfigVersion `json:"configVersion"`
+
+	// pinnedImageSets holds the desired pinned image sets that this node should pin and pull.
```
yuqi-zhang, line 89: Maybe specify that this is just the name, and how this should be set by a user (I think that's the expected workflow)

yuqi-zhang (commented):
Did a first pass, some comments (mostly historical questions from alpha1 days)

machineconfiguration/v1/types_machineconfignode.go
isabella-janssen: My impression was this field is intentionally general, as it can refer to the name of different MCO managed objects. For MCN, it looks like this refers to both [node names](https://github.com/isabella-janssen/openshift-api/blob/369015f3699ec7d2bd0e59c129d250284a3c4b87/machineconfiguration/v1/types_machineconfignode.go#L74-L76) and [pool names](https://github.com/isabella-janssen/openshift-api/blob/369015f3699ec7d2bd0e59c129d250284a3c4b87/machineconfiguration/v1/types_machineconfignode.go#L78-L81).

machineconfiguration/v1/types_machineconfignode.go
isabella-janssen: Updated to be consistent with [mosc type](https://github.com/openshift/api/blob/7bb8a0201a5666537b835d2b7c27709e943ce302/machineconfiguration/v1/types_machineosconfig.go#L108-L112).

machineconfiguration/v1/types_machineconfignode.go
isabella-janssen: Updated.

machineconfiguration/v1/types_machineconfignode.go
isabella-janssen: Updated, but it can be future revised if it is still inadequate.

machineconfiguration/v1/types_machineconfignode.go
isabella-janssen: Here are the initial results I got from testing the process of applying a bad PIS then correcting it.

**Apply a PIS with images that _cannot_ be pulled:**
```
Status:
…
  Pinned Image Sets:
    Desired Generation:      1
    Last Failed Generation:  1
    Last Failed Generation Errors:
      failed to execute podman manifest inspect for "quay.io/rh-ee-ijanssen/machine-config-operator@sha256:65d3a308767b1773b6e3ead2ec1bcae499dde6ef085753d7e20e685f78841079": exit status 125
    Name:  test-pinned
```

**Apply a PIS with images that _can_ be pulled:**
```
Status:
…
  Pinned Image Sets:
    Current Generation:      2
    Desired Generation:      2
    Last Failed Generation:  0
    Name:                    test-pinned
```

To me it looks like there are a few notable initial findings (that I would still like to look into more):
- It appears that when there’s an error `Last Failed Generation` is populated and is synonymous with `Desired Generation`, but `Current Generation` is not populated.
- It looks like after an error the `Last Failed Generation` value is reset to 0 and the error message is cleared. In this case `Desired Generation` and `Current Generation` are populated and synonymous.
- Right now the `Last Failed Generation` and `Last Failed Generation Errors` fields don’t seem to be tracking history. If we want a history we will need to update how they are set/reset.
- In both cases (success & failure), it seems like the `... Generation` values are just duplicates, so they might be worth reconsidering how they are used.

machineconfiguration/v1/types_machineconfignode.go
isabella-janssen: More information about the intended purpose of the Spec.ConfigVersion.Desired field is described in the [original enhancement](https://github.com/openshift/enhancements/blob/2890cccf20ebcb94fce901f7afb170ca680aa2d9/enhancements/machine-config/machine-config-node.md?plain=1#L132). I'd like to leave this field as is for now and get some feedback from more people throughout the review process.

machineconfiguration/v1/types_machineconfignode.go
yuqi-zhang: Thanks for investigating! Given this context, I think I would lean towards not keeping it as history in a field like this. I suggest modifing this to surface the current error state if it is erroring, and if not, to clear it and keep it empty as a single field.

machineconfiguration/v1/types_machineconfignode.go
yuqi-zhang: @djoshy and I discussed this briefly as well. My instinct is that it would make more sense to have the spec reflect the desired config, and the status reflect the current config, to match the MCP/node annotations. (More concretely, drop the desired from MachineConfigNodeStatusMachineConfigVersion)

machineconfiguration/v1/types_machineconfignode.go
```diff
@@ -0,0 +1,278 @@
...
+	// MachineConfigNodeUpdateResumed describes a machine that has resumed normal processes
+	MachineConfigNodeUpdateResumed StateProgress = "Resumed"
+	// MachineConfigNodeUpdateCompatible the part of the preparing phase where the mco decides whether it can update
+	MachineConfigNodeUpdateCompatible StateProgress = "UpdateCompatible"
+	// MachineConfigNodeUpdateDrained describes the part of the inprogress phase where the node drains
+	MachineConfigNodeUpdateDrained StateProgress = "Drained"
+	// MachineConfigNodeUpdateFilesAndOS describes the part of the inprogress phase where the nodes file and OS config change
+	MachineConfigNodeUpdateFilesAndOS StateProgress = "AppliedFilesAndOS"
+	// MachineConfigNodeUpdateCordoned describes the part of the completing phase where the node cordons
+	MachineConfigNodeUpdateCordoned StateProgress = "Cordoned"
```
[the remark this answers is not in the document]
isabella-janssen: These fields are flipped to `Unknown` if the node fails to cordon or uncordon. Alternatively, they flip to `True` once the node cordon or uncordon is complete. As with all MCN statuses, when a node is fully updated (desired config = current config), these statuses are listed as `False`.

1. [Cordon failed](https://github.com/openshift/machine-config-operator/blob/2320deabaac93c8bae12d4d8c35842965720bac8/pkg/controller/drain/drain_controller.go#L393-L402)
2. [Uncordon failed](https://github.com/openshift/machine-config-operator/blob/2320deabaac93c8bae12d4d8c35842965720bac8/pkg/controller/drain/drain_controller.go#L313-L327)
3. [Cordon successfully completed](https://github.com/openshift/machine-config-operator/blob/2320deabaac93c8bae12d4d8c35842965720bac8/pkg/controller/drain/drain_controller.go#L408-L415)
4. [Uncordon successfully completed](https://github.com/openshift/machine-config-operator/blob/2320deabaac93c8bae12d4d8c35842965720bac8/pkg/controller/drain/drain_controller.go#L329-L336)

How often does a node fail to cordon/uncordon? Is it helpful to have this level of granularity for understanding why a node fails to update?

machineconfiguration/v1/types_machineconfignode.go
```diff
@@ -0,0 +1,278 @@
...
+	// MachineConfigNodeUpdatePostActionComplete describes a machine that has executed its post update action
+	MachineConfigNodeUpdatePostActionComplete StateProgress = "UpdatePostActionComplete"
+	// MachineConfigNodeUpdateComplete describes a machine that has completed the core parts of an upgrade
+	MachineConfigNodeUpdateComplete StateProgress = "UpdateComplete"
+	// MachineConfigNodeUpdated describes a machine that has a matching desired and current config after executing an update
+	MachineConfigNodeUpdated StateProgress = "Updated"
+	// MachineConfigNodeUpdateResumed describes a machine that has resumed normal processes
+	MachineConfigNodeUpdateResumed StateProgress = "Resumed"
+	// MachineConfigNodeUpdateCompatible the part of the preparing phase where the mco decides whether it can update
+	MachineConfigNodeUpdateCompatible StateProgress = "UpdateCompatible"
```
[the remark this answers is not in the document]
isabella-janssen: The `MachineConfigNodeUpdateCompatible` phase is the only child phase of the `MachineConfigNodeUpdatePrepared` phase. Therefore, the reported status of `UpdatePrepared` is always a mirror of `UpdateCompatible`. These phases report `Unknown` when the updates are not compatible and `True` when the updates are fine to proceed.

- [Compatibility check failed due to reconcilable error](https://github.com/openshift/machine-config-operator/blob/2320deabaac93c8bae12d4d8c35842965720bac8/pkg/daemon/update.go#L1096-L1105)
- [Compatibility check failed due to error calculating post config change actions](https://github.com/openshift/machine-config-operator/blob/2320deabaac93c8bae12d4d8c35842965720bac8/pkg/daemon/update.go#L1133-L1151)
- [Compatibility check succeeds](https://github.com/openshift/machine-config-operator/blob/2320deabaac93c8bae12d4d8c35842965720bac8/pkg/daemon/update.go#L1177-L1185)

Since the prepared phase only encapsulates the compatibility check, there would be no loss in status update granularity, from my perspective, if the `MachineConfigNodeUpdateCompatible` phase is removed.

machineconfiguration/v1/types_machineconfignode.go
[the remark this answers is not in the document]
yuqi-zhang: I'll try to continue any discussion via https://github.com/openshift/api/pull/2201 instead, apologies for splitting the conversation. Just to cap this point out,

> How often does a node fail to cordon/uncordon? Is it helpful to have this level of granularity for understanding why a node fails to update?

This almost never happens (only if the API itself is inaccessible), and the MCO generally retries and drowns any temporary error out almost immediately. The failures we see here is (very rarely) on the opposite side, where the MCD thinks its uncordoned the node, but was not able to successfully do so, leaving the state of the node as cordoned until manual intervention. For these cases, the best we could determine was that there was some weird API behaviour.

Alternatively, if the intent of these states is to bridge the existing gap where a cordon on the node came from not the MCO, they may be worth it to keep. The idea being that if a node was manually cordoned or cordoned by another component, but the MCN's MachineConfigNodeUpdateUncordoned is true, then we know the MCC had successfully done so and the current cordon's ownership is us.

isabella-janssen:
/close

Closing in favor of #2255.

PR #2183, opened 2025-01-30
no-jira: Validate AWS resource tag keys with aws: prefix

patrickdillon:
AWS docs indicate that tag keys cannot be prefix with aws:. See:

Using this key prefix leads to an AWS API error indicating the prefix
is reserved for AWS system usage. This commit adds API validation
and ratecheting tests, as the key was previously allowed.

This was originally included in #2124 but cannot land due to a bug addressed by https://github.com/openshift/kubernetes/pull/2167, so 

Depends on #2124 
Depends on https://github.com/openshift/kubernetes/pull/2167

### DIFF
--- a/config/v1/tests/infrastructures.config.openshift.io/AAA_ungated.yaml
+++ b/config/v1/tests/infrastructures.config.openshift.io/AAA_ungated.yaml
@@ -1527,3 +1527,31 @@ tests:
                 value: value*
             type: AWS
       expectedStatusError: "invalid AWS resource tag value. The string can contain only the set of alphanumeric characters, space (' '), '_', '.', '/', '=', '+', '-', ':', '@'"
+    - name: Should not be able to create an aws resourcetag with aws prefix in key
+      initial: |
+        apiVersion: config.openshift.io/v1
+        kind: Infrastructure
+        spec:
+          platformSpec:
+            aws: {}
+            type: AWS
+      updated: |
+        apiVersion: config.openshift.io/v1
+        kind: Infrastructure
+        spec:
+          platformSpec:
+            aws: {}
+            type: AWS
+        status:
+          controlPlaneTopology: HighlyAvailable
+          cpuPartitioning: None
+          infrastructureTopology: HighlyAvailable
+          platform: AWS
+          platformStatus:
+            aws:
+              region: us-east-1
+              resourceTags:
+              - key: aws:key
+                value: value with space
+            type: AWS
+      expectedStatusError: "the prefix 'aws:' is reserved for AWS system usage and cannot be used at the beginning of a key"
--- a/config/v1/tests/infrastructures.config.openshift.io/AAA_ungated.yaml
+++ b/config/v1/tests/infrastructures.config.openshift.io/AAA_ungated.yaml
@@ -1,6 +1,8 @@
 apiVersion: apiextensions.k8s.io/v1 # Hack because controller-gen complains if we don't have this
 name: "Infrastructure"
 crdName: infrastructures.config.openshift.io
+featureGates:
+- -AWSClusterHostedDNS
 tests:
   onCreate:
     - name: Should be able to create a minimal Infrastructure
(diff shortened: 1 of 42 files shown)


patrickdillon:
/hold
for dependencies

JoelSpeed:
@patrickdillon Did you intend to move this PR forward?

patrickdillon:
> @patrickdillon Did you intend to move this PR forward?

Yes. I see now the dependent work has merged. So I can rebase this; I am tied up with a conference this week so I will try to get this once I'm back next week.

JoelSpeed:
@patrickdillon Just a reminder to move this one along when you have a moment

patrickdillon:
/hold cancel

@JoelSpeed finally ready, thanks!

config/v1/tests/infrastructures.config.openshift.io/AAA_ungated.yaml
```diff
@@ -1837,3 +1837,31 @@ tests:
                 value: value*
             type: AWS
       expectedStatusError: "invalid AWS resource tag value. The string can contain only the set of alphanumeric characters, space (' '), '_', '.', '/', '=', '+', '-', ':', '@'"
+    - name: Should not be able to create an aws resourcetag with aws prefix in key
```
JoelSpeed, line 1840: Please also add a ratcheting test (see readme in the tests dir) to prove that existing values are not affected by this change. You should show that
* Existing entries persist when a new item is added to the list
* Additional bad entries are not allowed when there is an existing bad entry
* The bad entry can be updated to a valid entry

config/v1/tests/infrastructures.config.openshift.io/AAA_ungated.yaml
```diff
@@ -1837,3 +1837,31 @@ tests:
...
+          infrastructureTopology: HighlyAvailable
+          platform: AWS
+          platformStatus:
+            aws:
+              region: us-east-1
+              resourceTags:
+              - key: aws:key
+                value: value with space
+            type: AWS
+      expectedStatusError: "the prefix 'aws:' is reserved for AWS system usage and cannot be used at the beginning of a key"
```
JoelSpeed: Nit, missing final new line char

config/v1/tests/infrastructures.config.openshift.io/AAA_ungated.yaml
patrickdillon: Ok, I added the ratcheting validation tests, but am stuck on the first one (the other two work ok). In order to allow existing bad values, I updated the validation in https://github.com/openshift/api/pull/2183/commits/34f8d72eb69ec24785ba7dafe08a989e55a6f774 but apparently this is invalid, as the tests fail with:

```
Invalid value: "self == oldSelf || !self.startsWith('aws:')": oldSelf cannot be used on the uncorrelatable portion of the schema within spec.validation.openAPIV3Schema.properties[status].properties[platformStatus].properties[aws].properties[resourceTags]
```
I see in [these k8s docs](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/) that this means the rule cannot be applied:

> Errors will be generated on CRD writes if a schema node contains a transition rule that can never be applied, e.g. "oldSelf cannot be used on the uncorrelatable portion of the schema within path".

I'm not sure how to resolve this issue. Any insight?

patrickdillon: @JoelSpeed claude helped me figure out that I needed to change ResourceTags to listType=map, but now the actual validation of the aws: prefix does not seem to be working

```
 [config.openshift.io/v1, Resource=infrastructures][ClusterProfiles=Hypershift,SelfManagedHA][FeatureSet="Default"][FeatureGate=-AWSClusterHostedDNSInstall][File=0000_10_config-operator_01_infrastructures-Default.crd.yaml] Infrastructure On Update [It] Should not be able to create an aws resourcetag with aws prefix in key
/Users/padillon/go/src/github.com/openshift/api/tests/generator.go:345

  [FAILED] Expected an error, got nil
  In [It] at: /Users/padillon/go/src/github.com/openshift/api/tests/generator.go:321 @ 10/14/25 22:00:52.111
```
Stuck on this.

config/v1/types_infrastructure.go
```diff
@@ -543,6 +545,7 @@ type AWSResourceTag struct {
 	// key sets the key of the AWS resource tag key-value pair. Key is required when defining an AWS resource tag.
 	// Key should consist of between 1 and 128 characters, and may
 	// contain only the set of alphanumeric characters, space (' '), '_', '.', '/', '=', '+', '-', ':', and '@'.
+	// Key must not be start with 'aws:'.
```
patrickdillon: will move this comment on next update

patrickdillon:
@JoelSpeed these fields are supposed to be immutable. Should i just add validation to enforce that?

config/v1/tests/infrastructures.config.openshift.io/AAA_ungated.yaml
patrickdillon: I went ahead and side-stepped the issue by validating for immutability. IMHO that makes more sense, as post-install updates are not supported. On the other hand, I'm not sure if there are issues with making a field immutable post-GA; the intention was always for these fields to be immutable, it just wasn't validated.

config/v1/types_infrastructure.go
```diff
@@ -503,6 +503,7 @@ type AWSPlatformSpec struct {
 }
 
 // AWSPlatformStatus holds the current status of the Amazon Web Services infrastructure provider.
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.resourceTags) && !has(self.resourceTags) || has(oldSelf.resourceTags) && has(self.resourceTags)",message="resourceTags may only be configured during installation"
```
JoelSpeed: Equivalent but cheaper?
```suggestion
// +kubebuilder:validation:XValidation:rule="has(oldSelf.resourceTags) == has(self.resourceTags)",message="resourceTags may only be configured during installation"
```

config/v1/tests/infrastructures.config.openshift.io/AAA_ungated.yaml
JoelSpeed: We still need a ratcheting test to show that anyone who already has a value with `aws:` in the prefix won't be broken by this change

patrickdillon: Oh right, that makes sense: so we don't brick the rest of the aws platform status from being updated. Because I made resourceTags immutable, we can't cover the three expected cases in your original message. Does the check added in d58295417a7b9e29243df7f4919592dc22ad0a1e look good/sufficient? Or what else would you have in mind.

JoelSpeed: Yep, that covers it, thanks

patrickdillon:
/hold cancel

@JoelSpeed I just noticed this never merged due to essentially a rebase conflict. I rebased and reran `make update`. PTAL

JoelSpeed:
@patrickdillon Some generation appears to be out of date here

patrickdillon:
> @patrickdillon Some generation appears to be out of date here

ACK. This PR really doesn't want to merge. I went to rebase and regenerate, but it looks like `make update` is broken on master. I opened a thread in the api review channel.

patrickdillon:
/hold cancel

Not broken at all, I was just confused.

Pushed the newly generated files

config/v1/types_infrastructure.go
```diff
@@ -581,9 +583,11 @@ type AWSResourceTag struct {
 	// key sets the key of the AWS resource tag key-value pair. Key is required when defining an AWS resource tag.
 	// Key should consist of between 1 and 128 characters, and may
 	// contain only the set of alphanumeric characters, space (' '), '_', '.', '/', '=', '+', '-', ':', and '@'.
+	// Key must not be start with 'aws:'.
```
JoelSpeed: Coderabbit found this one tbf
```suggestion
	// Key must not start with 'aws:'.
```

config/v1/types_infrastructure.go
```diff
@@ -581,9 +583,11 @@ type AWSResourceTag struct {
...
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=128
 	// +kubebuilder:validation:XValidation:rule=`self.matches('^[0-9A-Za-z_.:/=+-@ ]+$')`,message="invalid AWS resource tag key. The string can contain only the set of alphanumeric characters, space (' '), '_', '.', '/', '=', '+', '-', ':', '@'"
+	// +kubebuilder:validation:XValidation:rule=`!self.startsWith('aws:')`,message="the prefix 'aws:' is reserved for AWS system usage and cannot be used at the beginning of a key"
```
JoelSpeed, line 590: Do we need to consider casing on the AWS prefix?

config/v1/types_infrastructure.go
patrickdillon: fixed

config/v1/types_infrastructure.go
patrickdillon: In my opinion: no. The docs say literally `aws:`. We could argue that it is inconclusive, and test whether casing is relevant, but I don't think it's worth the time or worth holding this up. If we do want to pursue it, it would be better to pursue it as an installer bug, as this API should be used by users.

JoelSpeed:
Some code generation is out of date here, but otherwise LGTM

patrickdillon:
Hopefully this fixes it up

JoelSpeed:
`./hack/update-payload-crds.sh` needs running

patrickdillon:
@JoelSpeed as much as I'd like to get this merged, I view this as non critical so would lean toward waiting until the ack requirement is removed. not a strong opinion though...

JoelSpeed:
This needs regenerating now that the new OKD featureset is in

patrickdillon:
> This needs regenerating now that the new OKD featureset is in

Pushed! Watching tests in case I missed anything.

patrickdillon:
unit tests passed

PR #2184, opened 2025-01-30
no-jira: correct default CIDR for ipv6 join subnet in API descriptors

tthvo:
The correct default CIDR for ipv6 join subnet is `fd98::/64`. See reference [0] and [1].

However, API markers is showing the incorrect `fd98::/48`. For official documentation, release version >= 4.16 is correct, but 4.15 is not. See reference [2].

## Notes

This PR only fixes the API markers to avoid confusion when official docs (correct source) and API descriptor (i.e. `oc explain`) are showing different values.

## References

[0] https://github.com/openshift/cluster-network-operator/blob/9ee349a119c769bc390d6699b61e59094ddbc2c7/pkg/network/ovn_kubernetes.go#L73 (4.19)
[1] https://docs.openshift.com/container-platform/4.17/networking/networking_operators/cluster-network-operator.html (4.17 - latest)
[2] https://docs.openshift.com/container-platform/4.15/networking/cluster-network-operator.html (4.15)


patrickdillon:
Based on
https://github.com/openshift/cluster-network-operator/blob/9ee349a119c769bc390d6699b61e59094ddbc2c7/pkg/network/ovn_kubernetes.go#L73

LGTM but probably better to get someone from SDN to tag @pliurh can you confirm? should be quick

PR #2195, opened 2025-02-06
CCXDEV-10455: update insightsDataGather to v1alpha2 with GathererConfig

opokornyy:
This PR introduces v1alpha2 for insightsDataGather, adding GathererConfig struct to enable/disable gatherers.


config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,105 @@
...
+	// When omitted, this means no opinion and the platform is left to choose a reasonable default, which is subject to change over time.
+	// The current default is None.
+	// +optional
+	DataPolicy DataPolicy `json:"dataPolicy,omitempty"`
+	// gatherers is a list of gatherers configurations.
+	// The particular gatherers IDs can be found at https://github.com/openshift/insights-operator/blob/master/docs/gathered-data.md.
+	// Run the following command to get the names of last active gatherers:
+	// "oc get insightsoperators.operator.openshift.io cluster -o json | jq '.status.gatherStatus.gatherers[].name'"
+	// +optional
+	Gatherers []GathererConfig `json:"gatherers"`
```
opokornyy, line 58: This field was changed from the v1alpha1

ncaak (approved):
/lgtm

config/v1alpha2/Makefile
```diff
@@ -0,0 +1,3 @@
+.PHONY: test
+test:
+	make -C ../../tests test GINKGO_EXTRA_ARGS=--focus="config.openshift.io/v1alpha1"
```
JoelSpeed, line 3:
```suggestion
	make -C ../../tests test GINKGO_EXTRA_ARGS=--focus="config.openshift.io/v1alpha2"
```

config/v1alpha2/doc.go
```diff
@@ -0,0 +1,8 @@
+// +k8s:deepcopy-gen=package,register
+// +k8s:defaulter-gen=TypeMeta
+// +k8s:openapi-gen=true
+
+// +kubebuilder:validation:Optional
```
JoelSpeed, line 5: This is generally bad practice, please drop the package level setting of default optional vs required

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,115 @@
...
+	// status holds observed values from the cluster. They may not be overridden.
+	// +optional
+	Status InsightsDataGatherStatus `json:"status"`
+}
+
+type InsightsDataGatherSpec struct {
+	// gatherConfig spec attribute includes all the configuration options related to
+	// gathering of the Insights data and its uploading to the ingress.
+	// +optional
+	GatherConfig GatherConfig `json:"gatherConfig,omitempty"`
```
JoelSpeed, line 37: Did we ever work out if we needed any siblings to this field? I wonder if we could promote everything below this up one level?

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,115 @@
...
+	// When omitted, this means no opinion and the platform is left to choose a reasonable default, which is subject to change over time.
+	// The current default is ClearText.
+	// +optional
+	DataPolicy DataPolicy `json:"dataPolicy,omitempty"`
+	// gatherers is a list of gatherers configurations.
+	// The particular gatherers IDs can be found at https://github.com/openshift/insights-operator/blob/master/docs/gathered-data.md.
+	// Run the following command to get the names of last active gatherers:
+	// "oc get insightsoperators.operator.openshift.io cluster -o json | jq '.status.gatherStatus.gatherers[].name'"
+	// +kubebuilder:validation:MaxItems=100
+	// +listType=atomic
```
JoelSpeed, line 57: I'm surprised this is atomic, aren't the gatherers all uniquely named/id'd?

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,115 @@
...
+// gatherConfig provides data gathering configuration options.
+type GatherConfig struct {
+	// dataPolicy allows user to enable additional global obfuscation of the IP addresses and base domain
+	// in the Insights archive data. Valid values are "ClearText" and "ObfuscateNetworking".
+	// When set to ClearText the data is not obfuscated.
+	// When set to ObfuscateNetworking the IP addresses and the cluster domain name are obfuscated.
+	// When omitted, this means no opinion and the platform is left to choose a reasonable default, which is subject to change over time.
+	// The current default is ClearText.
+	// +optional
+	DataPolicy DataPolicy `json:"dataPolicy,omitempty"`
```
JoelSpeed, line 51: Does this godoc need to be updated now that this is a list and not a single item?

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,115 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	// spec holds user settable values for configuration
+	// +required
+	Spec InsightsDataGatherSpec `json:"spec"`
+	// status holds observed values from the cluster. They may not be overridden.
+	// +optional
+	Status InsightsDataGatherStatus `json:"status"`
+}
+
+type InsightsDataGatherSpec struct {
```
JoelSpeed, line 33: This spec has no required fields, what does it mean when the spec is empty?

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,115 @@
...
+	Enabled GathererState = "Enabled"
+)
+
+// dataPolicyOption declares valid data policy options
+// +kubebuilder:validation:Enum="";ClearText;ObfuscateNetworking
+type DataPolicyOption string
+
+// DataPolicy is a list of data policy options
+// +listType=set
+// +kubebuilder:validation:MaxItems=5
```
JoelSpeed, line 96: There are only two valid options for this value, so, max is 2?

We are typically avoiding the `listType` `set` rule in favour of using CEL with an atomic list, we've run into issues before

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,115 @@
...
+	// IP addresses and cluster domain name are obfuscated
+	ObfuscateNetworking DataPolicyOption = "ObfuscateNetworking"
+	// Gatherer state marked as disabled, which means that the gatherer will not run.
+	Disabled GathererState = "Disabled"
+	// Gatherer state marked as enabled, which means that the gatherer will run.
+	Enabled GathererState = "Enabled"
+)
+
+// dataPolicyOption declares valid data policy options
+// +kubebuilder:validation:Enum="";ClearText;ObfuscateNetworking
```
JoelSpeed, line 91: If this is an enum of values used in a list, why would a user add `""` to the list? Is that valid?

config/v1alpha2/types_insights.go
opokornyy: So if I understand it correctly you suggest using something like this?
```go
type InsightsDataGatherSpec struct {
	DataPolicy DataPolicy       `json:"dataPolicy,omitempty"`
	Gatherers  []GathererConfig `json:"gatherers,omitempty"`
}
```
I don't know about any other fields that should be added at the moment, so I would drop that extra level. WDYT @tremes @ncaak

config/v1alpha2/types_insights.go
opokornyy: AFAIK the InsightsDataGather CRD is an optional configuration of the Insigths Operator. If there are no values in the spec we will use defaults. Please correct me if I am wrong here @tremes

config/v1alpha2/types_insights.go
opokornyy: I was struggling with this because I initially had `+listType=set`, but the tests were failing since GathererConfig is not an atomic type. So, I iterated to +`listType=atomic`, which passed the tests. I'm not entirely sure what the correct approach is, but the gatherers do have unique names. Maybe we don’t need `+listType` at all?

config/v1alpha2/types_insights.go
opokornyy: Can you provide more information for the usage of CEL and atomic usage instead of `listType=set`? The DataPolicy should be a simple list of strings. However when I am thinking about it we will probably need to make sure that there are not both `ClearText` and `ObfuscateNetworking` at the same time defined -> `["ClearText", "ObfuscateNetworking"]`. We made it into a list because there is one more option right now `WorkloadNames`  and to keep it in sync with our ConfigMap. (I had forgotten to add the option here...)

config/v1alpha2/types_insights.go
opokornyy: I am not 100% why is it here, I have taken this from our `v1alpha1`. But I think we could probably get rid of that if I am not overlooking something WDYT @tremes @ncaak

config/v1alpha2/types_insights.go
ncaak: Hi there,

I think the original plan was having the configuration out of possible new uses for the type.

@JoelSpeed what are the advantages of changing this structure from the previous v1alpha1 version ?

JoelSpeed: Just one less level of indentation to be honest
```
spec:
  gatherConfig:
    dataPolicy: ...
    gatherers: ...
```
vs
```
spec:
  dataPolicy: ...
  gatherers: ...
```

If we don't think we will ever add a sibling alongside `gatherConfig`, then separating these into their own struct has little value, but if we might add more in the future, then happy to leave as is

It strikes me though, what would a resource do if it were just:
```
spec: {}
```

Should we require at least one property under spec to be present?

config/v1alpha2/types_insights.go
JoelSpeed: Will this object always exist, to show status? Or does something happen only when the resource exists, and we expect the customer to create the resource?

config/v1alpha2/types_insights.go
JoelSpeed: You want `+listType=map` with `+listMapKey=name` I think, each gatherer has a unique name?

config/v1alpha2/types_insights.go
JoelSpeed: To learn about CEL, take a look at https://kubernetes.io/docs/reference/using-api/cel/

Specifically here, Set has some issues with lists of objects (I know this is list of string), so we are staying away from it in general and using CEL to validate the uniqueness, you can add a rule like

```
self.exists_one(x, self.exists(y, x == y))
```

Which ensures uniqueness of objects within the list

config/v1alpha2/types_insights.go
tremes: humm I don't know to be honest. I don't think there's a special meaning behind that. Originally it was a string, so maybe it said that an empty string was accepted.

config/v1alpha2/types_insights.go
tremes: This was designed as a `cluster` singleton CR if I remember correctly. So it's created by the operator (CVO) and is used for the Insights configuration (by user). My understanding is that the status is reported in `insightsoperator.operators.openshift.io`

tremes: I think it was the general approach back then. I mean things that can be configured live in the `config.openshift.io` group and things that are reported by the operators live in the `operators.openshift.io`. That was the general consensus, or at least I understood it. 
I also remember that the `insightsoperator.operators.openshift.io` (the CR to report status) was immediately introduced as `v1` and I guess we weren't so sure about the config options so the `insightsdatagather` started as `v1alpha1`.

config/v1alpha2/types_insights.go
opokornyy: We are not sure if we will add anything in the future, so let's leave it as it is

config/v1alpha2/types_insights.go
opokornyy: Yes, gatherers have unique names

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,191 @@
...
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+//
+// InsightsDataGather provides data gather configuration options for the the Insights Operator.
+//
+// Compatibility level 4: No compatibility is provided, the API can change at any point for any reason. These capabilities should not be used by applications needing long term support.
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=insightsdatagathers,scope=Cluster
+// +kubebuilder:subresource:status
+// +openshift:api-approved.openshift.io=https://github.com/openshift/api/pull/1245
```
JoelSpeed, line 15: This will be 2195 now

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,191 @@
...
+	// +optional
+	GatherConfig GatherConfig `json:"gatherConfig,omitempty"`
+}
+
+type InsightsDataGatherStatus struct{}
+
+// gatherConfig provides data gathering configuration options.
+type GatherConfig struct {
+	// dataPolicy is an optional list of DataPolicyOptions that allows user to enable additional global obfuscation of the IP addresses and base domain in the Insights archive data.
+	// Valid values are ["ClearText"], ["ObfuscateNetworking"] and ["WorkloadNames"].
```
JoelSpeed, line 44: For consistency with other APIs
```suggestion
	// Valid values are ClearText, ObfuscateNetworking and WorkloadNames.
```

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,191 @@
...
+	// When set to ClearText the data is not obfuscated.
+	// When set to ObfuscateNetworking the IP addresses and the cluster domain name are obfuscated.
```
JoelSpeed, line 46: What about WorkloadNames?

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,191 @@
...
+	// When omitted, this means no opinion and the platform is left to choose a reasonable default, which is subject to change over time.
+	// The current default is "ClearText".
+	// +optional
+	DataPolicy DataPolicy `json:"dataPolicy,omitempty"`
```
JoelSpeed, line 50: What would it mean for the DataPolicy list to contain ClearText and ObfuscateNetworking? I suspect, that wouldn't be possible, and that actually, the ClearText option is no longer useful?

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,191 @@
...
+	// "oc get insightsoperators.operator.openshift.io cluster -o json | jq '.status.gatherStatus.gatherers[].name'"
+	// +kubebuilder:validation:MaxItems=100
+	// +listType=map
+	// +listMapKey=name
+	// +optional
+	Gatherers []GathererConfig `json:"gatherers,omitempty"`
+	// storage is an optional field that allows user to define persistent storage for gathering jobs to store the Insights data archive.
+	// If omitted, the gathering job will use ephemeral storage.
+	// +optional
+	StorageSpec *Storage `json:"storage,omitempty"`
```
JoelSpeed, line 65:
```suggestion
	Storage *Storage `json:"storage,omitempty"`
```

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,191 @@
...
+	// name is a string that follows the DNS1123 subdomain format.
+	// It must be at most 253 characters in length, and must consist only of lower case alphanumeric characters, '-' and '.', and must start and end with an alphanumeric character.
+	// +kubebuilder:validation:XValidation:rule="!format.dns1123Subdomain().validate(self).hasValue()",message="a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character."
+	// +kubebuilder:validation:MaxLength:=253
+	// +required
+	Name string `json:"name"`
+}
+
+// gathererConfig allows to configure specific gatherers
+type GathererConfig struct {
```
JoelSpeed, line 142: This API has existed for some time now, did we reach any consensus on potential additions to the GathererConfig? Is there any feedback that individual gatherers might need some additional configuration here?

config/v1alpha2/types_insights.go
opokornyy: Yes, ClearText should not be combined with any other DataPolicy. I will introduce CEL validation to allow the user to set ClearText explicitly. However, I am not sure about dropping it completely.

config/v1alpha2/types_insights.go
tremes: We don't have any additions at the moment. The main thing here is the ability to disable a specific gatherer.

config/v1alpha2/types_insights.go
tremes: We always want the default to be non-obfuscated. Maybe we don't need the `ClearText` option, but I think we should document the default state clearly.

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,194 @@
...
+	// Valid values are ClearText, ObfuscateNetworking and WorkloadNames.
+	// When set to ClearText the data is not obfuscated.
+	// When set to ObfuscateNetworking the IP addresses and the cluster domain name are obfuscated.
+	// When set to WorkloadNames the data from Deployment Validation Operator is obfuscated.
+	// When set to ClearText the other options are not allowed.
+	// When omitted, this means no opinion and the platform is left to choose a reasonable default, which is subject to change over time.
+	// The current default is "ClearText".
+	// +kubebuilder:validation:XValidation:rule="(self.exists(x, x == 'ClearText') && self.size() == 1) ||  !self.exists(x, x == 'ClearText')",message="ClearText must not be combined with other data policy options"
+	// +optional
+	DataPolicy DataPolicy `json:"dataPolicy,omitempty"`
```
everettraven, line 53: Any particular reason why this is a type alias rather than `[]DataPolicyOption`?

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,194 @@
...
+	// name is a string that follows the DNS1123 subdomain format.
+	// It must be at most 253 characters in length, and must consist only of lower case alphanumeric characters, '-' and '.', and must start and end with an alphanumeric character.
+	// +kubebuilder:validation:XValidation:rule="!format.dns1123Subdomain().validate(self).hasValue()",message="a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character."
+	// +kubebuilder:validation:MaxLength:=253
+	// +required
+	Name string `json:"name"`
+}
+
+// gathererConfig allows to configure specific gatherers
+type GathererConfig struct {
```
everettraven, line 145: What is the expected behavior if I do something like:
```yaml
...
- name: gathererone
  state: Disabled
- name: gathererone/functionone
  state: Enabled
...
```
?

Should this be an allowed configuration?

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,194 @@
...
+	// When set to ClearText the data is not obfuscated.
+	// When set to ObfuscateNetworking the IP addresses and the cluster domain name are obfuscated.
+	// When set to WorkloadNames the data from Deployment Validation Operator is obfuscated.
+	// When set to ClearText the other options are not allowed.
+	// When omitted, this means no opinion and the platform is left to choose a reasonable default, which is subject to change over time.
+	// The current default is "ClearText".
+	// +kubebuilder:validation:XValidation:rule="(self.exists(x, x == 'ClearText') && self.size() == 1) ||  !self.exists(x, x == 'ClearText')",message="ClearText must not be combined with other data policy options"
+	// +optional
+	DataPolicy DataPolicy `json:"dataPolicy,omitempty"`
+	// gatherers is an optional list of gatherers configurations.
```
everettraven, line 54: If I don't specify any gatherers, are they all enabled? disabled?

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,194 @@
...
+	// Gatherer consists of a lowercase letters only that may include underscores (_).
+	// Function consists of a lowercase letters only that may include underscores (_) and is separated from the gatherer by a forward slash (/).
+	// The particular gatherers can be found at https://github.com/openshift/insights-operator/blob/master/docs/gathered-data.md.
+	// Run the following command to get the names of last active gatherers:
+	// "oc get insightsoperators.operator.openshift.io cluster -o json | jq '.status.gatherStatus.gatherers[].name'"
+	// +kubebuilder:validation:MaxLength=256
+	// +kubebuilder:validation:XValidation:rule=`self.matches("^[a-z]+[_a-z]*[a-z]([/a-z][_a-z]*)?[a-z]$")`,message=`gatherer name must be in the format of {gatherer}/{function} where the gatherer and function are lowercase letters only that may include underscores (_) and are separated by a forward slash (/) if the function is provided`
+	// +required
+	Name string `json:"name"`
+	// state is an optional field that allows you to configure specific gatherer. Valid values are "Enabled" and "Disabled".
```
everettraven, line 158: From an earlier comment by Joel, it sounds like this API has been around for a little while. When we can I think we try to avoid `Enabled` and `Disabled` as they can be overloaded terms. I think it _can_ make sense in this case and the GoDoc describes what each state represents.

That being said, is there any other configuration option that could be more descriptive than `Enabled`/`Disabled`?

config/v1alpha2/types_insights.go
JoelSpeed: I'm leaning towards, there is no option to say `["ClearText"]`, and when the list is empty it means everything is cleartext, and we document that

Then, when the list is non-empty, there is always at least some obfuscation happening

config/v1alpha2/types_insights.go
JoelSpeed: I also found this to be a little confusing, can we just inline it?

config/v1alpha2/types_insights.go
tremes: This is a good one. I think this means that only `gathererone/functionone` is enabled so yes this can be allowed in my opinion.

config/v1alpha2/types_insights.go
tremes: all are enabled. That's the default we want.

config/v1alpha2/types_insights.go
tremes: I think not at the moment. We simply want to be able to disable some of the data points/gatherers.

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,194 @@
...
+	// When set to ObfuscateNetworking the IP addresses and the cluster domain name are obfuscated.
+	// When set to WorkloadNames the data from Deployment Validation Operator is obfuscated.
+	// When set to ClearText the other options are not allowed.
+	// When omitted, this means no opinion and the platform is left to choose a reasonable default, which is subject to change over time.
+	// The current default is "ClearText".
+	// +kubebuilder:validation:XValidation:rule="(self.exists(x, x == 'ClearText') && self.size() == 1) ||  !self.exists(x, x == 'ClearText')",message="ClearText must not be combined with other data policy options"
+	// +optional
+	DataPolicy DataPolicy `json:"dataPolicy,omitempty"`
+	// gatherers is an optional list of gatherers configurations.
+	// It can be used to enable or disable specific gatherers.
```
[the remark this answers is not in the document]
opokornyy: All are enabled by default, so this configuration should, for example, disable just a specific subset. If you specifically enable only a subset of gatherers, nothing will actually change. However, as you mentioned in one of your comments, if you disable `gathererOne` but enable `gathererOne/functionOne`, then only that specific gatherer function should run.

config/v1alpha2/types_insights.go
everettraven: It might be worth clarifying this intended behavior in the GoDoc for the slice of Gatherers, if it hasn't been updated already.

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,186 @@
...
+	Spec InsightsDataGatherSpec `json:"spec"`
+	// status holds observed values from the cluster. They may not be overridden.
+	// +optional
+	Status InsightsDataGatherStatus `json:"status"`
+}
+
+type InsightsDataGatherSpec struct {
+	// gatherConfig is an optional spec attribute that includes all the configuration options related to gathering of the Insights data and its uploading to the ingress.
+	// +optional
+	GatherConfig GatherConfig `json:"gatherConfig,omitempty"`
```
everettraven, line 36: Some open questions I'm imagining from a user perspective using something like `oc explain ...`:
- What does configuring this field allow me to do?
- What is the behavior if I leave this empty?

Additionally, should this have the `omitempty`? I know that for configuration APIs we try to have some form of discoverability. Having `omitempty` here means that an empty `GatherConfig` will serialize as something like:
```yaml
apiVersion: config.openshift.io/v1alpha2
kind: InsightsDataGather
metadata:
  name: cluster
spec: {}
```

Removing the omitempty will result in something like:
```yaml
apiVersion: config.openshift.io/v1alpha2
kind: InsightsDataGather
metadata:
  name: cluster
spec:
  gatherConfig: {}
```
That will show users with the default configuration in place of "no opinion" that there is a field they can use to do some configuration.

config/v1alpha2/types_insights.go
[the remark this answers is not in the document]
everettraven: Maybe this documentation should then also specify that any configuration here is _additive_ and does not override the default configuration of "all are enabled"? If this is the case, does it make sense to have an "enable" option if all the gatherers are enabled by default?

This also means that in order to specify that I only want a subset of gatherers enabled I _must_ specify every other gatherer as disabled. For example, if I only want to run 5/100 gatherers I have to, by hand, put the other 95 gatherers in here marked as disabled. Is this an acceptable user experience?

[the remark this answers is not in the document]
everettraven: If the example I had mentioned previously in https://github.com/openshift/api/pull/2195/files#r1983588864 is a valid configuration, it should also be called out explicitly that more specific configurations take priority over less specific configurations.

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,186 @@
...
+	// +kubebuilder:validation:MaxLength:=253
+	// +required
+	Name string `json:"name"`
+}
+
+// gathererConfig allows to configure specific gatherers
+type GathererConfig struct {
+	// name is the required name of a specific gatherer
+	// It may not exceed 256 characters.
+	// The format for the disabledGatherer should be: {gatherer}/{function} where the function is optional.
```
everettraven, line 140:
```suggestion
	// The format for a gatherer name is '{gatherer}/{function}' where the function is optional.
```
?

config/v1alpha2/types_insights.go
everettraven: If you just want to be able to disable gatherers, and the default is that all are enabled, does it make more sense to just have a slice of gatherer names that should be disabled?

For example:
```yaml
disabledGatherers:
  - foo/bar
  - baz
  ...
```
instead of:
```yaml
gatherers:
  - name: foo/bar
    state: Disabled
  - name: baz
    state: Disabled
  ...
```

config/v1alpha2/types_insights.go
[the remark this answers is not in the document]
opokornyy: To disable all gatherers, you need to specify only the subsets of gatherers. AFAIK there are about five of them, so you don’t need to create a list with 95 items—just five should be enough. Previously, with the `disabledGatherers` field, it was possible to use `disabledGatherers: all` to disable everything. I wonder if we need something similar here as well.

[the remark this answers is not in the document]
tremes: In general we need a simple way to completely disable the data gathering.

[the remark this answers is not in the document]
everettraven: What if you used a discriminated union to have different modes of operation?

Maybe something like:
```yaml
...
gatheringMode: Default | Custom | None
...
```

Default means the default configuration is used, which is set by the system. 

Custom means the user must provide an input on what gatherers they want to run.

None means no gatherers are run (i.e completely disabled).

Using this approach you could omit the custom configuration for now if there is no use case for it today and add it when there is a use case.

config/v1alpha2/types_insights.go
JoelSpeed: The map based approach was introduced to allow additional configuration for gatherers to be included in this list in the future.

If we decide in the future to add additional gather config for specific gatherers, and have a disabled list, then there's a disconnect between where the gatherer is enabled/disabled and the configuration of the gatherer itself

With the current pattern we can do something like below, where we disable config from being set when state is Disabled
```
gatherers:
- name: foo
  state: Enabled
  config:
    gatherer: foo
    foo:
      ... # specific options
```

config/v1alpha2/types_insights.go
opokornyy: Sorry, I mistakenly changed a different field. It should be fixed now

config/v1alpha2/types_insights.go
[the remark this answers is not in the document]
everettraven: That looks about how I imagined it in my head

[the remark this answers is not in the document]
everettraven: Might want some tweaks to field names though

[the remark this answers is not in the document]
everettraven: @JoelSpeed WDYT?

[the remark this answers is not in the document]
JoelSpeed: The discriminator value and the union member names must match, therefore it should probably be more like
```
gatherConfig:
  mode: Custom
  custom:
    gatherers:
    - name: ...
      state: Enabled
```

Nit, Enabled and Disabled, not Enable and Disable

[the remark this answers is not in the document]
opokornyy: I would maybe use `Enabled` / `Disabled` instead of `Default` / `None`

```yaml
gatheringMode: Enabled | Disabled | Custom
```

config/v1alpha2/types_insights.go
opokornyy: I will stick with the map-based approach here to be more flexible for future configuration.

config/v1alpha2/types_insights.go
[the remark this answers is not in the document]
opokornyy: Do we need to wrap the configuration into one more layer here? Can we omit the `gatherers` field in `custom` configuration or is it better to include that? 

It would look like this:
```yaml
gatherConfig:
  mode: Custom
  custom:
    - name: ...
      state: Enabled
```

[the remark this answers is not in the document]
everettraven: Any chance there are additional configuration options you would want to add to the `Custom` mode beyond the gatherers?

If not, I think your proposed one is OK. If you think you may need room for additional fields under a `Custom` configuration then I would stick with the nesting.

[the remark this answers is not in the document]
everettraven: _Generally_ I err on the side of caution and assume that I may need to add fields in the future unless I have a high level of confidence that I won't.

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,207 @@
...
+	// It is possible to disable an entire set of gatherers while allowing a specific function within that set.
+	// For example, you can disable "gathererOne" while enabling "gathererOne/functionOne."
+	// The particular gatherers IDs can be found at https://github.com/openshift/insights-operator/blob/master/docs/gathered-data.md.
+	// Run the following command to get the names of last active gatherers:
+	// "oc get insightsoperators.operator.openshift.io cluster -o json | jq '.status.gatherStatus.gatherers[].name'"
+	// +kubebuilder:validation:MaxItems=100
+	// +listType=map
+	// +listMapKey=name
+	// +optional
+	Custom []GathererConfig `json:"custom,omitempty"`
```
everettraven, line 72: It might be worth making it explicit in the GoDoc for this field that the configuration that is specified is what is used and no defaulting behavior takes place.

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,207 @@
...
+	// "oc get insightsoperators.operator.openshift.io cluster -o json | jq '.status.gatherStatus.gatherers[].name'"
+	// +kubebuilder:validation:MaxLength=256
+	// +kubebuilder:validation:XValidation:rule=`self.matches("^[a-z]+[_a-z]*[a-z]([/a-z][_a-z]*)?[a-z]$")`,message=`gatherer name must be in the format of {gatherer}/{function} where the gatherer and function are lowercase letters only that may include underscores (_) and are separated by a forward slash (/) if the function is provided`
+	// +required
+	Name string `json:"name"`
+	// state is an optional field that allows you to configure specific gatherer. Valid values are "Enabled" and "Disabled".
+	// When set to Enabled the gatherer will run.
+	// When set to Disabled the gatherer will not run.
+	// When omitted, this means no opinion and the platform is left to choose a reasonable default.
+	// The current default is Enabled.
```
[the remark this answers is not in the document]
opokornyy: I think it makes sense to make it required now.

config/v1alpha2/types_insights.go
opokornyy: I have reworked it with a custom struct, so you may check if the documentation is good enough. There should be no defaults anymore

config/v1alpha2/types_insights.go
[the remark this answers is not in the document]
opokornyy: There are no other config options at the moment that I am aware of, but I am not 100% sure that it won't change in the future.

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,207 @@
...
+	// Valid values are ObfuscateNetworking and WorkloadNames.
+	// When set to ObfuscateNetworking the IP addresses and the cluster domain name are obfuscated.
+	// When set to WorkloadNames the data from Deployment Validation Operator is obfuscated.
+	// When omitted no obfuscation is applied.
+	// +kubebuilder:validation:MaxItems=2
+	// +kubebuilder:validation:XValidation:rule="self.all(x, self.exists_one(y, x == y))",message="DataPolicy must not contain duplicates"
+	// +optional
+	DataPolicy []DataPolicyOption `json:"dataPolicy,omitempty"`
+	// mode is a required field that specifies the mode for gatherers.
+	// When set to Enabled, all gatherers are enabled.
```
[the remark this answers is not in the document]
everettraven: I think it is OK to use enabled and disabled as your terminology, but I would recommend updating your GoDoc to ensure that there isn't ambiguity as to the meaning of each mode that could cause confusion. Make it explicit what each term means for a user.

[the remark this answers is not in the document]
opokornyy: I have tried to improve the godoc, so you can check it now

everettraven (commented):
Aside from the open questions/threads I think this looks good. Tagging in @JoelSpeed to do a pass and get his thoughts on the open comments.

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,213 @@
...
+	// It may not exceed 100 items and each gatherer can be present only once.
+	// It is possible to disable an entire set of gatherers while allowing a specific function within that set.
+	// The particular gatherers IDs can be found at https://github.com/openshift/insights-operator/blob/master/docs/gathered-data.md.
+	// Run the following command to get the names of last active gatherers:
+	// "oc get insightsoperators.operator.openshift.io cluster -o json | jq '.status.gatherStatus.gatherers[].name'"
+	// +kubebuilder:validation:MaxItems=100
+	// +listType=map
+	// +listMapKey=name
+	// +required
+	Gatherers []GathererConfig `json:"gatherers"`
```
JoelSpeed, line 84: We have this list as a structured item so that we can introduce configuration for gatherers in the future potentially. If someone isn't using custom, are we to assume they are happy with the default configuration of the gatherer?

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,213 @@
...
+type InsightsDataGatherSpec struct {
+	// gatherConfig is an optional spec attribute that includes all the configuration options related to gathering of the Insights data and its uploading to the ingress.
+	// +optional
+	GatherConfig GatherConfig `json:"gatherConfig"`
+}
+
+type InsightsDataGatherStatus struct{}
+
+// gatherConfig provides data gathering configuration options.
+// +kubebuilder:validation:XValidation:rule="has(self.mode) && self.mode == 'Custom' ?  has(self.custom) : !has(self.custom)",message="custom is required when mode is Custom, and forbidden otherwise"
```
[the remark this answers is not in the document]
opokornyy: I wasn't sure about the naming, I have changed the `custom.gathereres` =>` custom.configs` .Can you please check if it's okay?

```yaml
dataPolicy: []
gatherers:
   mode: Enabled | Disabled | Custom 
   custom:
       configs:
           - name: gatherOne/funcOne
             state: Disabled
storage: {}
```

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,213 @@
...
+	// Valid values are ObfuscateNetworking and WorkloadNames.
+	// When set to ObfuscateNetworking the IP addresses and the cluster domain name are obfuscated.
+	// When set to WorkloadNames the data from Deployment Validation Operator is obfuscated.
+	// When omitted no obfuscation is applied.
+	// +kubebuilder:validation:MaxItems=2
+	// +kubebuilder:validation:XValidation:rule="self.all(x, self.exists_one(y, x == y))",message="DataPolicy must not contain duplicates"
+	// +optional
+	DataPolicy []DataPolicyOption `json:"dataPolicy,omitempty"`
+	// mode is a required field that specifies the mode for gatherers. Allowed values are Enabled, Disabled, and Custom.
+	// When set to Enabled, all gatherers wil run and gather data.
```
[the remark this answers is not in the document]
opokornyy: Yes, AFAIK there won't be added any that need to be explicitly enabled

config/v1alpha2/types_insights.go
opokornyy: Yes, we assume that most users are okay with using the default configuration, which now means running all gatherers. However, as Tomas mentioned, some users may want to disable a specific gatherer—this can be done with `custom` or to disable all gatherers - `mode: Disabled`. AFAIK, using a custom configuration is more or less a corner case for users, and if we add more configuration options, I think it will be the same—they will mostly stick with the default.

everettraven (approved):
/lgtm

config/v1alpha2/types_insights.go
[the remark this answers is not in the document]
JoelSpeed: I still think mode would make more sense as `All`, `None` instead of `Enabled`, `Disabled`, WDYT?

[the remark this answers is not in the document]
tremes: > I still think mode would make more sense as All, None instead of Enabled, Disabled, WDYT?

Sounds better to me too. Thank you.

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,213 @@
...
+type InsightsDataGatherStatus struct{}
+
+// gatherConfig provides data gathering configuration options.
+// +kubebuilder:validation:XValidation:rule="has(self.mode) && self.mode == 'Custom' ?  has(self.custom) : !has(self.custom)",message="custom is required when mode is Custom, and forbidden otherwise"
+type GatherConfig struct {
+	// dataPolicy is an optional list of DataPolicyOptions that allows user to enable additional obfuscation of the Insights archive data.
+	// It may not exceed 2 items and must not contain duplicates.
+	// Valid values are ObfuscateNetworking and WorkloadNames.
+	// When set to ObfuscateNetworking the IP addresses and the cluster domain name are obfuscated.
+	// When set to WorkloadNames the data from Deployment Validation Operator is obfuscated.
```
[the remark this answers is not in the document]
opokornyy: Updated, now it should be hopefully more user-friendly

config/v1alpha2/types_insights.go
```diff
@@ -0,0 +1,219 @@
...
+type GatherConfig struct {
+	// dataPolicy is an optional list of DataPolicyOptions that allows user to enable additional obfuscation of the Insights archive data.
+	// It may not exceed 2 items and must not contain duplicates.
+	// Valid values are ObfuscateNetworking and WorkloadNames.
+	// When set to ObfuscateNetworking the IP addresses and the cluster domain name are obfuscated.
+	// When set to WorkloadNames, the gathered data about cluster resources will not contain the workload names for your deployments. Resources UIDs will be used instead.
+	// When omitted no obfuscation is applied.
+	// +kubebuilder:validation:MaxItems=2
+	// +kubebuilder:validation:XValidation:rule="self.all(x, self.exists_one(y, x == y))",message="dataPolicy items must be unique"
+	// +optional
```
JoelSpeed, line 51: This will fix the broken schema checker
```suggestion
	// +listType=atomic
	// +optional
```

JoelSpeed:
LGTM once https://github.com/openshift/api/pull/2195#discussion_r2010029553 is sorted

JoelSpeed:
Could you please squash the commits, I don't think we need all the fixups in the history

PR #2200, opened 2025-02-12
CCXDEV-14850: insights add storage spec

opokornyy:


config/v1alpha1/types_insights.go
```diff
@@ -57,8 +56,29 @@ type GatherConfig struct {
 	// Run the following command to get the names of last active gatherers:
 	// "oc get insightsoperators.operator.openshift.io cluster -o json | jq '.status.gatherStatus.gatherers[].name'"
 	// An example of disabling gatherers looks like this: `disabledGatherers: ["clusterconfig/machine_configs", "workloads/workload_info"]`
+	// +kubebuilder:validation:MaxItems=100
+	// +kubebuilder:validation:items:MaxLength=256
```
everettraven, line 60: Adding these constraints are _technically_ a breaking change for users. It looks like this is a v1alpha1 API so it _should_ be okay to make this change since we only deploy v1alpha1 APIs in modes where upgrades are unsupported.

I would recommend updating the Godoc to also mention these new constraints.

While you are at it, are there any other validations that you can put in place? Do gatherers follow a specific format that you can use to ensure users aren't able to provide a string that could never evaluate to a valid gatherer name?

config/v1alpha1/types_insights.go
```diff
@@ -57,8 +56,29 @@ type GatherConfig struct {
...
 	// +optional
 	DisabledGatherers []string `json:"disabledGatherers"`
+	// storageSpec allows user to define persistent storage for on-demand gathering
+	// jobs to store the Insights data archive.
+	// If omitted, the gathering job will use ephemeral storage.
+	// +optional
+	StorageSpec StorageSpec `json:"storageSpec,omitempty"`
```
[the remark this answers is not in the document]
JoelSpeed: If the struct has no required fields, then yes, you can add it like this, but, `omitempty` won't work. In this case, you have a required field, and some optional field.

In this case I suspect you want this to be a pointer and omitempty, so that when someone does explicitly set the field, they are required to put the claim name in

The alternative, for discoverability, is to make the claim name optional and allow it to be the empty string, but then require that if the length of the mountpath is non-zero, then the claim name is also non empty

It's a bit more awkward

config/v1alpha1/types_insights.go
opokornyy: I didn't know this could be a breaking change, I added these only to pass linter check.

I would not add anything else, because we are planning to drop this field in the next version.

config/v1alpha1/types_insights.go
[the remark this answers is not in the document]
opokornyy: I will change it to pointer and `omitempty`

config/v1alpha1/types_insights.go
```diff
@@ -57,8 +56,29 @@ type GatherConfig struct {
...
+}
+
+type StorageSpec struct {
+	// persistentVolumeClaimName specifies the name of the PersistentVolumeClaim that will
+	// be used to store the Insights data archive.
```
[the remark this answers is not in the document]
opokornyy: I think the PVC must be in the `openshift-insights` namespace, I will add it to the godoc now. WDYT @tremes

config/v1alpha1/types_insights.go
```diff
@@ -57,8 +56,29 @@ type GatherConfig struct {
...
+	// +kubebuilder:validation:MaxLength=256
+	// +required
+	PersistentVolumeClaimName string `json:"persistentVolumeClaimName"`
+	// mountPath is the directory where the PVC will be mounted inside the Insights data gathering Pod.
+	// If omitted, the path that is used to store the Insights data archive by Insights
+	// operator will be used instead. By default, the path is /var/lib/insights-operator.
```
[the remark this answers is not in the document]
opokornyy: No, I don't expect it to change, I have added the `+default` tag @tremes

config/v1alpha1/types_insights.go
```diff
@@ -57,8 +56,29 @@ type GatherConfig struct {
...
+	// +kubebuilder:validation:MaxLength=1024
+	// +optional
+	MountPath string `json:"mountPath,omitempty"`
```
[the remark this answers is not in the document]
opokornyy: Tbh I am not aware of any, it only needs to be a valid path and based on what I have seen in kubernetes doc it must not contain `:`. Is that something we should validate with CEL?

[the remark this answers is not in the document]
everettraven: I would say yes. Anything we can do to limit the input to the valid set of values rather than allowing free form strings is beneficial, especially because it will reject at admission time rather than a controller putting an error in a status condition that may not be discovered as easily.

config/v1alpha1/types_insights.go
everettraven: Removing existing fields is also a breaking change :) - I think the same logic applies with this being v1alpha1 and thus only existing in unsupported, no upgrade path, cluster modes so it should be OK to remove.

cc @JoelSpeed for a sanity check here.

config/v1alpha1/types_insights.go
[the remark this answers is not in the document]
opokornyy: I have added the validation to check that

config/v1alpha1/types_insights.go
JoelSpeed: Adding limits like this to alpha types is ok, as when we go to v1, we will forget the v1alpha1 existed

If we are dropping fields between v1alpha1 and v1 that's also acceptable, since there's no upgrade path

So we can be more relaxed here than we would otherwise be normally about potentially breaking changes

everettraven (commented):
One additional comment after looking over the most recent changes. Other than that I think this looks good.

config/v1alpha1/types_insights.go
```diff
@@ -57,8 +57,38 @@ type GatherConfig struct {
...
+	// storageSpec is an optional field that allows user to define persistent storage for on-demand gathering
+	// jobs to store the Insights data archive.
+	// If omitted, the gathering job will use ephemeral storage.
+	// +optional
+	StorageSpec *StorageSpec `json:"storageSpec,omitempty"`
+}
+
+type StorageSpec struct {
+	// persistentVolumeClaimName is required field that specifies the name of the PersistentVolumeClaim that will
+	// be used to store the Insights data archive. The persistenVolumeClaim must be created in the openshift-insights namespace.
```
tremes, line 73: > The persistenVolumeClaim must be created in the openshift-insights namespace.

are the permissions the main reason for this? I guess we can't mount a PVC from different namespace right?

opokornyy: Yes, I think you can not mount a PVC that is not in the same namespace as the pod

config/v1alpha1/types_insights.go
```diff
@@ -57,10 +57,37 @@ type GatherConfig struct {
...
+	// +kubebuilder:validation:XValidation:rule="!self.contains(':')",message="mountPath must not contain a colon"
+	// +default="/var/lib/insights-operator"
+	// +optional
+	MountPath string `json:"mountPath,omitempty"`
+}
+
+// persistenVolumeClaimName is a string that follows the DNS1123 subdomain format.
+// +kubebuilder:validation:XValidation:rule="!format.dns1123Subdomain().validate(self).hasValue()",message="a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character."
+// +kubebuilder:validation:MaxLength:=253
+type PersistentVolumeClaimName string
```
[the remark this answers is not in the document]
opokornyy: I have reworked it to use the struct instead of a string, but the usage feels weird to me, because now we would have a field `StorageSpec.PersistentVolumeClaimName.Name`. Is that okay to you? I was thinking about renaming the `PersistentVolumeClaimName` to for example `PersistentVolumeClaim` or `PersistentVolumeClaimRef`, but then its weird to have the `mountPath` not included in the same struct. WDYT? @everettraven

opokornyy:
> One additional comment after looking over the most recent changes. Other than that I think this looks good.

What should I do about the error in tests? It adds a new required field, but it is wrapped in an optional struct, so IMO it should be fine right?

```
ERROR: NoNewRequiredFields: crd/datagathers.insights.openshift.io version/v1alpha1 field/^.spec.storageSpec.persistentVolumeClaimName is new and may not be required
````

**Comment by everettraven** [comment]
> > One additional comment after looking over the most recent changes. Other than that I think this looks good.
> 
> What should I do about the error in tests? It adds a new required field, but it is wrapped in an optional struct, so IMO it should be fine right?
> 
> ```
> ERROR: NoNewRequiredFields: crd/datagathers.insights.openshift.io version/v1alpha1 field/^.spec.storageSpec.persistentVolumeClaimName is new and may not be required
> ```

Looking at the failure, it looks like this check is safe to override. I don't have permissions to do this, so it'll be up to Joel to do that when he gives a review.

File: config/v1alpha1/types_insights.go (hunk 6 above)
Comment (everettraven) [inline config/v1alpha1/types_insights.go:73]:
```suggestion
	// be used to store the Insights data archive. The PersistentVolumeClaim must be created in the openshift-insights namespace.
```

config/v1alpha1/types_insights.go
```diff
@@ -57,8 +57,38 @@ type GatherConfig struct {
...
+	// If omitted, the gathering job will use ephemeral storage.
+	// +optional
+	StorageSpec *StorageSpec `json:"storageSpec,omitempty"`
+}
+
+type StorageSpec struct {
+	// persistentVolumeClaimName is required field that specifies the name of the PersistentVolumeClaim that will
+	// be used to store the Insights data archive. The persistenVolumeClaim must be created in the openshift-insights namespace.
+	// +required
+	PersistentVolumeClaimName PersistentVolumeClaimReference `json:"persistentVolumeClaimName"`
```
everettraven, line 75: This might be better as `PersistentVolumeClaim` to reduce repetition of the word "name" when used like:
```yaml
...
persistentVolumeClaimName:
  name: foo 
...
```

opokornyy: Done

config/v1alpha1/types_insights.go
```diff
@@ -57,8 +57,40 @@ type GatherConfig struct {
 	// Run the following command to get the names of last active gatherers:
 	// "oc get insightsoperators.operator.openshift.io cluster -o json | jq '.status.gatherStatus.gatherers[].name'"
 	// An example of disabling gatherers looks like this: `disabledGatherers: ["clusterconfig/machine_configs", "workloads/workload_info"]`
+	// +kubebuilder:validation:MaxItems=100
```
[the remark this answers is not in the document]
opokornyy: Currently, there are ~63 gatherers, and the number is growing. I added the limit only to pass the CI checks. Ideally, we would not set any limit here due to the increasing number of gatherers.

config/v1alpha1/types_insights.go
```diff
@@ -57,8 +57,40 @@ type GatherConfig struct {
...
+	// +kubebuilder:validation:items:MaxLength=256
```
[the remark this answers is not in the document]
opokornyy: We are not enforcing any format for a gatherer names.

config/v1alpha1/types_insights.go
```diff
@@ -57,8 +57,40 @@ type GatherConfig struct {
...
+type StorageSpec struct {
+	// persistentVolumeClaim is required field that specifies the name of the PersistentVolumeClaim that will
+	// be used to store the Insights data archive. The PersistentVolumeClaim must be created in the openshift-insights namespace.
+	// +required
+	PersistentVolumeClaim PersistentVolumeClaimReference `json:"persistentVolumeClaim"`
+	// mountPath is an optional field specifying the directory where the PVC will be mounted inside the
+	// Insights data gathering Pod. If omitted, the path that is used to store the Insights data archive by Insights
+	// operator will be used instead. The path cannot exceed 1024 characters and defaults to "/var/lib/insights-operator".
+	// +kubebuilder:validation:MaxLength=1024
+	// +kubebuilder:validation:XValidation:rule="!self.contains(':')",message="mountPath must not contain a colon"
```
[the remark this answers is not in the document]
opokornyy: I found it in a kuberentes documentation in `VolumeMount v1 core` section, there is a `mountPath` field with a description that it only can not contain ":". 

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/

config/v1alpha1/types_insights.go
```diff
@@ -57,8 +57,40 @@ type GatherConfig struct {
...
+	// +optional
+	StorageSpec *StorageSpec `json:"storageSpec,omitempty"`
+}
+
+// storageSpec provides persistent storage configuration options for on-demand gathering jobs.
+type StorageSpec struct {
+	// persistentVolumeClaim is required field that specifies the name of the PersistentVolumeClaim that will
+	// be used to store the Insights data archive. The PersistentVolumeClaim must be created in the openshift-insights namespace.
+	// +required
+	PersistentVolumeClaim PersistentVolumeClaimReference `json:"persistentVolumeClaim"`
```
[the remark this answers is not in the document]
tremes: I don't think so. It has always been ephemeral by default up until now. The thing is that with the data gathering as job, it's difficult to explore the archive and the data (and we want to be transparent to the users and also would like to make testing easier) so that is the main reason for introducing this volume option.

Personally, I can't imagine anything other than PVC.

config/v1alpha1/types_insights.go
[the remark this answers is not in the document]
JoelSpeed: Every list must have a limit. At what rate is this increasing? If the limit were 500 do you think we would ever get to 500 gatherers?

config/v1alpha1/types_insights.go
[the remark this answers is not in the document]
JoelSpeed: What other fields does the gatherer name appear in across OpenShift APIs? 

If I were implementing a gatherer, where would I register it? I assume there's some code somewhere where I have to tell something central that my gatherer exists?

config/v1alpha1/types_insights.go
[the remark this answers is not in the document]
JoelSpeed: The suggested union approach above would be more explicit, and allow for potential future unforeseen changes more easily than the current approach. I would strongly advise we move to that union pattern here just in case

config/v1alpha1/types_insights.go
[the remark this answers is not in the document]
JoelSpeed: Cool, so long as we match the validation there, then that makes sense, thanks!

config/v1alpha1/types_insights.go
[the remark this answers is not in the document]
tremes: it doesn't appear anywhere. It's basically just a string defined in the operator's codebase. No need to register anything. The naming of the gatherers is a bit "organic" (for some historical reasons IMO) and I think we don't have any other means how to say I want to exclude/disable this gatherer XYZ. 

Historically there was this `clusterconfig` gatherer (and now there are some others) including bunch of gathering functions https://github.com/openshift/insights-operator/blob/master/pkg/gatherers/clusterconfig/clusterconfig_gatherer.go#L28 so the way to exclude this gatherer (or to say I don't want to provide this data) is e.g to say `clusterconfig` (exclude all of this) or you can say `clusterconfig/machine_autoscalers` (to exclude only the machine autoscaler data).  For these purposes, we refer to https://github.com/openshift/insights-operator/blob/master/docs/gathered-data.md, where the configuration ID is essentially a name. 

It would be nice to have some more formal process for this, but I am affraid it won't be easy to do it right now.

config/v1alpha1/types_insights.go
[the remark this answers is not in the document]
opokornyy: It is growing based on demand for new gatherers, so I cannot say exactly. However, we have discussed it as a team and believe that 100 is a reasonable amount, so we won’t need to tweak it anytime soon.

config/v1alpha1/types_insights.go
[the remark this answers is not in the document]
opokornyy: I have tried to match the proposed structure, so you can review it now.

config/v1alpha1/types_insights.go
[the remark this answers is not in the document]
everettraven: Taking a brief look at the linked docs, it looks like the gatherer names generally follow the format of `{gatherer}/{function}` where gatherer is a string of lowercase alpha characters and function is a string of lowercase alpha characters that may contain an underscore. the `/` character would be allowed as well to separate gatherer and gathering functions

Would it be worth at least enforcing this pattern for now so that users can't put something that will clearly be invalid based on the known set of gatherers?

You can loosen the validation in the future if needed to add a gatherer that might not fit that mold anymore, but introducing it at least rejects at admission time values that would obviously be an invalid gatherer name.

config/v1alpha1/types_insights.go
```diff
@@ -57,15 +55,68 @@ type GatherConfig struct {
...
 
 const (
 	// No data obfuscation
 	NoPolicy DataPolicy = "None"
 	// IP addresses and cluster domain name are obfuscated
 	ObfuscateNetworking DataPolicy = "ObfuscateNetworking"
+	// Ephemeral storage type
+	Ephemeral StorageType = "Ephemeral"
+	// PersistentVolumeClaim storage type
+	PersistentVolumeClaim StorageType = "PersistentVolumeClaim"
```
everettraven, line 119: It is generally recommended that constants mapping to an enum type alias follow the pattern `{TypeAlias}{Name}`. In this case:
```suggestion
	// StorageTypeEphemeral storage type
	StorageTypeEphemeral StorageType = "Ephemeral"
	// StorageTypePersistentVolumeClaim storage type
	StorageTypePersistentVolumeClaim StorageType = "PersistentVolumeClaim"
```

config/v1alpha1/types_insights.go
```diff
@@ -57,15 +55,68 @@ type GatherConfig struct {
...
+	// +optional
+	StorageSpec *StorageSpec `json:"storageSpec,omitempty"`
+}
+
+// storageSpec provides persistent storage configuration options for on-demand gathering jobs.
+type StorageSpec struct {
+	// type is a required field that specifies the type of storage that will be used to store the Insights data archive.
+	// Valid values are "PersistentVolumeClaim" and "Ephemeral". If the value is omitted, the default value is "Ephemeral".
+	// +optional
+	Type StorageType `json:"type,omitempty"`
```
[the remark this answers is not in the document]
opokornyy: I think that removing the `Ephemeral` option and making the `Type` field required would be the best approach. If the entire `StorageSpec` is missing, we will default to using Ephemeral

config/v1alpha1/types_insights.go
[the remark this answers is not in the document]
opokornyy: I have added a validation check that should suit our naming conventions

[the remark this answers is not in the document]
JoelSpeed: So my main concern he would be, this API is alpha now, which means things can change. But when this goes v1, it will be much harder for you to enforce a standard, without breaking the folks who have already used this API.

So, if you do want to start enforcing a standard (I'm not sure how much work that would be), lets make sure we do that prior to shipping this.

Setting a limit (256 chars) and the pattern as Bryce described looks from my cursory scan like it would be an low effort way to enforce this, but the question is, how many don't fit this pattern and need fixing? And is it feasible to rename some gatherer?

config/v1alpha1/types_insights.go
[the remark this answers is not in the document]
JoelSpeed: Ack, relaxing it in the future is possible if we hit the limit so I'm not too concerned, but if you did want to go bigger than 100 I think you have headroom in the cost budget to do that

config/v1alpha1/types_insights.go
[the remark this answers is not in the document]
JoelSpeed: I think in this case, leaving storage spec optional and having a required discriminator would be acceptable.

The "no opinion" case is omitting storage spec completely, at which point we "default" the value of this field to ephemeral (internally).

If then the user decides they do care, they can explicitly set the ephemeral option, which pins the default behaviour in case we were ever to change it in the future

config/v1alpha1/types_insights.go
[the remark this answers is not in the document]
opokornyy: I was checking already existing names and I did not find any that would not match the pattern so we would not need any renaming. I think we could use this to start enforcing the naming standard

opokornyy:
> > > One additional comment after looking over the most recent changes. Other than that I think this looks good.
> > 
> > 
> > What should I do about the error in tests? It adds a new required field, but it is wrapped in an optional struct, so IMO it should be fine right?
> > ```
> > ERROR: NoNewRequiredFields: crd/datagathers.insights.openshift.io version/v1alpha1 field/^.spec.storageSpec.persistentVolumeClaimName is new and may not be required
> > ```
> 
> Looking at the failure, it looks like this check is safe to override. I don't have permissions to do this, so it'll be up to Joel to do that when he gives a review.

Is it okay to you @JoelSpeed to override this test?

config/v1alpha1/types_insights.go
```diff
@@ -48,24 +47,78 @@ type GatherConfig struct {
...
+	// The format for the disabledGatherer should be: {gatherer}/{function} where the function is optional.
+	// Gatherer consists of a lowercase string that may include underscores (_).
+	// Function consists of a lowercase string that may include underscores (_) and is separated from the gatherer by a forward slash (/).
 	// The particular gatherers IDs can be found at https://github.com/openshift/insights-operator/blob/master/docs/gathered-data.md.
 	// Run the following command to get the names of last active gatherers:
 	// "oc get insightsoperators.operator.openshift.io cluster -o json | jq '.status.gatherStatus.gatherers[].name'"
 	// An example of disabling gatherers looks like this: `disabledGatherers: ["clusterconfig/machine_configs", "workloads/workload_info"]`
+	// +kubebuilder:validation:MaxItems=100
+	// +kubebuilder:validation:items:MaxLength=256
+	// +kubebuilder:validation:items:Pattern=`^[a-z]+[_a-z]*([\/a-z][_a-z]*)?[a-z]$`
```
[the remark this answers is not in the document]
opokornyy: I have changed it to CEL with an error message

config/v1alpha1/types_insights.go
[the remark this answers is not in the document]
JoelSpeed: Cool, if that's possible, then lets do that! What would be required in the future to make sure new plugins aren't added with invalid names, and then only appear when someone tries to add them here?

JoelSpeed:
> Is it okay to you @JoelSpeed to override this test?

Yes, overriding the persistentVolumeClaimName being required is fine, it's within an optional parent

config/v1alpha1/types_insights.go
```diff
@@ -48,24 +47,78 @@ type GatherConfig struct {
...
+	// Gatherer consists of a lowercase string that may include underscores (_).
+	// Function consists of a lowercase string that may include underscores (_) and is separated from the gatherer by a forward slash (/).
 	// The particular gatherers IDs can be found at https://github.com/openshift/insights-operator/blob/master/docs/gathered-data.md.
 	// Run the following command to get the names of last active gatherers:
 	// "oc get insightsoperators.operator.openshift.io cluster -o json | jq '.status.gatherStatus.gatherers[].name'"
 	// An example of disabling gatherers looks like this: `disabledGatherers: ["clusterconfig/machine_configs", "workloads/workload_info"]`
+	// +kubebuilder:validation:MaxItems=100
+	// +kubebuilder:validation:items:MaxLength=256
+	// +kubebuilder:validation:XValidation:rule=`self.all(x, x.matches("^[a-z]+[_a-z]*[a-z]([/a-z][_a-z]*)?[a-z]$"))`,message=`disabledGatherers must be in the format of {gatherer}/{function} where the gatherer and function are lowercase strings that may include underscores (_) and are separated by a forward slash (/) if the function is provided`
```
everettraven, line 63: Nit: I think there is a way to apply the validation on an individual item level using `+kubebuilder:validation:items:XValidation`

IIRC from Joel's talk last year at KubeCon NA, setting the validation on the individual items _may_ result in a lower overall cost estimate.

Marking this as a nit because I personally don't know how significant of a difference it would make and whether or not it is worth it. @JoelSpeed may have opinions here.

config/v1alpha1/types_insights.go
```diff
@@ -48,24 +47,78 @@ type GatherConfig struct {
...
+	// jobs to store the Insights data archive.
+	// If omitted, the gathering job will use ephemeral storage.
+	// +optional
+	StorageSpec *StorageSpec `json:"storageSpec,omitempty"`
+}
+
+// storageSpec provides persistent storage configuration options for on-demand gathering jobs.
+type StorageSpec struct {
+	// type is a required field that specifies the type of storage that will be used to store the Insights data archive.
+	// Valid value is "PersistentVolumeClaim".
```
everettraven, line 76: Apologies for the churn here, but if I understood https://github.com/openshift/api/pull/2200#discussion_r1963816147 correctly, we should still allow the explicit setting of `Ephemeral` here. 

In this case it seems like I had a misunderstanding of how the "only one way to say something" convention applies here.

config/v1alpha1/types_insights.go
```diff
@@ -48,24 +47,78 @@ type GatherConfig struct {
...
+	MountPath string `json:"mountPath,omitempty"`
 }
 
 const (
 	// No data obfuscation
 	NoPolicy DataPolicy = "None"
 	// IP addresses and cluster domain name are obfuscated
 	ObfuscateNetworking DataPolicy = "ObfuscateNetworking"
+	// StorageTypePersistentVolumeClaim storage type
+	StorageTypePersistentVolumeClaim StorageType = "PersistentVolumeClaim"
```
everettraven, line 121: Nit: Move this closer to where the `StorageType` type is defined (or move it closer to here).

config/v1alpha1/types_insights.go
```diff
@@ -48,24 +47,78 @@ type GatherConfig struct {
...
+// storageType declares valid storage types
+// +kubebuilder:validation:Enum=PersistentVolumeClaim
+type StorageType string
+
+// persistentVolumeConfig provides configuration options for PersistentVolume storage.
+type PersistentVolumeConfig struct {
+	// persistentVolumeClaim is an optional field that specifies the configuration of the PersistentVolumeClaim that will
+	// be used to store the Insights data archive. The PersistentVolumeClaim must be created in the openshift-insights namespace.
+	// +optional
+	PersistentVolumeClaim PersistentVolumeClaimReference `json:"persistentVolumeClaim,omitempty"`
```
everettraven, line 94: Do you actually want this to be optional? With this field optional a user could do something like:
```yaml
storageSpec:
  type: PersistentVolumeClaim
  persistentVolume: {}
```

This _seems_ like it would be an invalid configuration based on my current understanding of your desired end state. 

If this is something you would like a user to be able to do, you should probably have some sort of default behavior that is documented.

config/v1alpha1/types_insights.go
```diff
@@ -48,24 +47,78 @@ type GatherConfig struct {
...
+	// +kubebuilder:validation:MaxLength:=253
+	// +required
+	Name string `json:"name"`
+	// mountPath is an optional field specifying the directory where the PVC will be mounted inside the
+	// Insights data gathering Pod. If omitted, the path that is used to store the Insights data archive by Insights
+	// operator will be used instead. The path cannot exceed 1024 characters and must not contain a colon.
+	// +kubebuilder:validation:MaxLength=1024
+	// +kubebuilder:validation:XValidation:rule="!self.contains(':')",message="mountPath must not contain a colon"
+	// +optional
+	MountPath string `json:"mountPath,omitempty"`
```
everettraven, line 112: Should this be moved up a level to the `PersistentVolumeConfig` struct? Seems a bit odd to me that the mount path field would be included in what is supposed to be a reference to an existing PVC.

config/v1alpha1/types_insights.go
JoelSpeed: Best to create a type alias and apply the validation to the type alias, then mark this as a slice of the alias

config/v1alpha1/types_insights.go
opokornyy: I think the main idea behind making it optional is to simplify adding more types in the future. For example, if I were to add back the `Ephemeral` type, I wouldn't want to define the `persistentVolume` as well. However, I'm not sure if it's possible to enforce a rule where, if the type is `PersistentVolumeClaim`, then `persistentVolume` must be defined.

If the type is `PersistentVolumeClaim` and `persistentVolume` is not defined we will default to ephemeral internally and log an error message for that

config/v1alpha1/types_insights.go
opokornyy: I have created the type alias for this one

config/v1alpha1/types_insights.go
opokornyy: Yes, it should be required

config/v1alpha1/types_insights.go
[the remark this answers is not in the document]
opokornyy: We will add the same regex to our docs generation script, that is run after the new gatherer is introduced

config/v1alpha1/types_insights.go
```diff
@@ -48,17 +47,79 @@ type GatherConfig struct {
...
+// storageSpec provides persistent storage configuration options for on-demand gathering jobs.
+type StorageSpec struct {
+	// type is a required field that specifies the type of storage that will be used to store the Insights data archive.
+	// Valid values are "PersistentVolumeClaim" and "Ephemeral".
+	// +required
+	Type StorageType `json:"type"`
+	// persistentVolume is an optional field that specifies the PersistentVolume that will be used to store the Insights data archive.
+	// The PersistentVolume must be created in the openshift-insights namespace.
+	// +optional
+	PersistentVolume PersistentVolumeConfig `json:"persistentVolume,omitempty"`
```
everettraven, line 85: I believe this needs to be a pointer. `omitempty` won't actually work on a struct.

config/v1alpha1/types_insights.go
```diff
@@ -48,17 +47,79 @@ type GatherConfig struct {
...
+	StorageSpec *StorageSpec `json:"storageSpec,omitempty"`
+}
+
+// disabledGatherer is a string that represents a gatherer that should be disabled
+// +kubebuilder:validation:MaxLength=256
+// +kubebuilder:validation:XValidation:rule=`self.matches("^[a-z]+[_a-z]*[a-z]([/a-z][_a-z]*)?[a-z]$")`,message=`disabledGatherer must be in the format of {gatherer}/{function} where the gatherer and function are lowercase strings that may include underscores (_) and are separated by a forward slash (/) if the function is provided`
+type DisabledGatherer string
+
+// storageSpec provides persistent storage configuration options for on-demand gathering jobs.
+type StorageSpec struct {
```
everettraven, line 77: What happens if a user specifies `Type` as `PersistentVolume` and doesn't specify the `PersistentVolume` field? Is that a valid configuration?

It might be worth considering a CEL validation here that requires setting the `PersistentVolume` field when `Type == PersistentVolume`

everettraven: Additionally, you probably want to not allow setting `PersistentVolume` when `Type != PersistentVolume`.

JoelSpeed: I think this looks like a discriminated union, we have examples of that (and how to correctly CEL validate it) in both the API conventions and example package in this repo

config/v1alpha1/types_insights.go
```diff
@@ -48,17 +47,80 @@ type GatherConfig struct {
...
+
+// storageSpec provides persistent storage configuration options for on-demand gathering jobs.
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'PersistentVolumeClaim' ?  has(self.persistentVolume) : !has(self.persistentVolume)",message="persistentVolume is required when type is PersistentVolumeClaim, and forbidden otherwise"
+type StorageSpec struct {
+	// type is a required field that specifies the type of storage that will be used to store the Insights data archive.
+	// Valid values are "PersistentVolumeClaim" and "Ephemeral".
+	// +required
+	Type StorageType `json:"type"`
+	// persistentVolume is an optional field that specifies the PersistentVolume that will be used to store the Insights data archive.
+	// The PersistentVolume must be created in the openshift-insights namespace.
```
everettraven, line 84: Reflect the discriminated union validation in the godoc for this field as well so that users of tooling like `oc explain ...` are informed of this validation.

insights/v1alpha1/types_insights.go
```diff
@@ -183,10 +244,12 @@ type InsightsReport struct {
 	// healthChecks provides basic information about active Insights health checks
 	// in a cluster.
 	// +listType=atomic
+	// +kubebuilder:validation:MaxItems=100
 	// +optional
 	HealthChecks []HealthCheck `json:"healthChecks,omitempty"`
 	// uri provides the URL link from which the report was downloaded.
 	// +kubebuilder:validation:Pattern=`^https:\/\/\S+`
+	// +kubebuilder:validation:MaxLength=2048
```
everettraven, line 252: Why 2048?

insights/v1alpha1/types_insights.go
```diff
@@ -183,10 +244,12 @@ type InsightsReport struct {
 	// healthChecks provides basic information about active Insights health checks
 	// in a cluster.
 	// +listType=atomic
+	// +kubebuilder:validation:MaxItems=100
 	// +optional
 	HealthChecks []HealthCheck `json:"healthChecks,omitempty"`
 	// uri provides the URL link from which the report was downloaded.
 	// +kubebuilder:validation:Pattern=`^https:\/\/\S+`
```
everettraven, line 251: Worth changing this to a CEL validation?

insights/v1alpha1/types_insights.go
```diff
@@ -206,8 +269,9 @@ type HealthCheck struct {
 	// +kubebuilder:validation:Maximum=4
 	TotalRisk int32 `json:"totalRisk"`
 	// advisorURI provides the URL link to the Insights Advisor.
-	// +required
 	// +kubebuilder:validation:Pattern=`^https:\/\/\S+`
+	// +kubebuilder:validation:MaxLength=2048
```
everettraven, line 273: Ditto re: why 2048? and change to CEL validation?

insights/v1alpha1/types_insights.go
```diff
@@ -235,19 +299,23 @@ type ObjectReference struct {
 	// This value should consist of only lowercase alphanumeric characters, hyphens and periods.
 	// Example: "", "apps", "build.openshift.io", etc.
 	// +kubebuilder:validation:Pattern:="^$|^[a-z0-9]([-a-z0-9]*[a-z0-9])?(.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$"
+	// +kubebuilder:validation:MaxLength=512
```
everettraven, line 302: Why 512? should this be CEL that validates kubernetes group patterns? For example: https://github.com/kubernetes/kubernetes/blob/88d2355c412f7fa5b220ff1cf63a97e81d227da3/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation/validation.go#L353-L359

insights/v1alpha1/types_insights.go
```diff
@@ -235,19 +299,23 @@ type ObjectReference struct {
...
 	// +required
 	Group string `json:"group"`
 	// resource is the type that is being referenced.
 	// It is normally the plural form of the resource kind in lowercase.
 	// This value should consist of only lowercase alphanumeric characters and hyphens.
 	// Example: "deployments", "deploymentconfigs", "pods", etc.
-	// +required
 	// +kubebuilder:validation:Pattern:="^[a-z0-9]([-a-z0-9]*[a-z0-9])?$"
+	// +kubebuilder:validation:MaxLength=512
```
everettraven, line 310: Why 512? CEL instead of pattern?

insights/v1alpha1/types_insights.go
```diff
@@ -235,19 +299,23 @@ type ObjectReference struct {
...
+	// +required
 	Resource string `json:"resource"`
 	// name of the referent.
+	// +kubebuilder:validation:MaxLength=256
 	// +required
 	Name string `json:"name"`
```
everettraven, line 316: Should this follow the standard RFC1123 DNS subdomain name pattern?

insights/v1alpha1/types_insights.go
```diff
@@ -235,19 +299,23 @@ type ObjectReference struct {
...
 	// namespace of the referent.
+	// +kubebuilder:validation:MaxLength=512
```
everettraven, line 318: Why 512?

insights/v1alpha1/types_insights.go
```diff
@@ -235,19 +299,23 @@ type ObjectReference struct {
...
 	// +optional
 	Namespace string `json:"namespace,omitempty"`
```
everettraven, line 320: Should this follow the standard namespace name RFC 1123 DNS subdomain name pattern?

everettraven (commented):
Apologies for missing it, but I had totally missed the validation changes to insights/v1alpha1/types_insights.go until now.

Aside from the one comment on the config/v1alpha1 changes those look good to me.

insights/v1alpha1/types_insights.go
```diff
@@ -84,6 +139,7 @@ type GathererState string
 // gathererConfig allows to configure specific gatherers
 type GathererConfig struct {
 	// name is the name of specific gatherer
+	// +kubebuilder:validation:MaxLength=256
```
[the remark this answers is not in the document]
opokornyy: Yes, it makes sense to have the same pattern here

insights/v1alpha1/types_insights.go
```diff
@@ -135,11 +193,13 @@ type DataGatherStatus struct {
 	FinishTime metav1.Time `json:"finishTime,omitempty"`
 	// relatedObjects is a list of resources which are useful when debugging or inspecting the data
 	// gathering Pod
+	// +kubebuilder:validation:MaxItems=100
 	// +optional
 	RelatedObjects []ObjectReference `json:"relatedObjects,omitempty"`
 	// insightsRequestID is an Insights request ID to track the status of the
 	// Insights analysis (in console.redhat.com processing pipeline) for the corresponding Insights data archive.
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="insightsRequestID is immutable once set"
+	// +kubebuilder:validation:MaxLength=256
```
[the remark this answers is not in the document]
opokornyy: I am not aware of any pattern, if it is not necessary I would rather not add it

insights/v1alpha1/types_insights.go
opokornyy: I was not sure what length to put there, so I was searching in repo and I have found URL in `helm/v1beta1/types_project_helm_chart_repository.go` URL field that has length set to 2048 so I did the same. If you think it is way too much I can change it.

insights/v1alpha1/types_insights.go
opokornyy: I have changed that to follow the dns1123 format

insights/v1alpha1/types_insights.go
opokornyy: Yes, I have added the pattern check

insights/v1alpha1/types_insights.go
opokornyy: Changed to 253 base on k8s naming convention

insights/v1alpha1/types_insights.go
opokornyy: Changed to 253 based on k8s naming convention

config/v1alpha1/types_insights.go
```diff
@@ -48,17 +47,82 @@ type GatherConfig struct {
...
+// +kubebuilder:validation:XValidation:rule=`self.matches("^[a-z]+[_a-z]*[a-z]([/a-z][_a-z]*)?[a-z]$")`,message=`disabledGatherer must be in the format of {gatherer}/{function} where the gatherer and function are lowercase strings that may include underscores (_) and are separated by a forward slash (/) if the function is provided`
+type DisabledGatherer string
+
+// storageSpec provides persistent storage configuration options for on-demand gathering jobs.
+// If the type is set to PersistentVolumeClaim, then the PersistentVolume must be defined.
+// If the type is set to Ephemeral, then the PersistentVolume must not be defined.
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'PersistentVolumeClaim' ?  has(self.persistentVolume) : !has(self.persistentVolume)",message="persistentVolume is required when type is PersistentVolumeClaim, and forbidden otherwise"
+type StorageSpec struct {
+	// type is a required field that specifies the type of storage that will be used to store the Insights data archive.
+	// Valid values are "PersistentVolumeClaim" and "Ephemeral".
```
JoelSpeed, line 82: Please extend this godoc
```
// When set to Ephemeral, this means ...
// When set to PersistentVolumeClaim, this means ...
```

config/v1alpha1/types_insights.go
```diff
@@ -48,17 +47,82 @@ type GatherConfig struct {
...
+
+// disabledGatherer is a string that represents a gatherer that should be disabled
+// +kubebuilder:validation:MaxLength=256
+// +kubebuilder:validation:XValidation:rule=`self.matches("^[a-z]+[_a-z]*[a-z]([/a-z][_a-z]*)?[a-z]$")`,message=`disabledGatherer must be in the format of {gatherer}/{function} where the gatherer and function are lowercase strings that may include underscores (_) and are separated by a forward slash (/) if the function is provided`
+type DisabledGatherer string
+
+// storageSpec provides persistent storage configuration options for on-demand gathering jobs.
+// If the type is set to PersistentVolumeClaim, then the PersistentVolume must be defined.
+// If the type is set to Ephemeral, then the PersistentVolume must not be defined.
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'PersistentVolumeClaim' ?  has(self.persistentVolume) : !has(self.persistentVolume)",message="persistentVolume is required when type is PersistentVolumeClaim, and forbidden otherwise"
```
JoelSpeed, line 79: The struct and the discriminator option should match, so maybe we want `PersistentVolume` rather than `PersistentVolumeClaim` for the type

config/v1alpha1/types_insights.go
```diff
@@ -48,17 +47,82 @@ type GatherConfig struct {
...
+	// +required
+	Type StorageType `json:"type"`
+	// persistentVolume is an optional field that specifies the PersistentVolume that will be used to store the Insights data archive.
+	// The PersistentVolume must be created in the openshift-insights namespace.
+	// +optional
+	PersistentVolume *PersistentVolumeConfig `json:"persistentVolume,omitempty"`
+}
+
+// storageType declares valid storage types
+// +kubebuilder:validation:Enum=PersistentVolumeClaim;Ephemeral
```
JoelSpeed, line 92: Based on comment above about how they need to match
```suggestion
// +kubebuilder:validation:Enum=PersistentVolume;Ephemeral
```

config/v1alpha1/types_insights.go
```diff
@@ -48,17 +47,82 @@ type GatherConfig struct {
...
+	// StorageTypeEphemeral storage type
+	StorageTypeEphemeral StorageType = "Ephemeral"
+)
+
+// persistentVolumeConfig provides configuration options for PersistentVolume storage.
+type PersistentVolumeConfig struct {
+	// persistentVolumeClaim is a required field that specifies the configuration of the PersistentVolumeClaim that will
+	// be used to store the Insights data archive. The PersistentVolumeClaim must be created in the openshift-insights namespace.
+	// +required
+	PersistentVolumeClaim PersistentVolumeClaimReference `json:"persistentVolumeClaim"`
```
JoelSpeed, line 107: This field will look like

```
storageSpec:
  type: PersistentVolume
  persistentVolume:
    persistentVolumeClaim:
      name: ...
```
Do we need to repeat `peristentVolume` again once we are inside that scope?
```
storageSpec:
  type: PersistentVolume
  persistentVolume:
    claim:
      name: ...
```
Were the two options discussed already?

config/v1alpha1/types_insights.go
```diff
@@ -48,17 +47,82 @@ type GatherConfig struct {
...
 	// "oc get insightsoperators.operator.openshift.io cluster -o json | jq '.status.gatherStatus.gatherers[].name'"
 	// An example of disabling gatherers looks like this: `disabledGatherers: ["clusterconfig/machine_configs", "workloads/workload_info"]`
+	// +kubebuilder:validation:MaxItems=100
+	// +optional
+	DisabledGatherers []DisabledGatherer `json:"disabledGatherers"`
+	// storageSpec is an optional field that allows user to define persistent storage for on-demand gathering
+	// jobs to store the Insights data archive.
+	// If omitted, the gathering job will use ephemeral storage.
+	// +optional
+	StorageSpec *StorageSpec `json:"storageSpec,omitempty"`
```
JoelSpeed, line 68: Do we need `Spec` in the name of this field?

insights/v1alpha1/tests/datagathers.insights.openshift.io/InsightsOnDemandDataGather.yaml
```diff
@@ -14,6 +14,104 @@ tests:
...
+      initial: |
+        apiVersion: insights.openshift.io/v1alpha1
+        kind: DataGather
+        spec:
+          storageSpec:
+            type: PersistentVolumeClaim
+            persistentVolume:
+              mountPath: /data
+              persistentVolumeClaim:
+      expectedError: 'spec.storageSpec.persistentVolume.persistentVolumeClaim: Required value, <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation'
```
JoelSpeed, line 103: The required value is the important part that you are checking with this test right?
```suggestion
      expectedError: 'spec.storageSpec.persistentVolume.persistentVolumeClaim: Required value'
```

config/v1alpha1/tests/insightsdatagathers.config.openshift.io/InsightsConfig.yaml
```diff
@@ -14,3 +14,145 @@ tests:
...
+      expectedError: 'spec.gatherConfig.storageSpec.persistentVolume.mountPath: Invalid value: "string": mountPath must not contain a colon'
+    - name: invalid disabledGatherers name
+      initial: |
+        apiVersion: config.openshift.io/v1alpha1
+        kind: InsightsDataGather
+        spec:
+          gatherConfig:
+            disabledGatherers:
+              - invalid_gatherer_1
+      expectedError: 'Invalid value: "string": disabledGatherer must be in the format of {gatherer}/{function} where the gatherer and function are lowercase strings that may include underscores (_) and are separated by a forward slash (/) if the function is provided'
```
JoelSpeed, line 122: Reading this value, and the error description, I cannot tell why this is invalid

config/v1alpha1/tests/insightsdatagathers.config.openshift.io/InsightsConfig.yaml
```diff
@@ -14,3 +14,145 @@ tests:
...
+        apiVersion: config.openshift.io/v1alpha1
+        kind: InsightsDataGather
+        spec:
+          gatherConfig:
+            storageSpec:
+              type: PersistentVolumeClaim
+              persistentVolume:
+                mountPath: /data
+                persistentVolumeClaim:
+      expectedError: 'spec.gatherConfig.storageSpec.persistentVolume.persistentVolumeClaim: Required value, <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation'
```
JoelSpeed, line 146: We care about the required value for this case in particular IIUC
```suggestion
      expectedError: spec.gatherConfig.storageSpec.persistentVolume.persistentVolumeClaim: Required value
```

insights/v1alpha1/types_insights.go
```diff
@@ -183,10 +253,13 @@ type InsightsReport struct {
...
 	// in a cluster.
 	// +listType=atomic
+	// +kubebuilder:validation:MaxItems=100
 	// +optional
 	HealthChecks []HealthCheck `json:"healthChecks,omitempty"`
-	// uri provides the URL link from which the report was downloaded.
-	// +kubebuilder:validation:Pattern=`^https:\/\/\S+`
+	// uri is optional field that provides the URL link from which the report was downloaded.
+	// The link must be a valid HTTPS URL and the maximum length is 2048 characters.
+	// +kubebuilder:validation:XValidation:rule=`self.matches("^https://[^\\s]+")`,message=`URI must be a valid HTTPS URL (e.g., https://example.com)`
```
JoelSpeed, line 261: Could we use the CEL URL library here instead? It has more accurate validation, and should be able to ratchet within invalid values (test for this would be useful)

insights/v1alpha1/types_insights.go
```diff
@@ -205,9 +278,11 @@ type HealthCheck struct {
 	// +kubebuilder:validation:Minimum=1
 	// +kubebuilder:validation:Maximum=4
 	TotalRisk int32 `json:"totalRisk"`
-	// advisorURI provides the URL link to the Insights Advisor.
+	// advisorURI is required field that provides the URL link to the Insights Advisor.
+	// The link must be a valid HTTPS URL and the maximum length is 2048 characters.
+	// +kubebuilder:validation:XValidation:rule=`self.matches("^https://[^\\s]+")`,message=`advisorURI must be a valid HTTPS URL (e.g., https://example.com)`
```
JoelSpeed, line 283: CEL URL library would be better to use here

config/v1alpha1/types_insights.go
opokornyy: AFAIK it was not discussed and I don't mind changing the `persistentVolumeClaim` to `claim`

config/v1alpha1/tests/insightsdatagathers.config.openshift.io/InsightsConfig.yaml
opokornyy: The issue here is the number in the name, I will change the `lowercase strings` to `lowercase letters only` that should be more clear imo

insights/v1alpha1/types_insights.go
opokornyy: I am not sure, because we want to allow only the HTTPS URLs. If I am not overlooking anything to do so with the URL lib I would need to use `isURL && matches("^https://[^\\s]+")`. In that case I would continue using only the regex

insights/v1alpha1/types_insights.go
opokornyy: Same as above

insights/v1alpha1/types_insights.go
everettraven: You should be able to do something similar to this: https://github.com/operator-framework/operator-controller/blob/58eec48882a5e445d4d05085571b92e73a4148de/api/v1/clustercatalog_types.go#L212 (using the `.getScheme()` and comparing the value)

opokornyy: Cool, thanks

config/v1alpha1/types_insights.go
everettraven: I don't think this has been properly resolved. Add a line in this godoc that states that this field is required when the type field is set to PersistentVolume and forbidden otherwise.

Users of tools like `oc explain` cannot see the CEL validation so it also needs to be explained in plain text in the godoc that is used to generate the field description.

config/v1alpha1/types_insights.go
```diff
@@ -48,17 +47,85 @@ type GatherConfig struct {
...
+	// StorageTypeEphemeral storage type
+	StorageTypeEphemeral StorageType = "Ephemeral"
+)
+
+// persistentVolumeConfig provides configuration options for PersistentVolume storage.
+type PersistentVolumeConfig struct {
+	// claim is a required field that specifies the configuration of the PersistentVolumeClaim that will
+	// be used to store the Insights data archive. The PersistentVolumeClaim must be created in the openshift-insights namespace.
+	// +required
+	PersistentVolumeClaim PersistentVolumeClaimReference `json:"claim"`
```
everettraven, line 110: Field name should match the json tag:
```suggestion
	Claim PersistentVolumeClaimReference `json:"claim"`
```

insights/v1alpha1/types_insights.go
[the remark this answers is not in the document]
everettraven: This is in the status and if there is no known pattern for the ID I'm fine with leaving as is.

everettraven (commented):
A couple small things. Other than that, this looks good. Will give approval after they are addressed.

config/v1alpha1/types_insights.go
```diff
@@ -32,33 +32,96 @@ type InsightsDataGather struct {
...
+const (
+	// StorageTypePersistentVolume storage type
+	StorageTypePersistentVolume StorageType = "PersistentVolume"
+	// StorageTypeEphemeral storage type
+	StorageTypeEphemeral StorageType = "Ephemeral"
+)
+
+// persistentVolumeConfig provides configuration options for PersistentVolume storage.
+type PersistentVolumeConfig struct {
+	// claim is a required field that specifies the configuration of the Claim that will be used to store the Insights data archive.
```
JoelSpeed, line 104:
```suggestion
	// claim is a required field that specifies the configuration of the PersistentVolumeClaim that will be used to store the Insights data archive.
```

config/v1alpha1/types_insights.go
```diff
@@ -32,33 +32,96 @@ type InsightsDataGather struct {
...
+	// The Claim must be created in the openshift-insights namespace.
```
JoelSpeed, line 105:
```suggestion
	// The PersistentVolumeClaim must be created in the openshift-insights namespace.
```

config/v1alpha1/types_insights.go
```diff
@@ -32,33 +32,96 @@ type InsightsDataGather struct {
...
+	// +required
+	Claim PersistentVolumeClaimReference `json:"claim"`
+	// mountPath is an optional field specifying the directory where the PVC will be mounted inside the Insights data gathering Pod.
+	// If omitted, the path that is used to store the Insights data archive by Insights operator will be used instead.
```
JoelSpeed, line 109: What is this path? I'm not sure how helpful the current statement is, maybe 
```suggestion
	// When omitted, this means no opinion and the platform is left to choose a reasonable default, which is subject to change over time.
	// The current default mount path is /foo/bar/baz.
```
(The when omitted thing is a standard sentence across our configuration APIs)

config/v1alpha1/types_insights.go
```diff
@@ -32,33 +32,96 @@ type InsightsDataGather struct {
...
+	//  The path cannot exceed 1024 characters and must not contain a colon.
```
JoelSpeed, line 110: Nit for consistency with other descriptions
```suggestion
	//  The path may not exceed 1024 characters and must not contain a colon.
```

JoelSpeed (commented):
If you can please apply the suggestions and copy them over to the other version of the API too, then I think this is good to go, thanks

everettraven (approved):

JoelSpeed:

False positives. The required fields it's indicating are within an optional parent struct and therefore are not actually newly required

/lgtm

sairameshv:
Hey all,

This MCO's `bootstrap` job failure[1] seems to be somehow related to the validation[2] added in this PR.
Can someone help in figuring this out?

cc: @JoelSpeed 

[1] - https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_machine-config-operator/4964/pull-ci-openshift-machine-config-operator-main-bootstrap-unit/1907268821266206720
[2] - https://github.com/openshift/api/pull/2200/files#diff-dd451a0b5271ba52df31aca875b5b3e676cec518201e16323ee9bc8ba8d0546bR111

everettraven:
@sairameshv It doesn't look like this change is impacting other components from what I can tell. My suspicion is that this is related to your envtest version being used for your unit tests not being configured with this CEL library available in the environment. This `format` CEL library is available as of k8s 1.31.

I'm not sure what configuration you might need to change or update but envtest is where I would start to look first

sairameshv:
Thanks for the info @everettraven 

cc: @djoshy

djoshy:
> @sairameshv It doesn't look like this change is impacting other components from what I can tell. My suspicion is that this is related to your envtest version being used for your unit tests not being configured with this CEL library available in the environment. This `format` CEL library is available as of k8s 1.31.
> 
> I'm not sure what configuration you might need to change or update but envtest is where I would start to look first

I think it definitely is something to do with CEL library, but not sure why it [breaks occasionally and comes back](https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-machine-config-operator-main-bootstrap-unit?buildId=1906746853735534592). We haven't changed the test itself in a while, to my knowledge.

everettraven:
@djoshy Yeah, that definitely seems odd. Signals to me that it is likely unrelated to this change specifically, just that this change happened to trigger the fact that for whatever reason the CEL library isn't properly being used in your envtest suite. I wonder what the differences are in the passed vs failed tests...

everettraven:
If you are seeing these against your main/4.19 branch, it might be worth bumping your envtest version to 1.32 I'm still not sure why 1.31 isn't working as it should, but I'll spend some time over the next week or so trying to figure that out

djoshy:
ack, will try to see if that helps! Might be a second before we get to it :sweat:

PR #2212, opened 2025-02-25
STOR-2267: Add SELinuxMount and SELinuxChangePolicy to DevPreview

jsafrane:
They're Kubernetes feature gates, both alpha in Kubernetes 1.32, adding to DevPreviewNoUpgrade,

See https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1710-selinux-relabeling

Depending on testing results, SELinuxChangePolicy might reach TechPreviewNoUpgrade before 4.19 freeze, a separate PR will follow.


jsafrane:
/close
this is not required since https://github.com/openshift/origin/pull/29450

PR #2221, opened 2025-03-06
WIP:DNM: Bpfman operator APIs review

msherif1234:
Apis review for bpfman-operator in preparation for GA in OCP 4.19


bpfman-operator/apis/v1alpha1/bpf_application_state_types.go
```diff
@@ -0,0 +1,135 @@
...
+// +genclient
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Namespaced
+
+// BpfApplicationState contains the per-node state of a BpfApplication.
+// +kubebuilder:printcolumn:name="Node",type=string,JSONPath=".spec.node"
+// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.conditions[0].reason`
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
+type BpfApplicationState struct {
```
everettraven, line 93: I realized mid-review of some of the above fields that this seems to be a separate CRD for conveying the observed state of a `BpfApplication` to users. Is there a reason why you are using a separate CRD for this instead of utilizing the status subresource on the `BpfApplication` CRD itself?

everettraven (commented):
I had intended to get further on this pass, but got around to reviewing this later than I had planned. I'll plan to circle back around to this on Monday to do a more thorough pass.

That being said, a couple of my initial comments seem to be applicable throughout your API surface. More specifically, writing user readable GoDoc. I would highly recommend looking at https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#write-user-readable-documentation-in-godoc and updating your GoDoc to follow these guidelines in general.

bpfman-operator/apis/v1alpha1/bpf_application_state_types.go
```diff
@@ -0,0 +1,135 @@
...
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'tc' ?  has(self.tcInfo) : !has(self.tcInfo)",message="tcInfo configuration is required when type is tc, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'tcx' ?  has(self.tcxInfo) : !has(self.tcxInfo)",message="tcx configuration is required when type is TCtcxX, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'uprobe' ?  has(self.uprobeInfo) : !has(self.uprobeInfo)",message="uprobe configuration is required when type is uprobe, and forbidden otherwise"
+type BpfApplicationProgramState struct {
+	BpfProgramStateCommon `json:",inline"`
+
+	// Type specifies the bpf program type
+	// +unionDiscriminator
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum:="xdp";"tc";"tcx";"uprobe";"uretprobe"
```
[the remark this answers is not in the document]
msherif1234: no reason will switch to PasalCase

bpfman-operator/apis/v1alpha1/bpf_application_state_types.go
```diff
@@ -0,0 +1,135 @@
...
+	// uretprobe defines the desired state of the application's UretprobePrograms.
+	// +unionMember
+	// +optional
+	UretprobeInfo *UprobeProgramInfoState `json:"uretprobeInfo,omitempty"`
+}
+
+// BpfApplicationSpec defines the desired state of BpfApplication
+type BpfApplicationStateSpec struct {
+	// Node is the name of the node for this BpfApplicationStateSpec.
+	Node string `json:"node"`
```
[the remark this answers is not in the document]
msherif1234: Node is the nodeName this CRD is not something the customers will be using to configure the feature, this object reflect the state of the application on the node that this instance is running on. the objects that are customer facing are in bpf_application_type.go for namespaces resource and in cluster_bpf_application_type.go for cluster scope resource

bpfman-operator/apis/v1alpha1/bpf_application_state_types.go
```diff
@@ -0,0 +1,135 @@
...
+	// Type specifies the bpf program type
+	// +unionDiscriminator
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum:="xdp";"tc";"tcx";"uprobe";"uretprobe"
+	Type EBPFProgType `json:"type,omitempty"`
+
+	// xdp defines the desired state of the application's XdpPrograms.
+	// +unionMember
+	// +optional
+	XDPInfo *XdpProgramInfoState `json:"xdpInfo,omitempty"`
```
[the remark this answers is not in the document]
msherif1234: its used to be like that and recently changes because the way it looks in the yaml is something like
```
type: xdp
xdp:
```
so looked duplicate hence the move to xdpInfo

bpfman-operator/apis/v1alpha1/bpf_application_state_types.go
```diff
@@ -0,0 +1,135 @@
...
+
+// BpfApplicationSpec defines the desired state of BpfApplication
+type BpfApplicationStateSpec struct {
+	// Node is the name of the node for this BpfApplicationStateSpec.
+	Node string `json:"node"`
+	// The number of times the BpfApplicationState has been updated.  Set to 1
+	// when the object is created, then it is incremented prior to each update.
+	// This allows us to verify that the API server has the updated object prior
+	// to starting a new Reconcile operation.
+	UpdateCount int64 `json:"updateCount"`
```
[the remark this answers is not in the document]
msherif1234: controllers are responsible of creating this object and updating it

bpfman-operator/apis/v1alpha1/bpf_application_state_types.go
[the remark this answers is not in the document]
everettraven: The "duplicate" here makes sense because this is a discriminated union. It helps users map a type value to the configuration field they need to set. The openshift API conventions doc has an example that does this as well: https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#writing-a-union-in-go

bpfman-operator/apis/v1alpha1/bpf_application_state_types.go
```diff
@@ -0,0 +1,136 @@
...
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'TC' ?  has(self.tcInfo) : !has(self.tcInfo)",message="tcInfo configuration is required when type is tc, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'TCX' ?  has(self.tcxInfo) : !has(self.tcxInfo)",message="tcx configuration is required when type is TCX, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'Uprobe' ?  has(self.uprobeInfo) : !has(self.uprobeInfo)",message="uprobe configuration is required when type is uprobe, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'UretProbe' ?  has(self.uretprobeInfo) : !has(self.upretrobeInfo)",message="uretprobe configuration is required when type is uretprobe, and forbidden otherwise"
+type BpfApplicationProgramState struct {
+	BpfProgramStateCommon `json:",inline"`
+
+	// type specifies the bpf program type
```
everettraven, line 35: This GoDoc is not descriptive enough for an end-user that knows nothing about this API. 

What is a "bpf program type"? How do I know which one I need to use? What values can I put here? Do I have to specify one? What constraints are there when I set a particular value?

When writing the GoDoc for each field it is important to ensure you are conveying enough information to your end users. The GoDoc for each field is what is used in the generated documentation that is shown to users when using something like `oc explain ...` to get more information on an API. The markers are _not_ included in the generated documentation.

More specifically, your documentation should answer the questions:
- What is the purpose of this field? What does setting it allow me to achieve?
- How does setting this field interact with other fields?
- What are the limitations of the field?
- Is the field optional or required?
    - If the field is optional, what happens if I don't set it?
    
To get you started, I would expect the GoDoc for this field to read something like:
```go
// type is a required field used to specify the type of the bpf program.
// The type dictates where in the kernel the program is executed.
// The Linux kernel may restrict or allow certain functionality depending on the program type.
// 
// Allowed values are XDP, TC, TCX, Uprobe, and UretProbe.
//
// When set to XDP, the program can attach to network devices and will be called on every incoming packet
// received by the network device. When using the XDP program type, the xdp field is required.
//
// When set to TC ....
// ...
```

In general, these GoDoc guidelines apply to _every_ field.

bpfman-operator/apis/v1alpha1/bpf_application_state_types.go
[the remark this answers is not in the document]
everettraven: Why are you using an entire CRD to communicate state of resources instead of the status subresource of your existing CRD types?

bpfman-operator/apis/v1alpha1/bpf_application_types.go
```diff
@@ -0,0 +1,101 @@
...
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'TC' ?  has(self.tcInfo) : !has(self.tcInfo)",message="tcInfo configuration is required when type is tc, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'TCX' ?  has(self.tcxInfo) : !has(self.tcxInfo)",message="tcxInfo configuration is required when type is tcx, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'Uprobe' ?  has(self.uprobeInfo) : !has(self.uprobeInfo)",message="uprobeInfo configuration is required when type is uprobe, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'UretProbe' ?  has(self.uretprobeInfo) : !has(self.upretrobeInfo)",message="uretprobe configuration is required when type is uretprobe, and forbidden otherwise"
+type BpfApplicationProgram struct {
+	// name is the name of the function that is the entry point for the BPF
+	// program
+	Name string `json:"name"`
+
+	// type specifies the bpf program type
```
everettraven, line 35: My prior comment on the `BpfApplicationProgramState` `type` field applies here.

bpfman-operator/apis/v1alpha1/bpf_application_types.go
```diff
@@ -0,0 +1,101 @@
...
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'TC' ?  has(self.tcInfo) : !has(self.tcInfo)",message="tcInfo configuration is required when type is tc, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'TCX' ?  has(self.tcxInfo) : !has(self.tcxInfo)",message="tcxInfo configuration is required when type is tcx, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'Uprobe' ?  has(self.uprobeInfo) : !has(self.uprobeInfo)",message="uprobeInfo configuration is required when type is uprobe, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'UretProbe' ?  has(self.uretprobeInfo) : !has(self.upretrobeInfo)",message="uretprobe configuration is required when type is uretprobe, and forbidden otherwise"
+type BpfApplicationProgram struct {
+	// name is the name of the function that is the entry point for the BPF
+	// program
+	Name string `json:"name"`
```
everettraven, line 33: Seems like this field is required. Should have the `+required` marker.

Are there any constraints for the name? Can it really be any arbitrary string value? Are there any validations you can put in place to reject invalid values at admission time?

bpfman-operator/apis/v1alpha1/bpf_application_types.go
```diff
@@ -0,0 +1,101 @@
...
+
+// BpfApplicationSpec defines the desired state of BpfApplication
+type BpfApplicationSpec struct {
+	BpfAppCommon `json:",inline"`
+
+	// programs is the list of bpf programs in the BpfApplication that should be
+	// loaded. The application can selectively choose which program(s) to run
+	// from this list based on the optional attach points provided.
+	// +kubebuilder:validation:MinItems:=1
+	Programs []BpfApplicationProgram `json:"programs,omitempty"`
```
everettraven, line 75: Is this meant to be an optional field? What happens if I create a `BpfApplication` and specify no programs?

bpfman-operator/apis/v1alpha1/bpf_application_types.go
```diff
@@ -0,0 +1,101 @@
...
+// BpfApplication is the Schema for the bpfapplications API
+// +kubebuilder:printcolumn:name="NodeSelector",type=string,JSONPath=`.spec.nodeselector`
+// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.conditions[0].reason`
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
+type BpfApplication struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   BpfApplicationSpec `json:"spec,omitempty"`
+	Status BpfAppStatus       `json:"status,omitempty"`
```
everettraven, line 92: Both these fields should have godoc and either a `+required` or `+optional` marker

bpfman-operator/apis/v1alpha1/bpf_application_types.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// +kubebuilder:object:root=true
+// BpfApplicationList contains a list of BpfApplications
+type BpfApplicationList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []BpfApplication `json:"items"`
```
everettraven, line 100: Field should have godoc and either a `+required` or `+optional` marker

bpfman-operator/apis/v1alpha1/cluster_bpf_application_types.go
```diff
@@ -0,0 +1,166 @@
...
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'KretProbe' ?  has(self.kretprobeInfo) : !has(self.kretprobeInfo)",message="kretprobeInfo configuration is required when type is kretprobe, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'Uprobe' ?  has(self.uprobeInfo) : !has(self.uprobeInfo)",message="uprobeInfo configuration is required when type is uprobe, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'UretProbe' ?  has(self.uretprobeInfo) : !has(self.uretprobeInfo)",message="uretprobeInfo configuration is required when type is uretprobe, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'TracePoint' ?  has(self.tracepointInfo) : !has(self.tracepointInfo)",message="tracepointInfo configuration is required when type is tracepoint, and forbidden otherwise"
+type ClBpfApplicationProgram struct {
+	// name is the name of the function that is the entry point for the BPF
+	// program
+	Name string `json:"name"`
```
everettraven, line 73: Same question as before about name on BpfApplicationProgram

bpfman-operator/apis/v1alpha1/cluster_fentry_program_types.go
```diff
@@ -0,0 +1,56 @@
...
+	// +kubebuilder:validation:MaxItems=1
+	// +kubebuilder:default:={}
+	Links []ClFentryAttachInfo `json:"links"`
+}
+
+// ClFentryLoadInfo contains the program-specific load information for Fentry
+// programs
+type ClFentryLoadInfo struct {
+	// function is the name of the function to attach the Fentry program to.
+	Function string `json:"function"`
```
everettraven, line 35: Can this be any arbitrary string value? Any validation you can put in place?

bpfman-operator/apis/v1alpha1/cluster_fentry_program_types.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// ClFentryAttachInfo indicates that the Fentry program should be attached to
+// the function identified in ClFentryLoadInfo. The only valid value for Attach
+// is true.
+type ClFentryAttachInfo struct {
+	// +kubebuilder:validation:Enum=true
+	Attach bool `json:"attach"`
```
everettraven, line 43: Don't use a boolean. Use an enum on a string type.

bpfman-operator/apis/v1alpha1/cluster_fentry_program_types.go
msherif1234: not really it is an ebpf program name so it can be any string

everettraven: so something like `!@ASjdhflaksjdfy@&!^*#@&^!` would be valid?

bpfman-operator/apis/v1alpha1/bpf_application_types.go
everettraven: Generally for lists we recommend having an upper bound to the number of entries. Is there a reasonable upper bound that can be specified here?

Additionally, what other constraints exist here? Can I have multiple program entries with the same `.name` value?

bpfman-operator/apis/v1alpha1/cluster_bpf_application_types.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// ClBpfApplicationSpec defines the desired state of BpfApplication
+type ClBpfApplicationSpec struct {
+	BpfAppCommon `json:",inline"`
+	// programs is the list of bpf programs in the BpfApplication that should be
+	// loaded. The application can selectively choose which program(s) to run
+	// from this list based on the optional attach points provided.
+	// +kubebuilder:validation:MinItems:=1
+	Programs []ClBpfApplicationProgram `json:"programs,omitempty"`
```
everettraven, line 139: Same questions as above here

bpfman-operator/apis/v1alpha1/cluster_fentry_program_types.go
```diff
@@ -0,0 +1,63 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// All fields are required unless explicitly marked optional
+// +kubebuilder:validation:Required
+package v1alpha1
```
everettraven, line 19: It is not recommended to have package level validations like this. It can lead to confusion and inflexibility. For example, you cannot add a new required field to a CRD without it being a breaking change. You should be explicit for each field.

bpfman-operator/apis/v1alpha1/cluster_fentry_program_types.go
everettraven: Are there any conventions, standards, or validations that exist _somewhere_ that enforces a particular pattern for the function names?

everettraven: If not a particular pattern, is there any reasonable upper bound for the length of the function name?

bpfman-operator/apis/v1alpha1/cluster_fentry_program_types.go
```diff
@@ -0,0 +1,63 @@
...
+
+// ClFentryProgramInfo defines the Fentry program details
+type ClFentryProgramInfo struct {
+	ClFentryLoadInfo `json:",inline"`
+	// Whether the program should be attached to the function.
+	// +optional
+	// +kubebuilder:validation:MaxItems=1
+	// +kubebuilder:default:={}
+	Links []ClFentryAttachInfo `json:"links"`
```
everettraven, line 28: Does this need to be a list? Will it ever be possible to specify multiple "links" for a `FentryProgram`? If I don't specify it, what is the default behavior?

bpfman-operator/apis/v1alpha1/cluster_fentry_program_types.go
```diff
@@ -0,0 +1,63 @@
...
+type ClFentryLoadInfo struct {
+	// function is the name of the function to attach the Fentry program to.
+	Function string `json:"function"`
+}
+
+type AttachType string
+
+const (
+	Attach  AttachType = "true"
+	Dettach AttachType = "false"
```
everettraven, line 42: A couple things here:
- We typically recommend a name format of `{alias}{constValue}`. So `Attach` would be `AttachTypeAttach`.
- These constant values don't match the values that a user is allowed to specify where this is used. A user can only input the string literals `Attach` and `Detach`, not `true` and `false`. You should have these values match what users can input so you can accurately do comparisons in code with these constants.

bpfman-operator/apis/v1alpha1/cluster_fentry_program_types.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+// ClFentryAttachInfo indicates that the Fentry program should be attached to
+// the function identified in ClFentryLoadInfo. The only valid value for Attach
+// is true.
+type ClFentryAttachInfo struct {
+	// +kubebuilder:validation:Enum==Attach;Dettach;
```
everettraven, line 49:
```suggestion
	// +kubebuilder:validation:Enum=Attach;Dettach;
```

bpfman-operator/apis/v1alpha1/cluster_fentry_program_types.go
```diff
@@ -0,0 +1,63 @@
...
+	Attach AttachType `json:"attach"`
```
everettraven, line 50: So from a user perspective the YAML for this field would look something like:
```yaml
attach: Attach | Detach
```

Is there a better name that can be used for this field? This is already in an `*AttachInfo` type so maybe something like `AttachmentMode` or just `Mode` is reasonable here?

Are there any configuration options that may be added in the future depending on the attachment mode?

bpfman-operator/apis/v1alpha1/cluster_bpf_application_types.go
```diff
@@ -0,0 +1,166 @@
...
+
+	// kprobeInfo defines the desired state of the application's KprobePrograms.
+	// +unionMember
+	// +optional
+	KprobeInfo *ClKprobeProgramInfo `json:"kprobeInfo,omitempty"`
+
+	// kretprobeInfo defines the desired state of the application's KretprobePrograms.
+	// +unionMember
+	// +optional
+	KretprobeInfo *ClKprobeProgramInfo `json:"kretprobeInfo,omitempty"`
```
everettraven, line 114: It looked like `ClKprobeProgramInfo` had some restrictions for `KretProbe` program types that may not apply to the `Kprobe` program type? Maybe this should use a separate type so that you can have distinct requirements for each type?

bpfman-operator/apis/v1alpha1/cluster_kprobe_program_types.go
```diff
@@ -0,0 +1,62 @@
...
+	Links []ClKprobeAttachInfo `json:"links"`
+}
+
+type ClKprobeAttachInfo struct {
+	// function to attach the kprobe to.
+	Function string `json:"function"`
+
+	// offset added to the address of the function for kprobe.
+	// The offset must be zero for kretprobes.
+	// TODO: Add a webhook to enforce kretprobe offset=0.
```
everettraven, line 36: Maybe this should be program type should use a distinct type? Maybe it wouldn't even make sense to have an `offset` field for this program type if it must be `0`?

bpfman-operator/apis/v1alpha1/cluster_tc_program_types.go
```diff
@@ -0,0 +1,100 @@
...
+// +kubebuilder:validation:Enum=unspec;ok;reclassify;shot;pipe;stolen;queued;repeat;redirect;trap;dispatcher_return
+type TcProceedOnValue string
+
+// ClTcProgramInfo defines the tc program details
+type ClTcProgramInfo struct {
+	// The list of points to which the program should be attached.  The list items
+	// are optional and may be udated after the bpf program has been loaded
+	// +optional
+	// +kubebuilder:default:={}
+	Links []ClTcAttachInfo `json:"links"`
```
everettraven, line 30: Same questions as on other program type links.

bpfman-operator/apis/v1alpha1/shared_types.go
```diff
@@ -0,0 +1,383 @@
...
+	Namespace string `json:"namespace"`
+
+	// pods indicate the target pods. This field must be specified, to select all pods use
+	// standard metav1.LabelSelector semantics and make it empty.
+	Pods metav1.LabelSelector `json:"pods"`
+
+	// containerNames indicate the Name(s) of container(s).  If none are specified, all containers in the
+	// pod are selected.
+	// +optional
+	ContainerNames *[]string `json:"containerNames,omitempty"`
```
everettraven, line 53: Any particular reason for this to be `*[]string` over `[]string`?

Also, is there any validation you can put in place on container names? I'm assuming Kubernetes container names fall under one of the requirements specified in https://kubernetes.io/docs/concepts/overview/working-with-objects/names

bpfman-operator/apis/v1alpha1/shared_types.go
```diff
@@ -0,0 +1,383 @@
...
+	// PullNever means that bpfman never pulls an image, but only uses a local image. Container will fail if the image isn't present
+	PullNever PullPolicy = "Never"
+	// PullIfNotPresent means that bpfman pulls if the image isn't present on disk. Container will fail if the image isn't present and the pull fails.
+	PullIfNotPresent PullPolicy = "IfNotPresent"
+)
+
+// ByteCodeSelector defines the various ways to reference bpf bytecode objects.
+type ByteCodeSelector struct {
+	// image used to specify a bytecode container image.
+	Image *ByteCodeImage `json:"image,omitempty"`
```
everettraven, line 152: The field definition here maps to the conventions for an optional field, but it is overridden by the package level required. Should this be optional or required?

bpfman-operator/apis/v1alpha1/shared_types.go
```diff
@@ -0,0 +1,383 @@
...
+
+	// path is used to specify a bytecode object via filepath.
+	Path *string `json:"path,omitempty"`
```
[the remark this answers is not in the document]
everettraven: Any validations you can put in place to ensure the user inputs a valid filepath at admission time?

bpfman-operator/apis/v1alpha1/shared_types.go
```diff
@@ -0,0 +1,383 @@
...
+	ImagePullSecret *ImagePullSecretSelector `json:"imagePullSecret,omitempty"`
+}
+
+// ImagePullSecretSelector defines the name and namespace of an image pull secret.
+type ImagePullSecretSelector struct {
+	// name of the secret which contains the credentials to access the image repository.
+	Name string `json:"name"`
+
+	// namespace of the secret which contains the credentials to access the image repository.
+	Namespace string `json:"namespace"`
```
everettraven, line 180: See https://kubernetes.io/docs/concepts/overview/working-with-objects/names for Kubernetes object name validations you can put in place to ensure they are a valid format at admission time

bpfman-operator/apis/v1alpha1/cluster_tc_program_types.go
```diff
@@ -0,0 +1,100 @@
...
+	// containers identifies the set of containers in which to attach the eBPF
+	// program. If Containers is not specified, the BPF program will be attached
+	// in the root network namespace.
+	// +optional
+	Containers *ClContainerSelector `json:"containers"`
+
+	// direction specifies the direction of traffic the tc program should
+	// attach to for a given network device.
+	// +kubebuilder:validation:Enum=ingress;egress
+	Direction string `json:"direction"`
```
everettraven, line 46: For an enum, this should use a type alias and have constants.

bpfman-operator/apis/v1alpha1/cluster_tc_program_types.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// priority specifies the priority of the tc program in relation to
+	// other programs of the same type with the same attach point. It is a value
+	// from 0 to 1000 where lower values have higher precedence.
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=1000
+	Priority int32 `json:"priority"`
```
everettraven, line 53: Does this priority need to be unique across multiple "links"?

bpfman-operator/apis/v1alpha1/cluster_tc_program_types.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// proceedOn allows the user to call other tc programs in chain on this exit code.
+	// Multiple values are supported by repeating the parameter.
+	// +optional
+	// +kubebuilder:validation:MaxItems=11
```
everettraven, line 58: Why 11?

bpfman-operator/apis/v1alpha1/cluster_tc_program_types.go
```diff
@@ -0,0 +1,100 @@
...
+	// +optional
+	// +kubebuilder:default:={}
+	Links []ClTcAttachInfoState `json:"links"`
+}
+
+type ClTcAttachInfoState struct {
+	AttachInfoStateCommon `json:",inline"`
+
+	// ifname is the Interface name to attach the tc program to.
+	IfName string `json:"ifName"`
```
everettraven, line 78: InterfaceName is a more descriptive name.

bpfman-operator/apis/v1alpha1/cluster_tc_program_types.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// Optional container pid to attach the tc program in.
+	// +optional
+	ContainerPid *int32 `json:"containerPid"`
```
everettraven, line 82: What does it mean if there is no container pid specified?

bpfman-operator/apis/v1alpha1/cluster_tracepoint_program_types.go
```diff
@@ -0,0 +1,52 @@
...
+	// are optional and may be updated after the bpf program has been loaded
+	// +optional
+	// +kubebuilder:default:={}
+	Links []ClTracepointAttachInfo `json:"links"`
+}
+
+type ClTracepointAttachInfo struct {
+	// name refers to the name of a kernel tracepoint to attach the
+	// bpf program to.
+	Name string `json:"name"`
```
everettraven, line 33: Any validations that can be put in place to ensure only valid kernel tracepoint names are specified?

bpfman-operator/apis/v1alpha1/cluster_uprobe_program_types.go
```diff
@@ -0,0 +1,89 @@
...
+
+type ClUprobeAttachInfo struct {
+	// function to attach the uprobe to.
+	// +optional
+	Function string `json:"function"`
+
+	// offset added to the address of the function for uprobe.
+	// +optional
+	// +kubebuilder:default:=0
+	Offset uint64 `json:"offset"`
```
everettraven, line 38: is there a reasonable upper bound you can set here?

bpfman-operator/apis/v1alpha1/cluster_uprobe_program_types.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// target is the Library name or the absolute path to a binary or library.
+	Target string `json:"target"`
```
everettraven, line 41: Any validations you can put in place to ensure a valid value is provided by a user at admission time?

bpfman-operator/apis/v1alpha1/cluster_xdp_program_types.go
```diff
@@ -0,0 +1,90 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// All fields are required unless explicitly marked optional
+// +kubebuilder:validation:Required
+package v1alpha1
+
+// +kubebuilder:validation:Enum=aborted;drop;pass;tx;redirect;dispatcher_return
```
everettraven, line 21: PascalCase. Constants

bpfman-operator/apis/v1alpha1/cluster_bpf_application_types.go
msherif1234: u mean `offset` value of `0` in case of kretprobe ? we were thinking of just using validation wh to ensure this or maybe we can do that with cel rule but other than that its the same?

bpfman-operator/apis/v1alpha1/cluster_fentry_program_types.go
dave-tucker: Rules for C identifers apply here. So must be `^[a-zA-Z][a-zA-Z0-9_]+`. 64 is a reasonable upperbound.

bpfman-operator/apis/v1alpha1/bpf_application_types.go
dave-tucker: Validation can be the same as https://github.com/openshift/api/pull/2221#discussion_r1991774122

bpfman-operator/apis/v1alpha1/cluster_bpf_application_types.go
everettraven: >u mean offset value of 0 in case of kretprobe ?

Yes

>we were thinking of just using validation wh to ensure this or maybe we can do that with cel rule but other than that its the same?

For now. Is it possible that the `kretprobe` case ends up requiring additional drift from `kprobe` in the future? If you go with using the same type, every change you make for one program type will inherently be added to the other. It is _usually_ better practice to have each one have their own distinct type so you can update each one individually as needed.

bpfman-operator/apis/v1alpha1/cluster_kprobe_program_types.go
everettraven: I think it serves the purpose better to have a separate type. Using a shared type ends up as more of a maintenance burden if the two implementations drift further apart in the future.

bpfman-operator/apis/v1alpha1/cluster_kprobe_program_types.go
```diff
@@ -0,0 +1,62 @@
...
+// +kubebuilder:validation:Required
+package v1alpha1
+
+// ClKprobeProgramInfo contains the information for the kprobe program
+type ClKprobeProgramInfo struct {
+	// The list of points to which the program should be attached.  The list items
+	// are optional and may be udated after the bpf program has been loaded
+	// +optional
+	// +kubebuilder:default:={}
+	Links []ClKprobeAttachInfo `json:"links"`
```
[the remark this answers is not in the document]
everettraven: Is there any reason to specify this program type without providing at least one link? Re: upper bounds, even if it depends on the kernel is there a reasonable value you can specify here? Would it make sense for a user to be able to specify 10,000 links?

bpfman-operator/apis/v1alpha1/cluster_tc_program_types.go
```diff
@@ -0,0 +1,100 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// All fields are required unless explicitly marked optional
+// +kubebuilder:validation:Required
+package v1alpha1
+
+// +kubebuilder:validation:Enum=unspec;ok;reclassify;shot;pipe;stolen;queued;repeat;redirect;trap;dispatcher_return
```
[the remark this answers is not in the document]
msherif1234: those enum are passed through to bpfmgr process as is, converting to PascalCase will require additional logic can this is an exception ?

bpfman-operator/apis/v1alpha1/shared_types.go
msherif1234: it can be specified via image or path ?

bpfman-operator/apis/v1alpha1/cluster_kprobe_program_types.go
msherif1234: done

bpfman-operator/apis/v1alpha1/bpf_application_state_types.go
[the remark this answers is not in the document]
msherif1234: done

bpfman-operator/apis/v1alpha1/cluster_bpf_application_types.go
msherif1234: create kretprobe type

bpfman-operator/apis/v1alpha1/bpf_application_state_types.go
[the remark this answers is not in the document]
dave-tucker: Just to expand on that a little...

1. A single BpfApplication object can create a lot of state! For example, it might load N eBPF probe(s) into every pod on every node.
1. When reconciling a BpfApplicationCRD things get complicated since there is one controller running on every node. Each controller is watching for BpfApplicationCRD changes. As every instance of the controller completes its reconciliation it would write back state to that BpfApplicationCRD, which then triggers another update... creating a bit of an "update storm".
1. Aside from ☝️ it wasn't clear how to cleanly layout the State within an `BpfApplicationCRD`. Could it be `map[string]State` where the key is the k8s node name where the agent is running?

aside: I would be very happy to not have *StateCRDs

bpfman-operator/apis/v1alpha1/cluster_xdp_program_types.go
```diff
@@ -0,0 +1,90 @@
...
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=1000
+	Priority int32 `json:"priority"`
+
+	// proceedOn allows the user to call other xdp programs in chain on this exit code.
+	// Multiple values are supported by repeating the parameter.
+	// +optional
+	// +kubebuilder:validation:MaxItems=6
+	// +kubebuilder:default:={pass,dispatcher_return}
+	ProceedOn []XdpProceedOnValue `json:"proceedOn"`
```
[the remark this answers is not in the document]
msherif1234: it turns out I can't use enum validation because it can multiple values enum validation will fail
```sh
Unsupported value: []interface {}{"pass", "dispatcher_return"}: supported values: "aborted", "drop", "pass", "tx", "redirect", "dispatcher_return",
```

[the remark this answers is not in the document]
JoelSpeed: You've added the marker in the wrong place if you got this error

Add the enum marker to the `XdpProceedOnValue` type directly

bpfman-operator/apis/v1alpha1/cluster_tc_program_types.go
msherif1234: This enum also passed through so it has to stay as "ingress" and "egress" can't change to "Ingress", "Egress", I will ad enum type

bpfman-operator/apis/v1alpha1/cluster_tc_program_types.go
msherif1234: no the under layer infra can handle links with the same pri

bpfman-operator/apis/v1alpha1/cluster_tc_program_types.go
JoelSpeed: Your API doesn't have to be identical to an upstream API, just convertible. 

There is nothing to stop you mapping Egress and Ingress to their upstream counterparts when you convert the configuration over.

bpfman-operator/apis/v1alpha1/cluster_bpf_application_types.go
msherif1234: added validation check to program name

bpfman-operator/apis/v1alpha1/shared_types.go
```diff
@@ -0,0 +1,383 @@
...
+// +kubebuilder:validation:MinProperties=1
+type InterfaceSelector struct {
+	// interfaces refers to a list of network interfaces to attach the BPF
+	// program to.
+	// +optional
+	Interfaces *[]string `json:"interfaces,omitempty"`
+
+	// primaryNodeInterface to attach BPF program to the primary interface on the node. Only 'true' accepted.
+	// +optional
+	PrimaryNodeInterface *bool `json:"primaryNodeInterface,omitempty"`
```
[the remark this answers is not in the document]
msherif1234: there could be secondary interface ?

bpfman-operator/apis/v1alpha1/cluster_tc_program_types.go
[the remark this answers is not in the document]
msherif1234: fixed

bpfman-operator/apis/v1alpha1/cluster_tc_program_types.go
msherif1234: I fixed it and did the mapping internally

bpfman-operator/apis/v1alpha1/cluster_xdp_program_types.go
msherif1234: done

bpfman-operator/apis/v1alpha1/cluster_xdp_program_types.go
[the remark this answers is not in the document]
msherif1234: yes fixed and all is good now thank you!!

bpfman-operator/apis/v1alpha1/cluster_tc_program_types.go
```diff
@@ -0,0 +1,100 @@
...
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=1000
+	Priority int32 `json:"priority"`
+
+	// proceedOn allows the user to call other tc programs in chain on this exit code.
+	// Multiple values are supported by repeating the parameter.
+	// +optional
+	// +kubebuilder:validation:MaxItems=11
+	// +kubebuilder:default:={pipe,dispatcher_return}
+	ProceedOn []TcProceedOnValue `json:"proceedOn"`
```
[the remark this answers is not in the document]
msherif1234: I think I can use `+kubebuilder:validation:UniqueItems:=true` to gurantee that will add it

bpfman-operator/apis/v1alpha1/bpf_application_state_types.go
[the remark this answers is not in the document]
everettraven: >A single BpfApplication object can create a lot of state! For example, it might load N eBPF probe(s) into every pod on every node.

Is it necessary for a user to have access to all this state? will they care? If you are mostly replicating the declarative configuration they have specified is it sufficient enough to just convey that you have successfully achieved their desired state?

What is the set of information that your users _need_ to know?

>When reconciling a BpfApplicationCRD things get complicated since there is one controller running on every node. Each controller is watching for BpfApplicationCRD changes. As every instance of the controller completes its reconciliation it would write back state to that BpfApplicationCRD, which then triggers another update... creating a bit of an "update storm"

Is it a technical requirement that you have a controller running on every node?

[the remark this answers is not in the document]
everettraven: >Aside from ☝️ it wasn't clear how to cleanly layout the State within an BpfApplicationCRD. Could it be map[string]State where the key is the k8s node name where the agent is running?

That is one way. You could use server-side apply as well to allow for multiple field managers. 

Not saying this is better, but another way I could think of is that you have the `BpfApplication` CRD be a higher level configuration option for users to specify the configuration across multiple nodes. The `BpfApplication` controller could then create `NodeBpfApplication` resources that are managed by the controllers running on the nodes. The `NodeBpfApplication` would specify a single node so that the controllers running on each node can identify which ones they are responsible for reconciling.

bpfman-operator/apis/v1alpha1/cluster_tc_program_types.go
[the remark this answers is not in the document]
msherif1234: tried different options as discussed over slack with no luck
https://redhat-internal.slack.com/archives/CE4L0F143/p1741298311519169

bpfman-operator/apis/v1alpha1/shared_types.go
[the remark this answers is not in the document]
msherif1234: but we aren't planning by default to findout non primary intf so its use is either primary or not hence the use of boolean, just for my education what is wrong with using booleans ?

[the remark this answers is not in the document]
everettraven: Generally we advise against using booleans because things may start out as boolean in nature but grow over time. For example, you aren't planning to do anything with non-primary interfaces _for now_. I imagine if a valid use case arises in the future to attach to a "secondary" interface you would want to be able to enable support for that?

Using a boolean backs you into a corner where you need to add yet another field to support that. Using an enum would allow you to add a new allowed enum value instead of adding another boolean field.

bpfman-operator/apis/v1alpha1/bpf_application_types.go
```diff
@@ -0,0 +1,104 @@
...
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'TC' ?  has(self.tc) : !has(self.tc)",message="tc configuration is required when type is tc, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'TCX' ?  has(self.tcx) : !has(self.tcx)",message="tcx configuration is required when type is TCX, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'UProbe' ?  has(self.uprobe) : !has(self.uprobe)",message="uprobe configuration is required when type is uprobe, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'URetProbe' ?  has(self.uretprobe) : !has(self.uretprobe)",message="uretprobe configuration is required when type is uretprobe, and forbidden otherwise"
+type BpfApplicationProgram struct {
+	// name is the name of the function that is the entry point for the BPF
+	// program
+	// +kubebuilder:validation:Pattern="^[a-zA-Z][a-zA-Z0-9_]+."
```
everettraven, line 33: General advice is to use CEL expressions to enforce regex matching now so that you can provide a more user-friendly message than returning a regex pattern to them on failed validations.

bpfman-operator/apis/v1alpha1/cluster_bpf_application_state_types.go
```diff
@@ -0,0 +1,167 @@
...
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'UProbe' ?  has(self.uprobe) : !has(self.uprobe)",message="uprobe configuration is required when type is uprobe, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'URetProbe' ?  has(self.uretprobe) : !has(self.uretprobe)",message="uretprobe configuration is required when type is uretprobe, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'TracePoint' ?  has(self.tracepoint) : !has(self.tracepoint)",message="tracepoint configuration is required when type is tracepoint, and forbidden otherwise"
+type ClBpfApplicationProgramState struct {
+	BpfProgramStateCommon `json:",inline"`
+	// type specifies the bpf program type
+	// +unionDiscriminator
+	// +required
+	// +kubebuilder:validation:Enum:="XDP";"TC";"TCX";"FEntry";"FExit";"KProbe";"KRetProbe";"UProbe";"URetProbe";"TracePoint"
+
```
everettraven, line 43:
```suggestion
```

bpfman-operator/apis/v1alpha1/cluster_tc_program_types.go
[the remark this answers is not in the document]
everettraven: I think the error you are seeing related to default may be because it is thinking you mean to literally set the string `"Pipe,DispatcherReturn"` instead of a slice containing 2 entries.

I think the default tag you are looking for is `+kubebuilder:default:=[Pipe,DispatcherReturn]`

bpfman-operator/apis/v1alpha1/shared_types.go
everettraven: So you want an option to use image _or_ path? If that is the case, that is not what your API currently states. It states that you must set _both_ image _and_ path.

bpfman-operator/apis/v1alpha1/bpf_application_types.go
msherif1234: can u pls share more info on how to do so with cel ?

everettraven: Something like https://github.com/openshift/api/blob/8a7efbfb531684ef0ff566a882096df03e0f499c/insights/v1alpha1/types_insights.go#L322

bpfman-operator/apis/v1alpha1/cluster_tc_program_types.go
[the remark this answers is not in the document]
msherif1234: this is my default `// +kubebuilder:default:={Pipe,DispatcherReturn}`

bpfman-operator/apis/v1alpha1/bpf_application_state_types.go
Billy99: I get the spirit of the request and I'll make a pass beefing up the descriptions. But I don't think we can or should give a full eBPF primer in the descriptions. Is it worth providing links to further reading in the description?

bpfman-operator/apis/v1alpha1/shared_types.go
msherif1234: missing optional ? it doesn't complain atm when I pass just the Image though ?

everettraven: You have a package level required set so I'm surprised that is working as optional. Generally when you want to do a "one or the other" option you would use a discriminated union. Some docs/guidelines on a discriminated union: https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#discriminated-unions

everettraven: Also, remove the package level required.

bpfman-operator/apis/v1alpha1/cluster_fentry_program_types.go
anfredette: Fentry and Fexit programs can only ever have one link.  However, I made this a slice also to be consistent with the other program types that have lists of links.  In this case, you can have a max of one link in the list.

For all program types, if a link does not exist in the list, it should not be attached, and if it was attached previously, but is no longer on the list, it will be detached.  The same is true for Fentry/Fexit.

bpfman-operator/apis/v1alpha1/bpf_application_state_types.go
everettraven: I would accept links to further reading as long as there is a decent base description. I agree that you shouldn't give a full eBPF primer in the descriptions, but you _should_ give at least a basic description of what each allowed value maps to. Think about the familiarity of your user base and include relevant information to answer the questions I shared above.

bpfman-operator/apis/v1alpha1/bpf_application_state_types.go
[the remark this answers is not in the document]
everettraven: I was thinking about this some more. If you are going to continue with this direction of separate state CRDs, you probably should not include a spec. Instead, have an empty spec and only ever populate the status with information for a user.

This would help prevent confusion from the user side that there are configurable options on this CRD.

bpfman-operator/apis/v1alpha1/cluster_fentry_program_types.go
everettraven: Gotcha. This should be explained in the documentation on the field so that users understand this limitation as well.

I would also recommend not making this a slice if it can never have more than a single link to make it clear to a user instead of them thinking that they can specify multiple due to it being a yaml list.

For example, when a user sees:
```yaml
links:
  - ...
```
they may think they can add more values here and then encounter an admission failure.

Instead, if they saw:
```yaml
link:
  field: value
```
it would be clear that this type only allows a single link. No workflow happens where a user attempts to add another link and encounters an admission failure.

Ultimately it is up to you if you would like to take this suggestion or not.

bpfman-operator/apis/v1alpha1/bpf_application_state_types.go
Billy99: I found the answer: `oc explain ...`
I also found [dev-guide/api-conventions.md](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md) and I'm reading through it.

Your example and the examples in the api-conventions.md have each new sentence on a different line and long sentences wrapping. In my terminal, the output seems to wrap at 80 characters and new lines in the comments are honored (see below). Is that just a setting in my terminal (looked through my terminal settings and don't see it) so making sure we don't need to manage line spacing in the GoDoc comments.

```
$ kubectl explain clusterbpfapplication.spec.programs
GROUP:      bpfman.io
KIND:       ClusterBpfApplication
VERSION:    v1alpha1

FIELD: programs <[]Object>


DESCRIPTION:
    programs is the list of eBPF programs in the ClusterBpfApplication that
    should be
    loaded in kernel memory.
    eBPF programs in this list will be loaded on the system based the
    nodeSelector.
    Even if eBPF program is loaded in kernel memory, it cannot be triggered
    until an
    attachment point is provided.
    :
```

everettraven: Generally we should try to respect the 80 characters, but this isn't something we have strictly enforced AFAIK. If you can put each sentence on a new line that is generally preferred. My example was just a general example, I wasn't expecting it to be taken literally.

bpfman-operator/apis/v1alpha1/shared_types.go
msherif1234: ok

bpfman-operator/apis/v1alpha1/shared_types.go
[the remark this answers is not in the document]
msherif1234: I added regex to validate the path

bpfman-operator/apis/v1alpha1/shared_types.go
```diff
@@ -0,0 +1,383 @@
...
+	Image *ByteCodeImage `json:"image,omitempty"`
+
+	// path is used to specify a bytecode object via filepath.
+	Path *string `json:"path,omitempty"`
+}
+
+// ByteCodeImage defines how to specify a bytecode container image.
+type ByteCodeImage struct {
+	// url is a valid container image URL used to reference a remote bytecode image.
+	Url string `json:"url"`
```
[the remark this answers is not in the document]
msherif1234: added url validation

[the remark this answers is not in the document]
everettraven: I don't think `isURL()` should be requiring the scheme to be `https`, but are you actually trying to validate a URL here or an image reference? 

A URL and an image reference are not the same thing.

bpfman-operator/apis/v1alpha1/shared_types.go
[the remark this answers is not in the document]
everettraven: Use a CEL expression (`kubebuilder:validation:XValidation`) instead of a raw regex. The CEL expression based validation allows you to return a more user friendly message on failed validation instead of just giving them a regex string.

When using a CEL expression, you should put a maximum length on the field as well to not exceed cost budgets.

bpfman-operator/apis/v1alpha1/shared_types.go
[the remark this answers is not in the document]
msherif1234: i did fail when I passed the image path from quay.io it only worked when I added https://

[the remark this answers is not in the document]
everettraven: You're right, I misread the underlying `net/url` method used by `isURL()`. That being said, do you want this to actually be a URL or an OCI image reference?

bpfman-operator/apis/v1alpha1/bpf_application_state_types.go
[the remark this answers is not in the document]
anfredette: > If you are mostly replicating the declarative configuration they have specified...
Actually, we are not replicating the declarative config.  The user specifies the high-level intent in the BpfApplication CRD, and then the bpfman agent on each node applies that intent to the given node.  For example, we have an option for the user to indicate that a program should be installed on the primary node interface.  The agent running on the node figures out what that interface is, and reflects the name (e.g., eth0) in the state CRD.  Or if there's a container selector, the agent finds the containers that match.  Also, we need to maintain the program IDs and link IDs so that we can unload/remove them when needed.  All these items are likely different on each node.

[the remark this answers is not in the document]
anfredette: Regarding your comment about putting the per node info in the status field, we just want to confirm.  Are recommending that we move all of the info currently in the Spec section to the Status section?  Are there any downsides to maintaining a large amount of info in the Status section?

bpfman-operator/apis/v1alpha1/bpf_application_types.go
anfredette: The purpose of a BpfApplication is to ask bpfman to load and possibly attach eBpf programs.  I don't see a use case for a BpfApplication CRD with no programs identified.  Regarding the bound, I don't know what an acceptable bound would be.  The same goes for links.

bpfman-operator/apis/v1alpha1/cluster_fentry_program_types.go
anfredette: I initially implemented it the way you suggested.  However, when we looked at it holistically, we thought it made more sense for the Fentry/Fexit "links" fields to be a slice just like all the other programs even though in these two cases, it can only hold one link.  As a side effect, it helps make the controller code more common between program types.  We'd prefer to keep it this way, but we can improve the comments to make it more clear.

bpfman-operator/apis/v1alpha1/cluster_uprobe_program_types.go
anfredette: The combination of target, function, and offset is used to define the location of an instruction at which the user wants to invoke the uProbe program.  So, the max offset would be the size of the largest allowable executable or library.  The theoretical limit would depend on a number of factors including things like file format, OS, and file system.  In any case, it would be pretty big, and I don't think it would be very helpful because in most cases it would need to be much smaller in practice.  If the user provides an offset that is out of bounds for the given target, the eBPF verifier will return an error.

bpfman-operator/apis/v1alpha1/bpf_application_state_types.go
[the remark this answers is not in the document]
everettraven: Yes, I would recommend that anything that is capturing state to convey to users goes into the status subresource. It would help prevent users from thinking that the resource is something they can modify. Generally users know not to mess with the status, but it is common practice that the spec is a user defined desired state.

AFAIK, there is no difference in maintaining the info in the status vs the spec. If it works for you today to store things in the spec, there is no reason I can think of as to why it wouldn't work just as well moved to the status.

bpfman-operator/apis/v1alpha1/bpf_application_types.go
everettraven: In that case, this should probably be a required field. For an upper bound, how many programs would you generally expect a user to specify? What number would surprise you as being a really high number?

bpfman-operator/apis/v1alpha1/cluster_kprobe_program_types.go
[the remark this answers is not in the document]
everettraven: I'm not familiar with eBPF, so I'm having trouble parsing how this relates to the questions I've asked in this thread.

Is there a number of links that you think would be fairly common for users to specify? What number would make you think that it is a _really_ high number of links?

Also, can I duplicate links or do they need to be unique?

bpfman-operator/apis/v1alpha1/cluster_uprobe_program_types.go
everettraven: How would that error manifest to users?

bpfman-operator/apis/v1alpha1/bpf_application_types.go
anfredette: Agreed, it should be required.  We'll fix that.

anfredette: I didn't want to arbitrarily limit it.  However, we'll try to come up with some upper bounds on what's reasonable.

bpfman-operator/apis/v1alpha1/cluster_kprobe_program_types.go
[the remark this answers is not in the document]
anfredette: My answer was in response to why it makes sense to load a program without any links - they may not be needed initially, but may added/modified later based on configuration or other events.

The links need to be unique.

As with programs, we'll come up with an upper bound.

bpfman-operator/apis/v1alpha1/cluster_uprobe_program_types.go
anfredette: 1. The BpfApplication CRD would have Status == "Error", and it would list which BpfApplicationState CRD was in error. 
2. The BpfApplicationState CRD in question would have Status == "Error"
3. The failed link would have LinkStatus == "AttachError" in the state CRD

bpfman-operator/apis/v1alpha1/bpf_application_types.go
everettraven: You can always increase it in the future if needed. You can never decrease it though.

If you _truly_ can't enforce a reasonable upper bound then that is fine, but usually there is a reasonable one that can be set

bpfman-operator/apis/v1alpha1/cluster_uprobe_program_types.go
everettraven: What is the impact to user workloads in this state?

anfredette: Other links that don't have errors will still be attached.  It's up to the user application to decide what to do when this happens.  Then, they may also need to figure out why what they are trying to do doesn't work.

eBPF is complicated, and the kernel isn't very helpful in telling you what's wrong when there's an error.  So, we will also log what we can about the error, but often it's just `EINVAL` and the eBPF application developer would need to use their knowledge of eBPF and the Linux kernel to figure out what's wrong.

everettraven: To clarify, by "workload" i'm meaning the non-ebpf applications that are running on a cluster via things like `Deployment`s

bpfman-operator/apis/v1alpha1/shared_types.go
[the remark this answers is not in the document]
msherif1234: oci image reference and I used regex to validate instead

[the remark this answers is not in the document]
everettraven: The pattern you put in place looks like it is correct for the tag of an OCI image reference.

I'd recommend getting a bit more detailed with what you do and do not allow here for image references. Use  https://github.com/operator-framework/operator-controller/blob/995dc2b34139ca62011a683a3ff42b1b38cf6636/api/v1/clustercatalog_types.go#L282-L330 as a reference for appropriately validating an OCI image reference. Specifically, this validates:
- Starts with a valid domain (`quay.io`, `my-custom.registry.io:8080`)
- Repository / Image name
- Tag
- Digests (`sha256`)

Feel free to filter this down to what you expect to support an be valid, but don't let this just be open ended when there are additional validations you can add to ensure your users are providing at least correctly shaped image reference strings

bpfman-operator/apis/v1alpha1/bpf_application_state_types.go
Billy99: @everettraven I have a bpfman-operator PR that made a pass through the ClusterBpfApplication CRD and updated all the field descriptions (see [PR#419](https://github.com/bpfman/bpfman-operator/pull/419). I also cut and pasted `kubectl explain` for each of the fields to help visualize the changes in a gist:
https://gist.github.com/Billy99/b871e60f04944d4b03c9c0106d2c8a43

I am getting ready to propagate my description changes to the other CRDs so hoping you could take a look and see if this what you had in mind before I propagate.

JoelSpeed (commented):
This API is huge!

I think we might want to tackle this in smaller pieces, and focus on one area, iterate on that, and apply the lessons learned then to other areas of the API rather than trying to review and correct the whole thing in one go.

I'd suggest that we also make sure to fix the linter issues. The linters enforce standards that help the API to become complete and have the validations that we would want to see for a comprehensive API.

The more you can self help on fixing the linter problems, the easier it becomes for us to tackle the trickier problems of API review.

For example, adding a `+optional` or `+required` as per your requirements to every field would then help us working out what does and doesn't serialize well at the moment which the current structures, and where we might need additional validations

anfredette:
@everettraven, Mohamed has left Red Hat, and we still have work to do on this PR.  Can someone give other members of the bpfman team permission to update this pr?  Thanks, Andre

JoelSpeed:
@anfredette Since the PR came from Mohamed's own repository, he is the only one who can give that access. You could either contact him and ask him to make you a collaborator on his fork of the repo, or you'll have to copy the branch and make a new PR

anfredette:
I’ll see if I can get in contact with Mohamed. Thanks, Andre

anfredette:
@msherif1234 can you add me and @frobware as collaborators on this pr?

msherif1234:
Done

JoelSpeed:
@msherif1234 @anfredette Is this PR still needed?

Billy99:
@frobware @andreaskaris - PTAL
Mohamed, Andre and Billy have moved off the project; Andy and Andreas are now managing.
Andy/Andreas have been talking about starting another round of reviews. My two cents is that this can be closed and a new PR started when they are ready, but they should chime in.

frobware:
> @frobware @andreaskaris - PTAL Mohamed, Andre and Billy have moved off the project; Andy and Andreas are now managing. Andy/Andreas have been talking about starting another round of reviews. My two cents is that this can be closed and a new PR started when they are ready, but they should chime in.

Given that I haven't been involved in this PR, unaware of the current state, I'd advocate we close and raise a new PR when we're ready for another review.

everettraven:
Closing based on input from @frobware and @Billy99 

/close

PR #2234, opened 2025-03-13
WIP: Add uid and extra claim mappings for external OIDC configuration

everettraven:
Adds:
- A new TechPreview feature-gate `ExternalOIDCWithUIDAndExtraMappings` (still needs to be linked to a proper EP for this. Maybe the existing OIDC one is sufficient?)
- Adds `uid` and `extra` fields to the `authentications.config.openshift.io` CRD's external OIDC provider configuration options, gated by the new `ExternalOIDCWithUIDAndExtraMappings` feature-gate

### DIFF
--- a/config/v1/types_authentication.go
+++ b/config/v1/types_authentication.go
@@ -262,6 +262,29 @@ type TokenClaimMappings struct {
 	// groups for the cluster identity.
 	// The referenced claim must use array of strings values.
 	Groups PrefixedClaimMapping `json:"groups,omitempty"`
+
+	// uid is an optional field for configuring the claim mapping
+	// used to construct the uid for the cluster identity.
+	//
+	// When using uid.claim to specify the claim it must be a single string value.
+	// When using uid.expression the expression must result in a single string value.
+	//
+	// When omitted, this means the user has no opinion and the platform
+	// is left to choose a default, which is subject to change over time.
+	// The current default is to use the 'sub' claim.
+	// +optional
+	// +openshift:enable:FeatureGate=ExternalOIDCWithUIDAndExtraMappings
+	UID UIDClaimMapping `json:"uid,omitempty"`
+
+	// extra is an optional field for configuring the mappings
+	// used to construct the extra attribute for the cluster identity.
+	// When omitted, no extra attributes will be present on the cluster identity.
+	// key values for extra mappings must be unique.
+	// +optional
+	// +openshift:enable:FeatureGate=ExternalOIDCWithUIDAndExtraMappings
+	// +listType=map
+	// +listMapKey=key
+	Extra []ExtraMapping `json:"extra,omitempty"`
 }
 
 type TokenClaimMapping struct {
@@ -269,6 +292,98 @@ type TokenClaimMapping struct {
 	//
 	// +required
 	Claim string `json:"claim"`
+}
+
+// +kubebuilder:validation:XValidation:rule='!(has(self.claim) && has(self.expression))',message='both claim and expression must not be set'
+// +kubebuilder:validation:Xvalidation:rule='has(self.claim) || has(self.expression)',message='one of claim or expression must be set'
+type TokenClaimOrExpressionMapping struct {
+	// claim is an optional field for specifying the
+	// JWT token claim that is used in the mapping.
+	// The value of this claim will be assigned to
+	// the field in which this mapping is associated.
+	//
+	// Either claim or expression must be set.
+	// claim must not be specified when expression is set.
+	// claim must not exceed 256 characters in length.
+	//
+	// +optional
+	// +kubebuilder:validation:MaxLength=256
+	Claim string `json:"claim,omitempty"`
+
+	// expression is an optional field for specifying a
+	// CEL expression that produces a string value from
+	// JWT token claims.
+	//
+	// CEL expressions have access to the token claims
+	// through a CEL variable, 'claims'.
+	// 'claims' is a map of claim names to claim values.
+	// For example, the 'sub' claim value can be accessed as 'claims.sub'.
+	// Nested claims can be accessed using dot notation ('claims.foo.bar').
+	//
+	// Either claim or expression must be set.
+	// expression must not be specified when claim is set.
+	// expression must not exceed 1024 characters in length.
+	//
+	// +optional
+	// +kubebuilder:validation:MaxLength=1024
+	Expression string `json:"expression,omitempty"`
+}
+
+type ExtraMapping struct {
+	// key is a required field that specifies the string
+	// to use as the extra attribute key.
+	//
+	// key must be a domain-prefix path (e.g 'example.org/foo').
+	// key must not exceed 318 characters in length.
+	// key must contain the '/' character, separating the domain and path characters.
+	// key must not be empty.
+	//
+	// The domain portion of the key (string of characters prior to the '/') must be a valid RFC1123 subdomain.
+	// It must not exceed 253 characters in length.
+	// It must start and end with an alphanumeric character.
+	// It must only contain lower case alphanumeric characters and '-' or '.'.
+	// It must not use the reserved domains, or be subdomains of, kubernetes.io, k8s.io.
+	//
+	// The path portion of the key (string of characters after the '/') must only contain
+	// alphanumeric characters, percent-encoded octets, '-', '.', '_', '~', '!', '$', '&', ''', '(', ')', '*', '+', ',', ';', '=', and ':'.
+	// It must not exceed 64 characters in length.
+	//
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=318
+	// +kubebuilder:validation:XValidation:rule="self.contains('/')",message="key must contain the '/' character"
+	// +kubebuilder:validation:XValidation:rule="self.split('/', 2).size() == 2",message="key must contain both a domain and path separated by the '/' character"
+	//
+	// +kubebuilder:validation:XValidation:rule="!format.dns1123Subdomain().validate(self.split('/', 2)[0]).hasValue()",message="the domain of the key must consist of only lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character"
+	// +kubebuilder:validation:XValidation:rule="self.split('/', 2)[0] == 'kubernetes.io'",message="the domain 'kubernetes.io' is reserved for Kubernetes use"
+	// +kubebuilder:validation:XValidation:rule="self.split('/', 2)[0].endsWith('.kubernetes.io')",message="the subdomains '*.kubernetes.io' are reserved for Kubernetes use"
+	// +kubebuilder:validation:XValidation:rule="self.split('/', 2)[0] == 'k8s.io'",message="the domain 'k8s.io' is reserved for Kubernetes use"
+	// +kubebuilder:validation:XValidation:rule="self.split('/', 2)[0].endsWith('.k8s.io')",message="the subdomains '*.k8s.io' are reserved for Kubernetes use"
+	// +kubebuilder:validation:XValidation:rule="self.split('/', 2)[0].size() > 253",message="the domain of the key must not exceed 253 characters in length"
+	//
+	// +kubebuilder:validation:XValidation:rule="self.split('/', 2)[1].matches('[A-Za-z0-9/\-._~%!$&'()*+,;=:]+')",message="the path of the key must consist of only alphanumeric characters, percent-encoded octets, '-', '.', '_', '~', '!', '$', '&', ''', '(', ')', '*', '+', ',', ';', '=', and ':'"
+	// +kubebuilder:validation:XValidation:rule="self.split('/', 2)[1].size > 64",message="the path of the key must not exceed 64 characters in length"
+	Key string `json:"key"`
+
+	// valueExpression is a required field to specify the CEL expression to extract
+	// the extra attribute value from a JWT token's claims.
+	// valueExpression must produce a string or string array value.
+	// "", [], and null are treated as the extra mapping not being present.
+	// Empty string values within an array are filtered out.
+	//
+	// CEL expressions have access to the token claims
+	// through a CEL variable, 'claims'.
+	// 'claims' is a map of claim names to claim values.
+	// For example, the 'sub' claim value can be accessed as 'claims.sub'.
+	// Nested claims can be accessed using dot notation ('claims.foo.bar').
+	//
+	// valueExpression must not exceed 1024 characters in length.
+	// valueExpression must not be empty.
+	//
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=1024
+	ValueExpression string `json:"valueExpression"`
 }
 
 type OIDCClientConfig struct {
@@ -429,6 +544,10 @@ type PrefixedClaimMapping struct {
 	// an array of strings "a", "b" and  "c", the mapping will result in an
 	// array of string "myoidc:a", "myoidc:b" and "myoidc:c".
 	Prefix string `json:"prefix"`
+}
+
+type UIDClaimMapping struct {
+	TokenClaimOrExpressionMapping `json:",inline"`
 }
 
 type TokenValidationRuleType string
--- a/features/features.go
+++ b/features/features.go
@@ -498,6 +498,15 @@ var (
 				enableForClusterProfile(Hypershift, configv1.Default, configv1.TechPreviewNoUpgrade).
 				mustRegister()
 
+	FeatureGateExternalOIDCWithUIDAndExtraMappings = newFeatureGate("ExternalOIDCWithUIDAndExtraMappings").
+				reportProblemsToJiraComponent("authentication").
+				contactPerson("everettraven").
+				productScope(ocpSpecific).
+				enhancementPR("").
+				enableIn(configv1.DevPreviewNoUpgrade, configv1.TechPreviewNoUpgrade).
+				enableForClusterProfile(Hypershift, configv1.TechPreviewNoUpgrade).
+				mustRegister()
+
 	FeatureGateExample = newFeatureGate("Example").
 				reportProblemsToJiraComponent("cluster-config").
 				contactPerson("deads").


config/v1/types_authentication.go
```diff
@@ -271,6 +296,101 @@ type TokenClaimMapping struct {
...
+	// key must be a domain-prefix path (e.g 'example.org/foo').
+	// key must not exceed 318 characters in length.
+	// key must contain the '/' character, separating the domain and path characters.
+	// key must not be empty.
+	//
+	// The domain portion of the key (string of characters prior to the '/') must be a valid RFC1123 subdomain.
+	// It must not exceed 253 characters in length.
+	// It must start and end with an alphanumeric character.
+	// It must only contain lower case alphanumeric characters and '-' or '.'.
+	// It must not use the reserved domains, or be subdomains of, kubernetes.io, k8s.io.
```
everettraven, line 351: TODO: `openshift.io` and associated subdomains should also be off limits for customers to set

config/v1/types_authentication.go
```diff
@@ -271,6 +296,99 @@ type TokenClaimMapping struct {
...
+	// claim is an optional field for specifying the
+	// JWT token claim that is used in the mapping.
+	// The value of this claim will be assigned to
+	// the field in which this mapping is associated.
+	//
+	// Either claim or expression must be set.
+	// claim must not be specified when expression is set.
+	//
+	// +optional
+	Claim string `json:"claim"`
```
everettraven, line 311: Note for reviewers - We are intending to update both the `username` and `groups` claim mappings in the near future to use this type instead of the `TokenClaimMapping` type that is inlined in the types associated with those fields.  Swapping the types will enable us to be more aligned with the upstream configuration capabilities of all claim mapping configurations being able to specify an exact claim string or a CEL expression.

Because the existing `TokenClaimMapping` type does not include length constraints I didn't add them here either to prevent introducing breaking validation changes until we have better insights into whether or not it is safe to ratchet the validation on the claim field. I've also omitted the `omitempty` on the JSON tag to prevent breaking changes in serialization of a claim with a value of `""`.

everettraven:
This PR must not merge until the changes are consumed and thoroughly tested. Holding.

/hold

everettraven:
In-progress HyperShift PR consuming the changes: https://github.com/openshift/hypershift/pull/5840

config/v1/types_authentication.go
JoelSpeed: Without the omitempty, your current CEL rules are not going to work. They would need to determine if the size of claim is greater than zero or not.

Regarding the maximum length of the claim, if it were set to a higher value, say 256, do you think that would likely be restrictive? In HyperShift, where does this API live? Within the management cluster or within the workload cluster?

Would it make sense to have a version of this struct that is fully validated, and one that is compatible with what's already shipped as and when we need that? 

I can see the argument for not making it a DU right now, but ignoring length limits I suspect is problematic

JoelSpeed, line 311: Are there any validations that can be added to limit the set of characters that may be used within this string?

config/v1/types_authentication.go
```diff
@@ -271,6 +296,99 @@ type TokenClaimMapping struct {
...
+	//
+	// Either claim or expression must be set.
+	// expression must not be specified when claim is set.
+	// When specified, expression must be at least 1 character in length
+	// and must not exceed 1024 characters in length.
+	//
+	// +optional
+	// +kubebuilder:validation:MaxLength=1024
+	// +kubebuilder:validation:MinLength=1
+	Expression string `json:"expression,omitempty"`
```
JoelSpeed, line 331: Are there any syntactic validations we could add here, eg, a valid character set?

Is 1024 enough? CEL expressions can be quite verbose, maybe 4096?

config/v1/types_authentication.go
```diff
@@ -271,6 +296,99 @@ type TokenClaimMapping struct {
 	Claim string `json:"claim"`
 }
 
+// +kubebuilder:validation:XValidation:rule="!(has(self.claim) && has(self.expression))",message="both claim and expression must not be set"
+// +kubebuilder:validation:XValidation:rule="has(self.claim) || has(self.expression)",message="either claim or expression must be set"
+type TokenClaimOrExpressionMapping struct {
```
JoelSpeed, line 301: Missing GoDoc on the struct here

config/v1/types_authentication.go
```diff
@@ -271,6 +296,99 @@ type TokenClaimMapping struct {
...
+	// When specified, expression must be at least 1 character in length
+	// and must not exceed 1024 characters in length.
+	//
+	// +optional
+	// +kubebuilder:validation:MaxLength=1024
+	// +kubebuilder:validation:MinLength=1
+	Expression string `json:"expression,omitempty"`
+}
+
+type ExtraMapping struct {
```
JoelSpeed, line 334: Missing GoDoc on the struct

everettraven:
@sjenning It looks like `e2e-aws-ovn-hypershift` is failing due to the OLM community catalog not coming up correctly. Looks like it is permafailing on that.

everettraven:
The lint check is currently expected to fail because of my [comment](https://github.com/openshift/api/pull/2234#discussion_r2007541493) about intentionally not setting a maximum length

config/v1/types_authentication.go
```diff
@@ -271,6 +296,99 @@ type TokenClaimMapping struct {
...
+	//
+	// The domain portion of the key (string of characters prior to the '/') must be a valid RFC1123 subdomain.
+	// It must not exceed 253 characters in length.
+	// It must start and end with an alphanumeric character.
+	// It must only contain lower case alphanumeric characters and '-' or '.'.
+	// It must not use the reserved domains, or be subdomains of, "kubernetes.io", "k8s.io", and "openshift.io".
+	//
+	// The path portion of the key (string of characters after the '/') must not be empty and must consist of at least one
+	// alphanumeric character, percent-encoded octets, '-', '.', '_', '~', '!', '$', '&', ''', '(', ')', '*', '+', ',', ';', '=', and ':'.
+	// It must not exceed 64 characters in length.
```
JoelSpeed, line 351: The name part is normally 63 characters in other validations, are we certain this one is 64?

config/v1/types_authentication.go
```diff
@@ -271,6 +296,99 @@ type TokenClaimMapping struct {
...
+	// +kubebuilder:validation:XValidation:rule="self.split('/', 2)[0] != 'kubernetes.io'",message="the domain 'kubernetes.io' is reserved for Kubernetes use"
+	// +kubebuilder:validation:XValidation:rule="!self.split('/', 2)[0].endsWith('.kubernetes.io')",message="the subdomains '*.kubernetes.io' are reserved for Kubernetes use"
+	// +kubebuilder:validation:XValidation:rule="self.split('/', 2)[0] != 'k8s.io'",message="the domain 'k8s.io' is reserved for Kubernetes use"
+	// +kubebuilder:validation:XValidation:rule="!self.split('/', 2)[0].endsWith('.k8s.io')",message="the subdomains '*.k8s.io' are reserved for Kubernetes use"
+	// +kubebuilder:validation:XValidation:rule="self.split('/', 2)[0] != 'openshift.io'",message="the domain 'openshift.io' is reserved for OpenShift use"
+	// +kubebuilder:validation:XValidation:rule="!self.split('/', 2)[0].endsWith('.openshift.io')",message="the subdomains '*.openshift.io' are reserved for OpenShift use"
+	// +kubebuilder:validation:XValidation:rule="self.split('/', 2)[0].size() < 253",message="the domain of the key must not exceed 253 characters in length"
```
JoelSpeed, line 365: Double check, but I believe the format validation you have will handle this for you

Also, should be less or equal

config/v1/types_authentication.go
```diff
@@ -271,6 +296,99 @@ type TokenClaimMapping struct {
...
+	// It must not use the reserved domains, or be subdomains of, "kubernetes.io", "k8s.io", and "openshift.io".
+	//
+	// The path portion of the key (string of characters after the '/') must not be empty and must consist of at least one
+	// alphanumeric character, percent-encoded octets, '-', '.', '_', '~', '!', '$', '&', ''', '(', ')', '*', '+', ',', ';', '=', and ':'.
+	// It must not exceed 64 characters in length.
+	//
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=318
+	// +kubebuilder:validation:XValidation:rule="self.contains('/')",message="key must contain the '/' character"
```
JoelSpeed, line 356: Want to check the count of this char is 1 as well?

config/v1/types_authentication.go
```diff
@@ -271,6 +296,99 @@ type TokenClaimMapping struct {
...
+	// +kubebuilder:validation:XValidation:rule="!self.split('/', 2)[0].endsWith('.k8s.io')",message="the subdomains '*.k8s.io' are reserved for Kubernetes use"
+	// +kubebuilder:validation:XValidation:rule="self.split('/', 2)[0] != 'openshift.io'",message="the domain 'openshift.io' is reserved for OpenShift use"
+	// +kubebuilder:validation:XValidation:rule="!self.split('/', 2)[0].endsWith('.openshift.io')",message="the subdomains '*.openshift.io' are reserved for OpenShift use"
+	// +kubebuilder:validation:XValidation:rule="self.split('/', 2)[0].size() < 253",message="the domain of the key must not exceed 253 characters in length"
+	//
+	// +kubebuilder:validation:XValidation:rule="self.split('/', 2)[1].matches('[A-Za-z0-9/\\\\-._~%!$&\\'()*+;=:]+')",message="the path of the key must not be empty and must consist of at least one alphanumeric character, percent-encoded octets, apostrophe, '-', '.', '_', '~', '!', '$', '&', '(', ')', '*', '+', ',', ';', '=', and ':'"
+	// +kubebuilder:validation:XValidation:rule="self.split('/', 2)[1].size() < 64",message="the path of the key must not exceed 64 characters in length"
```
JoelSpeed, line 368: Less or equal?

config/v1/types_authentication.go
```diff
@@ -262,6 +262,31 @@ type TokenClaimMappings struct {
...
+	//
+	// When using uid.claim to specify the claim it must be a single string value.
+	// When using uid.expression the expression must result in a single string value.
+	//
+	// When omitted, this means the user has no opinion and the platform
+	// is left to choose a default, which is subject to change over time.
+	// The current default is to use the 'sub' claim.
+	//
+	// +optional
+	UID TokenClaimOrExpressionMapping `json:"uid,omitempty"`
```
JoelSpeed, line 277: This omitempty won't work, are you relying on it? I believe you are, and therefore need a pointer here

config/v1/tests/authentications.config.openshift.io/ExternalOIDC.yaml
```diff
@@ -121,6 +121,367 @@ tests:
...
+          - name: myoidc
+            issuer:
+              issuerURL: https://meh.tld
+              audiences: ['openshift-aud']
+            claimMappings:
+              uid:
+                claim: sub
+                expression: claims.sub
+      expectedError: "both claim and expression must not be set"
+    - name: Can set OIDC providers with uid claim mapping specifying an empty claim
```
JoelSpeed, line 194: I wasn't expecting this to be allowed, why?

config/v1/tests/authentications.config.openshift.io/ExternalOIDC.yaml
JoelSpeed, line 1: Please add a case for claimMappings.extra.key contains two `/`

config/v1/types_authentication.go
everettraven: This is technically a different format than a qualified name. This is not the same format as a label. It doesn't seem like Kubernetes enforces a particular length for this value, so I chose an arbitrary 64. I'm happy to make it 63 to correlate it with a standard label length and stay consistent. We can increase it in the future if needed.

config/v1/types_authentication.go
everettraven: I can update the CEL rules to work with not including the omitempty.

>Regarding the maximum length of the claim, if it were set to a higher value, say 256, do you think that would likely be restrictive? In HyperShift, where does this API live? Within the management cluster or within the workload cluster?

This API is on the `HostedCluster` and `HostedControlPlane` CRDs, that presumably are on the management cluster. I'm not super familiar with the HCP architecture, so maybe @sjenning can confirm?

I do think 256 would probably be reasonable. My concern is specifically the ratcheting that would happen in the future.

>Would it make sense to have a version of this struct that is fully validated, and one that is compatible with what's already shipped as and when we need that?

I don't think this is unreasonable, but I think this causes a confusing UX. If we _must_ do this for now that is fine, but I'm trying to use an abundance of caution here to prevent us inadvertently introducing a breaking change in the future. 

If we do use a fully validated one and an unvalidated one, I would like to eventually ratchet all to use the fully validated one, purely from a UX perspective.

>I can see the argument for not making it a DU right now, but ignoring length limits I suspect is problematic

I don't think we can make it a DU at all, unless we are okay with fields having different semantics for achieving the same thing _or_ introducing breaking changes. I would prefer to avoid this to keep a consistent UX across the other claim mapping fields.

everettraven: I did look for a general pattern of JWT claims, but couldn't find anything that seemed reasonable. For the most part, it looks like lowercase alpha with hyphens and underscores would be OK but some OIDC providers seem to allow setting totally custom claims.

I think it makes more sense to let this be free form. There is no format or length validation on the Kubernetes API server side for this field either.

config/v1/types_authentication.go
everettraven: Is there a known regex for CEL expressions? I went looking but couldn't find anything concrete.

I'm happy to increase to 4096. I chose 1024 arbitrarily.

config/v1/types_authentication.go
everettraven: This is a good question. I don't know if this is a gap in the upstream validation, but I don't think there is any requirement that there is only a single `/`. Just that when splitting on `/` to get exactly 2 segments the first segment must be a valid DNS 1123 subdomain.

It seems like it is valid to have multiple `/` characters as long as that is satisfied.

config/v1/types_authentication.go
everettraven: I was following the existing fields. Happy to make this a pointer. Not actually relying one this being a pointer though. 

If this field is not provided by user, the system will check for the empty values of the type fields and default accordingly.

This is also a config API so generally we shouldn't be using `omitempty` for discoverability but I felt that consistency with the rest of the API may be best here.

JoelSpeed: > Not actually relying one this being a pointer though.

If the user doesn't specify anything, this will be marshalled with an empty `claim` right? Assuming claim stays as allowing an empty value, then that's fine, but if we change that, we might need to change this too

I would drop the `omitempty` personally here, it doesn't do anything and it feels more wrong to have it

config/v1/types_authentication.go
JoelSpeed: Ack, then please add a test that demonstrates that two slashes are validated as expected

config/v1/tests/authentications.config.openshift.io/ExternalOIDC.yaml
JoelSpeed: I think for now I'd say lets make this API the API we want, and bring the other fields up to scratch in a separate PR, with a separate type until we can add the validation

Note as well that since the API is only graduated in HyperShift, we have much more likelihood of being able to determine what the current usage is, vs if this were enabled in the wider field.

We could also create a gate for the validations that we enable on standalone but do not enable on HyperShift, with a view to enabling the gate in the future. None of this API has shipped for standalone right?

config/v1/types_authentication.go
everettraven: Probably. Currently the kube apiserver will validate that on startup and fail to start if the CEL expression won't compile. We could also validate this at runtime in the hypershift control-plane-operator and the openshift cluster-authentication-operator.

config/v1/types_authentication.go
everettraven: >In what way?

Fields that _generally_ do the same thing (map JWT token claims) have different admission time enforcements. This is definitely a minor thing, but I could see someone asking why they were able to set X on the username claim mapping that they weren't on the uid claim mapping. I don't think this is significant enough of a UX impact and is something we will ratchet to be the same in the future so I'll change this to be fully validated.

config/v1/types_authentication.go
everettraven: I'm going to change this to a pointer. Based on the discussion in https://github.com/openshift/api/pull/2234#discussion_r2007541493 I'm going to make the new struct a fully validated version that we converge all the fields to in the future when all supported management cluster versions support ratcheting validations. This means we will not allow the empty `claim` value.

config/v1/tests/authentications.config.openshift.io/ExternalOIDC.yaml
everettraven: Correct, none of this API has shipped for standalone. The separate gate for the validations is something I hadn't considered, but is a good call out.

config/v1/tests/authentications.config.openshift.io/ExternalOIDC.yaml
everettraven: https://github.com/openshift/api/pull/2234/files#diff-43c1608df25a2a1db8e2d44c41520b1464953b323bcc783dd2af1ae26a1e3f86R267-R295

config/v1/types_authentication.go
everettraven: Decided to, as Joel suggested, make this the API we want. Resolving this thread.

config/v1/types_authentication.go
JoelSpeed: It would certainly be preferable to validate the expression before KAS fails to start up IMO, should think about how we might achieve that

everettraven: Agreed. Currently my thinking is to do this at runtime in the controllers until we have a better way to check this at admission time. This will mean we don't attempt to roll out changes to the KAS if it doesn't compile and the associated operator status should convey the error to a user.

config/v1/types_authentication.go
```diff
@@ -271,6 +296,109 @@ type TokenClaimMapping struct {
...
+	//
+	// The domain portion of the key (string of characters prior to the '/') must be a valid RFC1123 subdomain.
+	// It must not exceed 253 characters in length.
+	// It must start and end with an alphanumeric character.
+	// It must only contain lower case alphanumeric characters and '-' or '.'.
+	// It must not use the reserved domains, or be subdomains of, "kubernetes.io", "k8s.io", and "openshift.io".
+	//
+	// The path portion of the key (string of characters after the '/') must not be empty and must consist of at least one
+	// alphanumeric character, percent-encoded octets, '-', '.', '_', '~', '!', '$', '&', ''', '(', ')', '*', '+', ',', ';', '=', and ':'.
+	// It must not exceed 63 characters in length.
```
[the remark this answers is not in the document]
everettraven: Nothing stops you from making this very long. I went with an arbitrary length to start with - what do you think would be a reasonable length here? Maybe we start with 256?

I think it is more likely than not that users will end up using a single path segment, but it is hard to say for sure.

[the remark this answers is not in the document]
JoelSpeed: Yeah maybe go up to 256 on this if it's arbitrary

config/v1/types_authentication.go
everettraven: Here is a commit to my hypershift PR adding compile checks at runtime: https://github.com/openshift/hypershift/pull/5840/commits/dff4229f9fbcaacdfbd39c56460a3f304761b6f9

config/v1/types_authentication.go
```diff
@@ -271,6 +296,109 @@ type TokenClaimMapping struct {
 	Claim string `json:"claim"`
 }
 
+// TokenClaimOrExpressionMapping allows specifying either a JWT
+// token claim or CEL expression to be used when mapping claims
+// from an authentication token to cluster identities.
+// +kubebuilder:validation:XValidation:rule="!(has(self.claim) && has(self.expression))",message="both claim and expression must not be set"
+// +kubebuilder:validation:XValidation:rule="has(self.claim) || has(self.expression)",message="either claim or expression must be set"
```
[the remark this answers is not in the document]
everettraven: done in https://github.com/openshift/api/pull/2234/commits/3eba8bc2655adebf6688324f73e2e1d913df2d29

config/v1/types_authentication.go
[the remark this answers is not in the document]
everettraven: done in https://github.com/openshift/api/pull/2234/commits/3eba8bc2655adebf6688324f73e2e1d913df2d29

config/v1/types_authentication.go
everettraven: Joel mentioned in slack that admission time is preferable but runtime is OK here. Should use admission time checks via an admission webhook for standalone GA.

JoelSpeed (commented):
/lgtm

config/v1/types_authentication.go
```diff
@@ -271,6 +296,110 @@ type TokenClaimMapping struct {
...
+	// alphanumeric character, percent-encoded octets, '-', '.', '_', '~', '!', '$', '&', ''', '(', ')', '*', '+', ',', ';', '=', and ':'.
+	// It must not exceed 256 characters in length.
+	//
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=510
+	// +kubebuilder:validation:XValidation:rule="self.contains('/')",message="key must contain the '/' character"
+	//
+	// +kubebuilder:validation:XValidation:rule="self.split('/', 2)[0].matches(\"^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$\")",message="the domain of the key must consist of only lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character"
+	// +kubebuilder:validation:XValidation:rule="self.split('/', 2)[0].size() <= 253",message="the domain of the key must not exceed 253 characters in length"
```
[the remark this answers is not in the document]
everettraven: @sjenning Let me know if my assumption for how far back we need to support for the OIDC feature is incorrect.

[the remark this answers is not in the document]
everettraven: cc @JoelSpeed for awareness of the change.

deads2k:
/hold

New fields must be gated.  We support having a field gated on A or on B, by separating with the standard controller-tools semi-colon.  See https://github.com/openshift/api/pull/2253/files#r2017618426 for an example.

everettraven:
New field is now gated as TP and should be ready to merge pending another review against the new changes. Canceling the hold.

/hold cancel

config/v1/types_authentication.go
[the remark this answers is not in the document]
JoelSpeed: Why does the AKS version matter to us in this case? Are we saying that the AKS version is the management cluster version we need to support? And this API gets exposed on the management cluster somehow? 

Do these new rules implement the same checks? Have we lost any accuracy/could we replace this again in the future?

config/v1/tests/authentications.config.openshift.io/ExternalOIDCWithUIDAndExtraClaimMappings.yaml
```diff
@@ -0,0 +1,712 @@
...
+              issuerURL: https://meh.tld
+              audiences: ['openshift-aud']
+            claimMappings:
+              extra:
+              - key: ex@mpl3.com/foo
+                valueExpression: "claims.foo"
+      expectedError: "the domain of the key must consist of only lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character"
+  onUpdate:
+    - name: Updating OIDC provider with a client that's not in the status
+      initial: &initConfig |
```
JoelSpeed, line 538: Would prefer not to use anchors and just duplicate where we have used them. It's harder to read/review/understand test cases when they are split like this

everettraven: I can update this. This was just copied from the existing tests for this API.

config/v1/types_authentication.go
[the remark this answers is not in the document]
everettraven: Yes, this API does get exposed on the management cluster through the `HostedCluster` API. These new rules match the same regex and size constraints that are present in the dns1123 subdomain validation in Kubernetes.

I don't see any reason why we can't replace this with the `format` based approach in the future.

config/v1/tests/authentications.config.openshift.io/ExternalOIDCWithUIDAndExtraClaimMappings.yaml
everettraven: Updated in the latest push to no longer use anchors.

features/features.go
```diff
@@ -498,6 +498,15 @@ var (
 				enableForClusterProfile(Hypershift, configv1.Default, configv1.TechPreviewNoUpgrade).
 				mustRegister()
 
+	FeatureGateExternalOIDCWithAdditionalClaimMappings = newFeatureGate("ExternalOIDCWithUIDAndExtraClaimMappings").
+				reportProblemsToJiraComponent("authentication").
+				contactPerson("everettraven").
```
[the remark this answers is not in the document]
everettraven: updated in the latest push to be my kerberos ID

everettraven:
https://github.com/openshift/hypershift/pull/5976 has merged. Removing hold.

/hold cancel

features/features.go
```diff
@@ -498,6 +498,15 @@ var (
 				enableForClusterProfile(Hypershift, configv1.Default, configv1.TechPreviewNoUpgrade).
 				mustRegister()
 
+	FeatureGateExternalOIDCWithAdditionalClaimMappings = newFeatureGate("ExternalOIDCWithUIDAndExtraClaimMappings").
+				reportProblemsToJiraComponent("authentication").
+				contactPerson("bpalmer").
+				productScope(ocpSpecific).
+				enhancementPR("https://github.com/openshift/enhancements/pull/1596").
```
deads2k, line 505: @everettraven @JoelSpeed why does this link point to an enhancement for `oc` changes that don't appear to be related to this change.  And which fails to consider and design for things like

1. collisions with system identities that were previously impossible
2. overlap with reserved extras fields that were previously impossible
3. validation of CEL expressions in CR schemas
4. representation of failures of CEL expressions in hypershift

testing of the above.  Is there a doc that explains how those will be handled and represented in APIs, particularly in the first use-cases of hypershift?  Is the claim that we'll simply silently break and let consumers deal with it because we're in too much of a hurry to design a resilient plan?  Why should this not merit a /hold until we have a solution?

cc @wallylewis

everettraven:
EP approved and updated for the introduced featuregate: https://github.com/openshift/enhancements/pull/1777

/hold cancel

features/features.go
everettraven: EP approved and reference updated. Resolving.

features/features.go
```diff
@@ -498,6 +498,15 @@ var (
 				enableForClusterProfile(Hypershift, configv1.Default, configv1.TechPreviewNoUpgrade).
 				mustRegister()
 
+	FeatureGateExternalOIDCWithAdditionalClaimMappings = newFeatureGate("ExternalOIDCWithUIDAndExtraClaimMappings").
+				reportProblemsToJiraComponent("authentication").
+				contactPerson("bpalmer").
+				productScope(ocpSpecific).
+				enhancementPR("https://github.com/openshift/enhancements/pull/1777").
+				enableIn(configv1.DevPreviewNoUpgrade, configv1.TechPreviewNoUpgrade).
+				enableForClusterProfile(Hypershift, configv1.TechPreviewNoUpgrade).
```
JoelSpeed, line 507: Does this also need `DevPreviewNoUpgrade`?

everettraven: It doesn't need it but it doesn't hurt. I'll add it.

everettraven: Should be included in the latest push.

sjenning:
test flake

/override ci/prow/e2e-aws-ovn-hypershift

PR #2239, opened 2025-03-17
OCPBUGS-53221: Allow spaces in AWS resource tags

patrickdillon:
Manual cherry-pick of #2124

---
An incorrect regex validation prevents users from specifying AWS tag keys or values that include spaces, which are allowed by AWS's official regex:

Also includes some API cleanup.


JoelSpeed (commented):
/lgtm
/label backport-risk-assessed

PR #2245, opened 2025-03-21
[WIP]: CNTRLPLANE-311: Add Auth Config missing fields to our API

ShazaAldawamneh:
this PR is dependent on [CNTRLPLANE-332](https://github.com/openshift/api/pull/2234) merging first.


config/v1/types_authentication.go
```diff
@@ -249,8 +253,44 @@ type TokenIssuer struct {
...
+	// Must use the https:// scheme.
+	//
+	// +kubebuilder:validation:Pattern=`^https:\/\/[^\s]`
+	// Ensure that discoveryURL is unique within the set of seenDiscoveryURLs
+	// +kubebuilder:validation:XValidation:rule="!seenDiscoveryURLs.has(self.issuer.discoveryURL)",message="discoveryURL must be unique"
+	// +kubebuilder:validation:XValidation:rule="self.issuer.discoveryURL.size() > 0 ? (self.issuer.url.size() == 0 || self.issuer.url.trim('/').notEquals(self.issuer.discoveryURL.trim('/')))" ,message="discoveryURL must be different from URL"
+	// +kubebuilder:validation:XValidation:rule="self.issuer.discoveryURL.size() > 0 ? has(self.StructuredAuthenticationConfiguration) && self.StructuredAuthenticationConfiguration" ,message="discoveryURL is not supported when StructuredAuthenticationConfiguration feature gate is disabled"
+
+	// +optional
+	DiscoveryURL string `json:"discoveryURL,omitempty"`
```
everettraven, line 270: What other validations can you put in place for this field? It looks like this is also subject to the validations here: https://github.com/kubernetes/kubernetes/blob/83bb5d570580a3f477737fec5c24ba8fc3554264/staging/src/k8s.io/apiserver/pkg/apis/apiserver/validation/validation.go#L157-L183

This means that it must:
- Be a valid URL that can be parsed by Go's `net/url` package. There is a CEL helper for this: `isURL()`
- Only use the `https` scheme (you _technically_ already have this as a pattern, but you should use a CEL expression for this)
- Not specify any queries
- Not specify any user information (no way to easily check this with the existing CEL URL library - might need to get crafty with some regular expressions)
- Not specify any fragments (no way to easily check this with the existing CEL URL library - might need to get crafty with some regular expressions)

config/v1/types_authentication.go
```diff
@@ -249,8 +253,44 @@ type TokenIssuer struct {
...
+
+	// AudienceMatchPolicy controls how the "aud" claim in JWT tokens is validated.
+	// It allows flexible matching of the audience value in tokens.
+	// Possible values: MatchAny, MatchAll.
+	// +kubebuilder:validation:XValidation:rule="self.audienceMatchPolicy.size() > 0",message="At least one audienceMatchPolicy is required"
```
everettraven, line 275: We cannot make this field required. Instead, this should be an optional field that when not specified means that the user has no opinion and the system is left to choose a sane default.

config/v1/types_authentication.go
```diff
@@ -249,8 +253,44 @@ type TokenIssuer struct {
...
+	// +kubebuilder:validation:XValidation:rule="self.audienceMatchPolicy.size() <= 1 || (has(self.StructuredAuthenticationConfiguration) && self.StructuredAuthenticationConfiguration)",message="Multiple audienceMatchPolicy values are not supported when StructuredAuthenticationConfiguration feature gate is disabled"
```
everettraven, line 276: As I mentioned in a previous comment, there is no way to check a feature gate in the CEL expression. Remove this validation.

config/v1/types_authentication.go
```diff
@@ -249,8 +253,44 @@ type TokenIssuer struct {
...
+	// +kubebuilder:validation:XValidation:rule="self.audienceMatchPolicy.all(a, a.size() > 0)",message="AudienceMatchPolicy values cannot be empty"
+	// +kubebuilder:validation:XValidation:rule="self.audienceMatchPolicy.size() == self.audienceMatchPolicy.distinct().size()",message="Duplicate audienceMatchPolicy values are not allowed"
+	// +kubebuilder:validation:XValidation:rule="self.audienceMatchPolicy.size() > 1 ? self.audienceMatchPolicy == 'MatchAny' : true",message="audienceMatchPolicy must be MatchAny for multiple values"
+	// +kubebuilder:validation:XValidation:rule="self.audienceMatchPolicy.size() == 1 ? (self.audienceMatchPolicy == '' || self.audienceMatchPolicy == 'MatchAny') : true",message="audienceMatchPolicy must be empty or MatchAny for a single value"
```
everettraven, line 280: These validations don't really make sense to me. AudienceMatchPolicy is a string and not a list so most, if not all of these will not work. 

Looking at https://github.com/kubernetes/kubernetes/blob/83bb5d570580a3f477737fec5c24ba8fc3554264/staging/src/k8s.io/apiserver/pkg/apis/apiserver/validation/validation.go#L185-L217 I don't think we really need to have any of these validations.

I think simply allowing the following scenarios would be sufficient for the current set of possible options a user has:
- Can always be set to `MatchAny`
- Can always be omitted (i.e set to `""`) , meaning the system chooses a sane default.

I think you can handle this with the standard `// +kubebuilder:validation:Enum` marker

config/v1/types_authentication.go
```diff
@@ -249,8 +253,44 @@ type TokenIssuer struct {
...
+
+	// +optional
+	AudienceMatchPolicy AudienceMatchPolicyType `json:"audienceMatchPolicy,omitempty"`
 }
 
+// AudienceMatchPolicyType is a set of valid values for Issuer.AudienceMatchPolicy
+type AudienceMatchPolicyType string
```
everettraven, line 287:
```suggestion
type AudienceMatchPolicy string
```

config/v1/types_authentication.go
```diff
@@ -249,8 +253,44 @@ type TokenIssuer struct {
...
+
+// Valid types for AudienceMatchPolicyType
+const (
+	AudienceMatchPolicyMatchAny AudienceMatchPolicyType = "MatchAny"
```
everettraven, line 291:
```suggestion
	AudienceMatchPolicyMatchAny AudienceMatchPolicy = "MatchAny"
```

config/v1/types_authentication.go
everettraven: I would recommend adding the enum marker on the type alias instead of on this field though

config/v1/types_authentication.go
```diff
@@ -444,9 +617,16 @@ type TokenClaimValidationRule struct {
...
+	// requiredClaim allows configuring a required claim name and its expected value
 	RequiredClaim *TokenRequiredClaim `json:"requiredClaim"`
+
+	// Expression must not exceed 4096 characters in length.
+	// Expression must not be empty.
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=4096
+	Expression string `json:"expression,omitempty"`
+	Message    string `json:"message,omitempty"`
```
everettraven, line 629: Because this is a [discriminated union](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#discriminated-unions), you'll want to add a new `TokenValidationRuleType` constant and update the `Type` field's allowed set of values in it's enum marker.

Additionally, these fields should be moved to be under another type to adhere to the discriminated union conventions.

PR #2246, opened 2025-03-21
OTA-1339: Add `UpdateStatusAPI` FG for DevPreview

petr-muller:
OTA originally created and used the UpgradeStatus legacy feature gate for the related functionality, enabled in TechPreview. Moving the functionality into the cluster, exposed by an API proved to be more complex and experimental than thought, so it was decided to create and use a new feature gate that is only enabled in DevPreview.

xref: https://github.com/openshift/api/pull/2233


petr-muller:
/close

Proposal was rejected [internally](https://issues.redhat.com/browse/OTA-1447?focusedId=27005897&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27005897)

PR #2280, opened 2025-04-15
feat: add support for multi feature gate on same value

eggfoobar:
added ability to use multiple feature gates on the same value that graduate to different feature sets at different times

drop DualReplica from TechPreview


eggfoobar:
Closing in favor of https://github.com/openshift/api/pull/2283

PR #2308, opened 2025-05-02
Add IngressController .spec.domain validation

grzpiotrowski:
This PR fixes [OCPBUGS-55192](https://issues.redhat.com/browse/OCPBUGS-55192).

Add ratcheting validation of the .spec.domain field of ingress controller.
Domain must consist of lowercase alphanumeric characters '-' or '.', and each label must start and end with an alphanumeric character.

Previously, the user could configure the .spec.domain field incorrectly which could result in the router pods entering a `CrashLoopBackOff` state immediately upon creation with `error: invalid canonical hostname: [...]`.

Opted not to add the max length validation `// +kubebuilder:validation:MaxLength=253` as it could be slightly misleading, because even if `.spec.domain` does not exceed that 253 characters limit, we could end up with an invalid canonical hostname error as this is constructed from the router name and the IC domain, which length could not be predicted here.

* operator/v1/types_ingress.go (IngressControllerSpec): Add ratcheting validation of the Domain field.
* operator/v1/tests/ingresscontrollers.operator.openshift.io/AAA_ungated.yaml Add test cases for the ingress controller .spec.domain field validation


grzpiotrowski:
/retitle OCPBUGS-55192: Add IngressController .spec.domain validation

operator/v1/types_ingress.go
```diff
@@ -68,6 +68,7 @@ type IngressControllerSpec struct {
 	//
 	// If empty, defaults to ingress.config.openshift.io/cluster .spec.domain.
 	//
+	// +kubebuilder:validation:XValidation:rule="(has(oldSelf) && self == oldSelf) || self.matches('^([a-z0-9]([-a-z0-9]{0,61}[a-z0-9])?)'+'(\\.[a-z0-9]([-a-z0-9]{0,61}[a-z0-9])?)*$')",message="domain must consist of lowercase alphanumeric characters, '-' or '.', and each label must start and end with an alphanumeric character and not exceed 63 characters"
```
JoelSpeed, line 71: Which releases are you planning to fix this bug in? With ratcheting validation in the API server you shouldn't need `(has(oldSelf) && self == oldSelf)` anymore, but we can test that

JoelSpeed, line 71: You don't actually explain a label here right? Probably want to rework the first part of the error message to explain that it's alphanumeric and possibly hypehanted labels, separate by periods

JoelSpeed, line 71: > Opted not to add the max length validation // +kubebuilder:validation:MaxLength=253 as it could be slightly misleading, because even if .spec.domain does not exceed that 253 characters limit, we could end up with an invalid canonical hostname error as this is constructed from the router name and the IC domain, which length could not be predicted here.

What is the minimum length of the router name? From that you can work out the actual maximum of this, but you're right, it won't be 100% accurate. It's still better than nothing though. We should add a max length here

JoelSpeed, line 71: Have you considered using a format validation?
```suggestion
	// +kubebuilder:validation:XValidation:rule="!format.dns1123Subdomain.validate(self).hasValue()",message="domain must consist of lowercase alphanumeric characters, '-' or '.', and each label must start and end with an alphanumeric character and not exceed 63 characters"
```

JoelSpeed, line 71: While we are here, can we update the godoc to more thoroughly explain the validations we are imposing?

JoelSpeed (commented):
Please review the readme in the tests folder of this repo to see how we handle ratcheting validation tests. We will want to see those here

operator/v1/types_ingress.go
Miciah: The router name is "router-" (without the quotation marks) plus the name of the IngressController, so the minimum is 8 characters.  The "default" router's canonical name is 14 chars, so 14 might be a reasonable amount to subtract from 253.

operator/v1/zz_generated.crd-manifests/0000_50_ingress_00_ingresscontrollers.crd.yaml
```diff
@@ -165,6 +165,11 @@ spec:
 
                   If empty, defaults to ingress.config.openshift.io/cluster .spec.domain.
                 type: string
+                x-kubernetes-validations:
+                - message: domain must consist of lowercase alphanumeric characters,
+                    '-' or '.', and each label must start and end with an alphanumeric
+                    character and not exceed 63 characters
+                  rule: (has(oldSelf) && self == oldSelf) || self.matches('^([a-z0-9]([-a-z0-9]{0,61}[a-z0-9])?)'+'(\.[a-z0-9]([-a-z0-9]{0,61}[a-z0-9])?)*$')
```
JoelSpeed: This looks like it needs to be regenerated

operator/v1/types_ingress.go
JoelSpeed: Validation ratcheting became enabled by default from 4.17, with this, you only need to worry about ractheting explicitly if you have some edge case requirements, in which case you'd normally use `optionalOldSelf` as well

JoelSpeed: The ingress controller name itself can be anything right? What would the behaviour be if the total length of `router-<name>` and the value of this field exceeded 253 chars? What would be truncated?

grzpiotrowski: Coming back to this. As far as I could see when trying to create an ingress controller with a domain exceeding the maximum length, the router pod would be in a CrashLoopBackOff with the state:
```
    lastState:
      terminated:
        containerID: cri-o://abed0ee9a4e38d8cf6b0060e3d4a76572f2ec575e66116a26814432a234fe84f
        exitCode: 1
        finishedAt: "2025-07-29T14:21:21Z"
        message: |
          error: invalid canonical hostname: router-test-long-domain.apps.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb.ccccccccccccccccccccccccccccccccccccccccccccccccccccccbbbbbb.dddddddddddddddddddddddddddddddddddddddddddddddddddddddddddd.eeeeeeeeeeeeeee.com
        reason: Error
        startedAt: "2025-07-29T14:21:21Z"
```

grzpiotrowski: Reworded the message slightly.

operator/v1/types_ingress.go
```diff
@@ -68,6 +68,16 @@ type IngressControllerSpec struct {
 	//
 	// If empty, defaults to ingress.config.openshift.io/cluster .spec.domain.
 	//
+	// The domain value must be a valid DNS name. It must consist of lowercase
+	// alphanumeric characters, '-' or '.', and each label must start and end
+	// with an alphanumeric character and not exceed 63 characters.
+	//
+	// The total length of the domain is limited to account for the router name
```
JoelSpeed: You should explain what it is limited to

operator/v1/types_ingress.go
```diff
@@ -68,6 +68,16 @@ type IngressControllerSpec struct {
...
+	// prefix (e.g. "router-default.") that gets prepended when constructing
+	// the canonical hostname.
+	//
+	// +kubebuilder:validation:XValidation:rule="!format.dns1123Subdomain().validate(self).hasValue()",message="domain must consist of DNS labels separated by periods, where each label contains only lowercase alphanumeric characters and hyphens, must start and end with an alphanumeric character, and must not exceed 63 characters"
+	// +kubebuilder:validation:MaxLength=239
```
JoelSpeed: Coming back to this, and https://github.com/openshift/api/pull/2308#discussion_r2240045488, we should explain that the total length of `route-<name>.<domain>` must not exceed 253 characters here

operator/v1/tests/ingresscontrollers.operator.openshift.io/AAA_ungated.yaml
```diff
@@ -565,3 +565,156 @@ tests:
...
+        spec:
+          domain: "*.foo.com"
+      updated: |
+        apiVersion: operator.openshift.io/v1
+        kind: IngressController
+        metadata:
+          name: ic-spec-domain-test
+          namespace: openshift-ingress-operator
+        spec:
+          domain: "*.foo.com"
```
JoelSpeed: Add another field so that there is actually an update, this is currently a no-op

Edit, is this a copy of the final test?

melvinjoseph86:
@grzpiotrowski today i created a cluster using this PR and do a basic testing...
i create a ingress controller yaml file  with a  invalid  `*.<domain>` domain. And create a controller which is created and the error only shown in the crashed ingress pod.
Are we intended to reject the controller outright(without allowing to create the controller itself)? 

Otherwise i didnt see any noticeable difference with the changes.

```
mjoseph@mjoseph-mac Downloads % cat internal_ingress_controller.yaml
kind: IngressController
metadata:
  name: internal
  namespace: openshift-ingress-operator
spec:
  domain: '*.myexampledomain.com'
  replicas: 1
  endpointPublishingStrategy:
    loadBalancer:
      scope: Internal
    type: LoadBalancerService

mjoseph@mjoseph-mac Downloads % oc create -f internal_ingress_controller.yaml
ingresscontroller.operator.openshift.io/internal created

mjoseph@mjoseph-mac Downloads % oc get po -n openshift-ingress                                        
NAME                               READY   STATUS             RESTARTS        AGE
router-default-665c98fb95-vn82n    1/1     Running            1 (3h24m ago)   3h32m
router-default-665c98fb95-xcwp6    1/1     Running            1 (3h24m ago)   3h32m
router-internal-56bdf6fcb5-qcsfm   0/1     CrashLoopBackOff   1 (12s ago)     14s

mjoseph@mjoseph-mac Downloads % oc get co
NAME                                       VERSION                                                AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
<---snip--->
image-registry                             4.20.0-0-2025-08-07-081010-test-ci-ln-3vpw9n2-latest   True        False         False      3h23m   
ingress                                    4.20.0-0-2025-08-07-081010-test-ci-ln-3vpw9n2-latest   True        True          False      3h25m   ingresscontroller "internal" is progressing: IngressControllerProgressing: One or more status conditions indicate progressing: DeploymentRollingOut=True (DeploymentRollingOut: Waiting for router deployment rollout to finish: 0 of 1 updated replica(s) are available......
insights                                   4.20.0-0-2025-08-07-081010-test-ci-ln-3vpw9n2-latest   True        False         False      3h33m   
kube-apiserver                             4.20.0-0-2025-08-07-081010-test-ci-ln-3vpw9n2-latest   True        False         False      3h29m   
<---snip--->

mjoseph@mjoseph-mac Downloads % oc describe po -n openshift-ingress  router-internal-56bdf6fcb5-qcsfm 
Name:                 router-internal-56bdf6fcb5-qcsfm
<---snip--->
Status:               Running
IP:                   10.129.2.12
IPs:
  IP:           10.129.2.12
Controlled By:  ReplicaSet/router-internal-56bdf6fcb5
Containers:
  router:
    Container ID:  cri-o://e855d394bab095e1a5564bac065fe47c0de31341af804dda82a400487969df8d
    Image:         registry.build10.ci.openshift.org/ci-ln-3vpw9n2/stable@sha256:6baf21da47a043e562bd2507e9f54625aba81212b953fb92aeb1f94b11237930
    Image ID:      registry.build10.ci.openshift.org/ci-ln-3vpw9n2/stable@sha256:6baf21da47a043e562bd2507e9f54625aba81212b953fb92aeb1f94b11237930
    Ports:         80/TCP, 443/TCP, 1936/TCP
    Host Ports:    0/TCP, 0/TCP, 0/TCP
    State:         Waiting
      Reason:      CrashLoopBackOff
    Last State:    Terminated
      Reason:      Error
      Message:     error: invalid canonical hostname: router-internal.*.myexampledomain.com
<---snip--->
Events:
  Type     Reason          Age                 From               Message
  ----     ------          ----                ----               -------
  Normal   Scheduled       114s                default-scheduler  Successfully assigned openshift-ingress/router-internal-56bdf6fcb5-qcsfm to ip-10-0-111-201.us-west-1.compute.internal
  Warning  FailedMount     114s                kubelet            MountVolume.SetUp failed for volume "default-certificate" : secret "router-certs-internal" not found
  Warning  FailedMount     114s                kubelet            MountVolume.SetUp failed for volume "metrics-certs" : secret "router-metrics-certs-internal" not found
  Normal   AddedInterface  112s                multus             Add eth0 [10.129.2.12/23] from ovn-kubernetes
  Normal   Pulled          25s (x5 over 112s)  kubelet            Container image "registry.build10.ci.openshift.org/ci-ln-3vpw9n2/stable@sha256:6baf21da47a043e562bd2507e9f54625aba81212b953fb92aeb1f94b11237930" already present on machine
  Normal   Created         25s (x5 over 112s)  kubelet            Created container: router
  Normal   Started         25s (x5 over 112s)  kubelet            Started container router
  Warning  BackOff         8s (x17 over 111s)  kubelet            Back-off restarting failed container router in pod router-internal-56bdf6fcb5-qcsfm_openshift-ingress(bae1eb79-86d0-484c-9a99-9b2ad7f74100)
```
  
I also want to know whether i did tested the intended way.
I can see the same error in the 4.19 cluster also whether the fix is not there...

Miciah:
> i create a ingress controller yaml file with a invalid `*.<domain>` domain. And create a controller which is created and the error only shown in the crashed ingress pod.
> Are we intended to reject the controller outright(without allowing to create the controller itself)?

Yes, if you create or update an IngressController CR with a wildcard in `spec.domain`, the create/update should be rejected outright by the API admission.  

> I also want to know whether i did tested the intended way.
> I can see the same error in the 4.19 cluster also whether the fix is not there...

The IngressController CRD is vendored into openshift/cluster-ingress-operator and copied into cluster-ingress-operator's CVO manifests, so you will need a PR in openshift/cluster-ingress-operator with the vendor bump and updated CVO manifests in order to test the updated validation.

operator/v1/tests/ingresscontrollers.operator.openshift.io/AAA_ungated.yaml
everettraven: This does look like a duplicate of the final test. Lets remove this one.

operator/v1/types_ingress.go
everettraven: To make sure I'm understanding correctly, we are saying here that the canonical hostname may not exceed 253 characters and as such the full string `router-{IngressController.metadata.name}.{IngressController.spec.domain}` must not exceed 253 characters.

Is this correct?

grzpiotrowski: > To make sure I'm understanding correctly, we are saying here that the canonical hostname may not exceed 253 characters and as such the full string `router-{IngressController.metadata.name}.{IngressController.spec.domain}` must not exceed 253 characters.
> 
> Is this correct?

That is correct. I tweaked the godoc to describe it a bit better now.

Edit: On another look, there's probably a better way to word this than what I did there. WDYT?

grzpiotrowski: Right, I was running into some issues when I tried to add a `message` to the `MaxLength` validation.
I did try:
`// +kubebuilder:validation:XValidation:rule="size(self) <= 239",message="total length of the router-<icName>.<domain> must not exceed 253 characters"`
but then I saw the `verify-crd-schema` failing on the budget.

I edited the godoc to say more clearly the 253 char limit and where the 239 `MaxLength` comes from.

operator/v1/types_ingress.go
```diff
@@ -68,6 +68,17 @@ type IngressControllerSpec struct {
...
+	// The domain value must be a valid DNS name. It must consist of lowercase
+	// alphanumeric characters, '-' or '.', and each label must start and end
+	// with an alphanumeric character and not exceed 63 characters.
+	//
+	// The total length of the domain is limited to 239 characters to account for
+	// the router name (e.g. "router-default.") that gets prepended when constructing
+	// the canonical hostname of which the total maximum length is 253 characters.
+	//
+	// +kubebuilder:validation:XValidation:rule="!format.dns1123Subdomain().validate(self).hasValue()",message="domain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character"
+	// +kubebuilder:validation:XValidation:rule="self.split('.').all(label, !format.dns1123Label().validate(label).hasValue())",message="each DNS label must not exceed 63 characters"
```
grzpiotrowski: Now in the long meantime, and through one of the test cases, I realized that the `!format.dns1123Subdomain()` itself ([link](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L219-L228)) was not checking each label of the domain. So on its own it  would allow a domain with a label exceeding 63 char limit.
For this reason I added the `!format.dns1123Label()` which checks that ([link](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L188-L203)).

operator/v1/tests/ingresscontrollers.operator.openshift.io/AAA_ungated.yaml
grzpiotrowski: Right, I removed it now and left the last test to cover this scenario.

operator/v1/types_ingress.go
everettraven: I do think there is some better wording we can put here to better reflect that, but something I am also wondering is if we want a higher-level validation that can take into account both the `.metadata.name` and `.spec.domain` of the `IngressController` for the validation here.

So maybe at the top-level we add a validation that is something like:
```
!format.dns1123Subdomain().validate('router-' + self.metadata.name + '.' + self.spec.domain)
```

Then we can make sure that the `router-{name}.{domain}` is always a valid hostname without having to juggle the edge case of `router-{name}.{domain}` exceeding 253 characters if each is validated separately?

everettraven: By "top-level" I mean on this struct: https://github.com/openshift/api/blob/58dfa6dd85faa1c61f44f502697d8fcbf06a345f/operator/v1/types_ingress.go#L38

operator/v1/types_ingress.go
JoelSpeed: Great catch! We haven't got this same validation on other uses, so we probably need to check those as well CC @everettraven keep this in mind when reviewing please

operator/v1/types_ingress.go
```diff
@@ -68,6 +68,17 @@ type IngressControllerSpec struct {
 	//
 	// If empty, defaults to ingress.config.openshift.io/cluster .spec.domain.
 	//
+	// The domain value must be a valid DNS name. It must consist of lowercase
+	// alphanumeric characters, '-' or '.', and each label must start and end
+	// with an alphanumeric character and not exceed 63 characters.
+	//
+	// The total length of the domain is limited to 239 characters to account for
+	// the router name (e.g. "router-default.") that gets prepended when constructing
+	// the canonical hostname of which the total maximum length is 253 characters.
```
candita: I would put it this way:  
```suggestion
	// The total length of the domain name is limited to 253 characters,  inclusive of the 
	// prepended prefix starting with "router-".
```

operator/v1/types_ingress.go
grzpiotrowski: By the looks of it this validation is too expensive to pass. I'm checking if it would be possible to validate just the combined length of the fields at the object level and then leave the label and subdomain validations on the spec.domain field.

JoelSpeed: Which specific rule are you referring to? Is it
```
!format.dns1123Subdomain().validate('router-' + self.metadata.name + '.' + self.spec.domain)
```

Is this because `self.metadata.name` doesn't have a bound in the API schema? We should be able to use kubebuilder markers (or a yaml patch) to set a length limit on it and make that validation happy

grzpiotrowski: yes, that's the rule I was trying out. I'm not exactly sure on the reason. I'll try to set the length limit on the name.

As a side question:
Is there a way of running only selected tests locally? I'm using `make integration` but it's taking quite some time if I'm not lucky with the randomization of the tests.

JoelSpeed: I think you'll have to set it using a manual CRD override (have a look for existing examples, its just a subset of the schema)

If you `make -C operator/v1 test` you should be able to run just a subset of the tests

operator/v1/types_ingress.go
candita: > The router name is "router-" (without the quotation marks) plus the name of the IngressController, so the minimum is 8 characters. The "default" router's canonical name is 14 chars, so 14 might be a reasonable amount to subtract from 253.

@Miciah Don't we also need to cover custom ingress controllers, which could have any canonical name?  Why subtract any set number of chars?

candita:
> @grzpiotrowski today i created a cluster using this PR and do a basic testing... i create a ingress controller yaml file with a invalid `*.<domain>` domain. And create a controller which is created and the error only shown in the crashed ingress pod. Are we intended to reject the controller outright(without allowing to create the controller itself)?
> 
> Otherwise i didnt see any noticeable difference with the changes.
> 
> ```
> mjoseph@mjoseph-mac Downloads % cat internal_ingress_controller.yaml
> kind: IngressController
> apiVersion: operator.openshift.io/v1
> metadata:
>   name: internal
>   namespace: openshift-ingress-operator
> spec:
>   domain: '*.myexampledomain.com'
>   replicas: 1
>   endpointPublishingStrategy:
>     loadBalancer:
>       scope: Internal
>     type: LoadBalancerService
> 
> mjoseph@mjoseph-mac Downloads % oc create -f internal_ingress_controller.yaml
> ingresscontroller.operator.openshift.io/internal created
> 
> mjoseph@mjoseph-mac Downloads % oc get po -n openshift-ingress                                        
> NAME                               READY   STATUS             RESTARTS        AGE
> router-default-665c98fb95-vn82n    1/1     Running            1 (3h24m ago)   3h32m
> router-default-665c98fb95-xcwp6    1/1     Running            1 (3h24m ago)   3h32m
> router-internal-56bdf6fcb5-qcsfm   0/1     CrashLoopBackOff   1 (12s ago)     14s
> 
> mjoseph@mjoseph-mac Downloads % oc get co
> NAME                                       VERSION                                                AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
> <---snip--->
> image-registry                             4.20.0-0-2025-08-07-081010-test-ci-ln-3vpw9n2-latest   True        False         False      3h23m   
> ingress                                    4.20.0-0-2025-08-07-081010-test-ci-ln-3vpw9n2-latest   True        True          False      3h25m   ingresscontroller "internal" is progressing: IngressControllerProgressing: One or more status conditions indicate progressing: DeploymentRollingOut=True (DeploymentRollingOut: Waiting for router deployment rollout to finish: 0 of 1 updated replica(s) are available......
> insights                                   4.20.0-0-2025-08-07-081010-test-ci-ln-3vpw9n2-latest   True        False         False      3h33m   
> kube-apiserver                             4.20.0-0-2025-08-07-081010-test-ci-ln-3vpw9n2-latest   True        False         False      3h29m   
> <---snip--->
> 
> mjoseph@mjoseph-mac Downloads % oc describe po -n openshift-ingress  router-internal-56bdf6fcb5-qcsfm 
> Name:                 router-internal-56bdf6fcb5-qcsfm
> <---snip--->
> Status:               Running
> IP:                   10.129.2.12
> IPs:
>   IP:           10.129.2.12
> Controlled By:  ReplicaSet/router-internal-56bdf6fcb5
> Containers:
>   router:
>     Container ID:  cri-o://e855d394bab095e1a5564bac065fe47c0de31341af804dda82a400487969df8d
>     Image:         registry.build10.ci.openshift.org/ci-ln-3vpw9n2/stable@sha256:6baf21da47a043e562bd2507e9f54625aba81212b953fb92aeb1f94b11237930
>     Image ID:      registry.build10.ci.openshift.org/ci-ln-3vpw9n2/stable@sha256:6baf21da47a043e562bd2507e9f54625aba81212b953fb92aeb1f94b11237930
>     Ports:         80/TCP, 443/TCP, 1936/TCP
>     Host Ports:    0/TCP, 0/TCP, 0/TCP
>     State:         Waiting
>       Reason:      CrashLoopBackOff
>     Last State:    Terminated
>       Reason:      Error
>       Message:     error: invalid canonical hostname: router-internal.*.myexampledomain.com
> <---snip--->
> Events:
>   Type     Reason          Age                 From               Message
>   ----     ------          ----                ----               -------
>   Normal   Scheduled       114s                default-scheduler  Successfully assigned openshift-ingress/router-internal-56bdf6fcb5-qcsfm to ip-10-0-111-201.us-west-1.compute.internal
>   Warning  FailedMount     114s                kubelet            MountVolume.SetUp failed for volume "default-certificate" : secret "router-certs-internal" not found
>   Warning  FailedMount     114s                kubelet            MountVolume.SetUp failed for volume "metrics-certs" : secret "router-metrics-certs-internal" not found
>   Normal   AddedInterface  112s                multus             Add eth0 [10.129.2.12/23] from ovn-kubernetes
>   Normal   Pulled          25s (x5 over 112s)  kubelet            Container image "registry.build10.ci.openshift.org/ci-ln-3vpw9n2/stable@sha256:6baf21da47a043e562bd2507e9f54625aba81212b953fb92aeb1f94b11237930" already present on machine
>   Normal   Created         25s (x5 over 112s)  kubelet            Created container: router
>   Normal   Started         25s (x5 over 112s)  kubelet            Started container router
>   Warning  BackOff         8s (x17 over 111s)  kubelet            Back-off restarting failed container router in pod router-internal-56bdf6fcb5-qcsfm_openshift-ingress(bae1eb79-86d0-484c-9a99-9b2ad7f74100)
> ```
> 
> I also want to know whether i did tested the intended way. I can see the same error in the 4.19 cluster also whether the fix is not there...

@melvinjoseph86 I believe you would need to test a cluster-ingress-operator PR that included this API change in the go.mod in order for it to work.  There isn't one yet.  @grzpiotrowski can you create one that temporarily includes this commit?

Edit: also mentioned in https://github.com/openshift/api/pull/2308#issuecomment-3186650833.

operator/v1/types_ingress.go
Miciah: We always add the "router-" prefix when setting the canonical host name:
```go
		cName := "router-" + ci.Name + "." + ci.Status.Domain
		env = append(env,
			corev1.EnvVar{Name: "ROUTER_DOMAIN", Value: ci.Status.Domain},
			corev1.EnvVar{Name: "ROUTER_CANONICAL_HOSTNAME", Value: cName},
		)
```
Source: https://github.com/openshift/cluster-ingress-operator/blob/69ba323f05c84f72f11a0d086836b3f3412512d4/pkg/operator/controller/ingress/deployment.go#L623-L626

operator/v1/types_ingress.go
grzpiotrowski: I gave this another shot. I was struggling a little with the validation budget, but the latest is a working version.

I added the validation on the `IngressController` object, which checks for the `'router-' + IC.name + '.' + IC.spec.domain` max length of 253 characters.

Now I also needed to add a length limit on the ingress controller name, for the validation budget check to pass.

I left the use of `format.dns1123Subdomain().validate()` and `format.dns1123Label().validate()` functions on the `.spec.domain` field also for the validation budget reason, as combining it with 'router-' + IC.name + '.' + IC.spec.domain` at the IC object level was also not fulfilling it.

Since the .spec.domain field also needed a length limit to pass, I added a:
`kubebuilder:validation:MaxLength=253` there, but as a consequence this locked me to the use of the generic `Too long: may not be more than 253 bytes` message. Which could be slightly confusing, as the user would see it even if their `.spec.domain` is less than 253 characters, but combined with the router name would exceed it.

So that's why I'm not 100% happy with it in terms of the message the validation gives in such case. As it is the generic `Too long: may not be more than 253 bytes` coming from the `// +kubebuilder:validation:MaxLength=253`, which as far as I know cannot be customized, but I'd be happy to learn if it can.

Perhaps there is another way of doing it. There is a bit of a not-that-good-looking complexity here.
Could you please advise on how to approach it better?

Edit:
I got `ERROR: (v1) ^.spec.domain - maxLength: maximum constraint added when there was none previously : 253` which is not ideal.

operator/v1/types_ingress.go
grzpiotrowski: Thank you. With the latest push I added a revised godoc, which also mentions that the validation is in a way implementation dependent.

operator/v1/types_ingress.go
candita: /cc @Miciah

grzpiotrowski: Just to add, I revised the <=253 char validation message at the IngressController level a little. I was under the wrong impression that it cannot be different from the maxLength messsage on the .spec.domain field validation.

So part of the previous comment is outdated then.

In this case:
If one tries to put the `.spec.domain` value longer than 253 characters, they will see:
>Too long: may not be more than 253 bytes

coming from the maxLength validation on that field.

However if the`.spec.domain` field is under 253 characters but combined with the `'router-' + IC.name + '.' + IC.spec.domain` exceeds 253 characters, they will see:

>The combined 'router-' + IngressController.Name + '.' + .spec.domain cannot exceed 253 characters

This is covered by these two test cases: https://github.com/grzpiotrowski/api/blob/2aa449e1919bab6749da7b47104e8bb0a007b7dd/operator/v1/tests/ingresscontrollers.operator.openshift.io/AAA_ungated.yaml#L607-L626

operator/v1/types_ingress.go
```diff
@@ -35,11 +35,13 @@ import (
 //
 // Compatibility level 1: Stable within a major release for a minimum of 12 months or 3 minor releases (whichever is longer).
 // +openshift:compatibility-gen:level=1
+// +kubebuilder:validation:XValidation:rule="!has(self.spec.domain) || size('router-' + self.metadata.name + '.' + self.spec.domain) <= 253",message="The combined 'router-' + IngressController.Name + '.' + .spec.domain cannot exceed 253 characters"
 type IngressController struct {
 	metav1.TypeMeta `json:",inline"`
 
 	// metadata is the standard object's metadata.
 	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
+	// // +kubebuilder:validation:XValidation:rule="!has(self.name) || size(self.name) <= 63",message="metadata.name must not exceed 63 characters"
```
JoelSpeed: Is this correct? What about existing objects? Are they not validated as a DNS1123 subdomain?

JoelSpeed: I assume you meant?
```suggestion
	// +kubebuilder:validation:XValidation:rule="!has(self.name) || size(self.name) <= 63",message="metadata.name must not exceed 63 characters"
```

operator/v1/types_ingress.go
```diff
@@ -68,6 +70,20 @@ type IngressControllerSpec struct {
...
+	//
+	// The implementation may add a prefix such as "router-default-" to the domain
+	// when constructing the router canonical hostname. To ensure the resulting
+	// hostname does not exceed the DNS maximum length of 253 characters,
+	// the domain length additionally validated at the IngressController object
+	// level.
+	//
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:XValidation:rule="!format.dns1123Subdomain().validate(self).hasValue()",message="domain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character"
+	// +kubebuilder:validation:XValidation:rule="self.split('.').all(label, !format.dns1123Label().validate(label).hasValue())",message="each DNS label must not exceed 63 characters"
```
JoelSpeed: Does validating against the dns1123Label gain us anything here? Anything that isn't a valid label (apart from length) is validated in the previous rule. So here we just need to check that each label segment has length `<= 63` right?

operator/v1/types_ingress.go
grzpiotrowski: Right.. that I'd call a happy mistake that slipped in from my testing.
I'm a little surprised now, that without this limit on the IC name, we are within the validation budget. In the previous combinations of validations I did not see it passing without this in place.

JoelSpeed: Adding a rule to limit the size of the name won't affect the CEL estimated worst cost. You need to have the maxlength on the schema itself, which can be done today with a manual schema patch if we are hitting issues

operator/v1/types_ingress.go
grzpiotrowski: Right, I agree that it could be replaced with just a length check for each label, which would be cheaper.

operator/v1/types_ingress.go
grzpiotrowski: Removed it as it was not needed.

grzpiotrowski: I removed this rule, as it was not needed.

operator/v1/types_ingress.go
grzpiotrowski: Changed the label validation to only check for the max length of 63 chars, since the regex check is done in `format.dns1123Subdomain()` already as you mentioned.

operator/v1/types_ingress.go
```diff
@@ -68,6 +69,22 @@ type IngressControllerSpec struct {
...
+	// when constructing the router canonical hostname. To ensure the resulting
+	// hostname does not exceed the DNS maximum length of 253 characters,
+	// the domain length is additionally validated at the IngressController object
+	// level. For the maximum length of the domain value itself, the shortest
+	// possible variant of the prefix and the ingress controller name was considered
+	// for example "router-a."
+	//
+	// +kubebuilder:validation:MaxLength=244
+	// +kubebuilder:validation:XValidation:rule="!format.dns1123Subdomain().validate(self).hasValue()",message="domain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character"
+	// +kubebuilder:validation:XValidation:rule="self.split('.').all(label, size(label) <= 63)",message="each DNS label must not exceed 63 characters"
```
grzpiotrowski, line 87: So to summarize so far.
The current approach I have is to validate at the level of the `IngressController` struct, to check for the length (253 max) of the combined domain there, since the ingress controller name can vary, impacting how long the `.spec.domain` can be in consequence.

I have two validations of the dns domain on the `.spec.domain` field.
One that uses `format.dns1123Subdomain()` and the other to check if each label does not exceed 63 chars limit, since the function above doesn't include this.

I needed to set some max length on the `.spec.domain` field, as to satisfy the validation budget. It is (I think) somewhat superficial, since the real length restriction is checked including the `IC.name`. I set it to 244 chars, which would be the shortest `IC.name` (1 char), resulting in the `router-a.`  prefix (9 chars) for the router's canonical domain. Hence 244 chars max for the `.spec.domain`
Though the `format.dns1123Subdomain()` also checks for the max length of 253 chars, but this wouldn't come into play here.

operator/v1/types_ingress.go
```diff
@@ -35,6 +35,7 @@ import (
 //
 // Compatibility level 1: Stable within a major release for a minimum of 12 months or 3 minor releases (whichever is longer).
 // +openshift:compatibility-gen:level=1
+// +kubebuilder:validation:XValidation:rule="!has(self.spec.domain) || size('router-' + self.metadata.name + '.' + self.spec.domain) <= 253",message="The combined 'router-' + IngressController.Name + '.' + .spec.domain cannot exceed 253 characters"
```
JoelSpeed: Probably easier to understand as metadata.name no? Since that's the first we are checking?
```suggestion
// +kubebuilder:validation:XValidation:rule="!has(self.spec.domain) || size('router-' + self.metadata.name + '.' + self.spec.domain) <= 253",message="The combined 'router-' + metadata.name + '.' + .spec.domain cannot exceed 253 characters"
```

grzpiotrowski: thank you, done now.

grzpiotrowski:
About `verify-crdify`
```bash
ERROR: (v1) ^.spec.domain - maxLength: maximum constraint added when there was none previously : 244
```

Is this something to worry about or override?

everettraven:
> About `verify-crdify`
> 
> ```shell
> ERROR: (v1) ^.spec.domain - maxLength: maximum constraint added when there was none previously : 244
> ```
> 
> Is this something to worry about or override?

In this case, it is likely overridable. IIRC this is adding validation for something that would have never been valid in the first place.

grzpiotrowski:
Tested manually in a cluster:

```bash
$ oc create -f -<<'EOF'
kind: IngressController
metadata:
  name: internal
  namespace: openshift-ingress-operator
spec:
  domain: '*.myexampledomain.com'
  replicas: 1
  endpointPublishingStrategy:
    loadBalancer:
      scope: Internal
    type: LoadBalancerService
EOF
The IngressController "internal" is invalid: spec.domain: Invalid value: "string": domain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character
```

```bash
$ oc create -f -<<'EOF'                                                            
kind: IngressController
metadata:
  name: ic-name-containing-exactly-40-characters
  namespace: openshift-ingress-operator
spec:
  domain: 'this-domain.has-208-characters.which-on.its-own-would-not-exceed.the-limit-of-253-chars.but-combined-with-the-ingress-controller-name.with-40-chars.and-the-router-prefix.ends-up-as-a-too-long.canonical-domain'
  replicas: 1
  endpointPublishingStrategy:
    loadBalancer:
      scope: Internal
    type: LoadBalancerService
EOF
The IngressController "ic-name-containing-exactly-40-characters" is invalid: <nil>: Invalid value: "object": The combined 'router-' + metadata.name + '.' + .spec.domain cannot exceed 253 characters
```

```bash
$ oc create -f -<<'EOF'
kind: IngressController
metadata:
  name: a       
  namespace: openshift-ingress-operator
spec:
  domain: 'this-domain.has-exactly-245-characters.for-the-purpose-of-testing.the-spec-domain-field-length-validation.it-exceeds-the-limit-set-on-the-spec-domain-field.by-one-character-to-test-the-error.message-from-this-validation.otherwise-it-is-valid.com'
  replicas: 1
  endpointPublishingStrategy:
    loadBalancer:
      scope: Internal
    type: LoadBalancerService
EOF
The IngressController "a" is invalid: 
* spec.domain: Too long: may not be more than 244 bytes
* <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation
```

```bash
$ oc create -f -<<'EOF'                                                            
kind: IngressController
metadata:
  name: internal
  namespace: openshift-ingress-operator
spec:
  domain: 'foo.this-label-exceeds-63-characters-for-the-purpose-of-testing-the-domain-validation.com'
  replicas: 1
  endpointPublishingStrategy:
    loadBalancer:
      scope: Internal
    type: LoadBalancerService
EOF                          
The IngressController "internal" is invalid: spec.domain: Invalid value: "string": each DNS label must not exceed 63 characters
```

```bash
$ oc create -f -<<'EOF'                                                            
kind: IngressController
metadata:
  name: ic-name-containing-exactly-40-characters
  namespace: openshift-ingress-operator
spec:
  domain: '*.this-domain.has-210-characters.which-on.its-own-would-not-exceed.the-limit-of-253-chars.but-combined-with-the-ingress-controller-name.with-40-chars.and-the-router-prefix.ends-up-as-a-too-long.canonical-domain'
  replicas: 1
  endpointPublishingStrategy:
    loadBalancer:
      scope: Internal
    type: LoadBalancerService
EOF
The IngressController "ic-name-containing-exactly-40-characters" is invalid: 
* <nil>: Invalid value: "object": The combined 'router-' + metadata.name + '.' + .spec.domain cannot exceed 253 characters
* spec.domain: Invalid value: "string": domain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character
```

```bash
$ oc create -f -<<'EOF'
kind: IngressController
metadata:
  name: internal                                
  namespace: openshift-ingress-operator
spec:
  domain: 'exampledomain.com'
  replicas: 1                               
  endpointPublishingStrategy:
    loadBalancer:
      scope: Internal
    type: LoadBalancerService
EOF                          
ingresscontroller.operator.openshift.io/internal created
```

melvinjoseph86:
Tested along with https://github.com/openshift/cluster-ingress-operator/pull/1295.
the fix is working and the ingress controller is creating with wrong values in .spec.domain

mjoseph@mjoseph-mac Downloads % oc get clusterversion
NAME      VERSION                                                AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.21.0-0-2025-10-17-143658-test-ci-ln-gk30bwk-latest   True        False         45m     Cluster version is 4.21.0-0-2025-10-17-143658-test-ci-ln-gk30bwk-latest

mjoseph@mjoseph-mac Downloads % oc create -f - <<EOF
heredoc> kind: IngressController
metadata:
  name: internal
  namespace: openshift-ingress-operator
spec:
  domain: '*.myexampledomain.com'
  replicas: 1
  endpointPublishingStrategy:
    loadBalancer:
      scope: Internal
    type: LoadBalancerService
heredoc> EOF
The IngressController "internal" is invalid: spec.domain: Invalid value: "string": domain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character
mjoseph@mjoseph-mac Downloads % 

mjoseph@mjoseph-mac Downloads % oc create -f -<<'EOF'                                                            
kind: IngressController
metadata:
  name: ic-name-containing-exactly-40-characters
  namespace: openshift-ingress-operator
spec:
  domain: 'this-domain.has-208-characters.which-on.its-own-would-not-exceed.the-limit-of-253-chars.but-combined-with-the-ingress-controller-name.with-40-chars.and-the-router-prefix.ends-up-as-a-too-long.canonical-domain'
  replicas: 1
  endpointPublishingStrategy:
    loadBalancer:
      scope: Internal
    type: LoadBalancerService
EOF
The IngressController "ic-name-containing-exactly-40-characters" is invalid: <nil>: Invalid value: "object": The combined 'router-' + metadata.name + '.' + .spec.domain cannot exceed 253 characters
mjoseph@mjoseph-mac Downloads %


mjoseph@mjoseph-mac Downloads % oc create -f -<<'EOF'
kind: IngressController
metadata:
  name: a
  namespace: openshift-ingress-operator
spec:
  domain: 'this-domain.has-exactly-245-characters.for-the-purpose-of-testing.the-spec-domain-field-length-validation.it-exceeds-the-limit-set-on-the-spec-domain-field.by-one-character-to-test-the-error.message-from-this-validation.otherwise-it-is-valid.com'
  replicas: 1
  endpointPublishingStrategy:
    loadBalancer:
      scope: Internal
    type: LoadBalancerService
EOF
The IngressController "a" is invalid: 
* spec.domain: Too long: may not be more than 244 bytes
* <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation
mjoseph@mjoseph-mac Downloads % 
mjoseph@mjoseph-mac Downloads % 


mjoseph@mjoseph-mac Downloads % oc create -f -<<'EOF'                                                            
kind: IngressController
metadata:
  name: internal
  namespace: openshift-ingress-operator
spec:
  domain: 'foo.this-label-exceeds-63-characters-for-the-purpose-of-testing-the-domain-validation.com'
  replicas: 1
  endpointPublishingStrategy:
    loadBalancer:
      scope: Internal
    type: LoadBalancerService
EOF
The IngressController "internal" is invalid: spec.domain: Invalid value: "string": each DNS label must not exceed 63 characters
mjoseph@mjoseph-mac Downloads % 
mjoseph@mjoseph-mac Downloads % 

mjoseph@mjoseph-mac Downloads % oc create -f -<<'EOF'                                                            
kind: IngressController
metadata:
  name: ic-name-containing-exactly-40-characters
  namespace: openshift-ingress-operator
spec:
  domain: '*.this-domain.has-210-characters.which-on.its-own-would-not-exceed.the-limit-of-253-chars.but-combined-with-the-ingress-controller-name.with-40-chars.and-the-router-prefix.ends-up-as-a-too-long.canonical-domain'
  replicas: 1
  endpointPublishingStrategy:
    loadBalancer:
      scope: Internal
    type: LoadBalancerService
EOF
The IngressController "ic-name-containing-exactly-40-characters" is invalid: 
* <nil>: Invalid value: "object": The combined 'router-' + metadata.name + '.' + .spec.domain cannot exceed 253 characters
* spec.domain: Invalid value: "string": domain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character
mjoseph@mjoseph-mac Downloads % 

/label qe-approved
/verified by @mjoseph

JoelSpeed (commented):
/lgtm

JoelSpeed:

The complaint is about adding a new maximum length constraint, but that is a property that ratchets, so I'm happy we are ok doing this

CC @everettraven we might want to teach crdify to ignore (or have an option to ignore) ratchetable changes

PR #2346, opened 2025-05-28
OCPBUGS-56773: disable psa for 4.19

ibihim:
## What

Disable PSA feature.

## Why

We did not yet prepare the rollout of PSA enforcement.


deads2k:
I'd expect the bug to also link the peer of https://github.com/openshift/hypershift/pull/5566.  That's this line: https://github.com/openshift/hypershift/blob/release-4.19/control-plane-operator/controllers/hostedcontrolplane/kas/config.go#L119

How would HCP honor this without that change?

ibihim:
Added a reference to the HyperShift PR.

PR #2350, opened 2025-05-30
Remove DataGather state print column

opokornyy:


everettraven (approved):
IIRC this field was removed in favor of standard status conditions to avoid a `Phase`-like field.

/lgtm

PR #2396, opened 2025-07-08
MCO-1805: MCO-1806: Add ManagedBootImagesCPMS feature gate & CPMS type to ManagedBootImages API

djoshy:
This PR adds:

- A new feature gate for [MCO-1007](https://issues.redhat.com/browse/MCO-1007); whose goal is to support boot image updates to `ControlPlaneMachineSets`
- A new `MachineManagerMachineSetsResourceType` enum for CPMS so they can be opted in for updates via the MachineConfiguration API object.


operator/v1/types_machineconfiguration.go
```diff
@@ -194,12 +195,14 @@ const (
...
 // type to be registered.
-// +kubebuilder:validation:Enum:="machinesets"
+// +kubebuilder:validation:Enum:="machinesets";"controlplanemachinesets"
 type MachineManagerMachineSetsResourceType string
 
 const (
 	// MachineSets represent the MachineSet resource type, which manage a group of machines and belong to the Openshift machine API group.
 	MachineSets MachineManagerMachineSetsResourceType = "machinesets"
+	// ControlPlaneMachineSets represent the ControlPlaneMachineSets resource type, which manage a group of control-plane machines and belong to the Openshift machine API group.
+	ControlPlaneMachineSets MachineManagerMachineSetsResourceType = "controlplanemachinesets"
```
djoshy, line 441: Question: Is there a way to only enable this value of enum on the feature gate?

everettraven: Yes there is. You would change the usage of `+kubebuilder:validation:Enum` to the following:
- `+openshift:validation:FeatureGateAwareEnum:featureGate="",enum="machinesets"`
- `+openshift:validation:FeatureGateAwareEnum:featureGate="ManagedBootImagesCPMS",enum="machinesets";"controlplanemachinesets"`

djoshy: Thanks, I've updated the PR. PTAL when you get a chance (:

features/features.go
```diff
@@ -393,6 +393,14 @@ var (
 						enableIn(configv1.DevPreviewNoUpgrade, configv1.TechPreviewNoUpgrade).
 						mustRegister()
 
+	FeatureGateManagedBootImagesCPMS = newFeatureGate("ManagedBootImagesCPMS").
+						reportProblemsToJiraComponent("MachineConfigOperator").
+						contactPerson("djoshy").
+						productScope(ocpSpecific).
+						enhancementPR("https://github.com/openshift/enhancements/pull/1761").
```
JoelSpeed: The linked EP explicitly calls out not targeting CPMS. Has there been design discussion of the impacts of enabling boot image updates on CPMS?

operator/v1/tests/machineconfigurations.operator.openshift.io/ManagedBootImages+ManagedBootImagesCPMS.yaml
```diff
@@ -0,0 +1,119 @@
...
+        spec:
+          logLevel: Normal
+          operatorLogLevel: Normal
+    - name: Should be able to create an empty ManagedBootImages configuration knob
+      initial: |
+        apiVersion: operator.openshift.io/v1
+        kind: MachineConfiguration
+        spec:
+          managedBootImages:
+            machineManagers: []
```
JoelSpeed, line 26: Why do we allow this?

operator/v1/tests/machineconfigurations.operator.openshift.io/ManagedBootImages+ManagedBootImagesCPMS.yaml
```diff
@@ -0,0 +1,119 @@
...
+                  mode: All
+      expected: |
+        apiVersion: operator.openshift.io/v1
+        kind: MachineConfiguration
+        spec:
+          logLevel: Normal
+          operatorLogLevel: Normal
+          managedBootImages:
+            machineManagers:
+              - resource: controlplanemachinesets
```
JoelSpeed, line 54: CPMS is a singleton within the cluster, perhaps we want to validate a specific selection (All?) to be required when this value is `controlplanemachinesets`?

features/features.go
djoshy: Ah, yes I can update that, will open a PR for it so the reference here can be corrected. We were asked by [Service Delivery folks to bump the priority](https://issues.redhat.com/browse/MCO-1007?focusedId=27526241&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27526241) for this and we had initially this slated for TechPreview in 4.21. Some recent developments [pushed Azure to 4.21](https://issues.redhat.com/browse/MCO-1672?focusedId=27560698&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27560698), so I decided to pull this into 4.20. Since CPMS do not use [marketplace AMIs/images](https://issues.redhat.com/browse/MCO-1007?focusedId=27527179&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27527179), this should be hopefully just re-using a lot of the existing implementaton for GCP/AWS management.

operator/v1/tests/machineconfigurations.operator.openshift.io/ManagedBootImages+ManagedBootImagesCPMS.yaml
djoshy: I think our original idea was that it would improve discovery: https://github.com/openshift/api/pull/1672#discussion_r1457537681 

Currently, it is used to explicitly disable updates in 4.18, so an auto opt-in does not take place on upgrade to 4.19.

operator/v1/tests/machineconfigurations.operator.openshift.io/ManagedBootImages+ManagedBootImagesCPMS.yaml
djoshy: Oh, interesting, I did not know that! Yes, I can update the validation here. It will also simplify the reconciliation loop in the MCO controller.

features/features.go
JoelSpeed: I'm sure from your side it's easy to get the CPMS updated, but there's a big difference between CPMS and MachineSets that needs to be discussed, primarily, that when you update the CPMS, it _could_ trigger a complete control plane replacement, that is potentially not desirable depending on when it happens, or even, at all in some cases. I think this is worth bringing to an architecture call, and perhaps even bringing some SD opinionated folks along

operator/v1/tests/machineconfigurations.operator.openshift.io/ManagedBootImages+ManagedBootImagesCPMS.yaml
JoelSpeed: So you are making a distinction now between omitted and the empty list? The API wasn't designed with this in mind and I'm not sure how you'd actually be doing that?

features/features.go
djoshy: Thanks for the context, I agree with your concerns. I will be happy to bring it to the next arch call.

JoelSpeed: I've gone into more detail on the issue you linked, hoping to trigger some discussion with SD, lets see if they respond

djoshy: ack, thanks!

operator/v1/tests/machineconfigurations.operator.openshift.io/ManagedBootImages+ManagedBootImagesCPMS.yaml
djoshy: > The API wasn't designed with this in mind and I'm not sure how you'd actually be doing that?

Yeah, it's not pretty :sweat:  and it is only meant as stop-gap < 4.18 since we have an explicit `None` option in 4.19+. I check if [the spec list exists](https://github.com/openshift/machine-config-operator/blob/c3d7b9bf94f0b6190054fefa74900cf4337dd421/pkg/operator/sync.go#L2255), if omitted, the list object would be nil and the MCO considers that to be no opinion.

JoelSpeed: What you are trying to achieve doesn't work. Go decoding/encoding won't tell the difference between a persisted `[]` and the field being omitted completely. Take a look at the output of https://go.dev/play/p/xEYwvCwxqB3.

If you wanted to be able to tell the difference between those two states, you'd need the list to be a pointer (`*[]T`). 

As soon as a structured client writes to the object after the use has persisted `[]`, it will be stripped away again.

JoelSpeed: Oh wait, we don't have `omitempty`... that changes it slightly, but damn that is sketchy and fragile 👀 This is not a behaviour I would be comfortable relying on. Kubernetes doesn't have a concept of pointers, it has present, or not present. Lists have size generally, and we should not assume an empty list round trips.

[Sketchy playground](https://go.dev/play/p/1aZh1LBGx4Q)

djoshy: Understood, given that the empty list method is now in use by 4.18 ROSA/Managed clusters, what would you suggest is the path forward here? Our 4.18 docs recommend the empty list for disabling prior to an upgrade, and 4.19 docs recommend the None option. Should we do some sort of migration?

JoelSpeed: What would backporting the none option look like? And then updating the guidance to recommend against that?

Have you had any issues of anyone reporting problems yet? I guess the problems would occur on the 4.18 to 4.19 upgrade boundary right?

djoshy: > What would backporting the none option look like? And then updating the guidance to recommend against that?
- Adding the `None` value to the enum in the 4.18 API, so we'll need a 4.18.z bug.
- Update Openshift 4.18 docs to mention the addition of the `None` value.
- Ask SRE folks currently using empty list method to transition to the new enum value after upgrading. (This is potentially painful?)
- Add an upgrade guard for 4.19 that checks for the empty list and blocks upgrades to 4.19 if they're using the empty list instead of using an explicit `None` enum. We'd need to add an upgrade edge as well to enforce this check.

> Have you had any issues of anyone reporting problems yet? I guess the problems would occur on the 4.18 to 4.19 upgrade boundary right?

No, we haven't seen any problems so far in our upgrade testing or from our customers using the empty list method. I don't know if an upgrade would cause the issue necessarily: both the 4.18 and 4.19 controllers are currently interpreting the empty list as boot image updates are disabled(no `machineManagers` defined); only 4.19 explicitly checks for the `None` value. Thinking a bit more about it, I think 4.18 will work with the `None` option too without any additional MCO side changes. Our default is to return an empty selector [when any of our expected 4.18 enums](https://github.com/openshift/machine-config-operator/blob/release-4.18/pkg/controller/machine-set-boot-image/helpers.go#L89)(All/Partial) aren't matched.

I think this only becomes an issue if we lose the empty list between calls per your earlier comment.

djoshy:
With https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/368 merged, we should hopefully have all the prerequisites to proceed with this feature. I've rebased this PR to fix merge conflicts, and I have also added a new rule to only allow `All` or `None` for the `ControlPlaneMachineSet` resources.

operator/v1/types_machineconfiguration.go
```diff
@@ -172,15 +172,17 @@ type ManagedBootImages struct {
 	// +listMapKey=resource
 	// +listMapKey=apiGroup
 	// +kubebuilder:validation:MaxItems=5
+	// +kubebuilder:validation:XValidation:rule="self.exists(x, x.resource=='controlplanemachinesets') ? self.exists(x, x.selection.mode == 'All' || x.selection.mode == 'None'):true", message="Only All or None selection mode is permitted for ControlPlaneMachineSets"
```
djoshy: I tried to make this rule feature gate aware, but it seemed to be failing my integration test that guards against the Partial mode. Any ideas what I'm missing here?

I guess it's not 100% necessary since the enum itself is guarded with the feature gate, but would be nice to keep the rules consistent.

operator/v1/tests/machineconfigurations.operator.openshift.io/ManagedBootImages+ManagedBootImagesCPMS.yaml
djoshy: Left a question about this here: https://github.com/openshift/api/pull/2396#discussion_r2360629031

operator/v1/tests/machineconfigurations.operator.openshift.io/ManagedBootImages+ManagedBootImagesCPMS.yaml
djoshy: Sorry, this got lost a little bit, should we do any remediation here? We haven't seen any issues with it so far, but its still early days for 4.19, I guess.

features/features.go
```diff
@@ -379,6 +379,14 @@ var (
 						enableIn(configv1.DevPreviewNoUpgrade, configv1.TechPreviewNoUpgrade).
 						mustRegister()
 
+	FeatureGateManagedBootImagesCPMS = newFeatureGate("ManagedBootImagesCPMS").
+						reportProblemsToJiraComponent("MachineConfigOperator").
+						contactPerson("djoshy").
+						productScope(ocpSpecific).
+						enhancementPR("https://github.com/openshift/enhancements/pull/1818").
```
JoelSpeed, line 369: Shouldn't be tagged to an unmerged closed EP, merged or unmerged and open is ok

operator/v1/types_machineconfiguration.go
JoelSpeed: Can you update the documentation on the selection field to also explain this

JoelSpeed: You can move this rule to the MachineManager struct itself, it would make more sense there, you could also avoid the ternary by negating the first clause, either works though, whichever you think is more readable
```suggestion
	// +kubebuilder:validation:XValidation:rule="self.resource != 'controlplanemachinesets') || x.selection.mode == 'All' || x.selection.mode == 'None')", message="Only All or None selection mode is permitted for ControlPlaneMachineSets"
```

JoelSpeed: This should still be gated, can you add it back and push and then we can check that the files are generating correctly and see what's up with the integration tests

operator/v1/tests/machineconfigurations.operator.openshift.io/ManagedBootImages+ManagedBootImagesCPMS.yaml
JoelSpeed: I'd prefer to fix it if we can reason that it isn't going to break anyone

features/features.go
djoshy: reopened, will attempt to update once we have finalized API here

operator/v1/tests/machineconfigurations.operator.openshift.io/ManagedBootImages+ManagedBootImagesCPMS.yaml
djoshy: ack, opened a bug for this: https://issues.redhat.com/browse/OCPBUGS-61994

operator/v1/types_machineconfiguration.go
djoshy: Thanks, moving it makes a lot of sense! Yeah, I was between the two, but I think I'll go the negation path, it is a bit more readable after moving the rule down a level.

djoshy: I updated the docs under `selection.mode` since that seemed like a better spot, but I can pull it up to  `selection` if that is more appropriate!

djoshy: I added the rule change as [separate commit](https://github.com/openshift/api/pull/2396/commits/93e2839af8497d61e8f6bebaf5b6a02047426961#diff-02dfc16d777f639471da62a47ecf8eda210c2bec7a3a3378febf9810d6870782R180), hope that helps!

operator/v1/types_machineconfiguration.go
```diff
@@ -177,10 +177,12 @@ type ManagedBootImages struct {
 
 // MachineManager describes a target machine resource that is registered for boot image updates. It stores identifying information
 // such as the resource type and the API Group of the resource. It also provides granular control via the selection field.
+// +openshift:validation:FeatureGateAwareXValidation:featureGate=ManagedBootImagesCPMS,rule="self.resource != 'controlplanemachinesets' || self.selection.mode == 'All' || self.selection.mode == 'None'", message="Only All or None selection mode is permitted for ControlPlaneMachineSets"
```
JoelSpeed: For some reason the generator isn't adding this to the partial manifests, that's why the tests are failing at the moment 🤔

djoshy: The feature gate is created in this PR(generation for that was run in an earlier commit), that wouldn't cause an issue, would it?

JoelSpeed: Seems not, the `verify` passed which means the generation is up to date. So something isn't quite right here with the incantation (or perhaps there's a bug in the implementation)

JoelSpeed: This gate also relies on `ManagedBootImages`, since it's completely contained inside the `ManagedBootImages` gate. Switch to this (and similar on the enum)
```suggestion
// +openshift:validation:FeatureGateAwareXValidation:requiredFeatureGate=ManagedBootImages;ManagedBootImagesCPMS,rule="self.resource != 'controlplanemachinesets' || self.selection.mode == 'All' || self.selection.mode == 'None'", message="Only All or None selection mode is permitted for ControlPlaneMachineSets"
```

djoshy: Done, seems to be good now. I had to update the test file to include both as well.

djoshy:
/verified later

We'll verify this through https://github.com/openshift/machine-config-operator/pull/5332, which updates the MCO to consume this API.

PR #2410, opened 2025-07-18
Fix kubebuilder tools/envtest index for v1.33.2 (add v prefix)

damdo:
- first commit rectifies issue with the previous https://github.com/openshift/api/pull/2407 PR by adding a `v` prefix version in the index
- second commit adds requirement about the `v` prefix in the documentation + a check in the tool to ensure it is specified.

Thanks @theobarberbany for the report.

### DIFF
--- a/tools/publish-kubebuilder-tools/main.go
+++ b/tools/publish-kubebuilder-tools/main.go
@@ -44,6 +44,11 @@ func main() {
 		panic("output is required")
 	}
 
+	// We only expect Kubernetes versions that begin with the v prefix.
+	if !strings.HasPrefix(*version, "v") {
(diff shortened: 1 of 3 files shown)


tools/publish-kubebuilder-tools/main.go
```diff
@@ -44,6 +44,11 @@ func main() {
 		panic("output is required")
 	}
 
+	// We only expect Kubernetes versions that begin with the v prefix.
+	if !strings.HasPrefix(*version, "v") {
```
theobarberbany, line 48: We love validation <3

theobarberbany (commented):
/lgtm
/approve

PR #1844, opened 2024-04-02
ETCD-488: Move etcd hardware speed feature to GA

dusk125:

PR #2416, opened 2025-07-23
OCPBUGS-55661: Remove HardwareSpeed featuregate

tjungblu:
introduced in #1844, removing it again as it is the default since 4.18. There was no reference in CEO for that feature gate in the first place, so there was no removal PR on the other side.


operator/v1/zz_generated.crd-manifests/0000_12_etcd_01_etcds-Default.crd.yaml
```diff
@@ -46,19 +46,6 @@ spec:
             type: object
           spec:
             properties:
-              controlPlaneHardwareSpeed:
```
tjungblu: this certainly looks wrong, we still require the string field to be there in the CRD

tjungblu: fixed

geliu2016 (approved):
/label cherry-pick-approved

PR #2421, opened 2025-07-24
no-jira: Rename some Installer feature gates

patrickdillon:
This PR renames a few installer feature gates we are looking to promote soon. Installer feature gates should have "install" in the title so that [the test analyzer will analyze install success--not a specific e2e test.](https://github.com/openshift/api/blob/master/tools/codegen/cmd/featuregate-test-analyzer.go#L565-L568). We forgot that convention when creating these feature gates, so we need to rename them.


patrickdillon:
cc @barbacbd @sadasu @jianlinliu @gpei @yunjiang29

patrickdillon:
/hold

joel suggested I try instead the approach in #2423

patrickdillon:
/close
in favor of #2423

PR #2423, opened 2025-07-24
features: add installer replacement feature gates

patrickdillon:
Installer feature gates need to have "install" in the title for the test analyzer. This commit adds feature gates for AWSClusterHostedDNSInstall, GCPClusterHostedDNSInstall, and GCPCustomAPIEndpointsInstall to replace their counterparts without the -install suffix, which will be removed in a follow up PR.


sadasu:
https://github.com/openshift/api/pull/2300 depends on this and has to be rebased to use the new feature gate name after this PR merges.

PR #2424, opened 2025-07-28
Support AMD SEV-SNP on AWS

fangge1212:
AMD SEV-SNP is one of the confidential computing technologies. This commit adds support for AMD SEV-SNP on AWS, so users can utilize the confidential computing on the cluster nodes.

Upstream CAPA PR: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5605


machine/v1beta1/types_awsprovider.go
```diff
@@ -17,6 +17,9 @@ type AWSMachineProviderConfig struct {
 	AMI AWSResourceReference `json:"ami"`
 	// instanceType is the type of instance to create. Example: m4.xlarge
 	InstanceType string `json:"instanceType"`
+	// cpuOptions is the set of cpu options for the instance, where you can enable/disable AMD SEV-SNP on the instance.
+	// +optional
+	CpuOptions *CpuOptions `json:"cpuOptions,omitempty"`
 	// tags is the set of tags to add to apply to an instance, in addition to the ones
```
bgartzi, line 27: Hi @fangge1212, would it make sense to use a field named `ConfidentialComputing` (or something similar) that could be `Disabled` or `AmdSevSnp` instead?

In case AWS supports other confidential computing technologies in the future, that would be easier to extend than this approach IMO. As it would require adding another new field to the API instead of a new value to the enum.

fangge1212: Same as https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5598, waiting for more opinions before chaning it

machine/v1beta1/types_awsprovider.go
```diff
@@ -109,6 +112,31 @@ type AWSMachineProviderConfig struct {
...
+const (
+	// AmdSevSnpSpecificationEnabled means AMD SEV SNP is enabled for the instance.
+	AmdSevSnpSpecificationEnabled AmdSevSnpSpecification = "enabled"
+
+	// AmdSevSnpSpecificationDisabled means AMD SEV SNP is disabled for the instance.
+	AmdSevSnpSpecificationDisabled AmdSevSnpSpecification = "disabled"
+)
+
+// CpuOptions defines the cpu options for the instance.
+type CpuOptions struct {
```
everettraven: A couple questions to help me better understand the direction this should take:
- Because all fields are optional here, what happens if this is set to the zero value of the struct `{}`?
- Because there is no constraints here as to how many properties can be specified at once, what would it mean if in the future there are multiple options to configure and I configure more than one of them?

fangge1212: - If it is an empty struct, cpuOptions=nil will be passed to the AWS API
- I don't get your question. If there are multiple options to configure in the future, then we can configure multiple options.

everettraven: >I don't get your question. If there are multiple options to configure in the future, then we can configure multiple options

What I was trying to get at is similar to this question upstream: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5598#discussion_r2226010124

If you are only ever going to have mutually exclusive CPU options then using an enum for those options likely makes sense.

If you are going to have a _subset_ of mutually exclusive options but the rest can be specified together, then this structure is generally OK, but I'd think an enum is still reasonable for a single mutual exclusivity field.

Looking at https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-specify-cpu-options.html I'd expect something like:
```yaml
cpuOptions:
  cores: 4
  threadsPerCore: 1
  confidentialComputing: AmdSevSnp
```

to be reasonable (assuming that something like `confidentialComputing` would be mutually exclusive so you can't set something like AmdSevSnp _and_ TDX if it were to be added in the future)

fangge1212: Updated to:
```
cpuOptions:
  confidentialComputing: AMDSevSnp
```

machine/v1beta1/types_awsprovider.go
```diff
@@ -109,6 +112,33 @@ type AWSMachineProviderConfig struct {
...
+	AWSConfidentialComputePolicyDisabled AWSConfidentialComputePolicy = "Disabled"
+	// AWSConfidentialComputePolicyAWSSEVSNP sets AMD SEV-SNP as the VM instance's confidential computing technology of choice.
+	AWSConfidentialComputePolicySEVSNP AWSConfidentialComputePolicy = "AMDSevSnp"
+)
+
+// CPUOptions defines the cpu options for the instance.
+type CPUOptions struct {
+	// confidentialCompute specifies whether confidential computing should be enabled for the instance,
+	// and, if so, which confidential computing technology to use.
+	// If set to Disabled, the instance will not use confidential computing.
```
everettraven: "Disabled" is often times an overloaded term and we generally try to avoid using it where we can.

For scenarios like this, I generally go for "None" and would encourage that usage here as well.

IMO, `confidentialCompute: None` reads more intuitively than `confidentialCompute: Disabled`.

machine/v1beta1/types_awsprovider.go
```diff
@@ -109,6 +112,33 @@ type AWSMachineProviderConfig struct {
...
+	// If set to AMDSevSnp, the instance will be configured with AMD SEV-SNP.
+	// In this case, ensure the following conditions are met:
+	// 1) The selected instance type supports AMD SEV-SNP.
+	// 2) The selected AWS region supports AMD SEV-SNP.
+	// 3) The selected AMI supports AMD SEV-SNP.
+	// More details can be checked at https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/sev-snp.html
+	// If omitted, the platform will apply a default value — currently Disabled, but this may change over time.
```
everettraven:
```suggestion
	// When set to Disabled, the instance will not use confidential computing.
	// When set to AMDSevSnp, the instance will be configured with AMD SEV-SNP.
	// In this case, ensure the following conditions are met:
	// 1) The selected instance type supports AMD SEV-SNP.
	// 2) The selected AWS region supports AMD SEV-SNP.
	// 3) The selected AMI supports AMD SEV-SNP.
	// More details can be checked at https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/sev-snp.html
	// When omitted, this means no opinion and the platform is left to choose a reasonable default, which is subject to change without notice. The current default is Disabled.
```

machine/v1beta1/types_awsprovider.go
```diff
@@ -109,6 +112,33 @@ type AWSMachineProviderConfig struct {
...
+
+const (
+	// AWSConfidentialComputePolicyDisabled disables confidential compute for the AWS machine.
+	AWSConfidentialComputePolicyDisabled AWSConfidentialComputePolicy = "Disabled"
+	// AWSConfidentialComputePolicyAWSSEVSNP sets AMD SEV-SNP as the VM instance's confidential computing technology of choice.
+	AWSConfidentialComputePolicySEVSNP AWSConfidentialComputePolicy = "AMDSevSnp"
+)
+
+// CPUOptions defines the cpu options for the instance.
+type CPUOptions struct {
```
everettraven, line 133: We try to avoid things like `cpuOptions: {}` being valid inputs because it is usually semantically the same as just not specifying the field altogether.

Add the `+kubebuilder:validation:MinProperties=1` marker so that it is required for at least one property to be specified, making `{}` an invalid input.

machine/v1beta1/types_awsprovider.go
```diff
@@ -17,6 +17,9 @@ type AWSMachineProviderConfig struct {
 	AMI AWSResourceReference `json:"ami"`
 	// instanceType is the type of instance to create. Example: m4.xlarge
 	InstanceType string `json:"instanceType"`
+	// cpuOptions is the set of cpu options for the instance.
+	// +optional
```
everettraven, line 25: What happens if this field is not specified by a user?

machine/v1beta1/types_awsprovider.go
```diff
@@ -17,6 +17,9 @@ type AWSMachineProviderConfig struct {
 	AMI AWSResourceReference `json:"ami"`
 	// instanceType is the type of instance to create. Example: m4.xlarge
 	InstanceType string `json:"instanceType"`
+	// cpuOptions is the set of cpu options for the instance.
```
everettraven: I think we could improve the GoDoc here to be more descriptive.

https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#write-user-readable-documentation-in-godoc is a good starting point for the things that you should take into consideration and include in the documentation for a field.

machine/v1beta1/types_awsprovider.go
fangge1212: If unset, no CPU options are passed to the AWS platform and AWS default values are used.

machine/v1beta1/types_awsprovider.go
fangge1212: Updated

machine/v1beta1/types_awsprovider.go
fangge1212: Updated

machine/v1beta1/types_awsprovider.go
fangge1212: Updated

machine/v1beta1/types_awsprovider.go
fangge1212: Thanks. I updated the doc, please have a look again.

machine/v1beta1/types_awsprovider.go
```diff
@@ -109,6 +113,35 @@ type AWSMachineProviderConfig struct {
...
+// AWSConfidentialComputePolicy represents the confidential compute configuration for the instance.
+// +kubebuilder:validation:Enum=Disabled;AmdSevSnp
+type AWSConfidentialComputePolicy string
+
+const (
+	// AWSConfidentialComputePolicyNone disables confidential computing for the instance.
+	AWSConfidentialComputePolicyNone AWSConfidentialComputePolicy = "None"
+	// AWSConfidentialComputePolicySEVSNP enables AMD SEV-SNP as the confidential computing technology for the instance.
+	AWSConfidentialComputePolicySEVSNP AWSConfidentialComputePolicy = "AmdSevSnp"
+)
```
damdo, line 128: Let's follow the same approach we took for the GCP types: https://github.com/openshift/api/pull/2165/files#diff-9bd66f01f63b7ce0455fe993aa03dfe93964ac62b31278a6dbf3db2fcb7f792eR66-R71

In detail:
- Let's use `Enabled`/`Disabled` (rather than `None`)
- Let's fully spell out the `AmdSevSnp` to `AMDEncryptedVirtualizationNestedPaging`

machine/v1beta1/types_awsprovider.go
fangge1212: `+kubebuilder:validation:MinProperties=1` makes many tests failed:

machine/v1beta1/types_awsprovider.go
fangge1212: I updated to `Disabled;AMDEncryptedVirtualizationNestedPaging`
I don't think `Enabled` is a meaningfull value, because all the other values than `Disabled` implies `Enabled`

machine/v1beta1/types_awsprovider.go
fangge1212: When I add `+kubebuilder:validation:MinProperties=1`, kubeapilinter asked me to add tag `omitzero` and don't use make CPUOptions a pointer. I followed its prompt,  but this makes many tests in cluster-api-provider-aws failed:
```
    --- FAIL: TestSSHKeyName/SSH_key_name_is_nil_is_valid (0.01s)
        sshkeyname_test.go:89: ValidateCreate() error = AWSMachine.infrastructure.cluster.x-k8
       "machine-9zsqb" is invalid: spec.cpuOptions: Invalid value: 0: spec.cpuOptions in body should have at least 1 properties, wantErr false
```

machine/v1beta1/types_awsprovider.go
damdo: `Enabled` is normally for the default value (so in this case OpenShift engineers choose the default for the user) as you can see here on the GCP one:

`
	// When set to Enabled, the machine will be configured as a confidential computing instance with no preference on the confidential compute policy used. In this mode, the platform chooses a default that is subject to change over time. Currently, the default is to use AMD Secure Encrypted Virtualization.
`

Ideally we should try and keep consistency across providers for the same feature, so I'd be happy to see it added

bgartzi: Hi @damdo, @fangge1212!

The problem I see in the AWS case is that AFAICT, it doesn't support the default GCP case: AMD SEV. It only supports AMD SEV-SNP. In that case, we would end up defining `Enabled` as a valid case for AWS, but would have a different behavior than the one from GCP, configuring AMD SEV-SNP machines instead. In my opinion this would be misleading.

This is another discussion, but at the moment the `Enabled` option for GCP is redundant, as `AMDEncryptedVirtualization` achieves the same result. In my opinion, Deprecating `Enabled` in the GCP provider in the future would help getting rid of that redundancy and also having consistency across providers for the same feature, as you mention.

damdo: If we are thinking about deprecating `Enabled` then it is probably ok to leave it out from this one.
cc. @JoelSpeed

everettraven: We normally try to avoid using terminology like `Enabled` now because it is a term that is often overloaded and the meaning not very clear to users at first glance. If you are wanting a "platform chooses the default" state, we generally use the omission of an optional field as signal that the user has no opinion and that the platform should choose a default.

I could also see a case for requiring that a user specify a value and one of those values could be something like `PlatformDefault` that means "I don't care, I just want confidential computing to be enabled, I'll let OpenShift default it to something".

In this case, I think that would essentially look something like:
```yaml
cpuOptions:
  confidentialCompute: ""
```
(the "omitted" case, which is usually what we do in OpenShift)

OR

```yaml
cpuOptions:
  confidentialCompute: PlatformDefault
```

everettraven: We can also use the omission of the parent field to do defaulting behavior as well, so just omitting the `cpuOptions` field altogether could mean "no opinion on CPU options, platform defaults", which is the path I would prefer if that is semantically equivalent to either of the examples above.

fangge1212: The omission of the cpuOptions field or confidentialCompute field will be equivalent to ConfidentialComputeDisabled.
The current code doesn't support this option: "I don't care, I just want confidential computing to be enabled, I'll let OpenShift default it to something". Do we need this?

everettraven: >The current code doesn't support this option: "I don't care, I just want confidential computing to be enabled, I'll let OpenShift default it to something". Do we need this?

Is that behavior you want your users to be able to do? I wouldn't say this is a requirement on my end, but I do think that we should be clear on the intended user workflows to be able to design this API appropriately.

fangge1212: I don’t think `PlatformDefault` is very useful in practice. When a user installs a cluster, they already choose the instance type and region, which directly constrains the available confidential computing technologies. That means the platform can’t meaningfully pick a “default” technology on the user’s behalf.

machine/v1beta1/types_awsprovider.go
fangge1212: Updated back to "Disabled" to align with GCP

machine/v1beta1/types_awsprovider.go
JoelSpeed: `Enabled` was likely a mistake in the previous API. I'm ok with the following states for this API
* Omitted - means its disabled
* `Disabled` - I've explicitly disabled this as I don't want anyone else defaulting it for me
* `AMDEncryptedVirtualizationNestedPaging` - I specifically want this kind of option

machine/v1beta1/types_awsprovider.go
everettraven: I see. I just took a look and it looks like cluster-api-provider-aws is on Go 1.23.1 which does not contain the `omitzero` behavior that kube-api-linter is asking you to add. That support was added in Go 1.24.0

everettraven: Is it possible for you to bump to Go 1.24.z in cluster-api-provider-aws?

everettraven: Looks like that is in progress here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5624

machine/v1beta1/types_awsprovider.go
```diff
@@ -17,6 +17,10 @@ type AWSMachineProviderConfig struct {
 	AMI AWSResourceReference `json:"ami"`
 	// instanceType is the type of instance to create. Example: m4.xlarge
 	InstanceType string `json:"instanceType"`
+	// cpuOptions defines CPU-related settings for the instance, including the confidential computing policy.
+	// If unset, no CPU options will be passed to the AWS platform and AWS default CPU options will be applied.
```
everettraven: How do I know what the "AWS default CPU options" that will be applied are? Are these literally defaults AWS imposes on requests that don't specify these options, or are these defaulted elsewhere?

fangge1212: Thanks for the comment! To clarify: OpenShift does not set defaults for cpuOptions. If the field is unset, the `RunInstances` request is sent without a `CpuOptions` block, and AWS applies its own defaults for the chosen instance type. 
I’ll update the field description to make that clear.

fangge1212: Updated the description, please take a look again.

machine/v1beta1/types_awsprovider.go
```diff
@@ -17,6 +17,12 @@ type AWSMachineProviderConfig struct {
 	AMI AWSResourceReference `json:"ami"`
 	// instanceType is the type of instance to create. Example: m4.xlarge
 	InstanceType string `json:"instanceType"`
+	// cpuOptions defines CPU-related settings for the instance, including the confidential computing policy.
+	// If unset, no cpuOptions will be included in the API request to AWS, and the instance will use the default CPU options
+	// applied by AWS for the selected intance type.
+	// +kubebuilder:validation:MinProperties=1
```
JoelSpeed: Pop this one on the struct rather than the field please

machine/v1beta1/types_awsprovider.go
```diff
@@ -109,6 +115,35 @@ type AWSMachineProviderConfig struct {
...
+	// When set to Disabled, confidential computing will be disabled for the instance.
+	// When set to AMDEncrytedVirtualizationNestedPaging, AMD SEV-SNP will be used as the confidential computing technology for the instance.
+	// In this case, ensure the following conditions are met:
+	// 1) The selected instance type supports AMD SEV-SNP.
+	// 2) The selected AWS region supports AMD SEV-SNP.
+	// 3) The selected AMI supports AMD SEV-SNP.
+	// More details can be checked at https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/sev-snp.html
+	// When omitted, this means no opinion and the AWS platform is left to choose a reasonable default,
+	// which is subject to change without notice. The current default is Disabled.
+	// +kubebuilder:validation:Enum=Disabled;AMDEncrytedVirtualizationNestedPaging
```
JoelSpeed: Any reason to put this here, vs on the type definition on L119?

machine/v1beta1/types_awsprovider.go
fangge1212: Updated.

machine/v1beta1/types_awsprovider.go
fangge1212: Updated the description again to make it more concise: 
` When omitted, this means no opinion and the AWS platform is left to choose a reasonable default.`

machine/v1beta1/types_awsprovider.go
fangge1212: The placement of this line differs between `AWSNetworkInterfaceType` (within the struct) and `MarketType `(on the type definition). I was unsure which pattern to follow, so I selected one approach arbitrarily.

JoelSpeed:
API looks good, but cannot merge until the equivalent upstream API is merged into CAPA, please LMK once that is merged

damdo:
New CAPA upstream PR for this is here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5605

damdo:
The upstream PR is now unblocked, and reviews/merging can proceed
https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5605#issuecomment-3265151506

fangge1212:
/unhold
[Upstream pr](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5605 ) is merged

machine/v1beta1/types_awsprovider.go
JoelSpeed: Do we know what the default values are currently on AWS?

machine/v1beta1/types_awsprovider.go
```diff
@@ -109,6 +113,36 @@ type AWSMachineProviderConfig struct {
...
+	// AWSConfidentialComputePolicySEVSNP enables AMD SEV-SNP as the confidential computing technology for the instance.
+	AWSConfidentialComputePolicySEVSNP AWSConfidentialComputePolicy = "AMDEncryptedVirtualizationNestedPaging"
+)
+
+// CPUOptions defines CPU-related settings for the instance, including the confidential computing policy.
+// +kubebuilder:validation:MinProperties=1
+type CPUOptions struct {
+	// confidentialCompute specifies whether confidential computing should be enabled for the instance,
+	// and, if so, which confidential computing technology to use.
+	// Valid values are: Disabled, AMDEncryptedVirtualizationNestedPaging
```
everettraven: Nit: We generally also mention that an allowed value is `omitted` for optional enums.

In practice, omitting this field is for now an invalid configuration, but in the future if new fields are added omitting this field and specifying another one would be valid.

everettraven:
@fangge1212 Could you include a link to the associated PR that implements the validations outlined here in the validating webhook this API runs through?

machine/v1beta1/types_awsprovider.go
everettraven: Even if we don't have concrete defaults we can point to, it would be nice to include guidance on how an end-user could identify what the defaults for their configuration would be.

fangge1212:
> Could you include a link to the associated PR that implements the validations outlined here in the validating webhook this API runs through?

Previously, I added instance type validation for the specified confidential computing technology (currently only AMD SEV-SNP). However, the instance type list is hard-coded and needs to be updated manually, so I removed the validation. As a result, there is now no webhook validation for this API change.

everettraven:
>Previously, I added instance type validation for the specified confidential computing technology (currently only AMD SEV-SNP). However, the instance type list is hard-coded and needs to be updated manually, so I removed the validation. As a result, there is now no webhook validation for this API change.

AFAIK none of the validations that exist on the markers in the API here are actually enforced unless done through the webhook. I would expect us to have changes in the webhook to reject invalid configurations.

machine/v1beta1/types_awsprovider.go
fangge1212: Thanks, updated. Also updated the description for CPUOptions to make it clear: 
`+ // If provided, it must not be empty — at least one field must be set.
`

machine/v1beta1/types_awsprovider.go
fangge1212: The CpuOptions in AWS consists of three fields: 
- amdSevSnp - default value is Disabled.
- coreCount - default value depends on the instance type.
- threadsPerCore - default value depends on the instance type.
Refer to: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/cpu-options-supported-instances-values.html

In this PR, only amdSevSnp is exposed to users. I'm not entirely sure how best to describe this in the API documentation — should we include a link to the AWS documentation?

fangge1212:
> > Previously, I added instance type validation for the specified confidential computing technology (currently only AMD SEV-SNP). However, the instance type list is hard-coded and needs to be updated manually, so I removed the validation. As a result, there is now no webhook validation for this API change.
> 
> AFAIK none of the validations that exist on the markers in the API here are actually enforced unless done through the webhook. I would expect us to have changes in the webhook to reject invalid configurations.

Ah, I didn't know this. I will add validation in webhook

fangge1212:
> > > Previously, I added instance type validation for the specified confidential computing technology (currently only AMD SEV-SNP). However, the instance type list is hard-coded and needs to be updated manually, so I removed the validation. As a result, there is now no webhook validation for this API change.
> > 
> > 
> > AFAIK none of the validations that exist on the markers in the API here are actually enforced unless done through the webhook. I would expect us to have changes in the webhook to reject invalid configurations.
> 
> Ah, I didn't know this. I will add validation in webhook

When I added `+kubebuilder:validation:MinProperties=1` before, I had this error. It seems the validation on the marker is working?
```
    --- FAIL: TestSSHKeyName/SSH_key_name_is_nil_is_valid (0.01s)
        sshkeyname_test.go:89: ValidateCreate() error = AWSMachine.infrastructure.cluster.x-k8
       "machine-9zsqb" is invalid: spec.cpuOptions: Invalid value: 0: spec.cpuOptions in body should have at least 1 properties, wantErr false
```

everettraven:
That looks like it is an upstream test. @JoelSpeed knows the nuances here more than I do, but my understanding is that machine configurations for OpenShift are embedded as raw JSON in the Machine API and as such they must be programmatically evaluated

JoelSpeed:
Upstream, the types are real CRDs and so all of the validation markers apply. But, the `AWSProviderConfig` (and likewise for other platforms) in the Machine V1beta1 group downstream are not real CRDs, and are served via a `RawExtension` within the Machine CRD.

This means that they have no validation, and the only validation we can do is implemented as a Go ValidatingWebhookConfiguration which lives in the machine-api-operator repository.

This means the markers (while helpful and I appreciate being added even if they do nothing), will need to also be implemented manually in the webhook

machine/v1beta1/types_awsprovider.go
fangge1212: Added the link to AWS website

fangge1212:
> Upstream, the types are real CRDs and so all of the validation markers apply. But, the `AWSProviderConfig` (and likewise for other platforms) in the Machine V1beta1 group downstream are not real CRDs, and are served via a `RawExtension` within the Machine CRD.
> 
> This means that they have no validation, and the only validation we can do is implemented as a Go ValidatingWebhookConfiguration which lives in the machine-api-operator repository.
> 
> This means the markers (while helpful and I appreciate being added even if they do nothing), will need to also be implemented manually in the webhook

This is the draft PR(https://github.com/openshift/machine-api-operator/pull/1420/files) that adds webhook validation for the new parameter cpuOptions.
The only problem is that I can't distinguish between `cpuOptions not provided` and `provided but empty`(cpuOptions:{}), because both of them will be:
```
CPUOptions{
    ConfidentialCompute: "",
}
```
Should we change `CPUOptions` to pointer type in `AWSMachineProviderConfig`? Then if it is not provided, it will be nil.

everettraven:
>Should we change CPUOptions to pointer type in AWSMachineProviderConfig? Then if it is not provided, it will be nil.

Yes, this was a nuance of webhook based validation that I missed and is not something the linter is aware of. We can override the lint check if it fails on this.

@JoelSpeed and I have already started discussing how we can improve this linting behavior for APIs that would need the same validation behaviors.

fangge1212:
> > Should we change CPUOptions to pointer type in AWSMachineProviderConfig? Then if it is not provided, it will be nil.
> 
> Yes, this was a nuance of webhook based validation that I missed and is not something the linter is aware of. We can override the lint check if it fails on this.
> 
> @JoelSpeed and I have already started discussing how we can improve this linting behavior for APIs that would need the same validation behaviors.

After chaning CPUOptions to pointer type, linting fails:
```
# make lint
hack/golangci-lint.sh run --new-from-rev=master 
make[1]: Entering directory '/root/go/src/github.com/openshift/api/tools'
make[1]: Nothing to be done for 'kube-api-linter'.
make[1]: Leaving directory '/root/go/src/github.com/openshift/api/tools'
machine/v1beta1/types_awsprovider.go:25:2: optionalfields: field CPUOptions does not allow the zero value. The field does not need to be a pointer. (kubeapilinter)
	CPUOptions *CPUOptions `json:"cpuOptions,omitempty,omitzero"`
	^
1 issues:
* kubeapilinter: 1
make: *** [Makefile:47: lint] Error 1
```
Remove `omitzero`:
```
[root@dell-per740-78 api]# make lint
hack/golangci-lint.sh run --new-from-rev=master 
make[1]: Entering directory '/root/go/src/github.com/openshift/api/tools'
make[1]: Nothing to be done for 'kube-api-linter'.
make[1]: Leaving directory '/root/go/src/github.com/openshift/api/tools'
machine/v1beta1/types_awsprovider.go:25:2: optionalfields: field CPUOptions does not allow the zero value. It must have the omitzero tag. (kubeapilinter)
	CPUOptions *CPUOptions `json:"cpuOptions,omitempty"`
	^
1 issues:
* kubeapilinter: 1
make: *** [Makefile:47: lint] Error 1

```
How about I remove the marker "// +kubebuilder:validation:MinProperties=1" to make the linting pass?

everettraven:
>How about I remove the marker "// +kubebuilder:validation:MinProperties=1" to make the linting pass?

I don't think we want to remove that. We don't want to allow an empty struct (`{}`). I can override the linter failure if it is related to a nuance the linter isn't aware of, like this needing to be a pointer.

everettraven:
>Remove omitzero:

~I might be wrong here, but I think we also want to keep the `omitzero` so that the zero value of the struct is omitted during serialization as well.~

I was wrong here, `omitzero` will still serialize as `{}` when put on a pointer.

fangge1212:
When CPUOptions is a pointer type, I can't distinguish between the following configurations in webhook validation:
- provided but empty(CPUOptions:{})
- provided with zero value(CPUOptions:{ConfidentialCompute:""})
Both of them will be `CPUOptions:{ConfidentialCompute:""}`

When CPUOptions is a struct type, I can't distinguish between the following configurations in webhook validation:
- not provided
- provided but empty(CPUOptions:{})
- provided with zero value(CPUOptions:{ConfidentialCompute:""})
All of them will be `CPUOptions:{ConfidentialCompute:""}`

So there is no way to validate minProperties=1 for CPUOptions in webhook validation.

Updates:
I set both CPUOptions and ConfidentialCompute to pointer type, now I can distiguish between the three configurations. But then I got a new kubeapilinter error:
```
machine/v1beta1/types_awsprovider.go:147:2: optionalfields: field ConfidentialCompute does not allow the zero value. The field does not need to be a pointer. (kubeapilinter)
	ConfidentialCompute *AWSConfidentialComputePolicy `json:"confidentialCompute,omitempty"`
	^
```

everettraven:
>Updates:
I set both CPUOptions and ConfidentialCompute to pointer type, now I can distiguish between the three configurations. But then I got a new kubeapilinter error:

That seems like a reasonable error to override as well if you can't properly distinguish between the necessary states otherwise.

fangge1212:
> > Updates:
> > I set both CPUOptions and ConfidentialCompute to pointer type, now I can distiguish between the three configurations. But then I got a new kubeapilinter error:
> 
> That seems like a reasonable error to override as well if you can't properly distinguish between the necessary states otherwise.

@everettraven 
I have updated the pr with both CPUOptions and ConfidentialCompute set to pointer type, now I just need to wait for you to overwrite the lint rule, right?

everettraven:
@fangge1212 Yes, I'll override the failure and approve this PR once I am confident that the appropriate validations are in place on the webhook side. I left a comment on that draft PR regarding a case that looks to be considered valid that should not be based on the current API constraints.

machine/v1beta1/types_awsprovider.go
```diff
@@ -109,6 +116,37 @@ type AWSMachineProviderConfig struct {
...
+	// confidentialCompute specifies whether confidential computing should be enabled for the instance,
+	// and, if so, which confidential computing technology to use.
+	// Valid values are: Disabled, AMDEncryptedVirtualizationNestedPaging and omitted.
+	// When set to Disabled, confidential computing will be disabled for the instance.
+	// When set to AMDEncryptedVirtualizationNestedPaging, AMD SEV-SNP will be used as the confidential computing technology for the instance.
+	// In this case, ensure the following conditions are met:
+	// 1) The selected instance type supports AMD SEV-SNP.
+	// 2) The selected AWS region supports AMD SEV-SNP.
+	// 3) The selected AMI supports AMD SEV-SNP.
+	// More details can be checked at https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/sev-snp.html
```
everettraven, line 143: I forgot to ask this originally, but what happens if the user specifies AMD SEV-SNP on an instance type, region, or AMI that does _not_ support this?

Is there some feedback loop that enables a user to recover from this bad state?

everettraven:
Other than my last question above, this looks good and the webhook implements the validations I would have expected based on the current state of the API.

machine/v1beta1/types_awsprovider.go
fangge1212: AWS will return error messages in such situations.

- For invalid instance type
`api error InvalidParameterCombination: The specified instance type does not support AMD SEV-SNP. Specify a supported instance type and try again.`
- For invalid AMI
`api error InvalidParameterCombination: The specified AMI does not support a boot mode that is compatible with AMD SEV-SNP. Specify a compatible AMI and try again.`
- For invalid region
(I haven't tried this yet)

everettraven: How do those get returned to end users?

fangge1212: In cluster-api-provider-aws, errors from the EC2 RunInstances API (such as InvalidParameterCombination for unsupported instance types, AMIs, or regions) are caught and wrapped before being returned. These errors propagate up through the reconciliation loop until they surface in the AWSMachine (or higher-level resource) status conditions.

everettraven: I'm not as familiar with the CAPI work, so I want to make sure I clarify here - this means we should see a status condition populated in the `Machine` resource that user can clearly see what the problem is and make the necessary updates to that `Machine` resource to recover?

fangge1212: I'm also not familiar with how it works, I did a little research. The end user can see the error message in the terminal or install log. Here is the flow of  returning an error to the end user when the cluster creation fails is like this:

1. Installer → applies Cluster API manifests (Cluster, Machine, AWSMachine, etc.)

2. CAPA controller (specifically awsmachine_controller.go) runs createInstance() via the Reconcile loop.

3. If AWS returns an error (e.g., InvalidParameterValue: instance type not supported in region), it bubbles up:

```
instance, err = r.createInstance(...)
if err != nil {
    machineScope.Error(err, "unable to create instance")
    conditions.MarkFalse(machineScope.AWSMachine, infrav1.InstanceReadyCondition, infrav1.InstanceProvisionFailedReason, ...)
    return ctrl.Result{}, err
}
```


This sets the `AWSMachine` status condition` InstanceReady=False` with reason `InstanceProvisionFailed`.

The corresponding `Machine` object (in the Cluster API layer) reflects this failure in its status — its `Phase` becomes "Failed", and `Status.FailureMessage` is populated with the error message (e.g., “Invalid instance type”).

4. When the OpenShift Installer monitors the cluster creation (in pkg/infrastructure/clusterapi/clusterapi.go), it repeatedly calls:

`checkMachineReady(machine, requirePublicIP)
`

This function inspects the Machine status and decides whether it’s ready, still provisioning, or failed.

```
if machine.Status.Phase != "Provisioned" && machine.Status.Phase != "Running" {
    // Not ready yet, keep waiting
    return false, nil
} else if machine.Status.Phase == "Failed" {
    msg := ptr.Deref(machine.Status.FailureMessage, "machine.Status.FailureMessage was not set")
    return false, fmt.Errorf("machine %s failed to provision: %s", machine.Name, msg)
}
```
So if the Machine enters the Failed phase, this function returns an error — a Go error that directly includes the AWS failure reason (e.g., “machine failed to provision: InvalidParameterValue: Instance type m5.x99 not supported”).

5. The installer monitors Machines until all are “ready”. If any checkMachineReady() returns an error, the installer stops waiting and reports the failure in the terminal output — for example:

`level=error msg="Bootstrap failed to complete: machine master-0 failed to provision: Invalid instance type m5.x99"`


The installer then exits with a non-zero status, and the user sees this message in their terminal or install.log.

everettraven: Gotcha. So regardless the installer will always return the failure.

@JoelSpeed this is relatively new territory for me as I haven't reviewed a lot of these machine API changes - is there a preference as to failing at configuration read time over during cluster bootstrap? Do you know if we clean up any resources we may or may not have created if we fail bootstrapping? What is the impact to an end-user if we fail during cluster bootstrapping versus validation of the configuration?

JoelSpeed: > is there a preference as to failing at configuration read time over during cluster bootstrap?

Where possible yes, it is preferred to validate at config time, but with instance type capabilities, that often means fetching information from the cloud provider which may or may not be publicly exposed via an API, so we have to be best effort here.

> Do you know if we clean up any resources we may or may not have created if we fail bootstrapping?

The installer can clean up all resources when an install fails yes

> What is the impact to an end-user if we fail during cluster bootstrapping versus validation of the configuration?

We spun up most of a cluster and wasted ~30 minutes to 2 hours of their time. Generally early validation is obvious, quick, and easy to read. Where when the cluster fails in this way, it's generally not very obvious why. The installer will report that the Machine API cluster operator isn't available, but it will also report other COs as unavailable, so it may not be immediately obvious to folks that it was this config that got them stuck here.

everettraven: Thanks for the explanation, that helps me put this into perspective a bit better.

> but with instance type capabilities, that often means fetching information from the cloud provider which may or may not be publicly exposed via an API, so we have to be best effort here.

Looks like AWS has an API for instance type discovery based on features outlined in https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstanceTypes.html - seems useful for determining what we would need to here. Is there any history of the installer using these types of APIs for earlier validation?

I don't think it would be worth blocking this PR on unless we already have some precedent set here to use those provider APIs, but I would be curious if there is a trivial enough way for us to let users know that this config was what got them stuck.

everettraven (approved):
/lgtm

everettraven:
`lint` check is failing on two known issues due to this being an API that is validated via webhook and not via the standard CRD validation cycle. Our linter isn't aware of these nuances at the moment, so overriding.

/override ci/prow/lint

everettraven:
@fangge1212 over to you for getting the `verified` label

JoelSpeed:
/override ci/prow/lint

Go validated type, not openapi which is how the current configuration is configured

PR #2451, opened 2025-08-12
OKD-259: Support an "OKD" featureset to be enabled by default on OKD clusters

jatinsu:
This PR was largely bult on the work done by `@Prashanth684`

- This first pass only includes all featuregates which are default
- For OKD clusters, this would be the "default" featureset
- Allows for upgrades


config/v1/types_feature.go
```diff
@@ -53,8 +53,12 @@ var (
 	// your cluster may fail in an unrecoverable way.
 	CustomNoUpgrade FeatureSet = "CustomNoUpgrade"
 
+	// OKD turns on features for OKD. Turning this feature set ON is supported for OKD clusters, but NOT for OpenShift clusters
+	// this feature set on CANNOT BE UNDONE for OKD clusters and when enabled on OpenShift clusters it PREVENTS UPGRADES.
+	OKD FeatureSet = "OKD"
```
JoelSpeed: How do we prevent this from being an option on OCP clusters?

What options do we have for including things in OCP but not OKD? Does OKD ship as a different cluster profile perhaps?

features/features.go
```diff
@@ -40,7 +40,7 @@ var (
 					reportProblemsToJiraComponent("Management Console").
 					contactPerson("jhadvig").
 					productScope(ocpSpecific).
-					enableIn(configv1.Default, configv1.DevPreviewNoUpgrade, configv1.TechPreviewNoUpgrade).
+					enableIn(configv1.Default, configv1.OKD, configv1.DevPreviewNoUpgrade, configv1.TechPreviewNoUpgrade).
```
JoelSpeed, line 43: Is there ever a reason we would want to promote a feature to Default and not have it also in the OKD featureset? If there isn't, we should either make it so that the OKD featureset inherits from the Default feature set, or, add some sort of test that will flag when something is promoted without OKD being added

config/v1/types_feature.go
Prashanth684: yes we need to do this. similar to the model that the[ installer](https://github.com/openshift/installer/blob/main/pkg/types/installconfig_scos.go) and other repos follow, we can restrict it to be enabled only on OKD clusters.

features/features.go
Prashanth684: >  If there isn't, we should either make it so that the OKD featureset inherits from the Default feature set, or, add some sort of test that will flag when something is promoted without OKD being added
 
agree, I wanted to connect with you about that @JoelSpeed . outside of code reviewing and enforcing that it be added to the OKD featureset as well - I didn't think there were other methods. How do we inherit from the default featureset ?

config/v1/types_feature.go
JoelSpeed: Currently, you're going to generate an API that will be applied to OCP clusters, which means that in theory we have to either a) block this value from being set at admission time, b) prevent the option from even being included in the API schema

I would probably lean towards b, but that means creating separate payload manifests for OKD I believe 🤔 

We also currently have no way to create differences like that in the tooling apart from feature gates, but, that isn't a long term thing, so we would need to provide other options

features/features.go
JoelSpeed: > How do we inherit from the default featureset

It would need code changes to the generation of the featureset specific manifests, that's something we _could_ do, but it hasn't been done before

How certain are we that every OCP feature will always land in OKD?

config/v1/types_feature.go
Prashanth684: >  block this value from being set at admission time

can we do this in cluster config operator maybe? if the featureset being enabled is OKD and the image is not built for OKD (through the method above), we reject it?

features/features.go
Prashanth684: > How certain are we that every OCP feature will always land in OKD?

100% i would say. Every feature in the default featureset is also in OKD by default.

config/v1/types_feature.go
Prashanth684: But also - the other angle here: Why do we want to prevent it from being turned on for OCP? It will always be a subset of techpreviewNoUpgrade, the only difference being that it will be upgradeable only on OKD clusters (this we can control through CCO as done [here](https://github.com/Prashanth684/cluster-config-operator/commit/a27a96175fa236a4d1e530088d11217d4a99391d#diff-7f99c3f2350fa8ba8083db55d24066a6b7e241c9639342c3f7c18ba7e775e697L58-R60)). This will mean from an OCP perspective, it will be just another featureset. There are no OKD specific features (nor do i see any in the future).

JoelSpeed: > can we do this in cluster config operator maybe?

We used to validate the feature gate transitions and values in the API server directly, but this was removed in favour of CEL, see https://github.com/openshift/kubernetes/pull/1944/files

Config operator doesn't have any admission time validation ability

> Why do we want to prevent it from being turned on for OCP?

I don't think we want an extra feature set in OCP. We have TPNU, we have DPNU, we have CNU.

Enabling an OKD feature sat that isn't OKDNU in OCP sends the wrong signal IMO. All of our features sets that have existed in OCP have either been upgradable, or have had `NoUpgrade` in the name.

The OKD set can be achieved as a version of CNU if a customer so desired.

Prashanth684: > We used to validate the feature gate transitions and values in the API server directly, but this was removed in favour of CEL, see https://github.com/openshift/kubernetes/pull/1944/files

Would bringing back validation like this: https://github.com/openshift/kubernetes/pull/2420 make sense?

JoelSpeed: I don't think it's necessary to have in tree validations, but we do need to agree on the behaviour of a new featureset.

I'm particularly keen not to expose the feature set to non-okd clusters

I think we should be able to achieve this with the tools we have available to us today, but we may need to think about how the CVO handles certain resources by restricting files from certain cluster profiles

Prashanth684: > I'm particularly keen not to expose the feature set to non-okd clusters

Does this mean we don't even build the featureset related manifests for OCP ? Or do we build the manifests and just not expose that featureset to OCP? (not sure how this is even possible because we do the detection for whether a component is OKD at build time)

JoelSpeed: Do you want this featureset to be opt-in? Or on by default in all OKD clusters?

For HyperShift and OCP, they both have the `Default` featureset, but, these featuresets are not necessarily always identical. In fact, HyperShift for some time has been carrying additional gates that are not enabled in the standalone default set.

Personally, I think something like this would make more sense, but it is currently implemented using a cluster profile annotation that CVO uses to choose whether or not to install. I suspect if we were to work with the OTA team they'd have some options so that we could include one yaml file for OKD and one for standalone OCP

JoelSpeed: > Do you want this featureset to be opt-in? Or on by default in all OKD clusters?

This is still an interesting question

config/v1/types_feature.go
```diff
@@ -67,10 +71,11 @@ type FeatureGateSelection struct {
 	// Turning on or off features may cause irreversible changes in your cluster which cannot be undone.
 	// +unionDiscriminator
 	// +optional
-	// +kubebuilder:validation:Enum=CustomNoUpgrade;DevPreviewNoUpgrade;TechPreviewNoUpgrade;""
+	// +kubebuilder:validation:Enum=CustomNoUpgrade;DevPreviewNoUpgrade;TechPreviewNoUpgrade;OKD;""
 	// +kubebuilder:validation:XValidation:rule="oldSelf == 'CustomNoUpgrade' ? self == 'CustomNoUpgrade' : true",message="CustomNoUpgrade may not be changed"
 	// +kubebuilder:validation:XValidation:rule="oldSelf == 'TechPreviewNoUpgrade' ? self == 'TechPreviewNoUpgrade' : true",message="TechPreviewNoUpgrade may not be changed"
 	// +kubebuilder:validation:XValidation:rule="oldSelf == 'DevPreviewNoUpgrade' ? self == 'DevPreviewNoUpgrade' : true",message="DevPreviewNoUpgrade may not be changed"
+	// +kubebuilder:validation:XValidation:rule="oldSelf == 'OKD' ? self == 'OKD' : true",message="OKD may not be changed"
```
JoelSpeed: I'm expecting the the OKD featureset is a subset of TechPreviewNoUpgrade which is also a subset of DevPreviewNoUpgrade. Have you considered allowing a transition from OKD to either TP or DP?

jatinsu: > Please add a test somewhere that ensures that a feature is not enabled in Default without being enabled in OKD as well

Sure thing! I'll get on that
> Please also link the work that will prevent the OKD featureset being enabled in OCP clusters

Here's what we added for the validation: https://github.com/openshift/kubernetes/commit/87b2e70effe2995ece0ae04578a77cef32ce4013

> Do you want this featureset to be opt-in? Or on by default in all OKD clusters?

It won't be enabled by default, the feature will be opt in through the installer

> I'm expecting the the OKD featureset is a subset of TechPreviewNoUpgrade which is also a subset of DevPreviewNoUpgrade. Have you considered allowing a transition from OKD to either TP or DP?

I haven't, but @Prashanth684 would this be something that we would be interested in?

Prashanth684:
> Please add a test somewhere that ensures that a feature is not enabled in Default without being enabled in OKD as well

@JoelSpeed is there a way to enforce this going forward? maybe as part of `verify-feature-promotion` command, we check that the feature is promoted to OKD if not already?

Prashanth684:
> I'm expecting the the OKD featureset is a subset of TechPreviewNoUpgrade which is also a subset of DevPreviewNoUpgrade. Have you considered allowing a transition from OKD to either TP or DP?

It's the other way around - from TP -> OKD -> Default. so it's everything in default featureset + few select ones from TP (which is to be determined based on feature stability + teams wanting to onboard to OKD)

hack/verify-okd-featuregates.sh
```diff
@@ -0,0 +1,16 @@
...
+
+source "$(dirname "${BASH_SOURCE}")/lib/init.sh"
+
+# Build codegen when it's not present and not overridden for a specific file.
+if [ -z "${CODEGEN:-}" ];then
+  ${TOOLS_MAKE} codegen
+  CODEGEN="${TOOLS_OUTPUT}/codegen"
+fi
+
+"${CODEGEN}" verify-okd-featuregates
```
Prashanth684: we should call this as part of verify-feature-promotion so it is seen as an issue that needs to be fixed. if it runs as its own optional test, folks won't care about the failure. Or maybe we can make it a required test. cc @JoelSpeed on his thoughts.

config/v1/types_feature.go
Prashanth684: > Do you want this featureset to be opt-in? Or on by default in all OKD clusters?

The idea is to make it default for OKD clusters through the installer, but there is a downside to it that no other featuresets can be enabled - unless we relax the kubebuilder validation and allow it to be changed. Do you see any negative implications of doing that?

hack/update-protobuf.sh
```diff
@@ -16,7 +16,7 @@ if [[ "${GOPATH}/src/github.com/openshift/api" != "${SCRIPT_ROOT}" ]]; then
   exit 1
 fi
 
-if [[ "$(protoc --version)" != "libprotoc 23."* ]]; then
+if [[ "$(protoc --version)" != "libprotoc 3."* ]]; then
```
Prashanth684: remove

config/v1/types_feature.go
Prashanth684: @JoelSpeed what are your thoughts on the above? ^

Prashanth684: i was thinking we could relax the validation like this:
```
// +kubebuilder:validation:XValidation:rule="oldSelf == 'OKD' ? self != 'Default' : true",message="OKD cannot transition to Default"
```

JoelSpeed: I had previously suggested similar, in that we would make sure that the OKD featureset can transition to TP/DP profiles, so yep, +1 from me

config/v1/types_feature.go
JoelSpeed: > It won't be enabled by default, the feature will be opt in through the installer

This contradicts something @Prashanth684 mentioned on another thread https://github.com/openshift/api/pull/2451/files#r2461166939

> I haven't, but @Prashanth684 would this be something that we would be interested in?

Prasanth also suggested the same, so lets make this change

hack/verify-promoted-features-pass-tests.sh
```diff
@@ -35,4 +35,10 @@ if [ -z "${CODEGEN:-}" ];then
...
 
-"${CODEGEN}" featuregate-test-analyzer
\ No newline at end of file
+# There's 2 commands to run, so we need to disable errexit so both commands can run
+set +e
+
+"${CODEGEN}" featuregate-test-analyzer
+
+# Verifies that all featuregates enabled in the Default featureset are also enabled in the OKD featureset
+"${CODEGEN}" verify-okd-featuregates
```
JoelSpeed: Nit, you're missing the final new line char at the end of this file

hack/verify-promoted-features-pass-tests.sh
```diff
@@ -35,4 +35,10 @@ if [ -z "${CODEGEN:-}" ];then
   CODEGEN="${TOOLS_OUTPUT}/codegen"
 fi
 
-"${CODEGEN}" featuregate-test-analyzer
\ No newline at end of file
+# There's 2 commands to run, so we need to disable errexit so both commands can run
```
JoelSpeed: Does this now correctly fail the verify step if L41 fails but L44 passes?

payload-manifests/featuregates/featureGate-Hypershift-OKD.yaml
JoelSpeed, line 1: Does hypershift OKD exist?

tools/codegen/cmd/verify-okd-featuregates.go
```diff
@@ -0,0 +1,204 @@
...
+	kyaml "sigs.k8s.io/yaml"
+)
+
+type VerifyOKDFeatureGatesOptions struct {
+	In     io.Reader
+	Out    io.Writer
+	ErrOut io.Writer
+
+	FeatureSetManifestDir string
+}
```
JoelSpeed: All of this is exported, but, I don't see a need for it to be?

tools/codegen/cmd/verify-okd-featuregates.go
```diff
@@ -0,0 +1,204 @@
...
+
+func NewVerifyOKDFeatureGatesOptions(in io.Reader, out, errOut io.Writer) *VerifyOKDFeatureGatesOptions {
```
JoelSpeed: Does this need to be exported?

tools/codegen/cmd/verify-okd-featuregates.go
```diff
@@ -0,0 +1,204 @@
...
+	return &VerifyOKDFeatureGatesOptions{
+		In:                    in,
+		Out:                   out,
+		ErrOut:                errOut,
```
JoelSpeed: Have you based this on another command within the codegen util?

tools/codegen/cmd/verify-okd-featuregates.go
```diff
@@ -0,0 +1,204 @@
...
+		return fmt.Errorf("--featureset-manifest-path is required")
+	}
+	if _, err := os.ReadDir(o.FeatureSetManifestDir); err != nil {
+		return fmt.Errorf("--featureset-manifest-path cannot be read: %w", err)
+	}
+	return nil
+}
+
+func (o *VerifyOKDFeatureGatesOptions) AddFlags(flags *pflag.FlagSet) {
+	flags.StringVar(&o.FeatureSetManifestDir, "featureset-manifest-path", o.FeatureSetManifestDir, "path to directory containing the FeatureGate YAMLs for each FeatureSet,ClusterProfile tuple.")
```
JoelSpeed: Do we not already have similar flags on other commands? Is this named the same?

tools/codegen/cmd/verify-okd-featuregates.go
JoelSpeed, line 1: Have you considered just writing a Go test in the features directory? There's an exported function to return all of the feature gating details as a map, I think this would be more efficient than writing a codegen util here

config/v1/types_feature.go
Prashanth684: > This contradicts something @Prashanth684 mentioned on another thread https://github.com/openshift/api/pull/2451/files#r2461166939

The more i think about this - i feel CVO should be the one dictating what featureset gets installed for OKD, not the installer. CVO should also migrate OKD clusters with default featureset to OKD featureset. I tried an implementation with some CVO changes on installs and upgrades and it worked fine. https://gist.github.com/Prashanth684/c650c14b8c2b02d820e67fe75153c669 captures the major changes needed to have OKD featureset by default on install and upgrades.

payload-manifests/featuregates/featureGate-Hypershift-OKD.yaml
Prashanth684: yes although not very common, we do have folks in the community who deploy Hypershift clusters with OKD.

tools/codegen/cmd/verify-okd-featuregates.go
jatinsu: > Have you considered just writing a Go test in the features directory?

Sorry just to be clear, do you mean creating a file such as `okd_featuregates_test.go` under this [features/ ](https://github.com/openshift/api/tree/master/features) directory? Or by placing it in one of the files in the features directory. [I have implemented it](https://github.com/jatinsu/openshift-api/blob/features-verify-okd-default/features/okd_featuregates_test.go) in the features directory and it does seem to work.
image

Prashanth684: i think the idea of having it as a codegen command was so that we could enforce during feature promotion to default - to make sure it was also promoted to OKD. A test would achieve the same effect, but my worry is that it might be ignored if it is not part of the feature promotion test

hack/verify-promoted-features-pass-tests.sh
jatinsu: After looking at, you're right no it doesn't as it doesn't capture the L41 error code. I will copy what is done in the [golangci-lin.sh](https://github.com/openshift/api/blob/master/hack/golangci-lint.sh) file, and after doing that it seems to be capturing the error correctly. Thanks!

image

tools/codegen/cmd/verify-okd-featuregates.go
jatinsu: Yes, the beginning parts were mainly lifted from [featureset-markdown.go](https://github.com/openshift/api/blob/master/tools/codegen/cmd/featureset-markdown.go).

tools/codegen/cmd/verify-okd-featuregates.go
jatinsu: Yeah I took it from here https://github.com/openshift/api/blob/50e2ece149d768a1b1eadae3a5f07134dda7b2ce/tools/codegen/cmd/featureset-markdown.go#L80

Should I remove it/rename it?

tools/codegen/cmd/verify-okd-featuregates.go
jatinsu: No it doesn't, you're right I'll go ahead and make it private

config/v1/types_feature.go
JoelSpeed: Do folks need a way to opt out here? Are we certain forcing every OKD cluster onto this OKD featureset is the right thing to do?

tools/codegen/cmd/verify-okd-featuregates.go
JoelSpeed: Our tests are required on every PR, verify feature promotion is occasionally overridden, but the tests never are. IMO tests will be more reliable here

config/v1/types_feature.go
```diff
@@ -53,8 +53,12 @@ var (
 	// your cluster may fail in an unrecoverable way.
 	CustomNoUpgrade FeatureSet = "CustomNoUpgrade"
 
+	// OKD turns on features for OKD. Turning this feature set ON is supported for OKD clusters, but NOT for OpenShift clusters
+	// this feature set on CANNOT BE UNDONE for OKD clusters and when enabled on OpenShift clusters it PREVENTS UPGRADES.
```
JoelSpeed:
```suggestion
	// OKD turns on features for OKD. Turning this feature set ON is supported for OKD clusters, but NOT for OpenShift clusters.
	// Turning this feature set on CANNOT BE UNDONE for OKD clusters and when enabled on OpenShift clusters it PREVENTS UPGRADES.
```

I thought we would prevent it on OpenShift completely rather than just preventing upgrades?

jatinsu: yep you're correct. I edited this comment and I forgot to push the commit. I'll rebase and fix this comment

tools/codegen/cmd/verify-okd-featuregates.go
Prashanth684: sounds good! @jatinsu let's go with the approach of adding this check to the tests.

features/okd_featureset_parity_test.go
```diff
@@ -0,0 +1,53 @@
+package features
```
jatinsu, line 1: @JoelSpeed Would you mind taking a look at how I implemented the verification test? I'm assuming [this comment](https://github.com/openshift/api/pull/2451#discussion_r2486365340) meant placing it in this features directory. I think this is this last thing we need before a lgtm/approval, unless there's something I missed.

features/okd_featureset_parity_test.go
```diff
@@ -0,0 +1,53 @@
...
+			sort.Strings(missingList)
+
+			t.Errorf("ClusterProfile %q: OKD featureset is missing %d featuregate(s) that are enabled in Default:\n  - %s\n\nAll featuregates enabled in Default must also be enabled in OKD.",
+				clusterProfile,
+				missingInOKD.Len(),
+				strings.Join(missingList, "\n  - "),
+			)
+		}
+	}
+}
```
JoelSpeed: Nit, missing final new line char here

jatinsu:
> Can you push a DNM commit that shows that this test fails our CI?

Just did that, and the unit test [failed and denotes the right failure message](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_api/2451/pull-ci-openshift-api-master-unit/1996619090684284928). Is there anything else that you would like for us to test/anything else I missed in this PR?

JoelSpeed:
Looks good, if you drop that DNM commit I'll tag the PR, thanks

jatinsu:
Just did. Thanks for your patience for this PR!

PR #2460, opened 2025-08-26
OPNET-678: Add internalDNSRecords field

cybertron:
Adds a new field to the Infrastructure object that allows control over the behavior of the on-prem internal DNS records. It can be set to either Internal or External, where Internal is the default (and previous) behavior, and External suppresses generation of the internal records so they can be managed outside the cluster. Currently, External can only be used with a UserManaged loadbalancer.


config/v1/types_infrastructure.go
```diff
@@ -1026,6 +1033,15 @@ type BareMetalPlatformStatus struct {
 	// +optional
 	LoadBalancer *BareMetalPlatformLoadBalancer `json:"loadBalancer,omitempty"`
 
+	// internalDNSRecords determines whether we deploy with internal records enabled for
+	// api, api-int, and ingress.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=Enabled;Disabled
+	// +openshift:validation:featureGate=OnPremInternalDNSRecords
```
everettraven: AFAIK, this isn't a real marker. Let's remove this.

config/v1/types_infrastructure.go
```diff
@@ -1026,6 +1033,15 @@ type BareMetalPlatformStatus struct {
 	// +optional
 	LoadBalancer *BareMetalPlatformLoadBalancer `json:"loadBalancer,omitempty"`
 
+	// internalDNSRecords determines whether we deploy with internal records enabled for
+	// api, api-int, and ingress.
+	// +kubebuilder:validation:Optional
```
everettraven: This is essentially a duplicate of the `+optional` marker below. We prefer the use of the `+optional` marker so let's remove this one.

config/v1/types_infrastructure.go
```diff
@@ -1026,6 +1033,15 @@ type BareMetalPlatformStatus struct {
...
+	// +kubebuilder:validation:Enum=Enabled;Disabled
+	// +openshift:validation:featureGate=OnPremInternalDNSRecords
+	// +openshift:enable:FeatureGate=OnPremInternalDNSRecords
+	// +optional
+	InternalDNSRecords InternalDNSRecordsType `json:"internalDNSRecords"`
```
everettraven: Because this is an optional field and the zero value is invalid, this should have `omitempty`.

config/v1/types_infrastructure.go
```diff
@@ -183,6 +183,13 @@ const (
 	LoadBalancerTypeOpenShiftManagedDefault PlatformLoadBalancerType = "OpenShiftManagedDefault"
 )
 
+type InternalDNSRecordsType string
+
+const (
+	InternalDNSRecordsDisabled InternalDNSRecordsType = "Disabled"
+	InternalDNSRecordsEnabled InternalDNSRecordsType = "Enabled"
+)
```
everettraven: Normally we try to avoid the terminology `Enabled` and `Disabled` where possible because the terms can often be overloaded and cause confusion.

What if instead of naming the field this applies to `internalDNSRecords`, what if we named it something like `dnsRecordsPolicy` (or maybe `dnsRecordsType`? not sure which one is better) and we had `Internal` and `External` as the options?

config/v1/types_infrastructure.go
```diff
@@ -1026,6 +1033,15 @@ type BareMetalPlatformStatus struct {
 	// +optional
 	LoadBalancer *BareMetalPlatformLoadBalancer `json:"loadBalancer,omitempty"`
 
+	// internalDNSRecords determines whether we deploy with internal records enabled for
+	// api, api-int, and ingress.
```
everettraven: Please include validation constraints in the GoDoc here so that this is more end-user friendly. This is the text used in our generated API documentation and what users will see when they use something like `oc explain ...` so we should make sure it reads appropriately as end-user documentation.

Some good guidelines for things to take into consideration for inclusion in the GoDoc are here: https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#write-user-readable-documentation-in-godoc

config/v1/types_infrastructure.go
```diff
@@ -1026,6 +1033,15 @@ type BareMetalPlatformStatus struct {
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=Enabled;Disabled
+	// +openshift:validation:featureGate=OnPremInternalDNSRecords
+	// +openshift:enable:FeatureGate=OnPremInternalDNSRecords
+	// +optional
+	InternalDNSRecords InternalDNSRecordsType `json:"internalDNSRecords"`
+
```
everettraven: This looks like we have only added this to the `BareMetalPlatformStatus` type? Is this because the OpenShift installer will end up setting this value at install time?

everettraven: If I understand the EP correctly as well, this sounds like this should only be possible to set when `loadBalancer` is set to `UserManaged`?

Do we need some additional validation logic (maybe a CEL expression) to enforce that?

config/v1/types_infrastructure.go
cybertron: Heh, yeah I was throwing stuff at the wall when I had trouble with the feature gate.

config/v1/types_infrastructure.go
cybertron: Yes, this will be populated by the installer. I have a validation in the installer to ensure it isn't set when it shouldn't be, but I can move that here if it would be better.

I should also note that this is only a partial version of the change. Because these are per-platform types we'll need to apply the same change to the other on-prem platforms once we know what it should look like.

cybertron:
I think this latest revision should address all of the comments so far, except moving the validation to the api layer. If we ever make this modifiable after initial install we'll have to do that, but I'd just as soon defer that effort until/if it's needed. Willing to move it here if you'd prefer though.

config/v1/types_infrastructure.go
```diff
@@ -1035,6 +1042,18 @@ type BareMetalPlatformStatus struct {
 	// +optional
 	LoadBalancer *BareMetalPlatformLoadBalancer `json:"loadBalancer,omitempty"`
 
+	// DNSRecordsType determines whether records for api, api-int, and ingress
```
JoelSpeed:
```suggestion
	// dnsRecordsType determines whether records for api, api-int, and ingress
```

config/v1/types_infrastructure.go
```diff
@@ -1035,6 +1042,18 @@ type BareMetalPlatformStatus struct {
...
+	// are provided by the internal DNS service or externally. `Internal` configures
+	// DNS records in the internal service. `External` means no records will be
```
JoelSpeed: Internal allows DNS resolution for components within the cluster right? It's configuring coredns?

config/v1/types_infrastructure.go
```diff
@@ -1035,6 +1042,18 @@ type BareMetalPlatformStatus struct {
...
+	// provided and must be configured external to the cluster. `External` is only
+	// allowed when a user-managed loadbalancer is configured. When unset, the
+	// internal records will be provided.
+	// api, api-int, and ingress.
```
everettraven: For enum based fields we generally try to follow a pattern like:

> dnsRecordsType ...
> Allowed values are Internal, External, and omitted.
> When set to Internal, ...
> When set to External, ...
> When omitted, ...

config/v1/types_infrastructure.go
```diff
@@ -183,6 +183,13 @@ const (
 	LoadBalancerTypeOpenShiftManagedDefault PlatformLoadBalancerType = "OpenShiftManagedDefault"
 )
 
+type DNSRecordsType string
+
+const (
+	DNSRecordsExternal DNSRecordsType = "External"
+	DNSRecordsInternal DNSRecordsType = "Internal"
```
everettraven:
```suggestion
	DNSRecordsTypeExternal DNSRecordsType = "External"
	DNSRecordsTypeInternal DNSRecordsType = "Internal"
```

everettraven:
>If we ever make this modifiable after initial install we'll have to do that, but I'd just as soon defer that effort until/if it's needed. Willing to move it here if you'd prefer though.

Is this API currently immutable? How do we enforce that?

config/v1/types_infrastructure.go
cybertron: It's configuring a coredns instance on each host. This is separate from the coredns pod that is part of the dns-operator.

cybertron:
> Is this API currently immutable? How do we enforce that?

This field is only part of the Infrastructure status. There's nothing in the spec to allow it to be modified.

cybertron:
Oh shoot, I had some stale changes in my editor. I'll fix that.

config/v1/types_infrastructure.go
```diff
@@ -1035,6 +1042,19 @@ type BareMetalPlatformStatus struct {
 	// +optional
 	LoadBalancer *BareMetalPlatformLoadBalancer `json:"loadBalancer,omitempty"`
 
+	// dnsRecordsType determines whether records for api, api-int, and ingress
+	// are provided by the internal DNS service or externally.
+	// Allowed values are `Internal`, `External`, and omitted.
+	// When set to `Internal`, records are provided by the internal infrastructure
+	// When set to `External`, records are not provided by the internal infrastructure
```
everettraven, line 1094:
```suggestion
	// When set to `External`, records are not provided by the external infrastructure
```
?

config/v1/types_infrastructure.go
```diff
@@ -1035,6 +1042,19 @@ type BareMetalPlatformStatus struct {
...
+	// and must be configured by the user. This value may only be set when a
+	// user-managed loadbalancer is configured.
+	// When omitted, the behavior will be the same as `Internal`.
```
everettraven: For the omitted scenario we have some standard terminology we've used throughout OpenShift APIs:
>When omitted, this means the user has no opinion and the platform is left
>to choose reasonable defaults. These defaults are subject to change over time.
>The current default is {default}

config/v1/types_infrastructure.go
cybertron: There's a "not" in the external description that differentiates the two. I wrote it that way because ultimately this feature controls the behavior of the internal DNS.

everettraven: Ah, whoops - I read to quickly and missed the "not" - my apologies.

cybertron:
Rebased and marking ready for review

everettraven:
@cybertron It looks like this needs a rebase. Could you also update the PR description to provide some additional information as to what work with PR is associated with?

Thanks!

cybertron:
Okay, I think this one will finally pass all the tests. I had claude add a validation for the new field, which was a bit of an adventure but seems to be correct now. I'm still planning to validate the input in the installer too so that was mostly to make the AI review happy. Also added a Jira reference and description to the PR.

config/v1/types_infrastructure.go
```diff
@@ -1074,6 +1082,21 @@ type BareMetalPlatformStatus struct {
...
+	// are provided by the internal DNS service or externally.
+	// Allowed values are `Internal`, `External`, and omitted.
+	// When set to `Internal`, records are provided by the internal infrastructure
+	// When set to `External`, records are not provided by the internal infrastructure
+	// and must be configured by the user. This value may only be set when a
+	// user-managed loadbalancer is configured.
+	// When omitted, this means the user has no opinion and the platform is left
+	// to choose reasonable defaults. These defaults are subject to change over time.
+	// The current default is `Internal`.
+	// +kubebuilder:validation:Enum=Internal;External
```
everettraven: Let's move this tag to the actual `DNSRecordsType` alias type. Then this gets picked up on all the fields that use this type.

config/v1/types_infrastructure.go
```diff
@@ -1074,6 +1082,21 @@ type BareMetalPlatformStatus struct {
 	// +optional
 	LoadBalancer *BareMetalPlatformLoadBalancer `json:"loadBalancer,omitempty"`
 
+	// dnsRecordsType determines whether records for api, api-int, and ingress
+	// are provided by the internal DNS service or externally.
+	// Allowed values are `Internal`, `External`, and omitted.
+	// When set to `Internal`, records are provided by the internal infrastructure
+	// When set to `External`, records are not provided by the internal infrastructure
+	// and must be configured by the user. This value may only be set when a
+	// user-managed loadbalancer is configured.
```
everettraven: Could you update this to mention the explicit field and the value that must be set to satisfy this condition?

config/v1/zz_generated.crd-manifests/0000_10_config-operator_01_infrastructures-CustomNoUpgrade.crd.yaml
```diff
@@ -2226,6 +2247,26 @@ spec:
                           rule: 'self == oldSelf || (size(self) == 2 && isIP(self[0])
                             && isIP(self[1]) ? ip(self[0]).family() != ip(self[1]).family()
                             : true)'
+                      dnsRecordsType:
+                        allOf:
```
cybertron: Is this right? It looks weird, but this is what was generated when I moved the enum definition.

everettraven: That is not correct. You need to make sure you remove the enum marker on _all_ the fields that are using the type now. They will all inherit the marker from the type marker now.

everettraven: I should have been more specific in my previous comment - my apologies

cybertron: D'oh, of course. Completely forgot that this is duplicated across all on-prem platforms. Fixing.

config/v1/types_infrastructure.go
```diff
@@ -183,6 +183,16 @@ const (
 	LoadBalancerTypeOpenShiftManagedDefault PlatformLoadBalancerType = "OpenShiftManagedDefault"
 )
 
+// DNSRecordsType defines whether api, api-int, and ingress records are provided by
+// the internal DNS infrastructure or must be configured external to the cluster.
+// +kubebuilder:validation:Enum=Internal;External
```
JoelSpeed, line 188: Add this one too to help with accurate openapi schema generation
```suggestion
// +kubebuilder:validation:Enum=Internal;External
// +enum
```

config/v1/types_infrastructure.go
```diff
@@ -1074,6 +1085,20 @@ type BareMetalPlatformStatus struct {
 	// +optional
 	LoadBalancer *BareMetalPlatformLoadBalancer `json:"loadBalancer,omitempty"`
 
+	// dnsRecordsType determines whether records for api, api-int, and ingress
+	// are provided by the internal DNS service or externally.
+	// Allowed values are `Internal`, `External`, and omitted.
+	// When set to `Internal`, records are provided by the internal infrastructure
```
JoelSpeed: What does this actually mean? This doesn't really tell me much as an end user. Does it mean that components internally will be able to resolve the endpoints without further config?
```suggestion
	// When set to `Internal`, records are provided by the internal infrastructure.
```

config/v1/types_infrastructure.go
```diff
@@ -1074,6 +1085,20 @@ type BareMetalPlatformStatus struct {
...
+	// When set to `External`, records are not provided by the internal infrastructure
+	// and must be configured by the user. This value may only be set when
```
JoelSpeed: So the users have to provide their own external DNS solution that can resolve the DNS names of the api, api-int and ingress?

config/v1/types_infrastructure.go
cybertron: I will expand this to clarify what the user needs to do (or not do) in both cases.

config/v1/types_infrastructure.go
cybertron: Yes

cybertron:
/verified by @cybertron 

I've tested this by vendoring it into the affected projects in my local dev cluster.

cybertron:
/retest-required

This should have no effect on an aws cluster. Looks like maybe a quota issue?

cybertron:
/retest-required

As far as I can tell, this is failing on the following error:
```
2025-11-13T16:10:21.817654562Z I1113 16:10:21.817595       1 requests.go:294] Making HTTP GET request at: https://console.redhat.com/api/insights-results-aggregator/v2/cluster/30461bca-facb-4b6c-9cc0-eb7feda89f9b/request/5399d52611ba4030b31666bac1f43a03/status
2025-11-13T16:10:51.915983970Z I1113 16:10:51.915917       1 gather_commands.go:443] Received HTTP status code 504, retry 3/3 in 1m0s
```
Which appears to be a server-side issue on console.redhat.com.

cybertron:
/retest-required

And now we're failing on rate limiting from AWS.

cybertron:
/retest-required

I see this job finally passed last night, hopefully the issues have been fixed.

cybertron:
/retest-required

Failed on a totally different set of tests this time. :-/

PR #2468, opened 2025-09-03
MON-4359: Add Monitoring Capability

rexagod:
Add a "Monitoring" capability to support optional monitoring.

### DIFF
--- a/config/v1/types_cluster_version.go
+++ b/config/v1/types_cluster_version.go
@@ -278,7 +278,7 @@ const (
 )
 
 // ClusterVersionCapability enumerates optional, core cluster components.
-// +kubebuilder:validation:Enum=openshift-samples;baremetal;marketplace;Console;Insights;Storage;CSISnapshot;NodeTuning;MachineAPI;Build;DeploymentConfig;ImageRegistry;OperatorLifecycleManager;CloudCredential;Ingress;CloudControllerManager;OperatorLifecycleManagerV1
+// +kubebuilder:validation:Enum=openshift-samples;baremetal;marketplace;Console;Insights;Storage;CSISnapshot;NodeTuning;MachineAPI;Build;DeploymentConfig;ImageRegistry;OperatorLifecycleManager;CloudCredential;Ingress;CloudControllerManager;OperatorLifecycleManagerV1;OptionalMonitoring
 type ClusterVersionCapability string
 
 const (
@@ -399,6 +399,20 @@ const (
 	// Managers deployed on top of OpenShift. They help you to work with cloud
 	// provider API and embeds cloud-specific control logic.
 	ClusterVersionCapabilityCloudControllerManager ClusterVersionCapability = "CloudControllerManager"
+
+	// ClusterVersionCapabilityOptionalMonitoring manages the enablement of the optional components of the in-cluster monitoring stack.
+	// The optional components are:
+	// * Alertmanager
+	// * Alermanager User Workload Monitoring
+	// * Prometheus User Workload Monitoring
+	// * PrometheusOperator User Workload Monitoring
+	// * ThanosRuler User Workload Monitoring
+	//
+	// NOTE: The only supported use-case for this capability is to reduce the monitoring stack's resource usage by only
+	// supporting telemetry (see https://rhobs-handbook.netlify.app/products/openshiftmonitoring/telemetry.md). Turning
+	// this capability off will trigger CMO to not deploy the optional components, as well as any related manifests.
+	// However, all CRDs will persist as these are needed by the operator regardless of this capability's state.
+	ClusterVersionCapabilityOptionalMonitoring ClusterVersionCapability = "OptionalMonitoring"
 )
 
 // KnownClusterVersionCapabilities includes all known optional, core cluster components.
@@ -420,10 +434,11 @@ var KnownClusterVersionCapabilities = []ClusterVersionCapability{
 	ClusterVersionCapabilityCloudCredential,
 	ClusterVersionCapabilityIngress,
 	ClusterVersionCapabilityCloudControllerManager,
+	ClusterVersionCapabilityOptionalMonitoring,
 }
 
 // ClusterVersionCapabilitySet defines sets of cluster version capabilities.
-// +kubebuilder:validation:Enum=None;v4.11;v4.12;v4.13;v4.14;v4.15;v4.16;v4.17;v4.18;vCurrent
+// +kubebuilder:validation:Enum=None;v4.11;v4.12;v4.13;v4.14;v4.15;v4.16;v4.17;v4.18;v4.19;v4.20;vCurrent
 type ClusterVersionCapabilitySet string
 
 const (
@@ -478,6 +493,18 @@ const (
 	// OpenShift.  This list will remain the same no matter which
 	// version of OpenShift is installed.
 	ClusterVersionCapabilitySet4_18 ClusterVersionCapabilitySet = "v4.18"
+
+	// ClusterVersionCapabilitySet4_19 is the recommended set of
+	// optional capabilities to enable for the 4.19 version of
+	// OpenShift.  This list will remain the same no matter which
+	// version of OpenShift is installed.
+	ClusterVersionCapabilitySet4_19 ClusterVersionCapabilitySet = "v4.19"
+
+	// ClusterVersionCapabilitySet4_20 is the recommended set of
+	// optional capabilities to enable for the 4.20 version of
+	// OpenShift.  This list will remain the same no matter which
+	// version of OpenShift is installed.
+	ClusterVersionCapabilitySet4_20 ClusterVersionCapabilitySet = "v4.20"
 
 	// ClusterVersionCapabilitySetCurrent is the recommended set
 	// of optional capabilities to enable for the cluster's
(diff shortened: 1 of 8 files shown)


config/v1/types_cluster_version.go
```diff
@@ -399,6 +399,15 @@ const (
 	// Managers deployed on top of OpenShift. They help you to work with cloud
 	// provider API and embeds cloud-specific control logic.
 	ClusterVersionCapabilityCloudControllerManager ClusterVersionCapability = "CloudControllerManager"
+
+	// ClusterVersionCapabilityOptionalMonitoring manages the cluster monitoring stack which is responsible for gathering and
+	// processing metrics from the in-house and user workloads. The following CRDs are constitute this capability:
+	// - TODO
```
everettraven: Presumably we should not add this capability until we know the full scope of what this capability will and will not do?

config/v1/types_cluster_version.go
```diff
@@ -399,6 +399,15 @@ const (
...
+	//
+	// WARNING: This capability will drop all aforementioned CRDs, and may operational issues in the cluster.
+	// The only supported use-case for this capability is to reduce the monitoring stack's resource usage by only
+	// supporting telemetry.
```
everettraven: What operational issues may happen here?

config/v1/types_cluster_version.go
```diff
@@ -399,6 +399,15 @@ const (
 	// Managers deployed on top of OpenShift. They help you to work with cloud
 	// provider API and embeds cloud-specific control logic.
 	ClusterVersionCapabilityCloudControllerManager ClusterVersionCapability = "CloudControllerManager"
+
+	// ClusterVersionCapabilityOptionalMonitoring manages the cluster monitoring stack which is responsible for gathering and
+	// processing metrics from the in-house and user workloads. The following CRDs are constitute this capability:
```
everettraven: Is it only these CRDs that are removed when this capability is disabled? Are there default monitoring workloads that may no longer be running on the cluster when this is disabled?

config/v1/types_cluster_version.go
everettraven: What does "only supporting telemetry" mean here?

config/v1/types_cluster_version.go
```diff
@@ -479,6 +489,18 @@ const (
 	// version of OpenShift is installed.
 	ClusterVersionCapabilitySet4_18 ClusterVersionCapabilitySet = "v4.18"
 
+	// ClusterVersionCapabilitySet4_19 is the recommended set of
+	// optional capabilities to enable for the 4.18 version of
```
everettraven:
```suggestion
	// optional capabilities to enable for the 4.19 version of
```

config/v1/types_cluster_version.go
```diff
@@ -479,6 +489,18 @@ const (
...
+	// OpenShift.  This list will remain the same no matter which
+	// version of OpenShift is installed.
+	ClusterVersionCapabilitySet4_19 ClusterVersionCapabilitySet = "v4.19"
+
+	// ClusterVersionCapabilitySet4_20 is the recommended set of
+	// optional capabilities to enable for the 4.18 version of
```
everettraven:
```suggestion
	// optional capabilities to enable for the 4.20 version of
```

config/v1/types_cluster_version.go
```diff
@@ -600,6 +622,46 @@ var ClusterVersionCapabilitySets = map[ClusterVersionCapabilitySet][]ClusterVers
...
+		ClusterVersionCapabilityMachineAPI,
+		ClusterVersionCapabilityBuild,
+		ClusterVersionCapabilityDeploymentConfig,
+		ClusterVersionCapabilityImageRegistry,
+		ClusterVersionCapabilityOperatorLifecycleManager,
+		ClusterVersionCapabilityOperatorLifecycleManagerV1,
+		ClusterVersionCapabilityCloudCredential,
+		ClusterVersionCapabilityIngress,
+		ClusterVersionCapabilityCloudControllerManager,
+		ClusterVersionCapabilityOptionalMonitoring,
```
everettraven: Why should this be included in the 4.19 set if the capability does not exist in 4.19?

config/v1/types_cluster_version.go
```diff
@@ -600,6 +622,46 @@ var ClusterVersionCapabilitySets = map[ClusterVersionCapabilitySet][]ClusterVers
...
```
everettraven: Why should this be included in the 4.20 capability list if this capability does not exist in 4.20?

rexagod:
@everettraven Apologies! It seems there was some [development](https://github.com/openshift/cluster-monitoring-operator/pull/2675/commits) before I got a chance to update this after marking it ready.

I'll make changes to implicitly enable the capability for TechPreview only in 4.21.

config/v1/types_cluster_version.go
rexagod: Added context.

Operations relying the optional components, i.e., Alertmanager, and all UWM components will be disabled. For e.g., the "Alerts" tab under "Observe" will stop showing alerting (firing, pending, etc.).

Telemetry here refers to https://rhobs-handbook.netlify.app/products/openshiftmonitoring/telemetry.md that teams can use to gather telemetry metrics about their workloads from clusters.

config/v1/types_cluster_version.go
rexagod: Amended. Disabling the capability drops all optional monitoring components (AM and the UWM stack) and any manifests related to them. Do note that CRDs will persist as these are needed by the operator regardless of the capability's state.

config/v1/types_cluster_version.go
rexagod: Amended.

config/v1/types_cluster_version.go
rexagod: Dropped, targeting 4.21 TechPreview now.

config/v1/types_cluster_version.go
rexagod: Dropped, targeting 4.21 TechPreview now.

rexagod:
@everettraven I'm probably missing something but I expected to find something similar to the feature-gates workflow (annotations/tags and such) to enable this just for TechPreview. Could you please drop some pointers on how I can achieve that? Thanks!

everettraven:
@rexagod As far as I am aware, the process for tech preview capabilities is the same as adding any new enum value under TechPreviewNoUpgrade, meaning:
- An OpenShift Enhancement Proposal
- A new feature gate
- Adding the new enum using a feature-gate aware enum marker

I'm relatively new to capabilities for OpenShift so I read up on the original intention of them in https://github.com/openshift/enhancements/blob/df1fada1e045b2e81b677941414f3085c02945fd/enhancements/installer/component-selection.md 

With my current understanding of how capabilities work it should be an all-or-nothing type configuration. It seems like an anti-pattern to have a partial capability.

I'd like to see answered in the EP you write up:
- Why your cluster operator cannot be responsible for removing optional monitoring configurations
- What is and is not included in this capability and why it is okay to exclude some monitoring components and not others
- What other components may have a dependency on these now optional components

Thanks!

Edit to clarify: Re-reading the description of the capability, I'm not sure a capability is what you are actually looking for here. This seems more like an option that should be configured on and managed by the CMO itself.

rexagod:
👋🏼 I'm a bit confused about the EP and feature-gate part, since https://github.com/openshift/enhancements/blob/df1fada1e045b2e81b677941414f3085c02945fd/enhancements/installer/component-selection.md#how-to-implement-a-new-capability doesn't mention those?

> What is and is not included in this capability and why it is okay to exclude some monitoring components and not others

I've tried to answer this in the godoc for the capability but essentially this allows us to tune down the resource consumption of the in-cluster monitoring stack by not deploying optional components.

> Why your cluster operator cannot be responsible for removing optional monitoring configurations
> Re-reading the description of the capability, I'm not sure a capability is what you are actually looking for here. This seems more like an option that should be configured on and managed by the CMO itself.

We chose to go with capabilities to support configuring these at install time for clusters that are only concerned with forwarding telemetry metrics.

> What other components may have a dependency on these now optional components

The optional components in question are Alertmanager and all UserWorkloadMonitoring components managed by CMO. The latter is opt-out by default, so no dependencies there. For AM, as I mentioned [earlier](https://github.com/openshift/api/pull/2468#discussion_r2424940382), this could impact any areas that expect an AM presence 100% of the time (so far, monitoring-plugin, which we'll fix down the line to hide the "Alerts" tab when AM is absent).

everettraven:
>I'm a bit confused about the EP and feature-gate part, since https://github.com/openshift/enhancements/blob/df1fada1e045b2e81b677941414f3085c02945fd/enhancements/installer/component-selection.md#how-to-implement-a-new-capability doesn't mention those?

This document looks like it hasn't been updated in the last couple years. Our processes have changed throughout this time such that any API change, especially to a v1 API, must go behind a feature gate in TechPreviewNoUpgrade. Creating a feature gate requires an OpenShift enhancement proposal. This is meant to ensure that we:
- Don't inadvertently ship something as GA without a backing implementation
- Don't GA something without appropriate regression testing

This is all meant to ensure we are maintaining a quality bar for changes that go into the product.

>I've tried to answer this in the godoc for the capability but essentially this allows us to tune down the resource consumption of the in-cluster monitoring stack by not deploying optional components.

I appreciate you taking the time to answer some of this in the GoDoc. I'm still not quite sure I have an understanding from reading that why some pieces of the monitoring stack we deploy by default is considered optional and why others are not.

> We chose to go with capabilities to support configuring these at install time for clusters that are only concerned with forwarding telemetry metrics.

My understanding of capabilities is that they are all-or-nothing type mechanisms to tell the ClusterVersionOperator whether or not to stamp $thing out onto the cluster, not a way to tell cluster operators what they should and should not be doing. I don't think this partial use case is one that the "capability" concept was intended to support.

I'm not sure off the top of my head if there is another way to achieve what you are looking for at install time, but maybe @JoelSpeed would be more familiar with the options you have here.

I've got a few questions here that having an EP would be helpful to field these discussion:
- Why does a user need to configure this at install time?
- What is the impact to an end user if they can only configure this option as a "day 2" operation?
- Should users be able to enable/disable the optional pieces of the monitoring stack at will?

config/v1/types_cluster_version.go
```diff
@@ -399,6 +399,20 @@ const (
...
+	// ClusterVersionCapabilityOptionalMonitoring manages the enablement of the optional components of the in-cluster monitoring stack.
+	// The optional components are:
+	// * Alertmanager
+	// * Alermanager User Workload Monitoring
+	// * Prometheus User Workload Monitoring
+	// * PrometheusOperator User Workload Monitoring
+	// * ThanosRuler User Workload Monitoring
+	//
+	// NOTE: The only supported use-case for this capability is to reduce the monitoring stack's resource usage by only
+	// supporting telemetry (see https://rhobs-handbook.netlify.app/products/openshiftmonitoring/telemetry.md). Turning
```
simonpasquier, line 412: I would rather link to the official docs: https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/support/remote-health-monitoring-with-connected-clusters

config/v1/types_cluster_version.go
```diff
@@ -399,6 +399,20 @@ const (
 	// Managers deployed on top of OpenShift. They help you to work with cloud
 	// provider API and embeds cloud-specific control logic.
 	ClusterVersionCapabilityCloudControllerManager ClusterVersionCapability = "CloudControllerManager"
+
+	// ClusterVersionCapabilityOptionalMonitoring manages the enablement of the optional components of the in-cluster monitoring stack.
+	// The optional components are:
+	// * Alertmanager
+	// * Alermanager User Workload Monitoring
```
simonpasquier, line 406: should we say "All user-defined monitoring components"?

PR #2463, opened 2025-09-01
WIP: Monitoring API prometheus k8s config API

marioferh:

PR #2461, opened 2025-08-29
MON-4029: Add KubeStateMetrics api

marioferh:
after: https://github.com/openshift/api/pull/2322

PR #2481, opened 2025-09-17
MON-4031: Add prometheusOperatorConfig API

marioferh:


config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -416,6 +421,74 @@ type MetricsServerConfig struct {
...
+	// Allowed values are error, warn, info, and debug.
+	// When set to error, only errors will be logged.
+	// When set to warn, both warnings and errors will be logged.
+	// When set to info, general information, warnings, and errors will all be logged.
+	// When set to debug, detailed debugging information will be logged.
+	// When omitted, this means no opinion and the platform is left to choose a reasonable default, that is subject to change over time.
+	// The current default value is `info`.
+	// +optional
+	// +kubebuilder:validation:MaxLength=10
+	LogLevel *string `json:"logLevel,omitempty"`
```
everettraven: Should be an enum.

Is setting this to `""` any different than omitting the field?

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -416,6 +421,74 @@ type MetricsServerConfig struct {
...
+	// nodeSelector defines the nodes on which the Pods are scheduled
+	// nodeSelector is optional.
+	//
+	// When omitted, this means the user has no opinion and the platform is left
+	// to choose reasonable defaults. These defaults are subject to change over time.
+	// The current default value is `kubernetes.io/os: linux`.
+	// +optional
+	// +kubebuilder:validation:MinProperties=1
+	// +kubebuilder:validation:MaxProperties=10
```
everettraven, line 451: Include the constraints in the godoc please.

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -416,6 +421,74 @@ type MetricsServerConfig struct {
...
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+	// resources defines the compute resource requests and limits for the Prometheus Operator container.
+	// This includes CPU, memory and HugePages constraints to help control scheduling and resource usage.
+	// When not specified, defaults are used by the platform. Requests cannot exceed limits.
+	// This field is optional.
+	// More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
+	// +optional
+	Resources *v1.ResourceRequirements `json:"resources,omitempty"`
```
everettraven: Presumably this should follow the pre-existing pattern used in https://github.com/openshift/api/pull/2461/files#diff-0b485b16eab292251411851da7d69701a168005a26dbdb76e93ca4f27a91d002R510 ?

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -94,6 +94,11 @@ type ClusterMonitoringSpec struct {
 	// When omitted, this means no opinion and the platform is left to choose a reasonable default, which is subject to change over time.
 	// +optional
 	MetricsServerConfig MetricsServerConfig `json:"metricsServerConfig,omitempty,omitzero"`
+	// prometheusOperatorConfig is an optional field that can be used to configure the Prometheus Operator component.
+	// Specifically, it can configure how the Prometheus Operator instance is deployed, pod scheduling, and resource allocation.
+	// When omitted, this means no opinion and the platform is left to choose a reasonable default, which is subject to change over time.
+	// +optional
+	PrometheusOperatorConfig PrometheusOperatorConfig `json:"prometheusOperatorConfig,omitempty,omitzero"`
```
everettraven, line 101: How does this configuration relate to the configuration proposed in https://github.com/openshift/api/pull/2463?

config/v1alpha1/types_cluster_monitoring.go
marioferh: right, wrong c&p

config/v1alpha1/types_cluster_monitoring.go
marioferh: done

config/v1alpha1/types_cluster_monitoring.go
marioferh: correct. Done

config/v1alpha1/types_cluster_monitoring.go
marioferh: This is Prometheus Operator, the other one is Prometheus config. Of course they are related but they have different configs.

everettraven: Is the Prometheus config used by the PrometheusOperator?

Would it make sense to co-locate the configurations under a top-level `prometheus` field?

marioferh: Not directly . Prometheus Config is use by Prometheus. PrometheusOperator manages Prometheus instances, a
Alertmanagare, etc.

everettraven: So what configures the Prometheus instances created by the Prometheus Operator to use the Prometheus Config?

marioferh: I am not sure what is the question. Prometheus Operator manages Prometheus instances. Prometheus instances need a config, that config is PrometheusK8sConfig

everettraven: I'm trying to better understand the relationship between the two to determine whether or not it makes sense to group them under a common spec field.

For example:
```yaml
kind: ClusterMonitoring
spec:
  userDefined:
  mode: "Disabled"
  prometheus:
    operatorConfig:
      ...
    instanceConfig:
      ...
```

I'm understanding that the PrometheusOperatorConfig would be read by the monitoring operator to configure the deployment of the PrometheusOperatorConfig.

Who would be reading and acting on the PrometheusK8sConfig?

marioferh: - Who would be reading and acting on the PrometheusK8sConfig?
It's CMO, not PrometheusOperator

everettraven: What will the CMO _do_ with the PrometheusK8sConfig?

marioferh: CMO takes PrometheusK8sConfing configmap and create a CR. 
PrometheosOperator takes that CR and configure Prometheus.

I can understand your idea but PrometheusOperator manages all these components and I it's not a good idea to have all fields inside PrometheusOperator.
```
A core feature of the Prometheus Operator is to monitor the Kubernetes API server for changes to specific objects and ensure that the current Prometheus deployments match these objects. The Operator acts on the following [Custom Resource Definitions (CRDs)](https://kubernetes.io/docs/tasks/access-kubernetes-api/extend-api-custom-resource-definitions/):

    Prometheus, which defines a desired Prometheus deployment.

    PrometheusAgent, which defines a desired Prometheus deployment, but running in Agent mode.

    Alertmanager, which defines a desired Alertmanager deployment.

    ThanosRuler, which defines a desired Thanos Ruler deployment.

    ServiceMonitor, which declaratively specifies how groups of Kubernetes services should be monitored. The Operator automatically generates Prometheus scrape configuration based on the current state of the objects in the API server.

    PodMonitor, which declaratively specifies how group of pods should be monitored. The Operator automatically generates Prometheus scrape configuration based on the current state of the objects in the API server.

    Probe, which declaratively specifies how groups of ingresses or static targets should be monitored. The Operator automatically generates Prometheus scrape configuration based on the definition.

    ScrapeConfig, which declaratively specifies scrape configurations to be added to Prometheus. This CustomResourceDefinition helps with scraping resources outside the Kubernetes cluster.

    PrometheusRule, which defines a desired set of Prometheus alerting and/or recording rules. The Operator generates a rule file, which can be used by Prometheus instances.

    AlertmanagerConfig, which declaratively specifies subsections of the Alertmanager configuration, allowing routing of alerts to custom receivers, and setting inhibit rules.
```

everettraven: So to make sure I am following along, the CMO will:
- Deploy the PrometheusOperator based on the `PrometheusOperatorConfig`
- Create `Prometheus` CRs using the configurations provided in `PrometheusK8sConfig`. Does this apply to _all_ Prometheus CRs?

While these are two distinct things, they are both inherently related to how the CMO handles prometheus configuration on the cluster.

>I can understand your idea but PrometheusOperator manages all these components and I it's not a good idea to have all fields inside PrometheusOperator. 

I'm not suggesting that we put all the fields under `PrometheusOperatorConfig`, I'm suggesting we use a shared parent field named `prometheus` that can have sibling fields for configuring the Prometheus Operator itself and, separately, configuring the individual Prometheus instance configurations. This way, if you want to add additional configuration options related to prometheus in the future, you don't have to add _another_ `Prometheus*` field.

marioferh: > 
> - Deploy the PrometheusOperator based on the PrometheusOperatorConfig
> - Create Prometheus CRs using the configurations provided in PrometheusK8sConfig. Does this apply to all Prometheus CRs?

Correct

> I'm not suggesting that we put all the fields under PrometheusOperatorConfig, I'm suggesting we use a shared parent field named prometheus that can have sibling fields for configuring the Prometheus Operator itself and, separately, configuring the individual Prometheus instance configurations. This way, if you want to add additional configuration options related to prometheus in the future, you don't have to add another Prometheus* field.

But they are different things, the are related but from my point of view and how CMO works it makes no sense.
https://github.com/prometheus-operator/prometheus-operator
https://github.com/prometheus/prometheus

@danielmellado @simonpasquier any thoughts?

danielmellado: I would keep @marioferh approach. While I understand @everettraven concern about API organization, the reality is that prometheusOperatorConfig and prometheusK8sConfig are solving different problems and IMHO  having them under a shared parent would actually make things more confusing.

Think about it from an operator's perspective: when you're configuring prometheusOperatorConfig, you're basically saying "how should we deploy and run the Prometheus Operator pods themselves" - stuff like resource limits, node scheduling, log levels. But when you're dealing with prometheusK8sConfig, you're configuring "what should the actual Prometheus servers do" - scraping rules, storage, retention policies, etc. Again, I think mixing them together would be confusing.

Plus, we already have a working pattern with alertmanagerConfig and metricsServerConfig that users understand. Why break that consistency for a theoretical future problem?

If we do end up with too many prometheus fields later, I'm totally happy to revisit the structure, but I think that for now the separation actually makes the API clearer and more intuitive.

@simonpasquier wdyt?

everettraven: >If we do end up with too many prometheus fields later, I'm totally happy to revisit the structure

If you go with a distributed field approach, you have to maintain that essentially forever or go through a pretty painful process to refine the structure once you've promoted the API to `v1`.

>Why break that consistency for a theoretical future problem?

I'm not sold that doing something like:
```yaml
prometheusConfig:
  operator:
    ...
  servers:
    ...
```

breaks that consistency, but if you folks feel strongly that users will have a better experience with multiple `prometheus*Config` fields I won't block it.

If you don't think you'll ever have more than the two fields for the operator and servers respectively, this probably isn't that big of a deal.

>Think about it from an operator's perspective: when you're configuring prometheusOperatorConfig, you're basically saying "how should we deploy and run the Prometheus Operator pods themselves" - stuff like resource limits, node scheduling, log levels. But when you're dealing with prometheusK8sConfig, you're configuring "what should the actual Prometheus servers do" - scraping rules, storage, retention policies, etc. Again, I think mixing them together would be confusing

I think the example above still considers this perspective and difference of field responsibilities. Except now you have a dedicated umbrella field that captures everything related to the configuration of the "Prometheus stack" (operator and the servers). 

Again, if you folks feel strongly that this doesn't make sense and users would have a better experience with the currently implemented approach I won't stop it from being done, but it must be clear to users what each `prometheus*Config` field is responsible for and when they should/should not be specifying the fields.

everettraven:
@marioferh When this PR is ready for another round of review, please remove the `lifecycle/stale` label and ping me for a review.

Thank you!

marioferh:
> @marioferh When this PR is ready for another round of review, please remove the `lifecycle/stale` label and ping me for a review.
> 
> Thank you!

Done.

About this: https://github.com/openshift/api/pull/2481#discussion_r2391324182

I think current approach is the better one

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -405,7 +410,90 @@ type MetricsServerConfig struct {
...
+	//      request: 40Mi
+	//      limit: null
+	// Maximum length for this list is 10.
+	// Minimum length for this list is 1.
+	// +optional
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=10
+	// +kubebuilder:validation:MinItems=1
+	Resources []ContainerResource `json:"resources,omitempty"`
```
everettraven, line 475: Have we also considered enforcing a minimum request/limit for certain resources to ensure that the values don't get set too low for the operator to function properly?

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -405,7 +410,90 @@ type MetricsServerConfig struct {
...
+	//      request: 4m
+	//      limit: null
+	//    - name: memory
+	//      request: 40Mi
+	//      limit: null
+	// Maximum length for this list is 10.
+	// Minimum length for this list is 1.
+	// +optional
+	// +listType=map
+	// +listMapKey=name
```
everettraven, line 472: Document `name` uniqueness in the GoDoc please.

marioferh:
@CodeRabbit ignore

config/v1alpha1/types_cluster_monitoring.go
marioferh: done

config/v1alpha1/types_cluster_monitoring.go
marioferh: and don't think we have hard limits in prometheus. Maybe we can suggest a minimum resources but that depends on the use case.

everettraven: >Maybe we can suggest a minimum resources but that depends on the use case.

Making a suggestion in documentation seems reasonable if there isn't any concrete minimums we can enforce.

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -295,6 +300,7 @@ type ContainerResource struct {
 	// name of the resource (e.g. "cpu", "memory", "hugepages-2Mi").
 	// This field is required.
 	// name must consist only of alphanumeric characters, `-`, `_` and `.` and must start and end with an alphanumeric character.
+	// name must be unique within the resources list.
```
everettraven: This comment should go on the fields that use this type in a list

everettraven:
The pending pipeline tests no longer exist and were sharded some time ago. The sharded versions of the tests are passing so this should be good to go.

marioferh:
> The pending pipeline tests no longer exist and were sharded some time ago. The sharded versions of the tests are passing so this should be good to go.
> 
> /override ci/prow/e2e-aws-serial /override ci/prow/e2e-aws-serial-techpreview

thanks

PR #2492, opened 2025-09-22
CORS-4221: Remove the OAUTH and STS endpoints

barbacbd:
** The GCP endpoints for oauth and sts should not be overridden using the Private Service Connect. While these appear to be services just like the others (ex compute) they are not to be overridden as per google. Remove the option to override these endpoints.


JoelSpeed:
This API is only available in tech preview. While normally removing an enum value is a breaking change, for tech preview we are able to make this change since we don't support upgrades

/lgtm
/verified bypass

This won't need verification

JoelSpeed:
/lgtm
/hold cancel

Removing enums in TP is not a breaking change we care about

JoelSpeed:
/verified bypass

Won't need verification on this TP removal, local integration tests are passing so the CRD is compatible with KAS still

barbacbd:
Closing in favor of https://github.com/openshift/api/pull/2576. Please see https://github.com/openshift/api/pull/2576#issue-3608882808

PR #2507, opened 2025-09-25
OCPBUGS-59939: ConsoleLink CRD has incorrect additionalPrinterColumns entry

krishagarwal278:
### Bug fix for master

The ConsoleLink CRD has an `additionalPrinterColumns` entry for `Menu` that is not valid. The correct column name is `Location`.

 - name: Menu
  type: string
  jsonPath: .spec.menu
should be changed to

 - name: Location
  type: string
  jsonPath: .spec.location


everettraven (approved):
/lgtm

PR #2512, opened 2025-09-30
Add: Support for specifying elliptic curves in TLS handshake

richardsonnick:
To support new option set default ssl curves.
Adds an option to TLSProfile config to add a list of supported groups.

See for more context https://github.com/openshift/cluster-ingress-operator/pull/1287


config/v1/types_tlssecurityprofile.go
```diff
@@ -213,6 +213,14 @@ type TLSProfileSpec struct {
...
 	// +listType=atomic
 	Ciphers []string `json:"ciphers"`
+	// curves is used to specify the elliptic curves that are used during
+	// the TLS handshake.  Operators may remove entries their operands do
+	// not support.  For example, to use X25519 and P-256 (yaml):
+	//
+	//   curves:
+	//     - X25519
+	//     - P-256
+	Curves []string `json:"curves.omitempty"`
```
everettraven:
```suggestion
	Curves []string `json:"curves,omitempty"`
```

Is the list of allowed elliptic curves static? Presumably the configuration here are going to be consumed by Go-based controllers and should probably be limited to the set in https://pkg.go.dev/crypto/tls#CurveID?

What happens if I do not supply this configuration value? Is there any defaulting that would occur?

What happens if I specify the same curve more than once?

How many curves can I specify? Does ordering of the curves matter?

richardsonnick: This setting is passed through HAProxy (used by OpenShift router) to OpenSSL (or whatever tls implementation it is using) via [ssl-bind-curves](https://www.haproxy.com/documentation/haproxy-configuration-manual/latest/#3.1-ssl-default-bind-curves). In the end the list is consumed by OpenSSL's [SSL_CTX_set1_groups_list](https://man.openbsd.org/SSL_CTX_set1_groups_list.3).

> Is the list of allowed elliptic curves static? 

The list of curves is static. I don't see anywhere in the API codebase that enforces other static lists (i.e. [TLSProfileSpec.Ciphers](https://github.com/openshift/api/blob/230d0e045316bf1c679f3be8ad4cb5c501ca899b/config/v1/types_tlssecurityprofile.go#L215), so I'm assuming this needs to be handled by the caller? The curves list is consumed by HAProxy [ssl-bind-curves](https://www.haproxy.com/documentation/haproxy-configuration-manual/latest/#3.1-ssl-default-bind-curves) which sets the supported groups in whatever TLS implementation HAProxy is using. 

> Presumably the configuration here are going to be consumed by Go-based controllers and should probably be limited to the set in https://pkg.go.dev/crypto/tls#CurveID?

In this case, the configuration is actually consumed by HAProxy, which uses OpenSSL's TLS implementation. I think it makes sense to use [go/crypto's list of curves](https://www.iana.org/assignments/tls-parameters/tls-parameters.xml#tls-parameters-8) for the static list. OpenSSL has the [same iana naming scheme](https://github.com/openssl/openssl/blob/3206bb708246a97b281133009a419fb7421971d9/ssl/t1_trce.c#L505). I could imagine that in the future other go based TLS implementations could use this curve list.

> What happens if I do not supply this configuration value? Is there any defaulting that would occur?

If the variable is omitted, then the default set of curves will be used. This default set depends on the underlying TLS implementation. **Omitting this variable will not impact existing configurations, they will just continue to use the defaults of the TLS implementation.**

> What happens if I specify the same curve more than once?

This depends on the tls implementation. By default OpenSSL will just set the curve once.

> How many curves can I specify? 

You can specify as many curves as you want (or as many curves are supported).

> Does ordering of the curves matter?

**Yes**, the order of curves matters. The curves ultimately end up as the TLS server's [supported_groups](https://datatracker.ietf.org/doc/html/rfc8446#section-4.1.1). This list of curves is compared against the supported_groups extension provided in the ClientHello message. The list is compared **in order** to the client's supported_groups. The first match is used. This is identitical to how the cipher suites are negotiated.

richardsonnick:
> Additionally, this is a change to a v1 API and requires that the new field is introduced behind a new feature gate.

Sorry for the over sight on this, first time contributing. I'll circle back with a more in depths review of this change.

config/v1/types_tlssecurityprofile.go
everettraven: >This setting is passed through HAProxy (used by OpenShift router) to OpenSSL (or whatever tls implementation it is using) via [ssl-bind-curves](https://www.haproxy.com/documentation/haproxy-configuration-manual/latest/#3.1-ssl-default-bind-curves). In the end the list is consumed by OpenSSL's [SSL_CTX_set1_groups_list](https://man.openbsd.org/SSL_CTX_set1_groups_list.3).

This particular API is used by multiple operators, not just the cluster-ingress-operator. We should include in the enhancement proposal for this new feature how it impacts other operators.

Looking through the types that use it, it looks like all OpenShift API server operators would be impacted and so would the machine configuration operator (for configuring kubelet):
- https://github.com/openshift/api/blob/230d0e045316bf1c679f3be8ad4cb5c501ca899b/config/v1/types_apiserver.go#L64
- https://github.com/openshift/api/blob/230d0e045316bf1c679f3be8ad4cb5c501ca899b/machineconfiguration/v1/types.go#L727

>The list of curves is static. I don't see anywhere in the API codebase that enforces other static lists (i.e. [TLSProfileSpec.Ciphers](https://github.com/openshift/api/blob/230d0e045316bf1c679f3be8ad4cb5c501ca899b/config/v1/types_tlssecurityprofile.go#L215), so I'm assuming this needs to be handled by the caller? The curves list is consumed by HAProxy [ssl-bind-curves](https://www.haproxy.com/documentation/haproxy-configuration-manual/latest/#3.1-ssl-default-bind-curves) which sets the supported groups in whatever TLS implementation HAProxy is using.

This API is pretty old and conventions have changed since it was originally introduced. We should ensure that we are following modern conventions, which means ensuring that we are including admission-time validations on fields to ensure end-users get a fast feedback loop on configuration of invalid values and prevents rollout of an invalid configuration.

> In this case, the configuration is actually consumed by HAProxy, which uses OpenSSL's TLS implementation. I think it makes sense to use [go/crypto's list of curves](https://www.iana.org/assignments/tls-parameters/tls-parameters.xml#tls-parameters-8) for the static list. OpenSSL has the [same iana naming scheme](https://github.com/openssl/openssl/blob/3206bb708246a97b281133009a419fb7421971d9/ssl/t1_trce.c#L505). I could imagine that in the future other go based TLS implementations could use this curve list.

https://pkg.go.dev/crypto/tls#CurveID seems like the canonical list that Go supports? If we want to support a wider list than what Go's standard library TLS implementations support, we need to detail how this impacts other OpenShift system components.

>If the variable is omitted, then the default set of curves will be used. This default set depends on the underlying TLS implementation. Omitting this variable will not impact existing configurations, they will just continue to use the defaults of the TLS implementation

We should explicitly mention this information in the GoDoc for the field. The GoDoc is used to generate user facing documentation, so including this information is helpful for end users to understand what happens when they do or do not configure the field.

>>What happens if I specify the same curve more than once?
>
>This depends on the tls implementation. By default OpenSSL will just set the curve once.

If it doesn't do anything, we should probably enforce some kind of uniqueness here then to prevent weird behaviors.

>Yes, the order of curves matters. The curves ultimately end up as the TLS server's [supported_groups](https://datatracker.ietf.org/doc/html/rfc8446#section-4.1.1). This list of curves is compared against the supported_groups extension provided in the ClientHello message. The list is compared in order to the client's supported_groups. The first match is used. This is identitical to how the cipher suites are negotiated.

We should explicitly include this in the documentation for the field.

config/v1/types_tlssecurityprofile.go
```diff
@@ -213,6 +213,14 @@ type TLSProfileSpec struct {
...
 	// +listType=atomic
 	Ciphers []string `json:"ciphers"`
+	// curves is used to specify the elliptic curves that are used during
+	// the TLS handshake.  Operators may remove entries their operands do
+	// not support.  For example, to use X25519 and P-256 (yaml):
+	//
+	//   curves:
+	//     - X25519
+	//     - P-256
+	Curves []string `json:"curves,omitempty"`
```
tmshort, line 223: My main concern here is the names used for these curves, as they are pretty much all over the place.

`X25519` is pretty standard, but I've also seen `ecdh_x25519`: (https://github.com/openshift/cluster-ingress-operator/pull/1287)

Then there's the P curves. Within golang, they are referred to as:
* `CurveP256`
* `CurveP384`
* `CurveP521`

https://wiki.mozilla.org/Security/Server_Side_TLS refers to them as:
* `prime256v1`
* `secp384r1`

(No mention of `secp521r1`)
Note: OpenSSL/LibreSSL/BoringSSL also use these terms
Note: `secp256r1` (usedin PR1287) is equivalent to `prime256v1` but `prime256v1` is more commonly used

This code seems to be using:
* `P-256`

(And possibly `P-384` and `P-521`)

We should pick one and use it all over openshift, so we don't need conversion functions all over the place.

richardsonnick:
This was implemented in this PR: https://github.com/openshift/api/pull/2583

PR #2556, opened 2025-10-30
WIP: remove kms encryption from TechPreview feature gate

flavianmissi:
we're reducing the batch size of the kms encryption feature to allow for incremental feature delivery.


flavianmissi:
with 4.21 feature development closed, we have dropped the idea of a dev preview.

/close

PR #2561, opened 2025-11-03
MCO-1962: Move skew enforcement feature gate to tech preview

djoshy:
As skew enforcement is considered a pre-requisite for bringing [dual stream support](https://issues.redhat.com/browse/OCPSTRAT-1150) to GA, we need to accelerate the timeline for this feature and start off with an MVP in tech preview.


djoshy:
/verified bypass

This is only a featureset change

PR #2564, opened 2025-11-04
OCPCLOUD-3214: Add ClusterAPI config

mdbooth:


operator/v1alpha1/types_clusterapi.go
```diff
@@ -0,0 +1,124 @@
...
+	// alphanumeric characters, '-' or '.', and must start and end with an
+	// alphanumeric character, with a maximum length of 253 characters.
+	// Example: "clusters.cluster.x-k8s.io"
+	//
+	// Items cannot be removed from this list once added.
+	//
+	// The maximum number of unmanagedCustomResourceDefinitions is 128.
+	//
+	// +optional
+	// +listType=set
```
chrischdi, line 61: Q: is this resource created by each actor or will there be one instance with co-ownership from different actors?

Example 1:
* 1 ClusterAPI object created by HyperShift
* 1 ClusterAPI object created by ACM

Example 2:
* 1 ClusterAPI object, some entries in .spec.unmanagedCustomResourceDefinitions are from HyperShift, others are from ACM.

operator/v1alpha1/types_clusterapi.go
```diff
@@ -0,0 +1,124 @@
...
+	// +required
+	Spec ClusterAPISpec `json:"spec,omitzero"`
+
+	// status defines the observed status of the capi-operator.
+	// +optional
+	Status ClusterAPIStatus `json:"status,omitzero"`
+}
+
+// ClusterAPISpec defines the desired configuration of the capi-operator.
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.unmanagedCustomResourceDefinitions) || has(self.unmanagedCustomResourceDefinitions)",message="unmanagedCustomResourceDefinitions cannot be unset once set"
```
chrischdi, line 43: We validate removal to not happen, do we also validate deletion of the object to not happen?

operator/v1alpha1/tests/clusterapis.operator.openshift.io/ClusterAPIMachineManagement.yaml
```diff
@@ -0,0 +1,329 @@
...
+          name: cluster
+        spec:
+          unmanagedCustomResourceDefinitions:
+            - a.b.c.d
+            - name-with-hyphens.domain-with-hyphens.io
+            - valid-name.example.io
+  onUpdate:
+    # TODO: Cannot test the XValidation rule "unmanagedCustomResourceDefinitions cannot be unset once set"
+    # because MinProperties=1 on ClusterAPISpec prevents creating spec: {} in the first place.
+    # The validation is still enforced at the API level via the XValidation rule.
```
chrischdi: Shouldn't this work? (I tried it)
```suggestion
    - name: Should not allow unsetting unmanagedCustomResourceDefinitions
      initial: |
        apiVersion: operator.openshift.io/v1alpha1
        kind: ClusterAPI
        metadata:
          name: cluster
        spec:
          unmanagedCustomResourceDefinitions:
            - clusters.cluster.x-k8s.io
            - machines.cluster.x-k8s.io
      updated: |
        apiVersion: operator.openshift.io/v1alpha1
        kind: ClusterAPI
        metadata:
          name: cluster
        spec:
          unmanagedCustomResourceDefinitions: null
      expectedError: "unmanagedCustomResourceDefinitions cannot be unset once set"
```

mdbooth: I don't think set to null is the same as unset.

However, I decided to go a different direction with this: I made `spec` optional. I think this should be the default for a ClusterOperator, tbh.

operator/v1alpha1/types_clusterapi.go
mdbooth: Interesting 🤔

I'm not convinced we can validate this here. I suspect we don't want to validate this at all, though: if the user is deleting ClusterOperators then they're already having a bad day. They've chosen this course of action and they're going to do it regardless. Whatever barriers we raise are just another in a succession of irritations that will be circumvented: disabling CVO, deleting VAPs, webhooks, etc. We can discourage the use of footguns, but ultimately we can't uninvent them.

operator/v1alpha1/types_clusterapi.go
mdbooth: You can't have duplicates in listType=set, so multiple actors would own the same entry if that happened.

Each actor simply needs to submit an SSA transaction with their own requirements without reference to any other actors.

operator/v1alpha1/types_clusterapi.go
```diff
@@ -0,0 +1,123 @@
...
+// +kubebuilder:subresource:status
+// +openshift:api-approved.openshift.io=https://github.com/openshift/api/pull/2564
+// +openshift:file-pattern=cvoRunLevel=0000_30,operatorName=cluster-api,operatorOrdering=01
+// +openshift:enable:FeatureGate=ClusterAPIMachineManagement
+
+// ClusterAPI provides configuration for the capi-operator.
+//
+// Compatibility level 4: No compatibility is provided, the API can change at any point for any reason. These capabilities should not be used by applications needing long term support.
+// +openshift:compatibility-gen:level=4
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.spec) || !has(oldSelf.spec.unmanagedCustomResourceDefinitions) || has(self.spec.unmanagedCustomResourceDefinitions)",message="unmanagedCustomResourceDefinitions cannot be unset once set"
```
yuqi-zhang: I think for validations, it's best to have it explicitly in the godoc as well if possible (so in this case, maybe just `unmanagedCustomResourceDefinitions cannot be unset once set` as a godoc line is fine?).

Side note, I thought that if we don't edit `unmanagedCustomResourceDefinitions` but instead delete the whole spec field after setting it at some point, `self.spec.unmanagedCustomResourceDefinitions` would maybe not evaluate correctly, but it seems that CEL can handle that null exception (and you have a test case above covering it)

operator/v1alpha1/types_clusterapi.go
```diff
@@ -0,0 +1,123 @@
...
+type ClusterAPI struct {
```
yuqi-zhang, line 23: So, like the other operator types, I assume this will be a singleton. Interestingly enough the only one to enforce it is the olm operator: https://github.com/openshift/api/blob/master/operator/v1alpha1/types_olm.go#L25

I think generally speaking it's a good pattern to have, so probably worth adding that? Unless I'm misunderstanding how this can be configured.

operator/v1alpha1/types_clusterapi.go
```diff
@@ -0,0 +1,123 @@
...
+	metav1.TypeMeta `json:",inline"`
+
+	// metadata is the standard object's metadata.
+	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
+	// +required
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// spec is the specification of the desired behavior of the capi-operator.
+	// +optional
+	Spec *ClusterAPISpec `json:"spec,omitempty"`
```
yuqi-zhang, line 33: Based on updated guidance, I think now we only pointers if there's a difference between the unset and empty struct, otherwise having it same as the status field below (not a pointer, omitzero) is the best option.

In this case, since "" and empty should both mean "no CRDs are unmanaged", I think this falls under that umbrella?

The other way we could maybe do this (and I don't know what is "better" conventions-wise) is to instead make this required, keep `unmanagedCustomResourceDefinitions` optional, move the `// +kubebuilder:validation:XValidation:rule="!has(oldSelf.spec) || !has(oldSelf.spec.unmanagedCustomResourceDefinitions) || has(self.spec.unmanagedCustomResourceDefinitions)",message="unmanagedCustomResourceDefinitions cannot be unset once set"` validation here to the spec, and remove the need for nil checking generally. This might be more future proof in case we want to add more spec fields in the future.

operator/v1alpha1/types_clusterapi.go
```diff
@@ -0,0 +1,123 @@
...
+
+	// status defines the observed status of the capi-operator.
+	// +optional
+	Status ClusterAPIStatus `json:"status,omitzero"`
+}
+
+// ClusterAPISpec defines the desired configuration of the capi-operator.
+type ClusterAPISpec struct {
+	// unmanagedCustomResourceDefinitions is a list of ClusterResourceDefinition (CRD)
```
yuqi-zhang, line 45: Presumably, this is only a set of CRDs that CAPI operator would use, so e.g. if I set "MachineConfigs" here, nothing would happen since CAPI doesn't manage MachineConfigs, nor can it tell CVO/MCO to not deploy/use some other api version.

Would there be a way for the user to discover what CRDs is able to be unmanaged here? Is that a big list or something we can add to the godoc/validator?

Reading this and https://github.com/openshift/enhancements/pull/1845 I had initially thought that this is a generic list you can apply to any CRD but that shouldn't be the case right?

operator/v1alpha1/types_clusterapi.go
```diff
@@ -0,0 +1,123 @@
...
+	// +listType=set
+	// +kubebuilder:validation:MaxItems=128
+	// +kubebuilder:validation:XValidation:rule="oldSelf.all(item, item in self)",message="items cannot be removed from unmanagedCustomResourceDefinitions list"
+	// +kubebuilder:validation:items:XValidation:rule="!format.dns1123Subdomain().validate(self).hasValue()",message="a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character."
+	// +kubebuilder:validation:items:MaxLength=253
+	UnmanagedCustomResourceDefinitions []string `json:"unmanagedCustomResourceDefinitions,omitempty"`
+}
+
+// ClusterAPIStatus describes the current state of the capi-operator.
+// +kubebuilder:validation:MinProperties=1
```
yuqi-zhang: Question: I assume only CAPI-operator would be updating the status of this object. Does the validation here help prevent a specific scenario?

(should be fine to have, just curious why this validation was added)

operator/v1alpha1/types_clusterapi.go
```diff
@@ -0,0 +1,123 @@
...
+type ClusterAPIStatus struct {
+	// targetConfigMaps is a list of ConfigMap names that the staging controller
```
yuqi-zhang: I'm probably missing some context but I'm not 100% sure what these two configmap status field entails. I assume these are here for an admin to see? Are these reflecting the status of some other action the admin took elsewhere? Or something that has to do with the CRDs they just defined in the spec? (Just from an ELI5 perspective the status fields are confusing, if there's some other way to add some more context I think that would be helpful)

yuqi-zhang (commented):
General structure looks fine, some questions inline

Also, based on Claude suggestion, all optional fields (spec, status, UnmanagedCustomResourceDefinitions, TargetConfigMaps, ActiveConfigMaps) should have a godoc line explaining what happens if the value is empty. (example: `When omitted, the capi-operator will use default configuration with all CRDs managed by the operator.` in the spec, assuming we keep spec optional)

operator/v1alpha1/types_clusterapi.go
mdbooth: I think this list is fundamentally un-validatable. The problem is we have a documented use case where the user is expected to add an entry for a CRD which:
* May not yet be present in the cluster
* Is not part of any manifests we currently know about

I'd be hesitant to add a heuristic which will inevitably turn out to be wrong, especially if it's also complex to implement. I'd prefer to leave this to the documentation.

operator/v1alpha1/types_clusterapi.go
mdbooth: I'll try setting this back to a zeroable struct, but I recall having previously flipped this so I have a suspicion there's a reason that doesn't work down the line somewhere.

operator/v1alpha1/types_clusterapi.go
mdbooth: Pretty sure I added it to make the linter happy without having to make Status a pointer.

operator/v1alpha1/types_clusterapi.go
mdbooth: I need to re-write this (again), but this is for internal state of the installer controller.

operator/v1alpha1/types_clusterapi.go
mdbooth: Hmm, I'm not sure that godoc lives here, though. This isn't where a user is going to to go looking for it; they're going to be looking at the `unmanagedCustomResourceDefinitions` field. It's only here because of a CEL implementation detail.

operator/v1alpha1/types_clusterapi.go
mdbooth: The linter complains:

```
operator/v1alpha1/types_clusterapi.go:35:2: optionalfields: field Spec has a valid zero value ({}), but the validation is not complete (e.g. min properties/adding required fields). The field should be a pointer to allow the zero value to be set. If the zero value is not a valid use case, complete the validation and remove the pointer. (kubeapilinter)
        Spec ClusterAPISpec `json:"spec,omitempty"`
```

I feel like the same could be achieved by omitting `omitempty` from the json marker, but the linter doesn't permit that either. I suspect we're trying to cover multiple use cases with this marshalling behaviour, so perhaps that would break something else?

For now the linter requires this to be a pointer 🤷

operator/v1alpha1/types_clusterapi.go
yuqi-zhang: Good point. I'm wondering if we should move this validation to the `ClusterAPISpec` object then, since it doesn't refer to anything outside of spec

operator/v1alpha1/types_clusterapi.go
yuqi-zhang: Will check this with the API team.

operator/v1alpha1/types_clusterapi.go
yuqi-zhang: I think that makes sense. By documentation do you mean part of the godoc here? Or as part of OCP documentation tracked elsewhere?

And if I were to provide CRDs which the capi-operator doesn't understand, would it error or silently skip over?

mdbooth: By documentation I think that means:
* a good summary in the godoc
* comprehensive product docs elsewhere

An unmanaged CRD which isn't in any capi-operator manifests will just be ignored.

operator/v1alpha1/types_clusterapi.go
yuqi-zhang: Based on discussions, the behaviour you're seeing is because there is no required field (in this case, only an optional `UnmanagedCustomResourceDefinitions`) in the spec, so the linter thinks that there is a difference between an empty spec {} and no spec.

The best way to handle this (given that we want to keep UnmanagedCustomResourceDefinitions optional, and leave this open to expanding more configuration spec fields in the future) is to set min properties (of 1) on the spec, which will indicate that you either should have an empty spec, or if you provide a spec, you must also provide at least 1 field. Then we can remove the pointer and add back omitzero.

i.e. with that, it would be fine to set
```
...
```

or
```
spec:
  unmanagedCustomResourceDefinitions:
    - aaa
```

But not
```
spec: {}
```

operator/v1alpha1/types_clusterapi.go
yuqi-zhang: Wanted to circle back on this, the general guidance is that validations should be on the lowest level object, which in this case should be spec. (and we can do this if we set spec to required as commented below).

If we do keep it here, we can use chaining to simplify it to something like: 

```
// +kubebuilder:validation:XValidation:rule="oldSelf.?spec.unmanagedCustomResourceDefinitions.size() == 0 ||    
  self.?spec.unmanagedCustomResourceDefinitions.size() > 0",message="unmanagedCustomResourceDefinitions cannot be  
  unset once set"
```

But I think that shouldn't be necessary if we move it.

operator/v1alpha1/types_clusterapi.go
yuqi-zhang: Bump on this as well, we discussed this scenario a bit more, and in this case, it makes more sense to keep it a pointer, but make it required, and set minproperties=0 on the type.

This way all clusterAPI config objects will require there to be a `spec: {}` which is an explicit indication of an empty spec. It also comes with the side effect of this field always existing, so we can move the spec validation down here.

You can ignore my previous comment, sorry for the confusion.

operator/v1alpha1/types_clusterapi.go
yuqi-zhang: I think this is only necessary if that are no required fields within the struct, much like the suggestion for spec above.

Since DesiredRevision and Revisions are required, I think this should be safe to remove?

operator/v1alpha1/types_clusterapi.go
```diff
@@ -0,0 +1,195 @@
...
+
+// RevisionName represents the name of a revision. The name must be between 1
+// and 255 characters long.
+// +kubebuilder:validation:MinLength=1
+// +kubebuilder:validation:MaxLength=255
+type RevisionName string
+
+// ClusterAPIStatus describes the current state of the capi-operator.
+// +kubebuilder:validation:MinProperties=1
+// +kubebuilder:validation:XValidation:rule="self.revisions.exists(r, r.name == self.desiredRevision && self.revisions.all(s, s.revision <= r.revision))",message="desiredRevision must be the name of the revision with the highest revision number"
```
yuqi-zhang, line 79: Is it the same controller setting Current/Desired and the Revisions list? I'm curious as to the validations here, since I assume any errors would be handled by the controller as well, whether those can actually drift?

operator/v1alpha1/types_clusterapi.go
```diff
@@ -0,0 +1,195 @@
...
+	// revision. It is written by the revision controller.
+	//
+	// The maximum number of revisions is 16.
+	// All revisions must have a unique name.
+	// All revisions must have a unique revision number.
+	// When adding a revision, the revision number must be greater than the highest revision number in the list.
+	// Revisions are immutable, although they can be deleted.
+	//
+	// +required
+	// +listType=atomic
```
yuqi-zhang, line 106: Since name (and revision number) are unique, we can probably turn this into a `listType=map` and use either as the key, and then we can remove the unique validation on that.

operator/v1alpha1/types_clusterapi.go
```diff
@@ -0,0 +1,195 @@
...
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=16
+	// +kubebuilder:validation:XValidation:rule="self.all(x, self.exists_one(y, x.name == y.name))",message="each revision must have a unique name"
+	// +kubebuilder:validation:XValidation:rule="self.all(x, self.exists_one(y, x.revision == y.revision))",message="each revision must have a unique revision number"
+	// +kubebuilder:validation:XValidation:rule="self.all(new, oldSelf.exists(old, old.name == new.name) || oldSelf.all(old, new.revision > old.revision))",message="new revisions must have a revision number greater than all existing revisions"
+	// +kubebuilder:validation:XValidation:rule="oldSelf.all(old, !self.exists(new, new.name == old.name) || self.exists(new, new == old))",message="revisions are immutable"
```
yuqi-zhang: nit: maybe change `existing revisions are immutable, but can be deleted` like the godoc above?

operator/v1alpha1/types_clusterapi.go
```diff
@@ -0,0 +1,195 @@
...
+	// name is the name of a revision.
+	// +required
+	Name RevisionName `json:"name,omitempty"`
+
+	// revision is a monotonically increasing number that is assigned to a revision.
+	// +required
+	// +kubebuilder:validation:Minimum=1
+	Revision int64 `json:"revision,omitempty"`
+
+	// contentID uniquely identifies the content of this revision.
```
yuqi-zhang, line 127: Could you add some more description on what the ID entails? I'm not sure I understand why name, revision, and a third ID are all required.

operator/v1alpha1/types_clusterapi.go
```diff
@@ -0,0 +1,195 @@
...
+	//
+	// +required
+	// +listType=atomic
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=32
+	Components []ClusterAPIInstallerComponent `json:"components,omitempty"`
+}
+
+// ClusterAPIInstallerComponent defines a component which will be installed by this revision.
+// +kubebuilder:validation:MaxProperties=1
```
yuqi-zhang: There is only 1 property in this struct. Why have a MaxProperties check here?

operator/v1alpha1/types_clusterapi.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// ClusterAPIInstallerComponentImage defines an image source for a component.
+type ClusterAPIInstallerComponentImage struct {
+	// ref is an image reference to the image containing the component manifests. The reference
+	// must be a valid image digest reference in the format host[:port][/namespace]/name@sha256:<digest>.
+	// The digest must be 64 characters long, and consist only of lowercase hexadecimal characters, a-f and 0-9.
+	// The length of the field must be between 1 to 447 characters.
+	// +required
+	Ref machineosconfigv1.ImageDigestFormat `json:"ref,omitempty"`
```
yuqi-zhang: Update: per API guidance, we try not to share types between different APIs, in case it would need to be updated later.

In this case, the suggestion is to make a duplicate copy of the ImageDigestFormat type in this API, instead of referencing the existing object.

operator/v1alpha1/types_clusterapi.go
```diff
@@ -0,0 +1,195 @@
...
+	Components []ClusterAPIInstallerComponent `json:"components,omitempty"`
+}
+
+// ClusterAPIInstallerComponent defines a component which will be installed by this revision.
+// +kubebuilder:validation:MaxProperties=1
+type ClusterAPIInstallerComponent struct {
+	// image defines an image source for a component. The image must contain a
+	// /capi-operator-installer directory containing the component manifests.
+	// +required
+	Image ClusterAPIInstallerComponentImage `json:"image,omitzero"`
```
yuqi-zhang, line 183: There seems to be quite a few levels of nesting. Would we be able to just lift out the Ref and Profile to here directly?

From a higher level question, all the user is expected to do is set a list of `UnmanagedCustomResourceDefinitions` in the spec, right? What is defining the components in a normal workflow? Is this mostly for status? Or is some controller setting this for other controllers to consume?

operator/v1alpha1/types_clusterapi.go
mdbooth: These are written by different controllers (the revision and installer controllers). The `desiredRevision` is technically redundant. I don't think the controller actually looks at it, except to compare it to currentRevision. All it's actually doing is installing the most recent revision.

If this validation failed, this would be a bug in one of those controllers. The only real purpose is to throw an error at the point that the bug occurs rather than later. If some other actor wrote a weird value there, the controllers should recover from it.

operator/v1alpha1/types_clusterapi.go
mdbooth: I considered listType=map here, but my expectation is that this list will only ever be written atomically by the revision controller. There's no reason it couldn't be listType=map, but I figured that atomic is a bit cheaper server-side, so we might as well leave it atomic as we don't expect to manage multiple actors.

We could always revisit this for v1 if it becomes an issue?

operator/v1alpha1/types_clusterapi.go
mdbooth: This is a tough one. On `contentID` specifically, I don't think so, no 😅

The reason is that this description genuinely does completely define the API contract: that if any content of the revision changes, i.e. what the installer would actually install, then this value will change.

I could write here that it's implemented as a hash of X, Y, and Z, but:
* Nothing ever checks that
* The API consumer can't rely on that
* I want to be able to change it as necessary without API impact

`revision` is a boxcutter implementation detail. We just need it.

`name` is intended to be a human-useful value. Technically we could use contentID everywhere it is used, I just figured that the format of `name` would make it easier to understand what's going on.

operator/v1alpha1/types_clusterapi.go
mdbooth: With the discriminated union it's hopefully a bit more obvious why this is nested.

Yeah, I'm not a huge fan of these things but they make @JoelSpeed happy 😜

I *suspect* that we will probably never add another member to the union, but given that the first draft had 2 it's far from impossible.

JoelSpeed:
Reviewed and updated this synchronously with @mdbooth 

/lgtm
/verified by integration testing added

PR #2570, opened 2025-11-06
Add test cases for all supported format markers

mdbooth:
Test cases justify the included recommendation not to use `ipv4`, `ipv6`, and `cidr`.


JoelSpeed (commented):
/lgtm

Thanks

PR #2576, opened 2025-11-10
CORS-4281: Remove the GCP Service Endpoints

barbacbd:
Removing the GCP Service Endpoints in favor of a new solution. The new solution will create a private hosted zone that will route traffic to the googleapi endpoints via an ip address created during the initialization of the private serivce connect endpoint. The cluster components no longer need the endpoint overrides, so the service endpoints can and should be removed to avoid confusion.


damdo:
@barbacbd verify jobs are failing, might be worth rebasing and checking what's wrong

damdo:
/lgtm

@everettraven are the CRD verifications failures expected? TY

everettraven:
@damdo @barbacbd Both `verify-crd-schema` and `verify-crdify` are failing in an anticipated way. They aren't aware of the nuance of TPNU vs GA states so these can be overridden.

That being said, `verify` is failing for a legitimate reason - it looks like the payload manifests were not updated appropriately. Looks like something needs to be regenerated so running `make update` and including the payload manifest changes in your commit should resolve that.

JoelSpeed:
/hold

@barbacbd and I had agreed that these should stay to follow our newer patterns for populating CCM configuration. Would like to catch up on what's changed before we move forward here

patrickdillon:
>  Would like to catch up on what's changed before we move forward here

We determined that CCM configuration is not needed for this feature. All GCP API resolution in the cluster is redirected to the private endpoint by installer-created DNS records. Fun fact, the DNS solution is actually the only way this will work for OpenShift, we learned, because overriding the endpoints does not support certain APIs (`iam`, `sts`). 

Nothing in OpenShift should be using the API defined here. 

> @barbacbd and I had agreed that these should stay to follow our newer patterns for populating CCM configuration

I gathered some loose details on the CCM configuration pattern; it sounded like the installer should not be laying down CCM configuration, which was new to me. We should discuss to better establish expectations and motivations.

config/v1/types_infrastructure.go
```diff
@@ -699,75 +699,6 @@ const (
...
-	// An example of a valid endpoint that overrides the Compute Service: "https://compute-myendpoint1.p.googleapis.com"
-	//
-	// +required
-	// +kubebuilder:validation:MaxLength=253
-	// +kubebuilder:validation:XValidation:rule="isURL(self)",message="must be a valid URL"
-	// +kubebuilder:validation:XValidation:rule="isURL(self) ? (url(self).getScheme() == \"https\") : true",message="scheme must be https"
-	// +kubebuilder:validation:XValidation:rule="url(self).getEscapedPath() == \"\" || url(self).getEscapedPath() == \"/\"",message="url must consist only of a scheme and domain. The url path must be empty."
-	URL string `json:"url"`
-}
-
```
JoelSpeed, line 770: We need to know what this looked like to have the complete tombstone, so this should be commented rather than removed, same for the string alias, I'm happy to remove the constants though

everettraven: Using this as a learning opportunity for tombstoning:

Does it matter what the underlying field structure looked like if we can never add a field with the parent serialized name again?

If I can never create a field with name `tombstonedField`, I inherently cannot create a new field with name `tombstonedField.subField` so why does having the complete view of what this looked like matter?

The only value I see that adding is for future readers of the tombstone - which if that is the reason, fine by me.

EDIT: I also now see value in keeping just the struct name tombstoned so we don't accidentally re-implement a  same named type, but my question still seems relevant for removing the fields of the child type.

JoelSpeed:
/hold cancel

Discussed with @barbacbd out of band

barbacbd:
/override ci/prow/verify-crdify ci/prow/verify-crd-schema

Copying from Joel above, these are expected to fail here.

vincentdephily:
This seems to have broken https://github.com/openshift/client-go/blob/master/config/applyconfigurations/config/v1/gcpserviceendpoint.go which still references `configv1.GCPServiceEndpointName`.  Do you know if there is a PR planed for `o/client-go`, or do you have advice for a fix ?

patrickdillon:
@vincentdephily thank you for your help!
https://github.com/openshift/client-go/pull/350
updated client-go

Can you check to see if the issue is now resolved?

PR #2583, opened 2025-11-13
NE-2751: Implement enhancement in OpenShift API to support for TLS curves in TLSProfile

davidesalerno:
This change will add a new Groups field to the TLSProfile specification as described into the [openshift/ enhancements#1894](https://github.com/openshift/enhancements/pull/1894)

This is required in order to support new PQC curves, we need a config for explicitly setting the supported elliptic curves algorithms ("curve suite") that are negotiated during the SSL/TLS handshake with ECDHE.

This PR is related to [openshift/cluster-ingress-operator#1287]( https://github.com/openshift/cluster-ingress-operator/pull/1287) and [openshift/router#678](https://github.com/openshift/router/pull/678)


config/v1/types_tlssecurityprofile.go
```diff
@@ -213,6 +213,18 @@ type TLSProfileSpec struct {
...
 	// +listType=atomic
 	Ciphers []string `json:"ciphers"`
+	// curves is used to specify the elliptic curves that are used during
+	// the TLS handshake.  Operators may remove entries their operands do
+	// not support.  For example, to use X25519 and P-256 (yaml):
+	//
+	//   curves:
+	//     - X25519
+	//     - P-256
+	// +optional
```
yuqi-zhang, line 225: optional fields should have godoc around what happens if the field is not set (i.e. what is the default behaviour)

config/v1/types_tlssecurityprofile.go
```diff
@@ -213,6 +213,18 @@ type TLSProfileSpec struct {
...
+	// +listType=atomic
+	// +kubebuilder:validation:MaxItems=20
```
yuqi-zhang: Could you help understand the two constraints here? Is there a list of valid curves that the API can validate against instead of arbitrarily allowing users to provide up to 20 20-length strings? What happens if the user provides a faulty curve?

machineconfiguration/v1/zz_generated.crd-manifests/0000_80_machine-config_01_kubeletconfigs-CustomNoUpgrade.crd.yaml
```diff
@@ -144,6 +144,21 @@ spec:
                           type: string
                         type: array
                         x-kubernetes-list-type: atomic
+                      curves:
```
yuqi-zhang: Since other objects (e.g. the kubeletconfig here) references tlsSecurityProfile type, would the curve be supported for all affected objects and controllers?

yuqi-zhang (commented):
As discussed in slack, let's create a featuregate and enhancement to attach to this.

sanchezl (changes requested):
## Add Default Curves to TLSProfiles

The `Curves` field has been added to the API schema, but **all entries in the `TLSProfiles` map should be updated to include default curves**. Currently, only the `Ciphers` field is populated in each profile.

### Recommended Default Curves

All three TLS profiles (`TLSProfileOldType`, `TLSProfileIntermediateType`, `TLSProfileModernType`) should include these curves by default:

```go
Curves: []string{
    "X25519",
    "P-256",
    "P-384",
    "X25519MLKEM768",
},
```

### Go TLS Constant to Configuration String Mapping

The PR should document the mapping between Go's `crypto/tls` constants and the configuration strings used in the `Curves` field:

| Go Constant | Configuration String | Description |
|-------------|---------------------|-------------|
| `tls.X25519` | `X25519` | Curve25519 (recommended) |
| `tls.CurveP256` | `P-256` | NIST P-256 (secp256r1) |
| `tls.CurveP384` | `P-384` | NIST P-384 (secp384r1) |
| `tls.X25519MLKEM768` | `X25519MLKEM768` | Post-Quantum Cryptography hybrid |

Given that the cipher names expected in this configuration are based on OpenSSL names, I suggest OpenSSL versions of the strings. OpenSSL has a few alias defined for some of the curves. I've picked the aliases that match the NIST names when I thought it was appropriate.

### Documentation Update Needed

The field documentation should include an example showing how to configure a **PQC-only** (Post-Quantum Cryptography) TLS profile. Users wanting to enforce PQC-only encryption would create a custom TLS profile:

```yaml
# Example: Force PQC-only encryption
kind: APIServer
spec:
  tlsSecurityProfile:
    type: Custom
    custom:
      ciphers:
        - TLS_AES_128_GCM_SHA256
        - TLS_AES_256_GCM_SHA384
        - TLS_CHACHA20_POLY1305_SHA256
      curves:
        - X25519MLKEM768  # PQC-only: only hybrid quantum-resistant curve
      minTLSVersion: VersionTLS13
```

This configuration ensures that only connections using the post-quantum hybrid key exchange can be established.

config/v1/types_tlssecurityprofile.go
davidesalerno: I added these constraints because they seem to be mandatory when I created this change otherwise I couldn't regenerate the CRD and so I introduced that based on the openssl values for this field.

I've just tried to remove them and regenerate the CRD and seems that they are not mandatory anymore so I'm going to remove them.

yuqi-zhang: Sorry, to clarify, the lists should have constraints. I was wondering based on what you originally had, whether there were up to 20 available curves. Based on https://github.com/openshift/api/pull/2583#pullrequestreview-3488383917 it seems we only want to support a certain set? If it's a well known list that's not too big, maybe we can add validation here that only valid ones can be set via the API.

config/v1/types_tlssecurityprofile.go
davidesalerno: I added a note to address this case. What do you think?

config/v1/types_tlssecurityprofile.go
davidesalerno: Based on what suggested by @sanchezl I introduced an enum type (`TLSCurves`) with a one-to-one mapping Go's crypto/tls library, containing only the allowed values and switched from `string` to `TLSCurve` for this array.
In this way each allowed value is well documented and there is a validation on the enum side.

machineconfiguration/v1/zz_generated.crd-manifests/0000_80_machine-config_01_kubeletconfigs-CustomNoUpgrade.crd.yaml
davidesalerno: yes, do you think there are some objects that needs some specific improvements for this change?

Miciah:
https://github.com/openshift/api/compare/58da6c8c1090bcc312e8f1bd3d8bca87cd47227d..eba1e60731dc58f0c314e2ef10ed38dd3f528c00 dropped X448.  Was that intentional, perhaps because Go's crypto/tls package doesn't include support?  Do we have agreement from stakeholders that the current set of curves is sufficient? https://github.com/openshift/api/blob/eba1e60731dc58f0c314e2ef10ed38dd3f528c00/config/v1/types_tlssecurityprofile.go#L205-L224

config/v1/types_tlssecurityprofile.go
```diff
@@ -213,6 +234,37 @@ type TLSProfileSpec struct {
...
 	Ciphers []string `json:"ciphers"`
+	// curves is used to specify the elliptic curves that are used during
+	// the TLS handshake.  Operators may remove entries their operands do
+	// not support.
+	//
+	// TLSProfiles Old, Intermediate, Modern are including by default the following
+	// curves: X25519, P-256, P-384, X25519MLKEM768
+	// TLSProfiles Custom do not include any curves by default.
+	// NOTE: since this field is optional, if no curves are specified, the default curves
+	// used by the underlying TLS library will be used.
```
Miciah:
```suggestion
	// The Old, Intermediate, Modern profiles include the following
	// curves: X25519, P-256, P-384, X25519MLKEM768.
	// The Custom profile does not include any curves by default.
	// NOTE: Since this field is optional, if no curves are specified, the default curves
	// used by the underlying TLS library will be used.
```
Is that a "SHOULD" be used, or a "MUST" be used?  That is, does a component violate the API contract if it simply does not allow ECDHE when `curves` is empty?

config/v1/types_tlssecurityprofile.go
```diff
@@ -213,6 +234,37 @@ type TLSProfileSpec struct {
...
+	//   tlsSecurityProfile:
+	//     type: Custom
+	//     custom:
+	//       ciphers:
+	// 	       - TLS_AES_128_GCM_SHA256
+	//         - TLS_AES_256_GCM_SHA384
+	//         - TLS_CHACHA20_POLY1305_SHA256
+	//       curves:
+	//         - X25519MLKEM768  # PQC-only: only hybrid quantum-resistant curve
+	//       minTLSVersion: VersionTLS13
```
Miciah: Does this render all right in the `oc explain` output?

config/v1/types_tlssecurityprofile.go
```diff
@@ -213,6 +234,37 @@ type TLSProfileSpec struct {
...
+	//
+	// +optional
+	// +listType=atomic
+	// +kubebuilder:validation:MaxItems=5
+	Curves []TLSCurve `json:"curves,omitempty"`
```
Miciah: Does it make sense to allow duplicates, or should we have some validation to prevent that (something like ```// +kubebuilder:validation:XValidation:rule=`self.all(x, self.exists_one(y, x == y))`,message="curves cannot contain duplicates"```)?

davidesalerno: I think it is a really good idea!

What do you think if we switch to `listType=set`

config/v1/types_tlssecurityprofile.go
davidesalerno: I think that this depends on the component desiderata: if it must always be PQC compliant so if `curves` is empty the component have to use a default curve in order to allow ECDHE and explicity declare which one into its documentation, otherwise if it is allowed for the component to work using a standard DHE , if `curves` is empty there is no violation if it does not allow ECDHE

davidesalerno:
> https://github.com/openshift/api/compare/58da6c8c1090bcc312e8f1bd3d8bca87cd47227d..eba1e60731dc58f0c314e2ef10ed38dd3f528c00 dropped X448. Was that intentional, perhaps because Go's crypto/tls package doesn't include support? Do we have agreement from stakeholders that the current set of curves is sufficient?
> 
> https://github.com/openshift/api/blob/eba1e60731dc58f0c314e2ef10ed38dd3f528c00/config/v1/types_tlssecurityprofile.go#L205-L224

@Miciah 
Yes, I removed X448 based on what suggested [here](https://github.com/openshift/api/pull/2583#pullrequestreview-3488383917) @sanchezl about the mapping with Go `crypto/tls`

config/v1/types_tlssecurityprofile.go
yuqi-zhang: Technically you don't need the `NOTE: since this field is optional,` in the godoc, since it will be explained as an optional field, but up to you if you want to highlight that. Otherwise looks fine to me

config/v1/types_tlssecurityprofile.go
yuqi-zhang: I think sets should work to ensure that only one entry of each TLSCurve exists

config/v1/types_tlssecurityprofile.go
yuqi-zhang: Good question, we generally don't have cross-field examples, but I guess in this case, you cannot specify arbitrary curves that the ciphers you have don't support?

Would anything validate that the curves list is correct for the ciphers you have? Or are all curves supported by all ciphers? Would there ever be a need to say, specify a subset of curves per cipher?

(This might be completely invalid, but as an example, you would want X25519MLKEM768 for TLS_CHACHA20_POLY1305_SHA256, and P-384 for TLS_AES_256_GCM_SHA384, and only those combinations)

machineconfiguration/v1/zz_generated.crd-manifests/0000_80_machine-config_01_kubeletconfigs-CustomNoUpgrade.crd.yaml
yuqi-zhang: Not sure on this one. For example, this would be an additional field you can specify when creating a kubeletconfig object, which then the MCO renders into a MachineConfig, which gets rolled out to disk.

Would the controller need to know that this is a special case field it needs to render when it renders the kubelet.conf?

config/v1/types_tlssecurityprofile.go
richardsonnick: IMO the user facing docs call out that the [custom TLS profile "can cause some problems"](https://docs.redhat.com/en/documentation/openshift_container_platform/4.20/html/security_and_compliance/tls-security-profiles#tls-profiles-understanding_tls-security-profiles) if misconfigured which is a good hint to the administrator that they should:
1. Know what they are doing
2. Validate their config before using the profile in a production environment

I think the best approach here is to offer reasonable valid defaults while keeping custom profiles as a "power user" option.

candita:
Replaces https://github.com/openshift/api/pull/2512

candita:
@davidesalerno please work with @richardsonnick to get a jira/rfe number associated with this PR.

config/v1/types_tlssecurityprofile.go
```diff
@@ -202,6 +202,27 @@ const (
 	TLSProfileCustomType TLSProfileType = "Custom"
 )
 
+// TLSCurve is a named curve identifier that can be used in TLSProfile.Curves.
+// There is a one-to-one mapping between these names and the curve IDs defined
+// in crypto/tls package based on IANA's "TLS Supported Groups" registry:
+// https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8
```
kannon92, line 170: Looking at cryto/tls, I see the following values.

```golang
type CurveID uint16

const (
	CurveP256          CurveID = 23
	CurveP384          CurveID = 24
	CurveP521          CurveID = 25
	X25519             CurveID = 29
	X25519MLKEM768     CurveID = 4588
	SecP256r1MLKEM768  CurveID = 4587
	SecP384r1MLKEM1024 CurveID = 4589
)

Why are we not including all the values?
```

kannon92: ref: https://github.com/golang/go/blob/master/src/crypto/tls/common.go#L145

davidesalerno: Ok so I think we can align to those values the available curve supported

yuqi-zhang:
Based on the discussions in https://github.com/openshift/enhancements/pull/1894, it sounds like this is potentially blocked on upstream kubelet/apiserver support for curves?

Would it make sense to introduce this type but only apply it for the TLS configuration in ingress? Or would we like to hold this until we can ensure all components would conform to this?

richardsonnick:
@yuqi-zhang I discussed this with some engineers internally and we decided that the upstream changes should not block this api change. However, we will need to do this at some point to ensure consistency.

kannon92:
> @yuqi-zhang I discussed this with some engineers internally and we decided that the upstream changes should not block this api change. However, we will need to do this at some point to ensure consistency.

I don't really understand how the API will plumb through to kubelet / apiserver if upstream components don't actually set the servers to use these TLS curves.

kannon92:
I had to revert TLS Curve support for Kueue because I can't really add support for configuring TLS curves if there isn't support for this in SecureServing as extended apiservers won't be protected.

yuqi-zhang:
Copying your comment from the slack thread:
```
[TLS 1.3 is non-configurable in golang ](https://github.com/golang/go/issues/29349)(i.e. curves, cipher suites, etc cannot be changed from the default)
The upstream configuration object for the kubelet config does not currently have TLS curve support (and we do not currently have plans to add this)
```

So based on that, we are ok with TechPreview only curve support for TLS 1.2, and only for ingress and apiserver. We don't allow TLS 1.3 + curve, or curve + anything kubelet related.

In that case, since this PR is currently adding curves to the tlsprofile in kubeletconfig as well, perhaps we should split those types such that tlsprofile in kubeletconfig doesn't have curve option, which I think is Kevin's concern.

Please let me know if any of that is incorrect, thanks!

kannon92:
My main concern is similar to @everettraven (https://github.com/openshift/enhancements/pull/1894#issuecomment-3671473044).

Kube-APIServer / Kubelet upstream components do not yet have the flags to set this on their respective servers. So unless I am missing something we would need to carry some kind of patch downstream that adds these CLI parameters so that the servers actually respect the API we set.

Otherwise we end up having an API that doesn't actually configure the kubernetes components to use the right tls curves.

richardsonnick:
"So based on that, we are ok with TechPreview only curve support for TLS 1.2, and only for ingress and apiserver. We don't allow TLS 1.3 + curve, or curve + anything kubelet related."

To my knowledge this comment is correct. We are forced to keep this as TechPreview until TLS1.3 curve specification support is provided upstream.

config/v1/types_tlssecurityprofile.go
```diff
@@ -202,6 +202,31 @@ const (
...
+	// TLSCurveP256 represents P-256 (secp256r1).
+	TLSCurveP256 TLSCurve = "P-256"
+	// TLSCurveP384 represents P-384 (secp384r1).
+	TLSCurveP384 TLSCurve = "P-384"
+	// TLSCurveP521 represents P-521 (secp521r1).
+	TLSCurveP521 TLSCurve = "P-521"
+	// TLSCurveX25519MLKEM768 represents X25519MLKEM768.
+	TLSCurveX25519MLKEM768 TLSCurve = "X25519MLKEM768"
+	// TLSCurveP256r1MLKEM1024 represents P256r1MLKEM1024 (secp256r1MLKEM1024).
+	TLSCurveP256r1MLKEM768 TLSCurve = "P256r1MLKEM768"
```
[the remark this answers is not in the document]
everettraven: Good bot.

I'm unclear as to which this constant is supposed to actually represent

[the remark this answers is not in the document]
everettraven: Also, this value doesn't seem to be represented in the allowed values based on the enum marker.

config/v1/types_tlssecurityprofile.go
```diff
@@ -202,6 +202,31 @@ const (
...
+	// TLSCurveP384r1MLKEM1024 represents P384r1MLKEM1024 (secp384r1MLKEM1024).
+	TLSCurveP384r1MLKEM1024 TLSCurve = "P384r1MLKEM1024"
+)
```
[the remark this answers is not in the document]
everettraven: Good bot.

This needs addressing.

config/v1/types_tlssecurityprofile.go
```diff
@@ -213,6 +238,38 @@ type TLSProfileSpec struct {
 	//
 	// +listType=atomic
 	Ciphers []string `json:"ciphers"`
+	// curves is used to specify the elliptic curves that are used during
+	// the TLS handshake.  Operators may remove entries their operands do
+	// not support.
```
everettraven:
```suggestion
	// curves is an optional field used to specify the elliptic curves that are used during
	// the TLS handshake.  Operators may remove entries their operands do
	// not support.
```

config/v1/types_tlssecurityprofile.go
```diff
@@ -213,6 +238,38 @@ type TLSProfileSpec struct {
...
+	//
+	// TLSProfiles Old, Intermediate, Modern are including by default the following
+	// curves: X25519, P-256, P-384, P-521, X25519MLKEM768, SecP256r1MLKEM1024, SecP384r1MLKEM1024.
```
everettraven: I would probably specify this information on the actual profiles themselves like is done for the existing ciphers field.

This keeps documentation location consistent and omits what seems like an unnecessary reference to a higher-order field.

config/v1/types_tlssecurityprofile.go
```diff
@@ -213,6 +238,38 @@ type TLSProfileSpec struct {
...
+	// TLSProfiles Custom do not include any curves by default.
```
everettraven: Same comment as above, this seems better suited to the higher-order field documentation.

config/v1/types_tlssecurityprofile.go
everettraven: Maybe the example is better suited to the higher-order custom field since it is showing a more holistic configuration example?

config/v1/types_tlssecurityprofile.go
```diff
@@ -213,6 +238,38 @@ type TLSProfileSpec struct {
...
+	// NOTE: since this field is optional, if no curves are specified, the default curves
+	// used by the underlying TLS library will be used.
```
everettraven: What do you think of saying something like:
> When omitted, this means no opinion and the platform is left to choose reasonable defaults which are subject to change over time and may be different per platform component depending on the underlying TLS libraries they use.

It generally follows our more standard "omission means no opinion" for configuration APIs while clearing stating that the default behavior depends on the underlying implementation per-component.

config/v1/types_tlssecurityprofile.go
```diff
@@ -213,6 +238,38 @@ type TLSProfileSpec struct {
...
+	//         - TLS_CHACHA20_POLY1305_SHA256
+	//       curves:
+	//         - X25519MLKEM768  # PQC-only: only hybrid quantum-resistant curve
+	//       minTLSVersion: VersionTLS13
+	//
+	// +optional
+	// +listType=set
+	// +kubebuilder:validation:MaxItems=5
+	// +openshift:enable:FeatureGate=TLSCurvesConfiguration
+	Curves []TLSCurve `json:"curves,omitempty"`
```
everettraven: Would there be any difference in setting this to `curves: []` and omitting the field altogether?

everettraven:
>To my knowledge this comment is correct. We are forced to keep this as TechPreview until TLS1.3 curve specification support is provided upstream.

Keeping it tech preview until support is provided by core components seems reasonable to me. I suspect we would like to get the Kubernetes API server and the Kubelet to support this configuration as part of this effort since both of those configuration APIs rely on this type and any changes are subsequently reflected as configuration options on those APIs as well.

I don't know that this needs to necessarily wait until TLS 1.3 curve configuration support if we are able to block configuring fields for now when minTLSVersion is set to 1.3.

yuqi-zhang:
There's also a few issues from coderabbit (inconsistent curve naming) and CI (lack of unit tests, code generation failures) that should be addressed.

Also had a similar question to https://github.com/openshift/enhancements/pull/1894#issuecomment-3785847877 . Should we split the APIs so the kubeletconfig one doesn't have curve configuration for now?

config/v1/types_tlssecurityprofile.go
```diff
@@ -157,6 +167,31 @@ const (
...
+// https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8
+//
+// +kubebuilder:validation:Enum=X25519;P-256;P-384;P-521;X25519MLKEM768;P256r1MLKEM768;P384r1MLKEM1024
+type TLSCurve string
+
+const (
+	// TLSCurveX25519 represents X25519.
+	TLSCurveX25519 TLSCurve = "X25519"
+	// TLSCurveP256 represents P-256 (secp256r1).
+	TLSCurveP256 TLSCurve = "P-256"
```
candita: Naming nit - please include the whole (well-known) curve name here and elsewhere, e.g.:

```suggestion
	// TLSCurveSecp256r1 represents P-256 (secp256r1).
	TLSCurveSecp256r1 TLSCurve = "secp256r1"
```

davidesalerno: Should we use the same approach for secp384r1 and secp521r1 too? 
And so renaming TLSCurveP384 -> TLSCurveSecp384r1 and TLSCurveP521 -> TLSCurveSecp521r1

alebedev87:
/unassign

It's mostly Miciah who's following up on this feature work.

config/v1/types_tlssecurityprofile.go
damdo: Heads up, those curves you are referencing @kannon92 are all available only from go 1.25.

In go 1.24 there is only a subset of those: https://github.com/golang/go/blob/96e4e2b1616c3c59577d48abcf2823bf1fdcd2e2/src/crypto/tls/common.go#L139-L154

Not sure what that changes anything in this scope @davidesalerno @JoelSpeed @everettraven

config/v1/types_ingress.go
```diff
@@ -330,3 +330,56 @@ const (
...
+type IngressTLSProfileSpec struct {
+	// type is the TLS security profile type.
+	// Valid values are: Old, Intermediate, Modern, Custom.
+	// When omitted, defaults to the platform's recommended security profile.
+	// +optional
+	Type TLSProfileType `json:"type,omitempty"`
+
+	// custom is a user-defined TLS security profile.
+	// +optional
+	// +kubebuilder:validation:MinProperties=0
```
JoelSpeed: Why min of 0 on this struct? It has a required field so you shouldn't need that?

config/v1/types_ingress.go
```diff
@@ -330,3 +330,56 @@ const (
...
+
+// IngressCustomTLSProfile defines the parameters for a custom TLS profile on Ingress.
+type IngressCustomTLSProfile struct {
+	// ciphers is the list of allowed cipher suites.
+	//
+	// +listType=atomic
+	// +kubebuilder:validation:MaxItems=32
+	// +kubebuilder:validation:items:MaxLength=32
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:items:MinLength=1
```
JoelSpeed: How did we reach these figures? Are we sure this will be ok for existing stored resources out there?

features/features.go
```diff
@@ -978,4 +978,12 @@ var (
 						enhancementPR("https://github.com/openshift/enhancements/pull/1465").
 						enableIn(configv1.DevPreviewNoUpgrade).
 						mustRegister()
+
+	FeatureGateTLSCurvesConfiguration = newFeatureGate("TLSCurvesConfiguration").
+						reportProblemsToJiraComponent("Networking").
+						contactPerson("davidesalerno").
+						productScope(ocpSpecific).
+						enhancementPR("https://github.com/openshift/enhancements/pull/1894").
+						enableIn(configv1.TechPreviewNoUpgrade).
```
JoelSpeed: TPNU gates must also be enabled in DPNU

config/v1/types_tlssecurityprofile.go
candita: Yes, here and elsewhere, please include the whole well-known curve name similar to the suggested change.

config/v1/types_ingress.go
davidesalerno: I reached these figures finding a value large enough to accommodate the Old TLS profile, which is the largest standard profile used for legacy compatibility.

config/v1/types_ingress.go
davidesalerno: This is right but unfortunately with the latest kube api linter checks I was forced to add it.
BTW I removed this specific change in order to split the  work to introduce the curves field into the general TLS Profile and the Ingress specific one

JoelSpeed: Are you sure? If the linter wants you to add this, then there's a bug. It shouldn't need a min properties on a struct with required fields, and, it shouldn't have asked you to put it here either, it's not effective here

config/v1/types_tlssecurityprofile.go
kannon92: Good call out.

config/v1/zz_generated.crd-manifests/0000_10_config-operator_01_apiservers-CustomNoUpgrade.crd.yaml
```diff
@@ -339,6 +375,12 @@ spec:
                         - VersionTLS13
                         type: string
                     type: object
+                    x-kubernetes-validations:
+                    - message: curves cannot be configured if minTLSVersion is VersionTLS13
+                        as the underlying Golang implementation does not support customization
+                        for TLS 1.3
+                      rule: 'has(self.minTLSVersion) && self.minTLSVersion == ''VersionTLS13''
+                        ? !has(self.curves) : true'
```
joelanford: Curves are configurable in Go with minTLS 1.3. It is only cipher suites that are not configurable with minTLS 1.3.

config/v1/types_tlssecurityprofile.go
everettraven: @damdo thanks for the callout. If we wanted to scope the default set of provided curves to only what is supported by the core platform components going with the Go 1.24 ones for now would probably make the most sense. It looks like kube-apiserver is still on 1.24: https://github.com/openshift/kubernetes/blob/9d521311f5fb67dc43f49eeb728ee2c80976835a/go.mod#L9

config/v1/types_tlssecurityprofile.go
```diff
@@ -157,6 +167,31 @@ const (
...
+	// TLSCurveSecp384r1 represents P-384 (secp384r1).
+	TLSCurveSecp384r1 TLSCurve = "P-384"
+	// TLSCurveSecp521r1 represents P-521 (secp521r1).
+	TLSCurveSecp521r1 TLSCurve = "P-521"
+	// TLSCurveX25519MLKEM768 represents X25519MLKEM768.
+	TLSCurveX25519MLKEM768 TLSCurve = "X25519MLKEM768"
+	// TLSCurveP256r1MLKEM768 represents P256r1MLKEM768 (secp256r1MLKEM768).
+	TLSCurveP256r1MLKEM768 TLSCurve = "P256r1MLKEM768"
+	// TLSCurveP384r1MLKEM1024 represents P384r1MLKEM1024 (secp384r1MLKEM1024).
+	TLSCurveP384r1MLKEM1024 TLSCurve = "P384r1MLKEM1024"
```
everettraven: I saw a previous comment that mentioned that we should use the fully qualified curve name for the values as well - I agree, especially if those are the curve names that our end-users will be most familiar with.

We follow PascalCase for enum values so I suspect we'd want something like:
```suggestion
	// TLSCurveSecp256r1 represents P-256 (secp256r1).
	TLSCurveSecp256r1 TLSCurve = "SecP256r1"
	// TLSCurveSecp384r1 represents P-384 (secp384r1).
	TLSCurveSecp384r1 TLSCurve = "SecP384r1"
	// TLSCurveSecp521r1 represents P-521 (secp521r1).
	TLSCurveSecp521r1 TLSCurve = "SecP521r1"
	// TLSCurveX25519MLKEM768 represents X25519MLKEM768.
	TLSCurveX25519MLKEM768 TLSCurve = "X25519MLKEM768"
	// TLSCurveP256r1MLKEM768 represents P256r1MLKEM768 (secp256r1MLKEM768).
	TLSCurveP256r1MLKEM768 TLSCurve = "SecP256r1MLKEM768"
	// TLSCurveP384r1MLKEM1024 represents P384r1MLKEM1024 (secp384r1MLKEM1024).
	TLSCurveP384r1MLKEM1024 TLSCurve = "SecP384r1MLKEM1024"
```

config/v1/types_tlssecurityprofile.go
```diff
@@ -168,6 +203,25 @@ type TLSProfileSpec struct {
...
+	// change over time and may be different per platform component depending on the underlying TLS libraries they use.
+	//
+	// For example, to use X25519 and P-256 (yaml):
+	//
+	//   curves:
+	//     - X25519
+	//     - P-256
+	//
+	// +optional
+	// +listType=set
```
everettraven: Let's document that entries to this list must be unique as part of the field documentation.

config/v1/types_tlssecurityprofile.go
everettraven: I don't think this question has been sufficiently answered.

I noticed you added a `+kubebuilder:validation:MinItems=0` marker - is it valid to `curves: []`? What is the difference between explicitly setting it to an empty list and omitting the field altogether?

config/v1/types_tlssecurityprofile.go
```diff
@@ -204,6 +259,9 @@ const (
 // configuration guidelines (2019-06-28) with TLS 1.3 cipher suites prepended for
 // forward compatibility. See: https://ssl-config.mozilla.org/guidelines/5.0.json
 //
+// TLSProfiles Old, Intermediate, Modern are including by default the following
+// curves: X25519, P-256, P-384, P-521, X25519MLKEM768
+//
```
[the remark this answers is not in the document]
everettraven: This looks to be a valid comment that needs to be addressed. Either update the actual implementation or this comment to match what the correct state should be please.

config/v1/zz_generated.crd-manifests/0000_10_config-operator_01_apiservers-CustomNoUpgrade.crd.yaml
davidesalerno: I removed the not needed validation

config/v1/types_tlssecurityprofile.go
davidesalerno: Unfortunately if I remove the minItems I receive the error:

`config/v1/types_tlssecurityprofile.go:224:2: minlength: field TLSProfileSpec.Curves must have a minimum items, add kubebuilder:validation:MinItems marker (kubeapilinter)
`
Is it this a kube api linter bug? How could we workaround it?

Should we put 1 because if it is present, we are expecting at least 1 value?

JoelSpeed: The question is more about behaviour.

As an end user, do you expect there to be a semantic difference between not setting `curves` at all, vs `curves: []`, what would an explicit zero length list mean differently from omitted?

If there's no semantic difference, the appropriate change here is `MinItems=1`

davidesalerno: From a user experience point of view I think it's better this option (`MinItems=1`) so I'll change the code accordingly

config/v1/types_tlssecurityprofile.go
davidesalerno: I restricted the curves to the 5 ones that are available in golang 1.24 

@damdo thanks for the call out

config/v1/tests/apiservers.config.openshift.io/TLSCurvePreferences.yaml
```diff
@@ -0,0 +1,311 @@
...
+        spec:
+          audit:
+            profile: Default
+          tlsSecurityProfile:
+            type: Custom
+            custom:
+              minTLSVersion: VersionTLS12
+              ciphers:
+                - TLS_AES_128_GCM_SHA256
+              curves: []
```
everettraven: If you just remove the field, is this allowed?

config/v1/types_tlssecurityprofile.go
```diff
@@ -157,6 +167,27 @@ const (
...
+
+const (
+	// TLSCurveX25519 represents X25519.
+	TLSCurveX25519 TLSCurve = "X25519"
+	// TLSCurveSecp256r1 represents P-256 (secp256r1).
+	TLSCurveSecp256r1 TLSCurve = "SecP256r1"
+	// TLSCurveSecp384r1 represents P-384 (secp384r1).
+	TLSCurveSecp384r1 TLSCurve = "SecP384r1"
+	// TLSCurveSecp521r1 represents P-521 (secp521r1).
+	TLSCurveSecp521r1 TLSCurve = "SecP521r1"
```
everettraven: Can we match the constant Go name and the values here?

```suggestion
	TLSCurveSecP256r1 TLSCurve = "SecP256r1"
	// TLSCurveSecP384r1 represents P-384 (secp384r1).
	TLSCurveSecP384r1 TLSCurve = "SecP384r1"
	// TLSCurveSecP521r1 represents P-521 (secp521r1).
	TLSCurveSecP521r1 TLSCurve = "SecP521r1"
```

Or if the lowercase 'p' is the appropriate pattern:

```suggestion
	TLSCurveSecp256r1 TLSCurve = "Secp256r1"
	// TLSCurveSecp384r1 represents P-384 (secp384r1).
	TLSCurveSecp384r1 TLSCurve = "Secp384r1"
	// TLSCurveSecp521r1 represents P-521 (secp521r1).
	TLSCurveSecp521r1 TLSCurve = "Secp521r1"
```

I'll defer to you as to which format is more appropriate

everettraven (commented):
Aside from a couple minor comments, this LGTM.

config/v1/tests/apiservers.config.openshift.io/TLSCurvePreferences.yaml
everettraven: It looks like it is. Lets add a test that removing the field altogether is allowed

everettraven (approved):
/lgtm

everettraven:
`verify-crd-schema` is failing due to new CRDs being generated for machine config types from their dependency on the `TLSProfile` types.

The failure is an existing issue that wouldn't normally block a PR. Overriding due to existing issue.

config/v1/types_tlssecurityprofile.go
```diff
@@ -142,6 +152,27 @@ const (
...
+	TLSCurveX25519 TLSCurve = "X25519"
+	// TLSCurveSecp256r1 represents P-256 (secp256r1).
+	TLSCurveSecP256r1 TLSCurve = "SecP256r1"
+	// TLSCurveSecP384r1 represents P-384 (secp384r1).
+	TLSCurveSecP384r1 TLSCurve = "SecP384r1"
+	// TLSCurveSecP521r1 represents P-521 (secp521r1).
+	TLSCurveSecP521r1 TLSCurve = "SecP521r1"
+	// TLSCurveX25519MLKEM768 represents X25519MLKEM768.
+	TLSCurveX25519MLKEM768 TLSCurve = "X25519MLKEM768"
+)
```
damdo, line 192: I'm no crypto expert to take my questions and assumptions with a grain of salt here.

Is there a specific reason why are going with IANA names for Curves here whereas for Ciphers we went with OpenSSL names?

My understanding is that Go uses IANA as a reference for both Ciphers and Curves: https://github.com/golang/go/blob/96e4e2b1616c3c59577d48abcf2823bf1fdcd2e2/src/crypto/tls/common.go#L139-L154

I was under the impression that for Ciphers we originally went with OpenSSL names (rather than IANA) so that non go network components like HAProxy, kube-rbac-proxy, ... & other components can use the OpenSSL names natively, whereas for go we can always put them through the `library-go`'s `crypto.OpenSSLToIANACipherSuites()` to get the IANA formatted ciphers names to use with go's tls.Config.

For these Curves instead it looks to me we are going the other way around?

Wouldn't it be more consistent to go with OpenSSL names by default in the API and then have a `crypto.OpenSSLToIANACurvePreferences()` to get the IANA formatted curve names?

@davidesalerno @everettraven @joelanford @kannon92 @sanchezl ?
Thanks

everettraven: I could be missing something, but I couldn't find a canonical list of OpenSSL curve names where there is a clear deviation from the IANA names.

I don't have any strong opinions here, but if there _is_ a clear difference I think sticking with the ciphers convention makes sense here.

candita: Curves are user-visible fields that we'd like to keep in the most common user-visible formatting, and IANA is the best source for that.  We don't need to keep consistent with cipher formatting if the format that was used there was not what we would pick today.

@davidesalerno please change the string constant using the same case as in https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8, it doesn't make sense to use camel case on user-input strings.  We may need to make sure that validation ignores case as well, if it doesn't already.

If this was the last item on the review list, can we please proceed asap with this PR?

damdo: We have had a longer internal conversation on this here: https://redhat-internal.slack.com/archives/C098FU5MRAB/p1770317849553639

Let's coordinate there @candita @davidesalerno @everettraven

config/v1/types_tlssecurityprofile.go
```diff
@@ -222,6 +276,12 @@ var TLSProfiles = map[TLSProfileType]*TLSProfileSpec{
 			"AES256-SHA",
 			"DES-CBC3-SHA",
 		},
+		Curves: []TLSCurve{
+			TLSCurveX25519,
+			TLSCurveSecP256r1,
+			TLSCurveSecP384r1,
+			TLSCurveX25519MLKEM768,
+		},
```
candita: In my opinion, it doesn't make sense for use to hardcode the exact same curve list to each profile.  The programs that use these configurations should hardcode the curve list.  We should allow the Custom profile to set whatever curves are needed if they aren't the default.

features/features.go
```diff
@@ -938,5 +938,13 @@ var (
 				enhancementPR("https://github.com/openshift/enhancements/pull/1933").
 				enableIn(configv1.DevPreviewNoUpgrade, configv1.TechPreviewNoUpgrade).
 				mustRegister()
+
+	FeatureGateTLSCurvePreferences = newFeatureGate("TLSCurvePreferences").
+						reportProblemsToJiraComponent("Networking").
```
Miciah: We should use a real component name:
```suggestion
						reportProblemsToJiraComponent("Networking / router").
```

config/v1/types_tlssecurityprofile.go
Miciah: Why should the curves list be handled differently to the ciphers list and minimum TLS version?  I expect that a curve could be removed from the Modern or even Intermediate profile if it were found to be unsafe.

config/v1/types_tlssecurityprofile.go
Miciah: To follow up on this thread, I notice we also now have relevant test cases for `curves: []` (which is invalid) and for `curves` omitted (which is valid, and does not get a default value set):
https://github.com/openshift/api/blob/b6786238ec55cda6f5d7d1e09a185ed7a5f9dbab/config/v1/tests/apiservers.config.openshift.io/TLSCurvePreferences.yaml#L74-L109
Thanks for adding those!

features/features.go
```diff
@@ -1041,4 +1033,12 @@ var (
 					enhancementPR("https://github.com/openshift/enhancements/pull/1933").
 					enable(inDevPreviewNoUpgrade(), inTechPreviewNoUpgrade()).
 					mustRegister()
+
+	FeatureGateTLSCurvePreferences = newFeatureGate("TLSCurvePreferences").
+						reportProblemsToJiraComponent("Networking").
+						contactPerson("davidesalerno").
+						productScope(ocpSpecific).
+						enhancementPR("https://github.com/openshift/enhancements/pull/1894").
+						enable(inDevPreviewNoUpgrade(), inTechPreviewNoUpgrade()).
```
everettraven, line 985: Just a note that as of more recently we are trying to ensure that enhancements are merged prior to a feature going into tech preview.

I'm not going to block on this because the EP and this PR was created prior to that newer requirement, but we should work on getting the EP merged as a signal of general agreement that the work is reasonable to do and put into tech preview.

features/features.go
everettraven: Looks like this still needs doing

config/v1/types_tlssecurityprofile.go
everettraven: I'm not opinionated here so I'll defer to others here. Let's try to get this resolved before merging though.

davidesalerno: @sanchezl  specifically asked for this change.

Anyway I agree with Miciah and we can improve this part handling curves in the same way of cipher list and so:
- Modern: X25519, X25519MLKEM768 (only the strongest)
- Intermediate: X25519, P-256, P-384, X25519MLKEM768 (current set)
- Old: X25519, P-256, P-384, P-521, X25519MLKEM768 (broadest compatibility)

config/v1/types_tlssecurityprofile.go
```diff
@@ -155,6 +186,26 @@ type TLSProfileSpec struct {
...
+	// For example, to use X25519 and secp256r1 (yaml):
+	//
+	//   curves:
+	//     - X25519
+	//     - secp256r1
+	//
+	// +optional
+	// +listType=set
+	// +kubebuilder:validation:MaxItems=5
+	// +kubebuilder:validation:MinItems=1
```
everettraven: Small thing I missed: These constraints should be explicitly mentioned in the GoDoc - end-users will never see these markers in the documentation that is generated from the APIs because the GoDoc is what is used for documentation generation and that process strips the markers from the documentation.

config/v1/types_tlssecurityprofile.go
everettraven: Sounds good to me.

Let's make sure this decision is appropriately reflected here then.

config/v1/tests/apiservers.config.openshift.io/TLSCurvePreferences.yaml
```diff
@@ -0,0 +1,349 @@
...
+        apiVersion: config.openshift.io/v1
+        kind: APIServer
+        spec:
+          tlsSecurityProfile:
+            type: Custom
+            custom:
+              minTLSVersion: VersionTLS12
+              ciphers:
+                - TLS_AES_128_GCM_SHA256
+              curves:
```
neverpanic: Please do not use curves for this, it's outdated. The correct name to use now is "groups". MLKEM768 is not an (elliptic) curve.

See also https://github.com/openshift/enhancements/pull/1894#pullrequestreview-3915222687

neverpanic (changes requested):

features/features.go
davidesalerno: Updated

candita:
On hold for now.
/hold

config/v1/tests/apiservers.config.openshift.io/TLSCurvePreferences.yaml
```diff
@@ -0,0 +1,349 @@
...
+        apiVersion: config.openshift.io/v1
+        kind: APIServer
+        spec:
+          tlsSecurityProfile:
+            type: Custom
+            custom:
+              minTLSVersion: VersionTLS12
+              ciphers:
+                - TLS_AES_128_GCM_SHA256
+              curves: []
```
ormergi: The distinction between curves nil vs. empty is confusing and unexpected.
From a consumer perspective its seem harder to work with and require more tests.
Please consider keeping the same semantics when curves is empty or nil.

config/v1/types_tlssecurityprofile.go
ormergi: FWIW, I think it should be OpenSSL form and consistent with other fields (min version, ciphers).
Although it doesnt serve well Go projects (upstream who cannot consume openshift/ modules), it seems a good enough middle ground.
This API cannot satisfy all consumers, today some component expect OpenSSL and tomorrow some new component expected different format.

config/v1/tests/apiservers.config.openshift.io/TLSCurvePreferences.yaml
everettraven: From what I recall, the semantics are the same.

This test is testing that `curves: []` is not allowed, ensuring that there is only one way to say "no curves" - which is omission of the `curves` field altogether.

If a user specifies the `curves` field, they must specify at least one item in the list.

This matches our API conventions guidance of having only one way to represent something so that there isn't contention as to what is the "right" way to configure something.

config/v1/tests/apiservers.config.openshift.io/TLSCurvePreferences.yaml
davidesalerno: I renamed the field and the feature gate too

config/v1/types_tlssecurityprofile.go
davidesalerno: I added the needed documentation

config/v1/tests/apiservers.config.openshift.io/TLSCurvePreferences.yaml
davidesalerno: @ormergi  this follows the standard Kubernetes `+optional` `+MinItems` pattern, and the semantic difference is by design: omitted means "platform chooses defaults", while empty is explicitly invalid.

openapi/generated_openapi/zz_generated.openapi.go
```diff
@@ -20947,6 +20967,26 @@ func schema_openshift_api_config_v1_TLSProfileSpec(ref common.ReferenceCallback)
...
+							},
+						},
+						SchemaProps: spec.SchemaProps{
+							Description: "curves is an optional field used to specify the elliptic curves that are used during the TLS handshake.  Operators may remove entries their operands do not support.\n\nWhen omitted, this means no opinion and the platform is left to choose reasonable defaults which are subject to change over time and may be different per platform component depending on the underlying TLS libraries they use. If specified, the list must contain at least one curve and each curve must be unique.\n\nFor example, to use X25519 and secp256r1 (yaml):\n\n  curves:\n    - X25519\n    - secp256r1",
+							Type:        []string{"array"},
+							Items: &spec.SchemaOrArray{
+								Schema: &spec.Schema{
+									SchemaProps: spec.SchemaProps{
+										Default: "",
+										Type:    []string{"string"},
```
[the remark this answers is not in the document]
davidesalerno: The OpenAPI in zz_generated.openapi.go is auto-generated by make update. If the enum isn't showing up, it's likely a limitation of the openapi-gen tooling or because the field is feature-gated.

config/v1/types_tlssecurityprofile.go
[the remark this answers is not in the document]
davidesalerno: At the moment this is based on a wrong assumption that order matters  (see EP  https://github.com/openshift/enhancements/pull/1894 )

config/v1/types_tlssecurityprofile.go
```diff
@@ -244,6 +311,12 @@ var TLSProfiles = map[TLSProfileType]*TLSProfileSpec{
 			"TLS_AES_256_GCM_SHA384",
 			"TLS_CHACHA20_POLY1305_SHA256",
 		},
+		Groups: []TLSGroup{
+			TLSGroupX25519,
+			TLSGroupSecP256r1,
+			TLSGroupSecP384r1,
+			TLSGroupX25519MLKEM768,
+		},
```
Miciah, line 346: It is confusing to compare these lists of groups/curves and this note from the godoc: https://github.com/openshift/api/blob/6026d0cd8b54b7dded1e246531e78e358f6e668f/config/v1/types_tlssecurityprofile.go#L10-L13

https://ssl-config.mozilla.org/guidelines/5.7.json does not list the X25519MLKEM768 curve/group:
```json
{
    "version": 5.7,
    "href": "https://ssl-config.mozilla.org/guidelines/5.7.json",
    "configurations": {
        "modern": {
            /* ... */
            "tls_curves": ["X25519", "prime256v1", "secp384r1"],
            /* ... */
        },
        "intermediate": {
            /* ... */
            "tls_curves": ["X25519", "prime256v1", "secp384r1"],
            /* ... */
        },
        "old": {
            /* ... */
            "tls_curves": ["X25519", "prime256v1", "secp384r1"],
            /* ... */
        }
    }
}
```

What is the source of truth for the list of groups/curves?  Is it Mozilla, NIST, or something else?  Should the note be updated to add a reference for where the list of groups/curves comes from?  

Also, should the godoc mention that X25519MLKEM768 may/should be ignored when FIPS is enabled?

candita: Note that I found out from the FIPS team that `X25519` and `X25519MLKEM768` cannot be used with FIPS, but `SecP256r1MLKEM768` and `SecP384r1MLKEM1024` can be used.  Unfortunately there is no support for the latter two in golang < 1.26.  Do we have to conclude that PQC is not supported on FIPS?

davidesalerno: Good catch, thanks for raising this.

The ciphers are sourced from Mozilla 5.7 guidelines, but the groups list is based on Go 1.24+'s crypto/tls default curve preferences (the defaults when Config.CurvePreferences is nil), which include X25519MLKEM768 as a post-quantum hybrid group. The IANA "TLS Supported Groups" registry is the authoritative reference for group identifiers.

I think that to fix this I can update the godoc in order to:
1. Clarify that ciphers come from Mozilla 5.7, while groups come from Go's crypto/tls defaults (with a link to https://pkg.go.dev/crypto/tls#CurveID)
2. Note that X25519MLKEM768 is not FIPS-approved and must be omitted/ignored by components running in FIPS mode

Let me know if this works for you too.

config/v1/types_tlssecurityprofile.go
```diff
@@ -23,6 +29,9 @@ type TLSSecurityProfile struct {
 	// old is a TLS profile for use when services need to be accessed by very old
 	// clients or libraries and should be used only as a last resort.
 	//
+	// The supported groups list includes by default the following groups:
+	// X25519, secp256r1, secp384r1, X25519MLKEM768.
```
candita, line 33: In go 1.25 this list should include `secp521r1` http://pkg.go.dev/crypto/tls@go1.25.9#CurveID

In go 1.26 curves should additionally include `SecP256r1MLKEM768` and
	`SecP384r1MLKEM1024`  https://pkg.go.dev/crypto/tls@go1.26.2#CurveID

config/v1/types_tlssecurityprofile.go
```diff
@@ -222,6 +289,12 @@ var TLSProfiles = map[TLSProfileType]*TLSProfileSpec{
 			"AES256-SHA",
 			"DES-CBC3-SHA",
 		},
+		Groups: []TLSGroup{
+			TLSGroupX25519,
+			TLSGroupSecP256r1,
+			TLSGroupSecP384r1,
+			TLSGroupX25519MLKEM768,
+		},
```
candita, line 312: We should discuss ordering the groups with the preferred PQC-safe group first, and carefully consider the ordering after that to pick either the most popular or the most secure groups.  The ordering should be preferential.

Perhaps we may even want to change the ordering for each different tlsSecurityProfile?

cc @Miciah for feedback.

config/v1/tests/apiservers.config.openshift.io/TLSCurvePreferences.yaml
ormergi: Thanks for explaining, I missed the minItem marker forcing at least one item. sorry for the noise.

config/v1/types_tlssecurityprofile.go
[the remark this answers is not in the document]
candita: @davidesalerno the order should matter.  Whatever is using the groups list should take the first one as the preferred group, with preference in descending order.  If they want to prefer PQC-safe groups, then use ordinary groups if PQC-safe isn't offered, the ordering is important.  Can you take a look into this change?

config/v1/types_tlssecurityprofile.go
```diff
@@ -187,16 +251,23 @@ const (
 
 // TLSProfiles contains a map of TLSProfileType names to TLSProfileSpec.
 //
-// These profiles are based on version 5.7 of the Mozilla Server Side TLS
-// configuration guidelines. See: https://ssl-config.mozilla.org/guidelines/5.7.json
+// The cipher lists in these profiles are based on version 5.7 of the Mozilla
+// Server Side TLS configuration guidelines.
+// See: https://ssl-config.mozilla.org/guidelines/5.7.json
```
candita: Should we be using the 5.8 version, at https://ssl-config.mozilla.org/guidelines/5.8.json ?

everettraven (approved):
Aside from the ordering and 5.8 vs 5.7 guidelines nuances that I don't have much experience or background with, this LGTM.

I'll defer to @candita and/or @Miciah for any strong opinions on these concerns.

/lgtm

config/v1/types_tlssecurityprofile.go
davidesalerno: @candita just to be on the same page and in order to avoid not needed changes is it fine to go with the following changes?

 - Groups reorder: `X25519MLKEM768` first (PQC-preferred), then `X25519`, `secp256r1`, `secp384r1`
  - `secp521r1` removed from default: not in Mozilla 5.8 (still available in Custom profiles via the enum)
  - 3 new ciphers to add to Old profile: `ECDHE-ECDSA-AES256-SHA384`, `ECDHE-RSA-AES256-SHA384`, `AES256-SHA256` - this is a little bit out of topic for the main purpose of this PR but if we would like to be aligned with 5.8 Mozilla TLS configuration guidelines we have to update them
  - All version references updated from 5.7 to 5.8
  
I have also to change `listType=set `to  `listType=atomic` and add uniqueness `XValidation` (a CEL validation rule `self.all(x, self.exists_one(y, x == y))`  to enforces uniqueness is needed)

features/features.go
candita: The EP is now merged.

config/v1/zz_generated.swagger_doc_generated.go
```diff
@@ -3005,6 +3005,7 @@ func (OldTLSProfile) SwaggerDoc() map[string]string {
 var map_TLSProfileSpec = map[string]string{
 	"":              "TLSProfileSpec is the desired behavior of a TLSSecurityProfile.",
 	"ciphers":       "ciphers is used to specify the cipher algorithms that are negotiated during the TLS handshake.  Operators may remove entries their operands do not support.  For example, to use DES-CBC3-SHA  (yaml):\n\n  ciphers:\n    - DES-CBC3-SHA",
+	"curves":        "curves is used to specify the elliptic curves that are used during the TLS handshake.  Operators may remove entries their operands do not support.\n\nWhen omitted, this means no opinion and the platform is left to choose reasonable defaults which are subject to change over time and may be different per platform component depending on the underlying TLS libraries they use.\n\nFor example, to use X25519 and P-256 (yaml):\n\n  curves:\n    - X25519\n    - P-256",
 	"minTLSVersion": "minTLSVersion is used to specify the minimal version of the TLS protocol that is negotiated during the TLS handshake. For example, to use TLS versions 1.1, 1.2 and 1.3 (yaml):\n\n  minTLSVersion: VersionTLS11",
 }
```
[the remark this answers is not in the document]
candita: @coderabbit do not review auto-generated code.

config/v1/types_tlssecurityprofile.go
candita: I noticed that the Mozilla server side TLS guidelines have already removed the `Old` profile, but we probably need to support it for some time.  https://wiki.mozilla.org/Security/Server_Side_TLS

I don't know why it still exists in the json.  It's worth asking the Mozilla community.

config/v1/types_tlssecurityprofile.go
```diff
@@ -155,6 +198,29 @@ type TLSProfileSpec struct {
 	// and are always enabled when TLS 1.3 is negotiated.
 	// +listType=atomic
 	Ciphers []string `json:"ciphers"`
+	// groups is an optional, ordered field used to specify the supported groups (formerly known as
+	// elliptic curves) that are used during the TLS handshake.  The order of the groups represents
+	// preference, with the most preferred group first.  Operators may remove entries their operands
+	// do not support.
```
ricardomaraschini: It seems like Go has an [internal preference](https://github.com/golang/go/blob/master/src/crypto/tls/common.go#L802-L806) and the order is ignored. From the Go code:

> The order of the list is ignored, and key exchange mechanisms are chosen
> from this list using an internal preference order. If empty, the default
> will be used.

ingvagabund: Worth mentioning Go (and possibly other language) based solutions can ignore the ordering.

davidesalerno: @ricardomaraschini @ingvagabund  

Good catch, thanks! 
You're right, Go's `CurvePreferences` acts as a filter/set of allowed groups rather than an ordered preference list. Go applies its own internal ordering via `curvePreferenceOrder()`. I'll update the documentation to clarify that the ordering is a preference hint that may not be honored by all implementations - Go-based components will use their internal preference order. 

This way consumers of other TLS stacks (like HAProxy/OpenSSL for the router) can still benefit from the ordering, while Go-based components understand the list is effectively a set of allowed groups.

Could this solution work for you?

candita: Fortunately, the default 1.25 ordering: [X25519MLKEM768, X25519, CurveP256, CurveP384, CurveP521](https://github.com/golang/go/blob/release-branch.go1.25/src/crypto/tls/defaults.go#L18-L25) coincides with the ordering in this PR, so only the comment needs to change.

```
Groups: []TLSGroup{
			TLSGroupX25519MLKEM768,
			TLSGroupX25519,
			TLSGroupSecP256r1,
			TLSGroupSecP384r1,
		},
```

config/v1/types_tlssecurityprofile.go
candita: Update on secp521r1: it is still listed for golang 1.25+, so I'm not sure about removing it.

candita: I think if we can get this merged now as-is, that we can follow up on P-521 in the future.  I wrote a briefing on why I think we should include it, shared at our standup channel, but it can always be included in a custom profile if necessary.

candita:
/lgtm

In the interest of a timely merge, any comment left over from me can be addressed in a followup PR. We have a few other items we need to add to the API anyway.

candita:
I am surprised my lgtm triggered testing.  I thought this still needed approval.  But one lgtm from @everettraven and one from me sent it into testing.  Will it automatically go into the merge pool if the tests succeed?

simonpasquier:
> I am surprised my lgtm triggered testing. I thought this still needed approval. But one lgtm from @everettraven and one from me sent it into testing. Will it automatically go into the merge pool if the tests succeed?

The PR was already `/approve` by @everettraven (https://github.com/openshift/api/pull/2583#pullrequestreview-4310366492) 2 weeks ago. The bot accepts `/lgtm` from anybody part of the OpenShift org. To be merged, the PR only needs someone issuing a `/verified` comment (again not restricted to the repository's owners AFAIK).

config/v1/types_tlssecurityprofile.go
```diff
@@ -155,6 +198,29 @@ type TLSProfileSpec struct {
...
+	//
+	//   groups:
+	//     - X25519
+	//     - secp256r1
+	//
+	// +optional
+	// +listType=atomic
+	// +kubebuilder:validation:MaxItems=7
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:XValidation:rule="self.all(x, self.exists_one(y, x == y))",message="groups cannot contain duplicates"
```
everettraven: Instead of having a CEL validation here, I think we can get away with this being as `+listType=set` since the type is an alias to a string.

That should enforce the uniqueness constraint that the CEL expression currently enforces.

config/v1/types_tlssecurityprofile.go
everettraven: I'd like to see the comment updated to reflect the actual potential behavior of components. Please update the comment to clarify the ordering/preference behavior here to prevent potential end-user confusion.

everettraven:
Removing my approval until my outstanding comments are addressed.

/approve cancel

everettraven:
Removing the hold based on @candita 's recent comment.

/hold cancel

everettraven (approved):
/lgtm
/approve

candita:
@everettraven does your team require an implementation or e2e test in order to get the verified flag?

candita:
/retitle [NE-2751](https://redhat.atlassian.net/browse/NE-2751): Implement enhancement in OpenShift API to support for TLS curves in TLSProfile

PR #2585, opened 2025-11-17
[WIP] OCPBUGS-62145: User response does not include list of groups user is part of

krishagarwal278:
## Summary

The User resource's `groups` field was returning `null` when not populated, causing issues with the Pipelines Operator's role-based approval functionality.

This change adds a default empty array for the `groups` field in the User API schema, ensuring it returns `[]` instead of `null` when the user is not a member of any groups.

  Changes:
  - Added `// +default=[]` marker to User.Groups field in user/v1/types.go
  - Regenerated OpenAPI schema with default empty array

Fixes: [OCPBUGS-62145 ](https://issues.redhat.com/browse/OCPBUGS-62145)

### DIFF
--- a/user/v1/types.go
+++ b/user/v1/types.go
@@ -36,6 +36,7 @@ type User struct {
 	// groups specifies group names this user is a member of.
 	// This field is deprecated and will be removed in a future release.
 	// Instead, create a Group object containing the name of this User.
+	// +default=[]
 	Groups []string `json:"groups" protobuf:"bytes,4,rep,name=groups"`
 }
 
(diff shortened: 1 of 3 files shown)


user/v1/types.go
```diff
@@ -36,6 +36,7 @@ type User struct {
 	// groups specifies group names this user is a member of.
 	// This field is deprecated and will be removed in a future release.
 	// Instead, create a Group object containing the name of this User.
+	// +default=[]
```
everettraven, line 39: This type is served via an aggregated apiserver and defaulting this here would have no affect AFAIK.

PR #2586, opened 2025-11-17
AGENT-1372: Rename NoRegistryClusterOperations to NoRegistryClusterInstall

rwsu:
Renamed the feature gate from NoRegistryClusterOperations to NoRegistryClusterInstall to better reflect that this feature gate controls installation rather than general cluster operations.

Installation includes initial cluster installation and adding nodes during day 2.

### DIFF
--- a/features/features.go
+++ b/features/features.go
@@ -799,7 +799,7 @@ var (
 				enableIn(configv1.DevPreviewNoUpgrade, configv1.TechPreviewNoUpgrade, configv1.Default).
 				mustRegister()
 
-	FeatureGateNoRegistryClusterOperations = newFeatureGate("NoRegistryClusterOperations").
+	FeatureGateNoRegistryClusterInstall = newFeatureGate("NoRegistryClusterInstall").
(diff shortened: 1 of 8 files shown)


features/features.go
```diff
@@ -799,7 +799,7 @@ var (
 				enableIn(configv1.DevPreviewNoUpgrade, configv1.TechPreviewNoUpgrade, configv1.Default).
 				mustRegister()
 
-	FeatureGateNoRegistryClusterOperations = newFeatureGate("NoRegistryClusterOperations").
+	FeatureGateNoRegistryClusterInstall = newFeatureGate("NoRegistryClusterInstall").
```
everettraven, line 802: Just as a sanity check - there is _nothing_ that relies on this today that may break due to this rename?

everettraven: I've answered my own question by looking at the existing test failures - those look unrelated so this should be OK as it has never gone beyond TPNU.

everettraven:
@rwsu over to you to get the verified label added.

PR #2587, opened 2025-11-18
AUTOSCALE-172: add feature gate for ProvisioningRequest CRD

elmiko:
this change adds a new feature gate that will be utilized by the cluster autoscaler operator.


elmiko:
this will be utilized by https://github.com/openshift/cluster-autoscaler-operator/pull/340

elmiko:
/verified by elmiko

i chatted with @everettraven about adding this.

PR #2591, opened 2025-11-20
OCPBUGS-61858: Use pointer type for HTTPKeepAliveTimeout in IngressController API

alebedev87:
This commit addresses the problem of the creation of the default IngressController without `HTTPKeepAliveTimeout` tuning option by the cluster ingress operator:
```
"IngressController.operator.openshift.io \"default\" is invalid: spec.tuningOptions.httpKeepAliveTimeout: Invalid value: \"string\": httpKeepAliveTimeout must be greater than or equal to 1 millisecond"
```

The CEL expression which checks the minimal value prevents the creation of `IngressController` with unspecified (zero value) `HTTPKeepAliveTimeout`.

Follow-up of https://github.com/openshift/api/pull/2547.

### DIFF
--- a/operator/v1/types_ingress.go
+++ b/operator/v1/types_ingress.go
@@ -1912,7 +1912,7 @@ type IngressControllerTuningOptions struct {
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=16
 	// +optional
-	HTTPKeepAliveTimeout metav1.Duration `json:"httpKeepAliveTimeout,omitempty"`
+	HTTPKeepAliveTimeout *metav1.Duration `json:"httpKeepAliveTimeout,omitempty"`
 
 	// tlsInspectDelay defines how long the router can hold data to find a
 	// matching route.
--- a/operator/v1/zz_generated.deepcopy.go
+++ b/operator/v1/zz_generated.deepcopy.go
@@ -2564,7 +2564,11 @@ func (in *IngressControllerTuningOptions) DeepCopyInto(out *IngressControllerTun
 		*out = new(metav1.Duration)
 		**out = **in
 	}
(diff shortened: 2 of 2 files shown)


alebedev87:
The problem was indeed in the non pointer type: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1297/pull-ci-openshift-cluster-ingress-operator-master-e2e-azure-operator/1991445858800898048.

operator/v1/types_ingress.go
```diff
@@ -1912,7 +1912,7 @@ type IngressControllerTuningOptions struct {
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=16
 	// +optional
-	HTTPKeepAliveTimeout metav1.Duration `json:"httpKeepAliveTimeout,omitempty"`
+	HTTPKeepAliveTimeout *metav1.Duration `json:"httpKeepAliveTimeout,omitempty"`
```
everettraven, line 1915: We try to avoid pointers where possible.

If the zero value is not a valid value, there isn't really any need for a pointer. It looks like `metav1.Duration` is a struct and not a type alias.

Does adding `omitzero` resolve this issue?

```suggestion
	HTTPKeepAliveTimeout metav1.Duration `json:"httpKeepAliveTimeout,omitempty,omitzero"`
```

everettraven: @alebedev87 reminded me that this needs to be backported to 4.16 which will not support the `omitzero` tag.

Pointer is correct here.

alebedev87:
As discussed with NI&D QE lead (@lihongan), the API change is fine to be validated by the CI.

/verified by ci-jobs

[Similar example](https://github.com/openshift/api/pull/2584#issuecomment-3555492780).

PR #2592, opened 2025-11-20
CONSOLE-4882: Support mailto: links in ConsoleLink href field

jhadvig:
Tested the change on a live cluster and worked as desired.

/assign @spadgett @JoelSpeed @everettraven


everettraven (approved):
Spoke with @jhadvig and this validation change is the only thing necessary to ensure that this works.

This is a very small and unlikely to break anything change.

/lgtm

PR #2630, opened 2025-12-17
Add prometheusConfig API

danielmellado:
This commit introduces a new API to be introduced as a part of the
migration in CMO from ConfigMap to CRDs

Signed-off-by: Mario Fernandez <mariofer@redhat.com>
Signed-off-by: Daniel Mellado <dmellado@fedoraproject.org>


danielmellado:
The verify-crd-schema and verify-crdify jobs are failing because we intentionally changed the volumeClaimTemplate field from the full corev1.PersistentVolumeClaim type to a simpler VolumeClaimConfig struct with just sizeInGiB and storageClassName.

This was done based on reviewer feedback to provide better admission-time validation and avoid exposing ~40 PVC fields that don't make sense for this use case.

Since this is a TechPreview API (v1alpha1), breaking changes are allowed, and both tools explicitly say it's appropriate to override when you're making intentional changes rather than introducing regressions.

@everettraven do you have any another proposal for this? Besides that, my plan is to split this into smaller changes to make it easier to work/review if that's ok for you.

cc @marioferh

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -416,6 +427,549 @@ type MetricsServerConfig struct {
...
+	// enforcedBodySizeLimitBytes enforces a body size limit (in bytes) for Prometheus scraped metrics.
+	// If a scraped target's body response is larger than the limit, the scrape will fail.
+	// This helps protect Prometheus from targets that return excessively large responses.
+	// The value is specified in bytes (e.g., 4194304 for 4MB, 1073741824 for 1GB).
+	// When omitted, the Cluster Monitoring Operator automatically calculates an appropriate
+	// limit based on cluster capacity. Set an explicit value to override the automatic calculation.
+	// Minimum value is 1 byte.
+	// Maximum value is 1099511627776 (1TB).
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=1099511627776
```
everettraven: Are these reasonable?

When might someone want to reasonably set this to 1 byte? When might they want to set this 1 terabyte?

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -416,6 +427,549 @@ type MetricsServerConfig struct {
...
+type AdditionalAlertmanagerConfig struct {
+	// name is a unique identifier for this Alertmanager configuration entry.
+	// This field is used as the map key for server-side apply, allowing multiple
+	// controllers (e.g., ACM, cluster admins) to independently manage their own entries.
+	// The name must be a valid DNS subdomain (RFC 1123): lowercase alphanumeric characters,
+	// hyphens, or periods, and must start and end with an alphanumeric character.
+	// Maximum length is 253 characters.
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:XValidation:rule="self.matches('^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$')",message="must be a valid DNS subdomain: lowercase alphanumeric, hyphens, or periods"
```
everettraven: I think there is an existing format utility for DNS 1123 subdomain names.

Example:
https://github.com/openshift/api/blob/bfa868a224015e94456731c1b5b0c849f901b417/insights/v1alpha1/types_insights.go#L111

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -416,6 +427,549 @@ type MetricsServerConfig struct {
...
+	// +kubebuilder:validation:XValidation:rule="self.startsWith('/')",message="pathPrefix must start with '/'"
+	// +kubebuilder:validation:XValidation:rule="!self.endsWith('/') || self == '/'",message="pathPrefix must not end with '/' (except for root '/')"
+	// +kubebuilder:validation:XValidation:rule="!self.contains('?') && !self.contains('#')",message="pathPrefix must not contain '?' or '#'"
+	// +optional
+	PathPrefix string `json:"pathPrefix,omitempty"`
+	// scheme defines the URL scheme to use when communicating with Alertmanager
+	// instances.
+	// Possible values are `HTTP` or `HTTPS`.
+	// When omitted, defaults to `HTTP`.
+	// +kubebuilder:validation:Enum=HTTP;HTTPS
```
everettraven: If you use a type alias, we generally prefer that this marker is placed on the type alias.

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -416,6 +427,549 @@ type MetricsServerConfig struct {
...
+	RemoteTimeoutSeconds int32 `json:"remoteTimeoutSeconds,omitempty"`
+	// writeRelabelConfigs is a list of relabeling rules to apply before sending data to the remote endpoint.
+	// When omitted, no relabeling is performed and all metrics are sent as-is.
+	// Maximum of 10 relabeling rules can be specified.
+	// Each rule must have a unique name.
+	// +optional
+	// +kubebuilder:validation:MaxItems=10
+	// +listType=map
+	// +listMapKey=name
+	WriteRelabelConfigs []RelabelConfig `json:"writeRelabelConfigs,omitempty"`
```
everettraven, line 731: Minimum length of 1 on this if `[]` has the same semantic meaning as omitted please.

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -416,6 +427,549 @@ type MetricsServerConfig struct {
...
+	// regex is the regular expression to match against the concatenated source label values.
+	// Must be a valid RE2 regular expression (https://github.com/google/re2/wiki/Syntax).
+	// When omitted, defaults to "(.*)" (matches everything).
+	// Must be between 1 and 1000 characters in length when specified.
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=1000
+	Regex string `json:"regex,omitempty"`
+	// targetLabel is the target label name where the result is written.
+	// Required for Replace and HashMod actions (enforced by validation).
```
everettraven: We don't need to mention that this is enforced by validation
```suggestion
	// Required for Replace and HashMod actions.
```

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -416,6 +427,549 @@ type MetricsServerConfig struct {
...
+	// Must be between 1 and 253 characters in length.
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:XValidation:rule="self.matches('^[a-zA-Z0-9._-]+$')",message="must contain only alphanumeric characters, '-', '_', or '.'"
+	Key string `json:"key,omitempty"`
+}
+
+// VolumeClaimConfig defines the configuration for a PersistentVolumeClaim used for storage.
+// +kubebuilder:validation:MinProperties=1
```
everettraven: Not necessary because you already have one field marked as required.

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -416,6 +427,549 @@ type MetricsServerConfig struct {
...
+}
+
+// Retention configures how long Prometheus retains metrics data and how much storage it can use.
+// +kubebuilder:validation:MinProperties=1
+type Retention struct {
+	// durationInDays specifies how many days Prometheus will retain metrics data.
+	// Prometheus automatically deletes data older than this duration.
+	// When omitted, the default is 15 days.
+	// Minimum value is 1 day.
+	// Maximum value is 365 days (1 year).
```
everettraven, line 913: Any use cases to retain metrics for longer than a year?

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -244,14 +257,12 @@ type AlertmanagerCustomConfig struct {
...
-	// configure the persistent volume claim, including storage class, volume
-	// size, and name.
+	// volumeClaimTemplate defines persistent storage for Alertmanager. Use this setting to
+	// configure the persistent volume claim, including storage class and volume size.
 	// If omitted, the Pod uses ephemeral storage and alert data will not persist
 	// across restarts.
-	// This field is optional.
 	// +optional
-	VolumeClaimTemplate *v1.PersistentVolumeClaim `json:"volumeClaimTemplate,omitempty"`
+	VolumeClaimTemplate VolumeClaimConfig `json:"volumeClaimTemplate,omitempty,omitzero"`
```
everettraven: Be cautious of changing this.

If this changes the serialization of the type you'll need to upversion this to v1alpha2.

I had forgotten that `*v1.PersistentVolumeClaim` had been used in the past so it may be OK to stay consistent. Before promoting to a v1 though I'd recommend converting to a custom type so you can enforce better validations at admission time.

everettraven (commented):
Overall, the structure of this seems to be shaping up nicely.

I've got a handful more questions/comments on the API changes as-is. Once those are addressed I can do another comprehensive review to look at the tests and things like that.

Thanks @danielmellado for opening this new PR - GitHub is much happier loading this one now :P

everettraven:
> The verify-crd-schema and verify-crdify jobs are failing because we intentionally changed the volumeClaimTemplate field from the full corev1.PersistentVolumeClaim type to a simpler VolumeClaimConfig struct with just sizeInGiB and storageClassName.
> 
> This was done based on reviewer feedback to provide better admission-time validation and avoid exposing ~40 PVC fields that don't make sense for this use case.
> 
> Since this is a TechPreview API (v1alpha1), breaking changes are allowed, and both tools explicitly say it's appropriate to override when you're making intentional changes rather than introducing regressions.
> 
> @everettraven do you have any another proposal for this? Besides that, my plan is to split this into smaller changes to make it easier to work/review if that's ok for you.



I left a comment on this but I'll explicitly reply here as well - if if changes the serialization of the type it is _not_ OK to do in v1alpha1 if this has shipped at all. Older clients that relied on that field being the previous type would fail to serialize a valid object into a request body for the updated version of the resource.

The correct way to make this change is to up-version from v1alpha1 to v1alpha2.

Re: splitting into smaller changes, I'll leave that up to you. As-is this PR is pretty big but I think fairly manageable now that we've gotten a large majority of the structural changes for this API sorted out. The previous PR was just getting unwieldy due to the large number of review comments on it.

config/v1alpha1/types_cluster_monitoring.go
marioferh: think practical values would be:
a minimum of 10 KB,
and anything over 1 GB doesn’t make sense.
But I’m not sure about futurecases,  special or corner cases.

config/v1alpha1/types_cluster_monitoring.go
marioferh: this is the same here:
https://github.com/openshift/api/blob/e20773b543649f62501fce0aeab829c13995831e/config/v1alpha1/types_cluster_monitoring.go#L254

and discussed before

config/v1alpha1/types_cluster_monitoring.go
marioferh: done, thx

config/v1alpha1/types_cluster_monitoring.go

config/v1alpha1/types_cluster_monitoring.go
marioferh: then should I remove minimum length 1 of every [] field?

config/v1alpha1/types_cluster_monitoring.go

config/v1alpha1/types_cluster_monitoring.go

config/v1alpha1/types_cluster_monitoring.go
marioferh: it could be, but prometheus is not done for long term retention.

marioferh (commented):
Some of the addressed comments are out of the scope of this PR prometheusConfig. If you agree I can do a new one to fix that.

config/v1alpha1/types_cluster_monitoring.go
everettraven: Yes, I had forgotten about that when I had originally suggested moving to a custom type. I do think it is still better to move to a custom type over time so you can enforce admission time validations that are not present on the `*v1.PersistentVolumeClaim` type but that is out of the scope of this PR.

I'd agree with reverting this back to `*v1.PersistentVolumeClaim` for the v1alpha1 version of the API.

config/v1alpha1/types_cluster_monitoring.go
everettraven: In general, we find it relatively safe to increase maximums and decrease minimums if absolutely necessary - especially while this API is still a v1alpha1 and only present in TechPreview clusters.

I'd start with a range that is practical + some reasonable buffer for potential unknown cases. If 10kB - 1GB is practical let's go with that.

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -416,6 +427,547 @@ type MetricsServerConfig struct {
...
+	// also not supported.
+	// By default, PromQL queries are not logged.
+	// Must be an absolute path starting with `/` or a simple filename without path separators.
+	// Must be between 1 and 255 characters in length.
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=255
+	// +kubebuilder:validation:XValidation:rule="self.startsWith('/') || !self.contains('/')",message="must be an absolute path starting with '/' or a simple filename without '/'"
+	// +kubebuilder:validation:XValidation:rule="!self.startsWith('/dev/') || self in ['/dev/stdout', '/dev/stderr', '/dev/null']",message="only /dev/stdout, /dev/stderr, and /dev/null are allowed as /dev/ paths"
+	QueryLogFile string `json:"queryLogFile,omitempty"`
```
everettraven, line 520: I don't recall if this has already been discussed, but would it make sense to add some lightweight set of valid filename characters via a regular expression and that it follows the correct format?

For example, I think that with the current rules something like `/////////////` would be considered a valid input but doesn't sound like it would actually be valid?

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -416,6 +427,547 @@ type MetricsServerConfig struct {
...
+	//    - name: cpu
+	//      request: 4m
+	//    - name: memory
+	//      request: 40Mi
+	// +optional
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=10
+	// +kubebuilder:validation:MinItems=1
+	Resources []ContainerResource `json:"resources,omitempty"`
```
everettraven, line 548: Would it make sense to have minimums enforced for specific resources to ensure that the container has enough resources to minimally function?

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -416,6 +427,547 @@ type MetricsServerConfig struct {
...
+	// staticConfigs is a list of statically configured Alertmanager endpoints in the form
+	// of `<host>:<port>`. Each entry must be a valid hostname, IPv4 address, or IPv6 address
+	// (in brackets) followed by a colon and a valid port number (1-65535).
+	// Examples: "alertmanager.example.com:9093", "192.168.1.100:9093", "[::1]:9093"
+	// At least one endpoint must be specified (minimum 1, maximum 10 endpoints).
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=10
+	// +kubebuilder:validation:items:MaxLength=255
+	// +kubebuilder:validation:items:XValidation:rule="isURL('http://' + self) && url('http://' + self).getHostname() != '' && url('http://' + self).getPort() != '' && int(url('http://' + self).getPort()) >= 1 && int(url('http://' + self).getPort()) <= 65535",message="must be a valid 'host:port' where host is a DNS name, IPv4, or IPv6 address (in brackets), and port is 1-65535"
+	// +listType=set
```
everettraven, line 661: Specify that each entry must be unique in the GoDoc please.

config/v1alpha1/types_cluster_monitoring.go
everettraven: No, I'm asking to add a minimum length of 1 to this field if `writeRelabelConfigs: []` is the same as omitting the field (which it reads that way based on the GoDoc).

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -416,6 +427,547 @@ type MetricsServerConfig struct {
...
+	// and the resulting string is then matched against the regular expression for
+	// the replace, keep, or drop actions.
+	// If a referenced label does not exist on a series, Prometheus substitutes an empty string.
+	// When omitted, the rule operates without extracting source labels (useful for actions like labelmap).
+	// Maximum of 10 source labels can be specified, each between 1 and 128 characters.
+	// +optional
+	// +kubebuilder:validation:MaxItems=10
+	// +kubebuilder:validation:items:MinLength=1
+	// +kubebuilder:validation:items:MaxLength=128
+	// +listType=set
```
everettraven, line 760: Specify uniqueness constraint in the GoDoc please.

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -416,6 +427,547 @@ type MetricsServerConfig struct {
...
+	SourceLabels []string `json:"sourceLabels,omitempty"`
```
everettraven, line 761: Add minimum length of 1 if `sourceLabels: []` is the same as omitting the field altogether.

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -416,6 +427,547 @@ type MetricsServerConfig struct {
...
+	// Must be a valid RE2 regular expression (https://github.com/google/re2/wiki/Syntax).
+	// When omitted, defaults to "(.*)" (matches everything).
+	// Must be between 1 and 1000 characters in length when specified.
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=1000
+	Regex string `json:"regex,omitempty"`
+	// targetLabel is the target label name where the result is written.
+	// Required for Replace and HashMod actions.
+	// When omitted for other actions (Keep, Drop, LabelMap, LabelDrop, LabelKeep), no target label is set.
```
everettraven: What happens if I specify this field for these other actions?

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -416,6 +427,547 @@ type MetricsServerConfig struct {
...
+	// +kubebuilder:validation:MaxLength=128
+	TargetLabel string `json:"targetLabel,omitempty"`
+	// replacement is the value against which a regex replace is performed if the
+	// regular expression matches. Regex capture groups are available (e.g., $1, $2).
+	// When omitted, defaults to "$1" (the first capture group).
+	// Setting to an empty string ("") explicitly clears the target label value.
+	// Must be at most 255 characters in length.
+	// +optional
+	// +kubebuilder:validation:MaxLength=255
+	Replacement *string `json:"replacement,omitempty"`
```
everettraven, line 795: Does this only take effect when `action: Replace` is set?

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -416,6 +427,547 @@ type MetricsServerConfig struct {
...
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:XValidation:rule="!format.dns1123Subdomain().validate(self).hasValue()",message="must be a valid DNS subdomain (lowercase alphanumeric characters, '-' or '.', start and end with alphanumeric)"
+	ServerName string `json:"serverName,omitempty"`
+	// certificateVerification determines the policy for TLS certificate verification.
+	// Allowed values are "Verify" (performs certificate verification, secure) and "SkipVerify" (skips verification, insecure).
+	// When omitted, defaults to "Verify" (secure certificate verification is performed).
+	// +optional
+	// +kubebuilder:default=Verify
```
everettraven: For configuration APIs, we try to avoid the use of this marker because this becomes a contractual default you can never change without it being a breaking change.

Instead, we try to have a controller responsible for recognizing the omission of this field and performing some default behavior.

We normally document that with the terminology:
> 	When omitted, this means no opinion and the platform is left to choose a reasonable default, which is subject to change over time.
The default value is ...

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -416,6 +427,547 @@ type MetricsServerConfig struct {
...
+	// Maximum of 10 source labels can be specified, each between 1 and 128 characters.
+	// +optional
+	// +kubebuilder:validation:MaxItems=10
+	// +kubebuilder:validation:items:MinLength=1
+	// +kubebuilder:validation:items:MaxLength=128
+	// +listType=set
+	SourceLabels []string `json:"sourceLabels,omitempty"`
+	// separator is the character sequence used to join source label values.
+	// Common examples: ";" (default), ",", "::", "|||".
+	// When omitted, defaults to ";" (semicolon).
```
everettraven: We should probably use our standard terminology here:
> 	When omitted, this means no opinion and the platform is left to choose a reasonable default, which is subject to change over time.
The default value is `;`.

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -416,6 +427,547 @@ type MetricsServerConfig struct {
...
+	// Must be between 1 and 5 characters in length when specified.
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=5
+	Separator string `json:"separator,omitempty"`
+	// regex is the regular expression to match against the concatenated source label values.
+	// Must be a valid RE2 regular expression (https://github.com/google/re2/wiki/Syntax).
+	// When omitted, defaults to "(.*)" (matches everything).
```
everettraven: We should probably use our standard terminology:
>When omitted, this means no opinion and the platform is left to choose a reasonable default, which is subject to change over time.
The default value is `(.*)` to match everything.

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -416,6 +427,547 @@ type MetricsServerConfig struct {
...
+	// Required for Replace and HashMod actions.
+	// When omitted for other actions (Keep, Drop, LabelMap, LabelDrop, LabelKeep), no target label is set.
+	// Must be between 1 and 128 characters in length when specified.
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=128
+	TargetLabel string `json:"targetLabel,omitempty"`
+	// replacement is the value against which a regex replace is performed if the
+	// regular expression matches. Regex capture groups are available (e.g., $1, $2).
+	// When omitted, defaults to "$1" (the first capture group).
```
everettraven: We should stay consistent with our standard terminology:
>When omitted, this means no opinion and the platform is left to choose a reasonable default, which is subject to change over time.
The default value is `$1` (the first capture group).

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -416,6 +427,547 @@ type MetricsServerConfig struct {
...
+	// +required
+	SizeInGiB int32 `json:"sizeInGiB,omitempty"`
+}
+
+// Retention configures how long Prometheus retains metrics data and how much storage it can use.
+// +kubebuilder:validation:MinProperties=1
+type Retention struct {
+	// durationInDays specifies how many days Prometheus will retain metrics data.
+	// Prometheus automatically deletes data older than this duration.
+	// When omitted, the default is 15 days.
```
everettraven: Standard terminology for consistency:
>When omitted, this means no opinion and the platform is left to choose a reasonable default, which is subject to change over time.
The default value is `15`.

config/v1alpha1/types_cluster_monitoring.go

config/v1alpha1/types_cluster_monitoring.go
marioferh: done, also test

config/v1alpha1/types_cluster_monitoring.go
marioferh: I don't think so, it depends on each use case. If the user is unsure about the minimum or maximum, they can always not specify it.

config/v1alpha1/types_cluster_monitoring.go
marioferh: done

config/v1alpha1/types_cluster_monitoring.go
marioferh: done

config/v1alpha1/types_cluster_monitoring.go
marioferh: thanks, done.

config/v1alpha1/types_cluster_monitoring.go
marioferh: done, thanks

config/v1alpha1/types_cluster_monitoring.go
marioferh: fixed comment and added validation

config/v1alpha1/types_cluster_monitoring.go
marioferh: thanks, done

config/v1alpha1/types_cluster_monitoring.go
marioferh: yes

config/v1alpha1/types_cluster_monitoring.go
marioferh: right, thanks

config/v1alpha1/types_cluster_monitoring.go

config/v1alpha1/types_cluster_monitoring.go

marioferh:
verify-crd-schema errors seems not related with prometheus APO

everettraven:
@marioferh @danielmellado Are we replacing this PR with #2653 ?

I want to make sure I'm continuing to review the correct PR.

If so, please close this one.

danielmellado:
Hi Bryce, I asked Mario to put it there as I couldn't even open the old one
now that I'm back from PTO and working back on it. I assume that it got too
overloaded as it happened in the past... my .02 is due to the verbosity of
coderabbitt. Thanks!

PR #2637, opened 2026-01-02
OCPSTRAT-2876: Enable major version segmentation of enabled feature gates

JoelSpeed:
This PR works in tandem with https://github.com/openshift/cluster-version-operator/pull/1282 and an as yet unbuilt config operator PR to allow us to generate different versions of feature gates based on the target major version of the CVO payload.

This will allow us to have manifests that we deploy only in 4, or only in 5, or in a combination of releases. Importantly, it means that we can start setting up multiple major versions in development from the same branches, and gate CVO manifests, and feature gates, based on which major version is built into the build stream.


features/util.go
```diff
@@ -45,15 +46,72 @@ var (
...
 
+	status []featureGateStatus
+
 	statusByClusterProfileByFeatureSet map[ClusterProfileName]map[configv1.FeatureSet]bool
 }
+type featureGateStatus struct {
+	version        sets.Set[uint64]
+	clusterProfile sets.Set[ClusterProfileName]
+	featureSets    sets.Set[configv1.FeatureSet]
+}
```
everettraven, line 133: Don't we have a case for different feature set enablement based on the cluster profile?

AFAICT, this change removes the nuance of being able to say that a feature is enabled in different feature sets per-cluster-profile. Instead, it makes all cluster profiles all-or-nothing in relation to the feature sets.

Is this actually something we want to change?

everettraven: Also, do we ever envision needing to have per-version configurations?

I.e would it ever make sense for me to say - in OpenShift 5 I want FeatureXYZ to be enabled in DPNU and TPNU, but in OpenShift 6 I want FeatureXYZ to be enabled in DPNU, TPNU, and Default feature sets?

features/features.go
```diff
@@ -1,46 +1,76 @@
...
-	return featureGates.DeepCopy(), nil
+
+	return enabledDisabled
 }
 
-func AllFeatureSets() map[ClusterProfileName]map[configv1.FeatureSet]*FeatureGateEnabledDisabled {
-	ret := map[ClusterProfileName]map[configv1.FeatureSet]*FeatureGateEnabledDisabled{}
+func AllFeatureSets() map[uint64]map[ClusterProfileName]map[configv1.FeatureSet]*FeatureGateEnabledDisabled {
+	// Generating this many versions future proofs us until at least 2040.
+	versions := sets.New[uint64](4, 5, 6, 7, 8, 9, 10)
```
everettraven: Do we actually want per version configurations like this?

What if instead we did version ranged configurations like `>=5` to represent that a feature is enabled in all OpenShift versions greater than or equal to major version 5?

I know feature gates aren't meant to stick around long term so we _realistically_ shouldn't have feature gates that stick around a while, but something like a semver version range seems like it would future proof us indefinitely?

features/features.go
```diff
@@ -294,215 +324,215 @@ var (
...
+				enable(inDevPreviewNoUpgrade()).
 				mustRegister()
 
 	FeatureGateNewOLM = newFeatureGate("NewOLM").
 				reportProblemsToJiraComponent("olm").
 				contactPerson("joe").
 				productScope(ocpSpecific).
 				enhancementPR(legacyFeatureGateWithoutEnhancement).
-				enableForClusterProfile(SelfManaged, configv1.DevPreviewNoUpgrade, configv1.TechPreviewNoUpgrade, configv1.Default, configv1.OKD).
+				enable(inClusterProfile(SelfManaged), inDefault(), inOKD(), inTechPreviewNoUpgrade(), inDevPreviewNoUpgrade()).
```
everettraven, line 462: I don't think these semantically mean the same thing anymore?

everettraven (commented):
A handful of high-level comments I had from a brief look at this PR.

Will wait for responses related to these comments before digging further into the PR.

features/util.go
JoelSpeed: > Don't we have a case for different feature set enablement based on the cluster profile?
>
> AFAICT, this change removes the nuance of being able to say that a feature is enabled in different feature sets per-cluster-profile. Instead, it makes all cluster profiles all-or-nothing in relation to the feature sets.
>
> Is this actually something we want to change?

This behaviour is preserved. With the featuregatestatus object we collect a list of individual enablements, so if you enable only for one cluster profile in a call to `enable`, then when we later come to evaluate, you can work out that we have different enablement for HCP vs OCP. Note that the generated output files have not changed, and we already have some gates that are enabled for different levels across HCP vs OCP

> Also, do we ever envision needing to have per-version configurations?

Yes. We are planning to build overlapping versions of OCP from the same branches in the future. E.g. there will be one version in the 4 series and one in the 5 series that share the same branch. In that case, we want to be able to handle some features being enabled in 5 differently than in 4. We also want to be able to build a "6" in the future from 5 branches, where we would be able to disabled features that are set for removal. With this change we can, early, create a feature that is disabled in 6 and mimic what that cluster would look like by creating a 6 build from a 5 branch

features/features.go
JoelSpeed: Are you sure, the generated output is still the same?

features/features.go
JoelSpeed: I can explore this, but I expect it would be YAGNI. Feature gates that have differing behaviours between versions will not be the norm, and should always be relatively short lived. I wouldn't expect them to survive the 3 years between major version cycles

For this to work, we also need CVO to support the same semantic, so I'll pose that question on the CVO PR too.

Adding logic here does complicate things though, as we could then end up with competing clauses inadvertently. Enabling in `<=4` and then `>=5` would negate each others effect right?

features/util.go
everettraven: >This behaviour is preserved. With the featuregatestatus object we collect a list of individual enablements, so if you enable only for one cluster profile in a call to enable, then when we later come to evaluate, you can work out that we have different enablement for HCP vs OCP. Note that the generated output files have not changed, and we already have some gates that are enabled for different levels across HCP vs OCP

Ah, yes - you are right. For some reason, I missed that this was an entry to a slice....

I'm going to blame recharge brain still coming up to speed on that one 😅

features/features.go
everettraven: You're right - I overlooked that the feature gate statuses were a slice.

features/features.go
everettraven: >Enabling in <=4 and then >=5 would negate each others effect right?

Maybe? I think it would depend on how you handle version ranges and I would expect that PR reviews would be used to identify and remedy contradiction scenarios. I think that that `<=4` and `>=5` could be valid if they have different feature set enablements.

In my head, I'm envisioning that semver-range-enablement would be a first-satisfied approach. So the first enablement criteria that successfully matches would be used - if none, it isn't enabled at all.

features/util.go
```diff
@@ -45,15 +46,72 @@ var (
...
 type featureGateBuilder struct {
 	name                string
 	owningJiraComponent string
 	responsiblePerson   string
 	owningProduct       OwningProduct
 	enhancementPRURL    string
 
+	status []featureGateStatus
+
 	statusByClusterProfileByFeatureSet map[ClusterProfileName]map[configv1.FeatureSet]bool
```
everettraven: Would we still need this field or does it become effectively unused?

JoelSpeed: Ahh yep, that's unused, will cleanup

features/features.go
```diff
@@ -1,46 +1,76 @@
...
+			}
+		}
+
+		if enabled {
+			enabledDisabled.Enabled = append(enabledDisabled.Enabled, FeatureGateDescription{
+				FeatureGateAttributes: configv1.FeatureGateAttributes{
+					Name: name,
+				},
+			})
+		} else {
```
everettraven, line 31: Non-blocking nit: `continue` in above `if enabled` block instead of an `else` follows standard early return control flows.

features/util.go
```diff
@@ -45,15 +46,103 @@ var (
...
+
+func withFeatureSet(featureSet configv1.FeatureSet) featureGateEnableOption {
+	return func(s *featureGateStatus) {
+		s.featureSets.Insert(featureSet)
+	}
+}
+
+func inDefault() featureGateEnableOption {
+	return withFeatureSet(configv1.Default)
+}
```
everettraven, line 102: The OKD test requires that all Default enabled feature gates are also in the OKD feature set - would it make sense to make `inDefault()` automatically apply both?

I'm not strongly opinionated here, but it would save some verbosity in the feature gate definitions and needing to be aware of that nuance when promoting gates.

features/util.go
everettraven: Looks like this still needs cleanup?

machineconfiguration/v1/zz_generated.crd-manifests/0000_80_machine-config_01_machineconfignodes-Hypershift-Default.crd.yaml
everettraven, line 1: Is this file removal expected? I didn't see anywhere that it was obvious that this change should have been generated, but I might have missed it. 

Maybe this is just a stale generation that needed to be cleaned up to combine the standalone and HCP default set PRs?

everettraven:
Overall, I think this looks pretty good.

features/util.go
JoelSpeed: I did wonder about this as well. I had a theory that in the future we may want to have gates where we don't then require them in OKD and Default. Primarily I'm thinking about the case where we would use gates to remove features, so they'd be disabled in DPNU/TPNU/OKD but enabled in Default clusters

For now I think we can cope with the verbosity

machineconfiguration/v1/zz_generated.crd-manifests/0000_80_machine-config_01_machineconfignodes-Hypershift-Default.crd.yaml
JoelSpeed: This is expected. Part of the rewrite of the logic is a slightly different merging algorithm. So now if we have the same content across standalone and hypershift at the same featurset, they get merged together.

So you'll see this is removed, but the `Default` featureset file gained an annotation to show that it deploys on HCP as well as OCP

everettraven (approved):
A handful of the coderabbit comments seem like reasonable things to address, but I also don't think they are necessarily blocking in my eyes.

I think that the current state shows that it is working as intended, and if we encounter edge cases and bugs we can follow up in the future.

I'll tag, but add a hold in case you want to comb through the coderabbit suggestions and take any action on them.

/lgtm
/hold

tests/crd_filter.go
```diff
@@ -51,8 +52,15 @@ func perTestRuntimeInfo(suitePath, crdName string, featureGates []string) (*PerT
...
 		}
 
-		featureSet := currCRD.Annotations["release.openshift.io/feature-set"]
-		if featureSet == "CustomNoUpgrade" {
+		featureSetAnnotation := currCRD.Annotations["release.openshift.io/feature-set"]
+		featureSets := sets.New[string]()
+		if featureSetAnnotation != "" {
+			for _, featureSet := range strings.Split(featureSetAnnotation, ",") {
+				featureSets.Insert(featureSet)
+			}
```
everettraven, line 60: Non-blocking nit: this explicit loop isn't necessary. Could do something like:

```suggestion
		featureSets.Insert(strings.Split(featureSetAnnotation, ",")...)
```

tests/crd_filter.go
```diff
@@ -138,6 +147,27 @@ func clusterProfilesFrom(annotations map[string]string) sets.String {
 	return ret
 }
 
+func versionsFrom(annotations map[string]string) sets.Set[uint64] {
+	var versionString string
+	for k, v := range annotations {
+		if strings.HasPrefix(k, "release.openshift.io/major-version") {
```
everettraven, line 153: Any particular reason we want to use `strings.HasPrefix` here instead of an explicit `==`? If we ever add a new annotation like `release.openshift.io/major-version-X` (where `X` is anything else), this would match that which seems like something we wouldn't want?

The likelihood of us doing that is probably low, but this logic could need to evolve over time and using an explicit equivalence check seems less brittle in this case.

tools/codegen/pkg/manifestmerge/filters.go
```diff
@@ -48,13 +52,273 @@ func AllKnownFeatureSets(payloadFeatureGatePath string) (sets.String, error) {
...
+}
+
+// Contains checks if a version is within this range
+func (vr VersionRange) Contains(version uint64) bool {
+	for _, v := range vr {
+		if v == version {
+			return true
+		}
+	}
+	return false
```
everettraven, line 103: This works fine as-is, so not blocking but https://pkg.go.dev/slices#Contains is something that would probably work here as a straightforward:
```go
return slices.Contains(vr, version)
```

Not entirely sure, but might be able to replace this entire method with using that instead.

tools/codegen/pkg/manifestmerge/filters.go
```diff
@@ -48,13 +52,273 @@ func AllKnownFeatureSets(payloadFeatureGatePath string) (sets.String, error) {
...
+	featureGateMap := map[string]interface{}{}
+	if err := kyaml.Unmarshal(fileBytes, &featureGateMap); err != nil {
+		return nil, fmt.Errorf("unable to parse featuregate %s: %w", filename, err)
+	}
+
+	uncastFeatureGate := unstructured.Unstructured{
+		Object: featureGateMap,
+	}
+
+	// Extract cluster profile from annotations
```
everettraven, line 145: Seems like a stale comment?

tools/codegen/pkg/manifestmerge/filters.go
```diff
@@ -48,13 +52,273 @@ func AllKnownFeatureSets(payloadFeatureGatePath string) (sets.String, error) {
...
+		if annotationValue == "false-except-for-the-config-operator" {
+			// Convert long cluster profile names to short names
+			switch annotationKey {
+			case "include.release.openshift.io/ibm-cloud-managed":
+				clusterProfile = "Hypershift"
+			case "include.release.openshift.io/self-managed-high-availability":
+				clusterProfile = "SelfManagedHA"
+			default:
+				clusterProfile = annotationKey // fallback to the key itself
+			}
```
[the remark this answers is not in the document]
everettraven: I don't disagree that a centralized definition for the translations here would be nice, but I wouldn't block on this.

JoelSpeed:
/retest

/verified by CI

This is all updating generated content so CI should be sufficient to verify this

PR #2642, opened 2026-01-08
OCPCLOUD-2998: Add synchronizedAPI to Machine and MachineSet status

RadekManak:
## Summary

  This PR adds a new `synchronizedAPI` status field to Machine and MachineSet resources to support migration cancellation in the Machine API Migration feature.

  ## New Field

  - **`status.synchronizedAPI`**: Tracks the API that was last successfully synchronized before `status.authoritativeAPI` transitioned to `Migrating`. Valid values are `MachineAPI` and `ClusterAPI`.

  ## Purpose

  Currently, we determine the direction of migration by checking the value of `spec.authoritativeAPI`. This works for the base case, but it doesn't work when a migration becomes stuck in `status.authoritativeAPI: Migrating` and the user, to unblock the migration, changes `spec.authoritativeAPI` back to the original value.

  At that point, we need to reliably determine which direction we were trying to migrate when we got stuck. The `synchronizedAPI` field captures this information, allowing us to:

  1. Identify the source API of the migration (where we came from)
  2. Detect when the user requests a rollback (`spec.authoritativeAPI` matches `status.synchronizedAPI` while in `Migrating` state)
  3. Apply the correct rollback procedure to undo any partial migration steps (e.g., unpausing the correct resources)

  Without this field, once the user changes `spec.authoritativeAPI` back, we lose the information about the original migration direction and cannot safely restore the system to its previous state.

  ## Behavior

  1. The migration controller sets `status.synchronizedAPI` to the current `status.authoritativeAPI` value **before** transitioning to `Migrating`
  2. During migration, `synchronizedAPI` remains unchanged, preserving the source API
  3. On successful migration completion, `synchronizedAPI` is updated to reflect the new authoritative API
  4. If the user sets `spec.authoritativeAPI` back to match `status.synchronizedAPI` while in `Migrating` state, the migration controller detects this as a cancellation request, unpauses the appropriate resources, and returns to the previous state

  ## Validation

  - Once set, the field cannot be cleared (immutable non-empty rule on Machine CRD)
  - Only valid values: `MachineAPI`, `ClusterAPI`
  
  ## Implementation

  The controller implementation is available at: https://github.com/openshift/cluster-capi-operator/pull/376


machine/v1beta1/types_machine.go
```diff
@@ -402,10 +412,22 @@ type MachineStatus struct {
...
 	// +optional
 	AuthoritativeAPI MachineAuthority `json:"authoritativeAPI,omitempty"`
 
+	// synchronizedAPI represents the API that is currently in sync with the state of the resource.
+	// Valid values are "MachineAPI" and "ClusterAPI".
+	// When omitted (empty value), the resource has not yet been reconciled by the migration controller.
+	// The migration controller sets `status.synchronizedAPI` to the value of the `status.authoritativeAPI` before it transitions to "Migrating".
+	// It is used to determine the source API of the migration.
+	// +kubebuilder:validation:XValidation:rule="oldSelf == ''|| self != ''",message="The synchronizedAPI field must not be set empty once it has a value."
+	// +kubebuilder:validation:Enum=MachineAPI;ClusterAPI
```
JoelSpeed: Probably better to add this on the type itself

machine/v1beta1/types_machine.go
```diff
@@ -402,10 +412,22 @@ type MachineStatus struct {
 	// When set to Migrating, the migration controller is currently performing the handover of authority from one API to the other.
 	// +kubebuilder:validation:Enum=MachineAPI;ClusterAPI;Migrating
 	// +kubebuilder:validation:XValidation:rule="self == 'Migrating' || self == oldSelf || oldSelf == 'Migrating'",message="The authoritativeAPI field must not transition directly from MachineAPI to ClusterAPI or vice versa. It must transition through Migrating."
+	// +kubebuilder:validation:XValidation:rule="oldSelf == ''|| self != ''",message="The authoritativeAPI field must not be set empty once it has a value."
```
JoelSpeed: What's the motivation on this one? Empty isn't actually a valid value so it would be rejected anyway?

Were you trying to make sure it cannot be removed once set? If so, you need a CEL rule on the parent struct, not on the field itself

machine/v1beta1/types_machine.go
```diff
@@ -402,10 +412,22 @@ type MachineStatus struct {
...
 	// +openshift:enable:FeatureGate=MachineAPIMigration
 	// +optional
 	AuthoritativeAPI MachineAuthority `json:"authoritativeAPI,omitempty"`
 
+	// synchronizedAPI represents the API that is currently in sync with the state of the resource.
+	// Valid values are "MachineAPI" and "ClusterAPI".
+	// When omitted (empty value), the resource has not yet been reconciled by the migration controller.
```
JoelSpeed: Omitted doesn't actually mean an empty value, it means the key isn't present

machine/v1beta1/types_machine.go
```diff
@@ -402,10 +412,22 @@ type MachineStatus struct {
...
+	// The migration controller sets `status.synchronizedAPI` to the value of the `status.authoritativeAPI` before it transitions to "Migrating".
+	// It is used to determine the source API of the migration.
```
JoelSpeed: Source of the last successful synchronization wasn't it?

machine/v1beta1/types_machine.go
RadekManak: I wanted to make sure a hypothetical error in the controller does not remove the value by accident. Removed this as it is not required.

RadekManak:
Added the suggested changes and rewrote the description to be simpler and without implementation details.

machine/v1beta1/types_machine.go
JoelSpeed: Having a rule to prevent it from being removed makes sense, it was just in the wrong place, lets add one to the parent struct that prevents this child from being removed in an update

JoelSpeed:
/lgtm

This is fine as is, we may still want the removal protection thing we discussed but can follow up for that one

RadekManak:
/verified by @RadekManak
Tested the happy path with https://github.com/openshift/cluster-capi-operator/pull/376

PR #2644, opened 2026-01-12
CONSOLE-4986: Add Guided Tour capability flag to Console API and vendor it into console-operator

krishagarwal278:
### Summary

The OpenShift Console currently renders the Guided Tour feature unconditionally. This has caused flakiness in Console E2E (Cypress) tests running in CI, as the guided tour can interfere with UI interactions and timing.

To address this and to give cluster administrators explicit control, the Console API should expose a Guided Tour capability flag that allows the feature to be enabled or disabled at the cluster level.
 
### Motivation
Reduce Console CI flakiness caused by Guided Tour overlays
Allow cluster admins to opt in/out of Guided Tours
Provide a consistent capability-based mechanism aligned with other Console features
Enable test environments to explicitly disable Guided Tours

### Scope
Introduce a new GuidedTourFeature capability in the Console API
Ensure the capability is configurable and surfaced to the Console
Vendor the API changes into console-operator

### DIFF
--- a/operator/v1/types_console.go
+++ b/operator/v1/types_console.go
@@ -107,6 +107,9 @@ const (
 
 	// gettingStartedBanner is the name of the 'Getting started resources' banner in the console UI Overview page.
 	GettingStartedBanner ConsoleCapabilityName = "GettingStartedBanner"
+
+	// guidedTour is the name of the 'Guided Tour' feature in console UI.
+	GuidedTour ConsoleCapabilityName = "GuidedTour"
 )
 
 // CapabilityState defines the state of the capability in the console UI.
@@ -134,8 +137,8 @@ type CapabilityVisibility struct {
 // Capabilities contains set of UI capabilities and their state in the console UI.
 type Capability struct {
 	// name is the unique name of a capability.
-	// Available capabilities are LightspeedButton and GettingStartedBanner.
-	// +kubebuilder:validation:Enum:="LightspeedButton";"GettingStartedBanner"
+	// Available capabilities are LightspeedButton, GettingStartedBanner, and GuidedTour.
+	// +kubebuilder:validation:Enum:="LightspeedButton";"GettingStartedBanner";"GuidedTour"
 	// +required
 	Name ConsoleCapabilityName `json:"name"`
 	// visibility defines the visibility state of the capability.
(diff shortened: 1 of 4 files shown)


operator/v1/types_console.go
```diff
@@ -107,6 +107,9 @@ const (
 
 	// gettingStartedBanner is the name of the 'Getting started resources' banner in the console UI Overview page.
 	GettingStartedBanner ConsoleCapabilityName = "GettingStartedBanner"
+
+	// guidedTourFeature is the name of the 'Guided Tour' feature in console UI.
+	GuidedTourFeature ConsoleCapabilityName = "GuidedTourFeature"
```
JoelSpeed: Why Feature in the name here? CC @jhadvig

jhadvig: @JoelSpeed cause its not a single element but rather a walkthrough the console, in which we are highlighting UI elements. I understand what the nameig could be better but I wasnt really sure which name would be abstract enough to capture the nature of the feature, other then `Feature`. 
That said I can imagine we keep it as simple as `GuidedTour`

JoelSpeed: Yeah I was wondering why not to keep it simple as `GuidedTour`, unless we think there's going to be a clash with that name being that simple in the future?

jhadvig: +1 for shorten it to `GuidedTour` 
@krishagarwal278 please update the PR

krishagarwal278: Yes definitely @jhadvig @JoelSpeed!

operator/v1/types_console.go
jhadvig:
```suggestion
	// Available capabilities are LightspeedButton, GettingStartedBanner, and GuidedTour.
```

krishagarwal278: yikes 😬 
just added 👍🏻

JoelSpeed:
CodeRabbit has a point [here](https://github.com/openshift/api/pull/2644#discussion_r2685918858), once that's fixed I think this is good to go

JoelSpeed:
@krishagarwal278 Please squash this down to a single commit and then I'll tag

JoelSpeed:
/remove-label tid/merge-method-squash

We don't support squash merges on this repo (in fact you're not supposed to use it on any openshift repo), please squash manually

krishagarwal278:
QE verification:
/assign @yanpzhan

JoelSpeed:

This PR doesn't actually affect the payload so no need to gate on the ack

PR #2645, opened 2026-01-12
CNTRLPLANE-1752: Add PKI API to config.openshift.io/v1alpha1

sanchezl:
## Summary

This PR introduces the `PKI` API (`config.openshift.io/v1alpha1`) behind the `ConfigurablePKI` feature gate (TechPreview / DevPreview). It provides a cluster-scoped singleton resource that allows administrators to configure the cryptographic parameters (key algorithm, key size, elliptic curve) used for internally-managed certificates — including signer (CA), serving (TLS), and client certificates — with per-category overrides on top of a required set of defaults.

## API Structure

```mermaid
graph TD
    PKI["<b>PKI</b><br/><i>cluster-scoped singleton</i>"]
    PKISpec["<b>spec</b>"]
    CertMgmt["<b>certificateManagement</b><br/><i>union</i>"]
    Mode["<b>mode</b><br/><i>Unmanaged | Default | Custom</i>"]
    Custom["<b>custom</b><br/><i>required when mode=Custom</i>"]
    Defaults["<b>defaults</b> <i>(required)</i>"]
    Signers["<b>signerCertificates</b> <i>(optional)</i>"]
    Serving["<b>servingCertificates</b> <i>(optional)</i>"]
    Client["<b>clientCertificates</b> <i>(optional)</i>"]
    KeyConfig["<b>key</b><br/><i>union: algorithm</i>"]
    RSA["<b>rsa</b><br/>keySize: 2048–8192<br/><i>(multiples of 1024)</i>"]
    ECDSA["<b>ecdsa</b><br/>curve: P256 | P384 | P521"]

    PKI --> PKISpec --> CertMgmt
    CertMgmt --> Mode
    CertMgmt --> Custom
    Custom --> Defaults
    Custom --> Signers
    Custom --> Serving
    Custom --> Client
    Defaults --> KeyConfig
    Signers -.->|"optional override"| KeyConfig
    Serving -.->|"optional override"| KeyConfig
    Client -.->|"optional override"| KeyConfig
    KeyConfig --> RSA
    KeyConfig --> ECDSA
```

## Modes

- **Unmanaged** — The cluster does not manage certificate parameters; an external system is expected to handle PKI.
- **Default** — The cluster uses its built-in default cryptographic settings (current behaviour).
- **Custom** — The administrator provides explicit cryptographic parameters via `custom`. A `defaults` block is required; optional per-category overrides (`signerCertificates`, `servingCertificates`, `clientCertificates`) take precedence when present.

## Example (Custom mode with ECDSA defaults, RSA override for signers)

```yaml
kind: PKI
metadata:
  name: cluster
spec:
  certificateManagement:
    mode: Custom
    custom:
      defaults:
        key:
          algorithm: ECDSA
          ecdsa:
            curve: P384
      signerCertificates:
        key:
          algorithm: RSA
          rsa:
            keySize: 4096
```

## Links

- Enhancement: https://github.com/openshift/enhancements/pull/1882
- Jira: [CNTRLPLANE-1752](https://issues.redhat.com/browse/CNTRLPLANE-1752)

### DIFF
--- a/hack/update-payload-crds.sh
+++ b/hack/update-payload-crds.sh
@@ -28,6 +28,7 @@ crd_globs="\
     operator/v1/zz_generated.crd-manifests/0000_80_machine-config_01_machineconfigurations*.crd.yaml
     config/v1alpha1/zz_generated.crd-manifests/0000_10_config-operator_01_clustermonitoring*.crd.yaml
     config/v1alpha1/zz_generated.crd-manifests/0000_10_config-operator_01_criocredentialproviderconfigs*.crd.yaml
+    config/v1alpha1/zz_generated.crd-manifests/0000_10_config-operator_01_pkis*.crd.yaml
     operator/v1/zz_generated.crd-manifests/*_storage_01_storages*.crd.yaml
     operator/v1/zz_generated.crd-manifests/*_csi-driver_01_clustercsidrivers*.crd.yaml
     insights/v1/zz_generated.crd-manifests/0000_10_insights_01_datagathers*.crd.yaml


config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,257 @@
...
+
+// PKIManagementPolicy determines whether components use hardcoded defaults (Unmanaged), follow
+// OpenShift best practices (Default), or use administrator-specified cryptographic parameters (Custom).
+// This provides flexibility for organizations with specific compliance requirements or security policies
+// while maintaining backwards compatibility for existing clusters.
+//
+// +kubebuilder:validation:XValidation:rule="self.type == 'Unmanaged' ? (!has(self.defaults) && !has(self.categories) && !has(self.overrides)) : true",message="defaults, categories, and overrides must not be set when type is Unmanaged"
+// +kubebuilder:validation:XValidation:rule="self.type == 'Default' ? (!has(self.defaults) && !has(self.categories) && !has(self.overrides)) : true",message="defaults, categories, and overrides must not be set when type is Default"
+// +union
+type PKIManagementPolicy struct {
```
sanchezl: I want to note that I am not using a discriminated union pattern here, instead it is more of a "parameterized" union pattern (at at least a kind of parameterized union where its all fields or none). 

This configuration models a "predefined behavior vs. customization" choice pattern, rather than mutually exclusive options with different configuration requirements (like platform-specific settings).

If we add more predefined behaviors (e.g., type: FIPS, type: CNSACompliant), they would be additional enum values with no configuration fields. They represent different default behaviors, not different configuration structures. 

If we need to extend what's configurable (e.g., adding certificate lifetime configuration), we'd add fields to PKIProfile and/or CertificateConfig. These fields would apply regardless of which predefined behavior or custom configuration is in use.

everettraven: I'm not sure I'm following why this is a better approach than a discriminated union pattern.

Discriminated unions are a common pattern in OpenShift APIs (so our users will be used to them) and they allow for discriminators with no associated configuration fields.

>If we add more predefined behaviors (e.g., type: FIPS, type: CNSACompliant), they would be additional enum values with no configuration fields. They represent different default behaviors, not different configuration structures.
>
>If we need to extend what's configurable (e.g., adding certificate lifetime configuration), we'd add fields to PKIProfile and/or CertificateConfig. These fields would apply regardless of which predefined behavior or custom configuration is in use.

It sounds like if we added any additional pre-defined behaviors any new configuration options we add would be not configurable by end-users with these pre-defined behaviors.

To me, this reads that the only time we anticipate ever being able to configure something is when `type` is set to `Custom` - in which case an end-user must supply some kind of configuration.

Is this accurate?

sanchezl: Right. This is a pick a predefined policy, or supply your own. There would be no such thing as a pre-defined policy with options. I think this is distinct from "there are different kinds of policies with different kinds of fields".

everettraven: > I think this is distinct from "there are different kinds of policies with different kinds of fields".

Is it though? I read this as there being different kinds of policies with different kinds of fields, just that most of the policies have no fields.

IMO, something like:
```yaml
type: Custom
custom:
  defaults: ...
  categories: ...
  overrides: ...
```

is more intuitive that the fields are only applicable when the type is `custom` than something like:
```yaml
type: Custom
defaults: ...
categories: ...
overrides: ...
```

I think becomes even more true when we've used this discriminated union approach for quite some time for OpenShift APIs and is one that users are going to be familiar with.

We try to aim for consistency, where possible, in how we structure our APIs to keep the experience consistent and familiar for our users - I'm not sold on breaking that consistency here yet.

everettraven: The currently proposed approach also lends itself to an ever growing list of CEL validations to maintain instead of having one to prevent setting custom values when the `type` field is not `Custom`.

config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,271 @@
...
+}
+
+// PKISpec holds the specification for PKI configuration.
+type PKISpec struct {
+	// certificateManagement specifies how PKI configuration is managed for internally-generated certificates.
+	// This controls the certificate generation approach for all OpenShift components that create
+	// certificates internally, including certificate authorities, serving certificates, and client certificates.
+	//
+	// +required
+	CertificateManagement PKICertificateManagement `json:"certificateManagement,omitzero"`
```
everettraven, line 39: Do we envision this API being expanded to do more than certificate management in the future?

config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,271 @@
...
+	// - Default: Use OpenShift-recommended best practices for certificate generation.
+	//   The specific parameters may evolve across OpenShift releases to adopt improved cryptographic standards.
+	//   In the initial release, this matches Unmanaged behavior for each component.
+	//   In future releases, this may adopt ECDSA or larger RSA keys based on industry best practices.
+	//   Recommended for most customers who want to benefit from security improvements automatically.
+	// + Default when installing a fresh cluster.
+	//
+	// - Custom: Administrator explicitly configures cryptographic parameters.
+	//   Use the custom field to specify certificate generation parameters.
+	// + Recommended for customers with specific compliance requirements or organizational PKI policies.
```
everettraven: Instead of using the list notation, we generally try to use formats like:
```go
// When set to Unmanaged, ....
//
// When set to Default, ...
//
// When set to Custom, ...
```

This helps us to avoid any weirdness in formatting of the generated documentation when using list notations.

config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,271 @@
...
+// +union
+type PKICertificateManagement struct {
+	// mode determines how PKI configuration is managed.
+	// Valid values are "Unmanaged", "Default", and "Custom".
+	//
+	// - Unmanaged: Components use their existing hardcoded certificate generation behavior, exactly as if this feature did not exist.
+	//   Each component generates certificates using whatever parameters it was using before this feature.
+	//   While most components use RSA 2048, some may use different parameters.
+	//   Use of this mode might prevent upgrading to the next major OpenShift release.
+	// + Default when upgrading from a version without this feature to ensure zero behavior change.
```
everettraven: Reading the description of `Unmanaged` I'm wondering if `Unmanaged` is the right terminology here.

I'm struggling a bit with naming suggestions, but something along the lines of `ComponentDefaults` seems more appropriate here since this is targeted at a mode where individual components of the platform are using their own defaults.

What do you think?

config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,271 @@
...
+	//
+	// - Default: Use OpenShift-recommended best practices for certificate generation.
+	//   The specific parameters may evolve across OpenShift releases to adopt improved cryptographic standards.
+	//   In the initial release, this matches Unmanaged behavior for each component.
+	//   In future releases, this may adopt ECDSA or larger RSA keys based on industry best practices.
+	//   Recommended for most customers who want to benefit from security improvements automatically.
+	// + Default when installing a fresh cluster.
```
everettraven: Following the same logic of there being a couple layers of "default" behavior, maybe naming this something like `PlatformDefault` makes more sense as applying the default broadly across all components in the "platform"?

config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,271 @@
...
+	// - Custom: Administrator explicitly configures cryptographic parameters.
+	//   Use the custom field to specify certificate generation parameters.
+	// + Recommended for customers with specific compliance requirements or organizational PKI policies.
+	//
+	// + When upgrading from a version without this feature:
+	// + - The PKI resource is created with mode: Unmanaged to ensure zero behavior change.
+	//
+	// + When installing a fresh cluster:
+	// + - If no PKI configuration is provided in install-config.yaml, mode: Default is used.
+	// + - If PKI configuration is provided in install-config.yaml, mode: Custom is used with the specified configuration.
```
everettraven: I don't think we need to include semantics about upgrading vs installing from a fresh cluster as part of the API documentation.

It reads to me as more suited towards developer documentation as opposed to end-user documentation. The GoDoc is used as our user-facing documentation for APIs.

config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,271 @@
...
+	//
+	// +required
+	// +kubebuilder:validation:Enum=Unmanaged;Default;Custom
```
everettraven: We generally recommend putting the enum marker on the type alias. It is effectively the same, but if the type alias is ever used in another API somewhere it will not automatically add the enum constraint unless the marker is on the type alias.

config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,271 @@
...
+// to create certificates. Configuration can be specified at three hierarchical levels:
+// defaults apply to all certificates, categories apply to certificate types (SignerCertificate,
+// ServingCertificate, ClientCertificate), and overrides apply to specific named certificates.
+// More specific levels take precedence over general ones.
+// +kubebuilder:validation:MinProperties=1
+type PKIProfile struct {
+	// defaults specifies the default certificate configuration
+	// for all certificates unless overridden by category or specific
+	// certificate configuration.
+	// If not specified, uses platform defaults (typically RSA 2048).
```
everettraven: Is this the actual platform defaults that may be subject to change over time, or are these the component-specific defaults?

I know they are currently one and the same, but IIUC they may not be in the future.

config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,271 @@
...
+	// curve specifies the elliptic curve for ECDSA keys.
+	// Valid values are "P256", "P384", and "P521".
+	// +required
+	// +kubebuilder:validation:Enum=P256;P384;P521
+	Curve ECDSACurve `json:"curve,omitempty"`
+}
+
+type CategoryCertificateConfig struct {
+	// category identifies the certificate category.
+	// Valid values are "SignerCertificate", "ServingCertificate", and "ClientCertificate".
```
everettraven: What does each of these values mean when set?

config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,271 @@
...
+	// +required
+	Certificate CertificateConfig `json:"certificate,omitzero"`
+}
+
+// CertificateOverride allows configuration of certificate parameters for specific named certificates.
+//
+// +kubebuilder:validation:XValidation:rule=`self.certificateName in ['admin-kubeconfig-signer','kubelet-bootstrap-kubeconfig-signer','kube-apiserver-localhost-signer','kube-apiserver-service-network-signer','kube-apiserver-lb-signer','root-ca','kube-apiserver-to-kubelet-signer','kube-control-plane-signer','aggregator-signer','kubelet-signer','aggregator-ca','etcd-signer','etcd-metrics-signer','service-ca','csr-signer-ca','kube-apiserver-localhost-server','kube-apiserver-service-network-server','kube-apiserver-lb-server','kube-apiserver-internal-lb-server','machine-config-server','ironic-server','etcd-peer-server','etcd-server','etcd-metrics-server','admin-kubeconfig-client','kubelet-client','kube-apiserver-to-kubelet-client','kube-control-plane-kube-controller-manager-client','kube-control-plane-kube-scheduler-client','aggregator-client','apiserver-proxy-client','journal-gatewayd-client']`,message="certificateName must be a well-known certificate name"
```
everettraven: If we have a known list of values that we want to enforce, we should use an enum instead of a CEL expression. CEL expressions should only be used when necessary, and the OpenAPI schema used for CRDs already has a way to enforce enum constraints.

config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,271 @@
...
+	ECDSA ECDSAKeyConfig `json:"ecdsa,omitzero"`
+}
+
+// RSAKeyConfig specifies parameters for RSA key generation.
+type RSAKeyConfig struct {
+	// keySize specifies the size of RSA keys in bits.
+	// Valid values are 2048, 3072, and 4096.
+	// +required
+	// +kubebuilder:validation:Enum=2048;3072;4096
+	// +kubebuilder:validation:Minimum=2048
```
everettraven, line 206: I don't think the minimum marker is required here when setting the enum constraint. I don't recall whether or not the linter picks this up for integer types though 🤔

config/v1alpha1/types_pki.go
everettraven: This comment applies to all enum constraint markers in this PR.

config/v1alpha1/types_pki.go
sanchezl: GoDoc prefixed with `// +` does not make it into the OpenAPI [documentation](https://github.com/openshift/api/pull/2645/changes/BASE..67fe390958a23a5900d53b5e46f83751248af67a#diff-6f26c8627c096ba26e30a864dbdb69e055980f8c8b704283d9a483fbb999404cR23912), but it does make it into the GoDoc. Is that acceptable, or should I still remove?

config/v1alpha1/types_pki.go
sanchezl: Short answer, No. 

Yet, I added the spec.certificateManagement as a nested field as a hedge against future PKI configuration that might be orthogonal to the management of the certificate generation, or maybe even just the existing structure of the API and might need a spec level addition here.

While I do not have a specific scenario, my worries were triggered by thinking about some possible future needs to:
1. key storage options
2. some overall compliance mode (such as FIPS, PQC, etc) that might constrain valid configurations.
3. maybe some CA bundle management options, where its not about generating certs, but how they are distributed/shared.

config/v1alpha1/types_pki.go
sanchezl: I think I was contemplating enum vs integer got stuck in between.
Updated the validation to be an integer multiple of 1024 with a max/min. Technically, it can be anything divisible by 8, but that seems like asking for trouble.

config/v1alpha1/types_pki.go
sanchezl: Updated GoDoc.

config/v1alpha1/types_pki.go
sanchezl: Updated.

config/v1alpha1/types_pki.go
sanchezl: I went ahead and removed it. I'll make sure it is in the enhancement.

config/v1alpha1/types_pki.go
sanchezl: Done.

config/v1alpha1/types_pki.go
sanchezl: I want to validate using a [parameterized ValidatingAdmissionPolicy](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/#parameter-resources). To that end I added the `PKICertificateDefinition` type and a ValidatingAdmissionPolicy. 

Please advise on where the [ValidatingAdmissionPolicy should reside](https://github.com/openshift/api/pull/2645/changes/c8a7ab3622a1cc40b5d209cb3c651213515975a2).

config/v1alpha1/types_pki.go
everettraven: >maybe some CA bundle management options, where its not about generating certs, but how they are distributed/shared.

This seems like the most plausible to me, especially with the [Pod Certificates](https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#pod-certificate-requests) and [Cluster Trust Bundles](https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#cluster-trust-bundles) features being worked on upstream.

Thanks for the clarification!

config/v1alpha1/types_pki.go
everettraven: This is a new one for me - I'm going to discuss this with other folks on the API review team to get some more thoughts on this.

A question - what happens if the VAP gets removed and an invalid value makes its way into the API?

config/v1alpha1/manifests/pki-certificate-override-validation.yaml
```diff
@@ -0,0 +1,80 @@
...
+  # Use PKICertificateDefinition resources as the parameter source
+  paramKind:
+    apiVersion: config.openshift.io/v1alpha1
+    kind: PKICertificateDefinition
+
+  # Validate each certificate override references a registered certificate name
+  validations:
+  # Skip validation if no overrides are present
+  - expression: "!has(object.spec.certificateManagement.custom.overrides) || size(object.spec.certificateManagement.custom.overrides) == 0"
+    reason: Skip
```
JoelSpeed: Should be under `matchConditions`
```
matchConditions:
- name: check-only-overrides
  expression: size(object.spec.?certificateManagement.custom.overrides.orValue('[]')) > 0
```

config/v1alpha1/manifests/pki-certificate-override-validation.yaml
```diff
@@ -0,0 +1,80 @@
...
+
+  # Build list of all valid certificate names from all PKICertificateDefinition resources
+  - expression: |
+      has(params) && has(params.spec) && has(params.spec.certificates) ?
+        params.spec.certificates.map(cert, cert.name) : []
```
JoelSpeed: What are you trying to achieve here? This appears to be trying to construct a list of names, did you want to create a variable? 

Validation expressions must map evaluate to a boolean

I think you want a variable to use later?

```
variables:
- name: certificates
  expression: params.spec.?certificates.orValue('[]')
```

I'm not convinced I understand why you want to map here so I've left that out

config/v1alpha1/manifests/pki-certificate-override-validation.yaml
```diff
@@ -0,0 +1,80 @@
...
+  - expression: |
+      !has(object.spec.certificateManagement.custom.overrides) ||
+      object.spec.certificateManagement.custom.overrides.all(override,
+        params.exists(p,
+          has(p.spec) && has(p.spec.certificates) &&
+          p.spec.certificates.exists(cert, cert.name == override.certificateName)
+        )
+      )
+    message: "certificateName in overrides must reference a certificate registered in a PKICertificateDefinition resource"
+    reason: Invalid
```
JoelSpeed: Ok, so lets make the overrides a variable too
```
variables:
- name: overrides
  expression: object.spec.?certificateManagement.custom.overrides.orValue('[]')
```

And then this expression can be
```
variables.overrides.all(override, variables.certificates.exists(certificate, certificate.name == override.certificateName)
```

config/v1alpha1/manifests/pki-certificate-override-validation.yaml
```diff
@@ -0,0 +1,80 @@
...
+
+  # Validate that certificate names follow DNS subdomain rules
+  - expression: |
+      !has(object.spec.certificateManagement.custom.overrides) ||
+      object.spec.certificateManagement.custom.overrides.all(override,
+        override.certificateName.matches('^[a-z0-9]([-a-z0-9]*[a-z0-9])?$')
+      )
+    message: "certificateName must be a valid DNS subdomain (lowercase alphanumeric with hyphens)"
+    reason: Invalid
```
JoelSpeed: This should be within the API itself, you don't need a VAP for this

config/v1alpha1/manifests/pki-certificate-override-validation.yaml
```diff
@@ -0,0 +1,80 @@
...
+    # This prevents blocking PKI resource creation before any components have registered
+    parameterNotFoundAction: Allow
+
+  # Match all PKI resources
+  matchResources:
+    namespaceSelector:
+      matchExpressions:
+      - key: kubernetes.io/metadata.name
+        operator: In
+        values: [""]  # Empty string matches cluster-scoped resources
```
JoelSpeed: Are you sure it's metadata.name and not metadata.namespace?

config/v1alpha1/manifests/pki-certificate-override-validation.yaml
everettraven, line 1: Taking a closer look at the validations you are trying to enforce here, I don't believe that a VAP is actually suitable for your use case.

I could be wrong but as far as I understand it, when using parameter references that reference multiple parameter resources the policy is evaluated one parameter resource at a time. I think this makes the validation you are trying to achieve (rejecting PKI resources with `.spec.certificateManagement.custom.overrides` that do _not_ have corresponding `PKICertificateDefinition` resource entries) not feasible because you won't have a list of the dependent resources to build your validation from.

If you truly want to pursue this approach of having validations that are dependent on the presence of values in another API you'll either need a validating admission webhook _or_ to update our KAS admission plugins to enforce this validation.

config/v1alpha1/types_pki.go
```diff
@@ -219,16 +219,14 @@ type CategoryCertificateConfig struct {
...
 type CertificateOverride struct {
 	// certificateName identifies a specific certificate to configure.
-	// The name must match a well-known certificate name in the cluster.
-	// Examples: "kube-apiserver-to-kubelet-signer", "kube-apiserver-localhost-server",
-	// "admin-kubeconfig-client", "etcd-signer", "service-ca"
+	// The name must match a certificate registered in a PKICertificateDefinition resource.
 	// +required
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
```
everettraven: We prefer the use of CEL validations (`+kubebuilder:validation:XValidation`) that enforces a regular expression instead of this pattern marker. Using CEL validations means we can return a more human-readable error message to our end users than returning `Invalid ... ^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`.

config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,275 @@
...
+	ECDSA ECDSAKeyConfig `json:"ecdsa,omitzero"`
+}
+
+// RSAKeyConfig specifies parameters for RSA key generation.
+type RSAKeyConfig struct {
+	// keySize specifies the size of RSA keys in bits.
+	// Valid values are multiples of 1024 from 2048 to 8192.
+	// +required
+	// +kubebuilder:validation:Minimum=2048
+	// +kubebuilder:validation:Maximum=8192
```
everettraven, line 207: What are the implications of specifying a 8192-bit key?

AFAIK, there are some significant performance/usage implications when using keys larger than 4096-bits.

What is the use case we are trying to solve for by allowing 8192-bit keys?

config/v1alpha1/types_pkicertificatedefinition.go
```diff
@@ -0,0 +1,110 @@
...
+
+	// status holds observed state
+	// +optional
+	Status PKICertificateDefinitionStatus `json:"status,omitempty"`
+}
+
+// PKICertificateDefinitionSpec defines certificates managed by a component.
+type PKICertificateDefinitionSpec struct {
+	// component identifies the operator or component managing these certificates.
+	// This should typically be the name of the operator (e.g., "etcd-operator", "kube-apiserver-operator").
```
everettraven: Let's add a validation that enforces correct formatting here.

config/v1alpha1/types_pkicertificatedefinition.go
```diff
@@ -0,0 +1,110 @@
...
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Component string `json:"component,omitempty"`
+
+	// certificates is a list of certificate definitions managed by this component.
+	// Each certificate must have a unique name within the cluster.
```
everettraven: Unique name within the cluster or a unique name for each component?

config/v1alpha1/types_pkicertificatedefinition.go
```diff
@@ -0,0 +1,110 @@
...
+	// +required
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=256
+	// +listType=map
+	// +listMapKey=name
+	Certificates []CertificateDefinition `json:"certificates,omitempty"`
+}
+
+// CertificateDefinition describes a single certificate managed by a component.
+// +kubebuilder:validation:XValidation:rule="self.name.matches('^[a-z0-9]([-a-z0-9]*[a-z0-9])?$')",message="name must be a valid DNS subdomain (lowercase alphanumeric with hyphens)"
```
everettraven: You can probably leverage the existing format library to enforce this: https://kubernetes.io/docs/reference/using-api/cel/#kubernetes-format-library.

This validation should also be on the field itself.

config/v1alpha1/types_pkicertificatedefinition.go
```diff
@@ -0,0 +1,110 @@
...
+type CertificateDefinition struct {
+	// name is the unique identifier for this certificate.
+	// This name is used in PKI.spec.overrides[].certificateName to configure this certificate.
+	// Must be a valid DNS subdomain (lowercase letters, numbers, and hyphens).
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
```
everettraven: Remove this in favor of the CEL validation you have above.

config/v1alpha1/types_pkicertificatedefinition.go
```diff
@@ -0,0 +1,110 @@
...
+	Name string `json:"name,omitempty"`
+
+	// category specifies the certificate category.
+	// This helps administrators understand the certificate's role and select appropriate
+	// cryptographic parameters.
+	// +required
+	Category CertificateCategory `json:"category,omitempty"`
```
everettraven: Re-explain any constraints and what they mean for this field in the GoDoc. This looks like it would be constrained by an enum.

config/v1alpha1/types_pkicertificatedefinition.go
```diff
@@ -0,0 +1,110 @@
...
+
+	// description provides a human-readable explanation of this certificate's purpose.
+	// Examples: "CA for etcd peer and server certificates", "Server certificate for API server localhost endpoint"
```
everettraven: Is there any validations we can add that helps to ensure the general shape of this string would be human readable? Maybe limiting to common characters and punctuation would be sufficient?

config/v1alpha1/types_pkicertificatedefinition.go
```diff
@@ -0,0 +1,110 @@
...
+	// conditions represent the latest available observations of the PKICertificateDefinition's state.
+	// +optional
+	// +listType=map
+	// +listMapKey=type
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// registeredAt is the timestamp when this definition was first successfully validated.
+	// +optional
+	RegisteredAt *metav1.Time `json:"registeredAt,omitempty"`
+}
```
everettraven: Does this API need a status?

IIUC, the purpose of this API is for components to advertise what certificates they create and manage for themselves - in which case having a spec-only API seems reasonable to me.

config/v1alpha1/manifests/pki-certificate-override-validation.yaml
sanchezl: I concur. I'm going to proceed without the custom overrides for now.

config/v1alpha1/types_pki.go
sanchezl: I based it on a [NIST spec ](https://csrc.nist.gov/news/2019/nist-publishes-sp-800-56b-revision-2)referred to me by some customers. It lists 8192 as the highest "reasonable" key size.

config/v1alpha1/types_pki.go
sanchezl: Updated

config/v1alpha1/types_pki.go
sanchezl: I prefer the term `Unmanaged` to emphasize that the PKI resource has no influence over the certificate parameters. I believe `ComponentDefaults` suggests some consideration for what those component defaults are, but they are currently merely historical values. Furthermore, the plan is to eliminate this option in version N+1, preventing an upgrade from a version N that is still `Unmanaged`.

config/v1alpha1/types_pki.go
sanchezl: We will make sure to test with various combinations of key sizes.

config/v1alpha1/types_pki.go
sanchezl: Noted.

config/v1alpha1/types_pki.go
sanchezl: This is where you would override the platform defaults.

```
spec:
  certificateManagement:
    mode: custom
    custom:
      defaults: // <---
         ....
```

everettraven: I understood that, my question here is more scoped to understand what the default behavior is if I do _not_ specify any defaults there.

It says "platform defaults", but IIUC there is two levels of indirection there for the initial state of the PR - `Unmanaged` meaning each component in the platform has its own default behavior and `PlatformDefaults` (or something like that) that means that a singular default behavior that we control for the entire platform and is configured on each component.

With the current GoDoc, it isn't quite clear to me which one is the actual default in this case.

config/v1alpha1/types_pki.go
everettraven: >We will make sure to test with various combinations of key sizes.

Can we make sure to add this specifically to the testing section of the EP and marking it as part of the graduation criteria?

config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,248 @@
...
+	// +kubebuilder:validation:MaxItems=3
+	Categories []CategoryCertificateConfig `json:"categories,omitempty"`
+}
+
+// CertificateConfig specifies configuration parameters for certificates.
+// +kubebuilder:validation:MinProperties=1
+type CertificateConfig struct {
+	// key specifies the cryptographic parameters for the certificate's key pair.
+	// +optional
+	Key KeyConfig `json:"key,omitempty,omitzero"`
```
everettraven: In a world where this could be omitted, what does it mean for this field to be omitted?

config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,248 @@
...
+
+// RSAKeyConfig specifies parameters for RSA key generation.
+type RSAKeyConfig struct {
+	// keySize specifies the size of RSA keys in bits.
+	// Valid values are multiples of 1024 from 2048 to 8192.
+	// +required
+	// +kubebuilder:validation:Minimum=2048
+	// +kubebuilder:validation:Maximum=8192
+	// +kubebuilder:validation:MultipleOf=1024
+	// +kubebuilder:default=2048
```
everettraven: Required with a default is an anti-pattern:

```suggestion
```

config/v1alpha1/types_pki.go
sanchezl: Ah, I understand now.

To address this issue, I made `defaults` a mandatory field on `PKIProfile`. This ensures that when the mode is set to `Custom`, the user must explicitly specify their certificate defaults. This prevents any implicit fallback to either `Unmanaged` or `Default` behavior.

Additionally, I renamed categories to `overrides` to make the precedence relationship with defaults more explicit.

Since `CertificateConfig.Key` is optional, I also added a CEL validation on `CustomPKIPolicy`. This validation enforces that the `custom.defaults.key` is specified when the mode is set to `Custom`.

config/v1alpha1/types_pki.go
sanchezl: When `CertificateConfig` is used as the `defaults`, the `key` is required when the mode is `Custom` (see [comment](https://github.com/openshift/api/pull/2645#discussion_r2824590753)).

When `CertificateConfig` is used in an override, the value of `key` falls back to the `defaults` defined in the same `PKIProfile`.

everettraven: Would it make sense to use a separate type for the default configuration where we can just make the field required instead of optional?

config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,279 @@
...
+type CertificateOverride struct {
+	// type determines what this override targets.
+	// Valid values are "Category".
+	//
+	// When set to Category, the override applies to all certificates of the
+	// specified category. The category field must be set.
+	//
+	// +required
+	// +unionDiscriminator
+	Type CertificateOverrideType `json:"type,omitempty"`
```
everettraven: You mention elsewhere that certificate overrides can also override specific named certificates, but this currently doesn't allow that. Does that need to be allowed or are you moving away from that approach?

config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,279 @@
...
+	// category specifies an override for a category of certificates.
+	// Required when type is Category, and forbidden otherwise.
+	//
+	// +optional
+	// +unionMember
+	Category *CategoryOverride `json:"category,omitzero"`
+
+	// certificate specifies the certificate configuration for this override.
+	// +required
+	Certificate CertificateConfig `json:"certificate,omitzero"`
```
everettraven: For discriminated unions, we generally try to avoid having co-located fields and instead prefer something like:
```yaml
type: Category
category:
  name: Signer
  certificate: ...
```

config/v1alpha1/types_pki.go
everettraven: I'm still wondering if ^ is a better approach here so that users perusing documentation generated from the OpenAPI schema of the API have a more clear distinction of the scenarios in which this field is explicitly required and when it is explicitly optional.

config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,279 @@
...
+	PKICertificateManagementModeDefault PKICertificateManagementMode = "Default"
+
+	// PKICertificateManagementModeCustom uses administrator-specified configuration.
+	PKICertificateManagementModeCustom PKICertificateManagementMode = "Custom"
+)
+
+// PKIProfile defines the certificate generation parameters that OpenShift
+// components use to create certificates. Overrides take precedence over defaults.
+//
+// +kubebuilder:validation:XValidation:rule="has(self.overrides) ? self.overrides.filter(o, o.type == 'Category').all(a, self.overrides.filter(o, o.type == 'Category').exists_one(b, a.category.name == b.category.name)) : true",message="each override of type Category must have a unique name"
```
everettraven: Make sure we explicitly document this constraint in the GoDoc for the appropriate field please.

config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,279 @@
...
+	// +optional
+	// +unionMember
+	Custom CustomPKIPolicy `json:"custom,omitzero"`
+}
+
+// CustomPKIPolicy contains administrator-specified cryptographic configuration.
+// Administrators must specify defaults for all certificates and may optionally
+// override specific categories of certificates.
+//
+// +kubebuilder:validation:XValidation:rule="has(self.defaults.key)",message="the default key algorithm configuration is required",fieldPath="defaults.key"
```
everettraven: If we do not go the route of using separate types for delineation here, let's make sure this requirement is explicitly documented in the GoDoc of an appropriate field.

config/v1alpha1/types_pki.go
sanchezl: I’ve decided to postpone the implementation of the named certificate override to a future version. I need to take some time to carefully work out the certificate name validation.

config/v1alpha1/types_pki.go
sanchezl: Moved the validation closer to the field and updated the Go doc.

config/v1alpha1/types_pki.go
sanchezl: I created a separate type with key required.

config/v1alpha1/types_pki.go
sanchezl: Done.

config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,293 @@
...
+
+	// overrides allows overriding certificate parameters for specific
+	// categories of certificates.
+	// Overrides take precedence over defaults.
+	// Each override of type Category must target a unique category name.
+	//
+	// +optional
+	// +listType=atomic
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=3
```
everettraven: Document the minimum and maximum length constraints in the GoDoc as well please.

config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,293 @@
...
+	// +unionDiscriminator
+	Mode PKICertificateManagementMode `json:"mode,omitempty"`
+
+	// custom contains administrator-specified cryptographic configuration.
+	// Use the defaults and overrides fields to specify certificate generation parameters.
+	// Required when mode is Custom, and forbidden otherwise.
+	//
+	// +optional
+	// +unionMember
+	// +kubebuilder:validation:MinProperties=1
```
everettraven: I don't think this actually does anything because the inlined type has a required field.

If you put this in place to make the linter happy, it might be a bug with the linter not properly traversing the inlined types as fields of a specific type.

Some of this confusion is why we generally prefer to avoid inlined types where possible, but I'm not strongly opinionated enough to block the use of an inlined type.

config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,293 @@
...
+
+// KeyConfig specifies cryptographic parameters for key generation.
+//
+// +kubebuilder:validation:XValidation:rule="has(self.algorithm) && self.algorithm == 'RSA' ?  has(self.rsa) : !has(self.rsa)",message="rsa is required when algorithm is RSA, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="has(self.algorithm) && self.algorithm == 'ECDSA' ?  has(self.ecdsa) : !has(self.ecdsa)",message="ecdsa is required when algorithm is ECDSA, and forbidden otherwise"
+// +union
+type KeyConfig struct {
+	// algorithm specifies the key generation algorithm.
+	// Valid values are "RSA" and "ECDSA".
+	// +required
```
everettraven, line 184: Add a description of what it means when each of these values is set here like:
```
// When set to RSA, ...
//
// When set to ECDSA, ...
```

config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,293 @@
...
+)
+
+// +kubebuilder:validation:Enum=Signer;Serving;Client
+type CertificateCategory string
+
+const (
+	CertificateCategorySigner  CertificateCategory = "Signer"
+	CertificateCategoryServing CertificateCategory = "Serving"
+	CertificateCategoryClient  CertificateCategory = "Client"
+)
```
everettraven, line 257: All exported types/constants should have a GoDoc associated with them

config/v1alpha1/types_pki.go
sanchezl: Updated

config/v1alpha1/types_pki.go
sanchezl: Updated

config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,328 @@
...
+type CertificateConfig struct {
+	// key specifies the cryptographic parameters for the certificate's key pair.
+	// Currently this is the only configurable parameter. When omitted in an
+	// overrides entry, the key configuration from defaults is used.
+	// +optional
+	Key KeyConfig `json:"key,omitempty,omitzero"`
+
+	// Future extensibility: fields like Lifetime, Rotation, Extensions
+	// can be added here without restructuring the API.
+}
```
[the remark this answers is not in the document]
everettraven: This is intentional so that when specifying an override field that setting it to `{}` is not a valid input. Currently, `key` is the only field, but there is a potential future where more than the `key` field exists, making it actually optional.

The current implementation is correct for the conventions we have.

config/v1alpha1/types_pki.go
```diff
@@ -0,0 +1,264 @@
...
+	// +required
+	// +kubebuilder:validation:Minimum=2048
+	// +kubebuilder:validation:Maximum=8192
+	// +kubebuilder:validation:MultipleOf=1024
+	KeySize int32 `json:"keySize,omitempty"`
+}
+
+// ECDSAKeyConfig specifies parameters for ECDSA key generation.
+type ECDSAKeyConfig struct {
+	// curve specifies the elliptic curve for ECDSA keys.
```
everettraven: Would it be worth clarifying that the names here are NIST representations?

Would it be worth following the same `When set to {Value}, ...` comment format to ensure we describe additional information about what each value corresponds to, preventing confusion for folks less familiar with particular curve name formats?

I like the comments you put on the constants that correspond to this field type because it includes an "also known as ..." blurb.

config/v1alpha1/types_pki.go
sanchezl: Updated enhancement.

config/v1alpha1/types_pki.go
sanchezl: Update Godoc.

config/v1alpha1/types_pki.go
sanchezl: Done.

everettraven:
@sanchezl It looks like the 1.35 rebase cause some minor updates that need to be made here for verify to pass.

Rebasing on the `master` branch latest changes and then running `PROTO_OPTIONAL=true make update` should fix this up

PR #2648, opened 2026-01-14
OCPBUGS-73855: Permit other machine managers when skew enforcement is in Automatic mode

djoshy:
While working on https://github.com/openshift/machine-config-operator/pull/5539 and https://github.com/openshift/machine-config-operator/pull/5428, I realized that the original skew validation rule should also permit MachineManagers of other types. This is because with https://github.com/openshift/machine-config-operator/pull/5539, the MCO will respect a admin defined Manager opinion, and then merge into the status with OCP's default opinion for the other unspecified MachineManagers. For example, GCP by default looks like this:
```
kind: MachineConfiguration
metadata:
  name: cluster
  namespace: openshift-machine-config-operator
spec:
status:
  managedBootImagesStatus:
    machineManagers:
      - resource: controlplanemachinesets
        apiGroup: machine.openshift.io
        selection:
          mode: None
      - resource: machinesets
        apiGroup: machine.openshift.io
        selection:
          mode: All
```
If an admin then sets an opt-in config for the CPMS manager, the MCO will set the status as:
```
kind: MachineConfiguration
metadata:
  name: cluster
  namespace: openshift-machine-config-operator
spec:
  managedBootImages:
    machineManagers:
      - resource: controlplanemachinesets
        apiGroup: machine.openshift.io
        selection:
          mode: All
status:
  managedBootImagesStatus:
    machineManagers:
      - resource: controlplanemachinesets
        apiGroup: machine.openshift.io
        selection:
          mode: All
      - resource: machinesets
        apiGroup: machine.openshift.io
        selection:
          mode: All
```
Now, if we add Automatic skew mode to the mix, we run into an issue:
```
kind: MachineConfiguration
metadata:
  name: cluster
  namespace: openshift-machine-config-operator
spec:
status:
  managedBootImagesStatus:
    machineManagers:
      - resource: controlplanemachinesets
        apiGroup: machine.openshift.io
        selection:
          mode: None
      - resource: machinesets
        apiGroup: machine.openshift.io
        selection:
          mode: All
  bootImageSkewEnforcementStatus:
    mode: Automatic         
    automatic:
      ocpVersion: "4.18.2"
      rhcosVersion: "416.94.202411201433-0" 
```
And then if we add a CPMS opinion, we would hope to see:

```
kind: MachineConfiguration
metadata:
  name: cluster
  namespace: openshift-machine-config-operator
spec:
  managedBootImages:
    machineManagers:
      - resource: controlplanemachinesets
        apiGroup: machine.openshift.io
        selection:
          mode: All
status:
  managedBootImagesStatus:
    machineManagers:
      - resource: controlplanemachinesets
        apiGroup: machine.openshift.io
        selection:
          mode: All
      - resource: machinesets
        apiGroup: machine.openshift.io
        selection:
          mode: All
  bootImageSkewEnforcementStatus:
    mode: Automatic         
    automatic:
      ocpVersion: "4.18.2"
      rhcosVersion: "416.94.202411201433-0" 
```

Unfortunately due to the current [validation rule](https://github.com/openshift/api/blob/master/operator/v1/types_machineconfiguration.go#L21) for the spec field, this  fails:
```
# machineconfigurations.operator.openshift.io "cluster" was not valid:
# * <nil>: Invalid value: "object": when skew enforcement is in Automatic mode, managedBootImages must contain a MachineManager opting in all MachineAPI MachineSets
#
```

We first noticed this on presubmit payload runs for CPMS tests: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/openshift-machine-config-operator-5547-periodics-e2e-gcp-mco-disruptive-techpreview-1of2/2011125283620392960; they fail as the test is prevented by the validation from setting a CPMS configuration. 

To fix this, I'm proposing that we break the old rule into two parts when skew enforcement is in Automatic mode:
1. If a MAPI MachineSet Manager is present in `spec.managedBootImages.machineManagers`, it must be set to `All`. A MAPI MachineSet Manager not being present is a valid state.
2. Guard against `spec.managedBootImages.machineManagers` to be defined as an empty list; this is an old way to opt the cluster out of boot image updates. When set, the MCO reflects an empty list to the status, and so this is something that we shouldn't permit while in Automatic mode. Alternatively, we could also consider always enforcing a length requirement if it is defined, regardless of skew enforcement mode.

I've also added tests for these validations targeting the BootImageSkewEnforcement and ManagedBootImagesCPMS feature gate. The validations for managedBootImagesStatus remains the same; the MCO should always ensure that there is an "All" MAPI MachineSet manager for platforms that support Automatic skew enforcement.

### DIFF
--- a/operator/v1/tests/machineconfigurations.operator.openshift.io/BootImageSkewEnforcement+ManagedBootImagesCPMS.yaml
+++ b/operator/v1/tests/machineconfigurations.operator.openshift.io/BootImageSkewEnforcement+ManagedBootImagesCPMS.yaml
@@ -1,0 +1,223 @@
+apiVersion: apiextensions.k8s.io/v1 # Hack because controller-gen complains if we don't have this
+name: "MachineConfiguration"
+crdName: machineconfigurations.operator.openshift.io
+featureGates:
+- BootImageSkewEnforcement
+- ManagedBootImagesCPMS
+tests:
+  onUpdate:
+    - name: Should permit an "All" CPMS boot image configuration when bootImageSkewEnforcementStatus.mode is Automatic
+      initial: |
+        apiVersion: operator.openshift.io/v1
+        kind: MachineConfiguration
+        spec: {}
+        status:
+          managedBootImagesStatus:
+            machineManagers:
+              - resource: machinesets
+                apiGroup: machine.openshift.io
+                selection:
+                  mode: All
+          bootImageSkewEnforcementStatus:
+            mode: Automatic         
+            automatic:
+              ocpVersion: "4.18.2"
+              rhcosVersion: "416.94.202411201433-0"
+      updated: |
+        apiVersion: operator.openshift.io/v1
+        kind: MachineConfiguration
+        spec: 
+          managedBootImages:
+            machineManagers:
+              - resource: controlplanemachinesets
+                apiGroup: machine.openshift.io
+                selection:
+                  mode: All
+        status:
+          managedBootImagesStatus:
+            machineManagers:
+              - resource: machinesets
+                apiGroup: machine.openshift.io
+                selection:
+                  mode: All
+          bootImageSkewEnforcementStatus:
+            mode: Automatic         
+            automatic:
+              ocpVersion: "4.18.2"
+              rhcosVersion: "416.94.202411201433-0"
+      expected: |
+        apiVersion: operator.openshift.io/v1
+        kind: MachineConfiguration
+        spec:
+          logLevel: Normal
+          operatorLogLevel: Normal  
+          managedBootImages:
+            machineManagers:
+              - resource: controlplanemachinesets
+                apiGroup: machine.openshift.io
+                selection:
+                  mode: All
+        status:
+          managedBootImagesStatus:
+            machineManagers:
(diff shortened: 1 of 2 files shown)


JoelSpeed (commented):
/lgtm

djoshy:
Hmm, the Azure payloads didn't launch due to an infra issue, but the original failures were only on platforms that already had boot images enabled by default(GCP, AWS) so it is not really relevant. For the GCP payload runs, the CPMS tests are now passing in a combined run with https://github.com/openshift/machine-config-operator/pull/5428 (skew enforcement MVP PR) [[1]](https://pr-payload-tests.ci.openshift.org/runs/ci/1c8b1200-f171-11f0-8878-79c2b5f63f90-0) [[2]](https://pr-payload-tests.ci.openshift.org/runs/ci/2091cd30-f171-11f0-862c-140aba58a64d-0)

djoshy:
Hmm, not sure why its still gating on the unsharded version of the suites. It seems to be waiting for some kind of pipeline condition?

JoelSpeed:

There's an issue with the pipeline controller currently, these cases don't actually exist here

PR #2650, opened 2026-01-15
SPLAT-2615: Added support for dynamic AWS dedicated hosts

vr4manta:
[SPLAT-2615](https://issues.redhat.com//browse/SPLAT-2615)

### Changes
- Added new dynamic dedicated host support

### Blocks
- https://github.com/openshift/cluster-capi-operator/pull/374
- https://github.com/openshift/machine-api-provider-aws/pull/172
- https://github.com/openshift/machine-api-operator/pull/1455

### Notes
For dynamic hosts, we are currently allowing admins to provide a set of tags to apply to the new host.  Other future enhancements for it can be contained in the new DynamicHostAllocation struct.
This PR is implementing the OCP equivalent to the changes in https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5631.


vr4manta:
After some reviews, gonna make some small changes to not add a new affinity to keep in line with AWS.

machine/v1beta1/types_awsprovider.go
```diff
@@ -460,15 +460,56 @@ const (
 	HostAffinityDedicatedHost HostAffinity = "DedicatedHost"
 )
 
+// AllocationStrategy selects how a dedicated host is provided to the system for assigning to the instance.
+// +kubebuilder:validation:Enum:=Provided;Dynamic
+type AllocationStrategy string
+
+const (
+	// AllocationStrategyUserProvided specifies system is to assign a provided dedicated host to instances.
+	AllocationStrategyUserProvided AllocationStrategy = "Provided"
```
saschagrunert: I see a mismatch between the variable name and value. I guess we should either use 
```suggestion
	// AllocationStrategyProvided specifies system is to assign a provided dedicated host to instances.
	AllocationStrategyProvided AllocationStrategy = "Provided"
```

or 
```suggestion
	// AllocationStrategyUserProvided specifies system is to assign a provided dedicated host to instances.
	AllocationStrategyUserProvided AllocationStrategy = "UserProvided"
```

machine/v1beta1/types_awsprovider.go
```diff
@@ -460,15 +460,56 @@ const (
 	HostAffinityDedicatedHost HostAffinity = "DedicatedHost"
 )
 
+// AllocationStrategy selects how a dedicated host is provided to the system for assigning to the instance.
+// +kubebuilder:validation:Enum:=Provided;Dynamic
+type AllocationStrategy string
+
+const (
+	// AllocationStrategyUserProvided specifies system is to assign a provided dedicated host to instances.
```
saschagrunert: How about:
```suggestion
	// AllocationStrategyUserProvided specifies that the system should assign instances to a user-provided dedicated host.
```

machine/v1beta1/types_awsprovider.go
```diff
@@ -460,15 +460,56 @@ const (
...
+	AllocationStrategyUserProvided AllocationStrategy = "Provided"
+
+	// AllocationStrategyDynamic specifies system to dynamically allocate a dedicated host to use for assignment to instances.
```
saschagrunert: How about:
```suggestion
	// AllocationStrategyDynamic specifies that the system should dynamically allocate a dedicated host for instances.
```

machine/v1beta1/types_awsprovider.go
```diff
@@ -460,15 +460,56 @@ const (
...
 	// The value must start with "h-" followed by either 8 or 17 lowercase hexadecimal characters (0-9 and a-f).
 	// The use of 8 lowercase hexadecimal characters is for older legacy hosts that may not have been migrated to newer format.
 	// Must be either 10 or 19 characters in length.
 	// +kubebuilder:validation:XValidation:rule="self.matches('^h-([0-9a-f]{8}|[0-9a-f]{17})$')",message="hostID must start with 'h-' followed by either 8 or 17 lowercase hexadecimal characters (0-9 and a-f)"
 	// +kubebuilder:validation:MinLength=10
 	// +kubebuilder:validation:MaxLength=19
-	// +required
+	// +optional
+	// +unionMember
 	ID string `json:"id,omitempty"`
```
saschagrunert, line 531: `ID` is already part of 4.21, right? If so, then it's a breaking change:

**4.21 will ship**:
```go
type DedicatedHost struct {
    // +required
    ID string `json:"id,omitempty"`
}
```

**4.22 will ship**:
```go
type DedicatedHost struct {
    // +required
    AllocationStrategy AllocationStrategy `json:"allocationStrategy,omitempty"`

    // +optional (changed from +required)
    ID string `json:"id,omitempty"`

    // +optional (new field)
    DynamicHostAllocation *DynamicHostAllocationSpec `json:"dynamicHostAllocation,omitempty"`
}
```

Means that a valid 4.21 manifest will be invalid in 4.22.

```yaml
dedicatedHost:
  id: "h-1234567890abcdef0"
```

Do we have a webhook or similar solution in the works for this?

machine/v1beta1/types_awsprovider.go
```diff
@@ -460,15 +460,56 @@ const (
...
+
 	// id identifies the AWS Dedicated Host on which the instance must run.
 	// The value must start with "h-" followed by either 8 or 17 lowercase hexadecimal characters (0-9 and a-f).
 	// The use of 8 lowercase hexadecimal characters is for older legacy hosts that may not have been migrated to newer format.
 	// Must be either 10 or 19 characters in length.
 	// +kubebuilder:validation:XValidation:rule="self.matches('^h-([0-9a-f]{8}|[0-9a-f]{17})$')",message="hostID must start with 'h-' followed by either 8 or 17 lowercase hexadecimal characters (0-9 and a-f)"
 	// +kubebuilder:validation:MinLength=10
 	// +kubebuilder:validation:MaxLength=19
-	// +required
+	// +optional
```
saschagrunert, line 529: We need to document now what happens when ommited:

```
// This field is required when allocationStrategy is Provided, and forbidden when allocationStrategy is Dynamic.
// When omitted, allocationStrategy must be set to Dynamic to enable automatic host allocation.
```

machine/v1beta1/types_awsprovider.go
```diff
@@ -460,15 +460,56 @@ const (
...
+	// +unionMember
 	ID string `json:"id,omitempty"`
+
+	// dynamicHostAllocation specifies tags to apply to a dynamically allocated dedicated host.
+	// This field is mutually exclusive with id and always allocates exactly one host.
```
saschagrunert: Same here, we need to document missing behavior:
```suggestion
	// This field is only allowed when allocationStrategy is Dynamic, and is mutually exclusive with id.
	// When specified, a dedicated host will be allocated with the provided tags applied.
	// When omitted (and allocationStrategy is Dynamic), a dedicated host will be allocated without any additional tags.
```

machine/v1beta1/types_awsprovider.go
```diff
@@ -460,15 +460,56 @@ const (
...
+	// +optional
+	// +unionMember
+	DynamicHostAllocation *DynamicHostAllocationSpec `json:"dynamicHostAllocation,omitempty"`
+}
+
+// DynamicHostAllocationSpec defines the configuration for dynamic dedicated host allocation.
+// This specification always allocates exactly one dedicated host per machine.
+type DynamicHostAllocationSpec struct {
+	// tags to apply to the allocated dedicated host.
```
saschagrunert: And for this one as well, document missing behavior:
```suggestion
	// tags specifies a set of key-value pairs to apply to the allocated dedicated host.
	// When omitted or empty, no additional user-defined tags will be applied to the allocated host.
```

machine/v1beta1/types_awsprovider.go
```diff
@@ -460,15 +460,56 @@ const (
...
 type DedicatedHost struct {
+	// allocationStrategy specifies if the dedicated host will be provided by the admin through the id field or if the host will be dynamically allocated.
+	// Valid values are Provided and Dynamic.
+	// The current default value is "Provided".
+	// When AllocationStrategy is set to Provided, an ID of the dedicated host to assign must be provided.
+	// When AllocationStrategy is set to Dynamic, a dedicated host will be allocated and used to assign instances.
+	// When AllocationStrategy is set to Dynamic, and DynamicHostAllocation is provided, a dedicated host will be allocated and the tags in DynamicHostAllocation will be assigned to that host.
+	// +required
+	// +unionDiscriminator
+	// +kubebuilder:default=Provided
```
saschagrunert: The `+kubebuilder:default=Provided` marker does not generate a default value in the OpenAPI schema because the field is marked as `+required`. The marker only adds documentation text, not an actual OpenAPI default.

So we could remove that line and change
```go
	// The current default value is "Provided".
```

to 

```go
	// This field is required.
```

machine/v1beta1/types_awsprovider.go
vr4manta: I'll fix that, i saw that late yesterday and haven't updated yet.

machine/v1beta1/types_awsprovider.go
vr4manta: ah ok.  I'll adjust that.

machine/v1beta1/types_awsprovider.go
vr4manta: ok

machine/v1beta1/types_awsprovider.go
vr4manta: Yes there is webhook changes, that PR is almost ready.  For backwards compatability, if its "" (emptry string) it is defaulting to UserProvided.

vr4manta: For this, I can change the field to be optional but have it default to UserProvided.  Just let me know what you prefer.

machine/v1beta1/types_awsprovider.go
```diff
@@ -454,21 +454,71 @@ type HostAffinity string
...
+// +kubebuilder:validation:XValidation:rule="has(self.dynamicHostAllocation) ? self.allocationStrategy == 'Dynamic' : true",message="dynamicHostAllocation is only allowed when allocationStrategy is Dynamic"
+// +union
 type DedicatedHost struct {
+	// allocationStrategy specifies if the dedicated host will be provided by the admin through the id field or if the host will be dynamically allocated.
+	// Valid values are UserProvided and Dynamic.
+	// This field is required.
+	// When AllocationStrategy is set to UserProvided, an ID of the dedicated host to assign must be provided.
+	// When AllocationStrategy is set to Dynamic, a dedicated host will be allocated and used to assign instances.
+	// When AllocationStrategy is set to Dynamic, and DynamicHostAllocation is configured, a dedicated host will be allocated and the tags in DynamicHostAllocation will be assigned to that host.
+	// +optional
```
saschagrunert: Documentation says "This field is required." but Marker Says `+optional`.
```suggestion
	// This field is optional and defaults to "UserProvided".
	// When AllocationStrategy is set to UserProvided, an ID of the dedicated host to assign must be provided.
	// When AllocationStrategy is set to Dynamic, a dedicated host will be allocated and used to assign instances.
	// When AllocationStrategy is set to Dynamic, and DynamicHostAllocation is configured, a dedicated host will be allocated and the tags in DynamicHostAllocation will be assigned to that host.
	// +optional
```

machine/v1beta1/types_awsprovider.go
```diff
@@ -454,21 +454,71 @@ type HostAffinity string
...
 	// The use of 8 lowercase hexadecimal characters is for older legacy hosts that may not have been migrated to newer format.
 	// Must be either 10 or 19 characters in length.
+	// This field is required when allocationStrategy is UserProvided, and forbidden when allocationStrategy is Dynamic.
+	// When omitted, allocationStrategy must be set to Dynamic to enable automatic host allocation.
 	// +kubebuilder:validation:XValidation:rule="self.matches('^h-([0-9a-f]{8}|[0-9a-f]{17})$')",message="hostID must start with 'h-' followed by either 8 or 17 lowercase hexadecimal characters (0-9 and a-f)"
 	// +kubebuilder:validation:MinLength=10
 	// +kubebuilder:validation:MaxLength=19
-	// +required
+	// +optional
+	// +unionMember
```
saschagrunert: I think the union discriminator values are incorrect. Controller-gen is generating `"ID"` and `"DynamicHostAllocation"` instead of the enum values `"UserProvided"` and `"Dynamic"`. This will cause union validation to fail at runtime.


```suggestion
	// +unionMember=UserProvided
```

machine/v1beta1/types_awsprovider.go
```diff
@@ -454,21 +454,71 @@ type HostAffinity string
...
 	ID string `json:"id,omitempty"`
+
+	// dynamicHostAllocation specifies tags to apply to a dynamically allocated dedicated host.
+	// This field is only allowed when allocationStrategy is Dynamic, and is mutually exclusive with id.
+	// When specified, a dedicated host will be allocated with the provided tags applied.
+	// When omitted (and allocationStrategy is Dynamic), a dedicated host will be allocated without any additional tags.
+	// +kubebuilder:validation:MinProperties=1
+	// +optional
+	// +unionMember
```
saschagrunert: Same for this:

```suggestion
	// +unionMember=Dynamic
```

machine/v1beta1/types_awsprovider.go
```diff
@@ -454,21 +454,71 @@ type HostAffinity string
...
+const (
+	// AllocationStrategyUserProvided specifies that the system should assign instances to a user-provided dedicated host.
+	AllocationStrategyUserProvided AllocationStrategy = "UserProvided"
+
+	// AllocationStrategyDynamic specifies that the system should dynamically allocate a dedicated host for instances.
+	AllocationStrategyDynamic AllocationStrategy = "Dynamic"
+)
+
 // DedicatedHost represents the configuration for the usage of dedicated host.
+// +kubebuilder:validation:XValidation:rule="self.allocationStrategy == 'UserProvided' ? has(self.id) : true",message="id is required when allocationStrategy is UserProvided, and forbidden otherwise"
```
saschagrunert: Message says "forbidden otherwise" but this rule doesn't enforce that (line 478 does). The validation is correct, just the message is confusing, so I'd suggest:

```suggestion
// +kubebuilder:validation:XValidation:rule="self.allocationStrategy == 'UserProvided' ? has(self.id) : true",message="id is required when allocationStrategy is UserProvided"
```

machine/v1beta1/types_awsprovider.go
```diff
@@ -454,21 +454,71 @@ type HostAffinity string
...
-	// +required
+	// +optional
+	// +unionMember
 	ID string `json:"id,omitempty"`
+
+	// dynamicHostAllocation specifies tags to apply to a dynamically allocated dedicated host.
+	// This field is only allowed when allocationStrategy is Dynamic, and is mutually exclusive with id.
+	// When specified, a dedicated host will be allocated with the provided tags applied.
+	// When omitted (and allocationStrategy is Dynamic), a dedicated host will be allocated without any additional tags.
+	// +kubebuilder:validation:MinProperties=1
```
saschagrunert: `MinProperties` marker on a `$ref` field is ignored. This marker only works on inline object schemas, not references:
```suggestion
```

The `MinProperties=1` on the Tags field (line 521) inside DynamicHostAllocationSpec is working correctly.

machine/v1beta1/types_awsprovider.go
```diff
@@ -454,21 +454,71 @@ type HostAffinity string
...
+	// allocationStrategy specifies if the dedicated host will be provided by the admin through the id field or if the host will be dynamically allocated.
+	// Valid values are UserProvided and Dynamic.
+	// This field is required.
+	// When AllocationStrategy is set to UserProvided, an ID of the dedicated host to assign must be provided.
+	// When AllocationStrategy is set to Dynamic, a dedicated host will be allocated and used to assign instances.
+	// When AllocationStrategy is set to Dynamic, and DynamicHostAllocation is configured, a dedicated host will be allocated and the tags in DynamicHostAllocation will be assigned to that host.
+	// +optional
+	// +unionDiscriminator
+	// +default="UserProvided"
+	AllocationStrategy *AllocationStrategy `json:"allocationStrategy,omitempty"`
```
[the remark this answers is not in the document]
saschagrunert: I'd say this is right to change.

openapi/generated_openapi/zz_generated.openapi.go
```diff
@@ -41175,17 +41176,45 @@ func schema_openshift_api_machine_v1beta1_DedicatedHost(ref common.ReferenceCall
...
+								"id":                    "ID",
+							},
+						},
+					},
 				},
-				Required: []string{"id"},
 			},
 		},
+		Dependencies: []string{
+			"github.com/openshift/api/machine/v1beta1.DynamicHostAllocationSpec"},
```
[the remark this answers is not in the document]
saschagrunert: This should get fixed with https://github.com/openshift/api/pull/2650#discussion_r2712253251 and https://github.com/openshift/api/pull/2650#discussion_r2712300503.

vr4manta:
> No integration tests exist for `DedicatedHost` union validation, can we add them in a location like `machine/v1beta1/tests/awsmachineproviderconfig.machine.openshift.io/DedicatedHost.testsuite.yaml`?

I can try.  In past experiences since the machine spec sections are not CRD validated, we've done these over in the MAO project via unit tests where the webhook does the actual validation.

openapi/openapi.json
```diff
@@ -23672,15 +23730,30 @@
...
+      },
+      "x-kubernetes-unions": [
+        {
+          "discriminator": "allocationStrategy",
+          "fields-to-discriminateBy": {
+            "dynamicHostAllocation": "DynamicHostAllocation",
+            "id": "ID"
+          }
+        }
+      ]
```
saschagrunert, line 23764: This should get fixed with https://github.com/openshift/api/pull/2650#discussion_r2712253251 and https://github.com/openshift/api/pull/2650#discussion_r2712300503.

openapi/openapi.json
```diff
@@ -23672,15 +23730,30 @@
...
       "type": "object",
-      "required": [
-        "id"
-      ],
       "properties": {
+        "allocationStrategy": {
+          "description": "allocationStrategy specifies if the dedicated host will be provided by the admin through the id field or if the host will be dynamically allocated. Valid values are UserProvided and Dynamic. This field is required. When AllocationStrategy is set to UserProvided, an ID of the dedicated host to assign must be provided. When AllocationStrategy is set to Dynamic, a dedicated host will be allocated and used to assign instances. When AllocationStrategy is set to Dynamic, and DynamicHostAllocation is configured, a dedicated host will be allocated and the tags in DynamicHostAllocation will be assigned to that host.",
+          "type": "string",
+          "default": "UserProvided"
+        },
```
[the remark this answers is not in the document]
saschagrunert: Is this coming from partially pushed code? Don't see the enum marker at `AllocationStrategy` anymore.

machine/v1beta1/types_awsprovider.go
saschagrunert: I think the current implementation addresses the raised concerns.

machine/v1beta1/types_awsprovider.go
vr4manta: Removing this causes:
```
INFO[2026-01-21T12:48:28Z] hack/golangci-lint.sh run --new-from-rev=a2874d62513ceb5fb0f3c4ac32434e6cdcf3c3ed 
Writing tools vendor version 98cf84638e2c240301582d2d655a662049ca99f5
machine/v1beta1/types_awsprovider.go:512:2: minlength: field DedicatedHost.DynamicHostAllocation type DynamicHostAllocationSpec must have a minimum properties, add kubebuilder:validation:MinProperties marker (kubeapilinter)
	DynamicHostAllocation *DynamicHostAllocationSpec `json:"dynamicHostAllocation,omitempty"`
	^
1 issues:
* kubeapilinter: 1
0 issues.
```
But I can move it above the struct so that should be fine.  Not sure if that is what you were wanting me to do.

machine/v1beta1/types_awsprovider.go
```diff
@@ -454,21 +454,71 @@ type HostAffinity string
...
 	// HostAffinityAnyAvailable lets the platform select any available dedicated host.
+
 	HostAffinityAnyAvailable HostAffinity = "AnyAvailable"
 
 	// HostAffinityDedicatedHost requires specifying a particular host via dedicatedHost.host.hostID.
 	HostAffinityDedicatedHost HostAffinity = "DedicatedHost"
 )
 
+// AllocationStrategy selects how a dedicated host is provided to the system for assigning to the instance.
+// +kubebuilder:validation:Enum:=UserProvided;Dynamic
```
saschagrunert, line 492: We may need this here
```suggestion
// +kubebuilder:validation:Enum:=UserProvided;Dynamic
// +enum
```

saschagrunert:
@vr4manta is this still WIP? The linter is not happy, though.

vr4manta:
> @vr4manta is this still WIP? The linter is not happy, though.

@saschagrunert yes still WIP.  That lint message pushed me to work on other PRs.  I'll return to this one today since its blocking 4 that will need this.

vr4manta:
@saschagrunert so the linter is complaining about a field that existed before my PR.  I can fix it by making the key field required in the struct, but it will affect other areas that use it.  I feel doing this would be not safe due to going from optional to required.  What would you like me to do for this lint error?

saschagrunert:
> @saschagrunert so the linter is complaining about a field that existed before my PR. I can fix it by making the key field required in the struct, but it will affect other areas that use it. I feel doing this would be not safe due to going from optional to required. What would you like me to do for this lint error?

Yeah the linter is new. Can you add `+required` to the `TagSpecification` fields?

```go
  // TagSpecification is the name/value pair for a tag
  type TagSpecification struct {
      // name of the tag.
      // This field is required and must be a non-empty string.
      // +required
      Name string `json:"name"`

      // value of the tag.
      // This field is required but may be an empty string.
      // +required
      Value string `json:"value"`
  }

```

vr4manta:
> > @saschagrunert so the linter is complaining about a field that existed before my PR. I can fix it by making the key field required in the struct, but it will affect other areas that use it. I feel doing this would be not safe due to going from optional to required. What would you like me to do for this lint error?
> 
> Yeah the linter is new. Can you add `+required` to the `TagSpecification` fields?
> 
> ```go
>   // TagSpecification is the name/value pair for a tag
>   type TagSpecification struct {
>       // name of the tag.
>       // This field is required and must be a non-empty string.
>       // +required
>       Name string `json:"name"`
> 
>       // value of the tag.
>       // This field is required but may be an empty string.
>       // +required
>       Value string `json:"value"`
>   }
> ```

@saschagrunert  Yes I can.  I Just wanted you to confirm that it is ok and not gonna break things.  Since you've approved, i'll make change.  I plan on leaving value optional since you can set a key w/ no value for tags and that works now.

machine/v1beta1/types_awsprovider.go
```diff
@@ -407,6 +409,26 @@ type AWSMachineProviderStatus struct {
...
+// DedicatedHostStatus defines the observed state of a dynamically allocated dedicated host
+// associated with an AWSMachine. This struct is used to track the ID of the dedicated host.
+// +kubebuilder:validation:MinProperties=1
+type DedicatedHostStatus struct {
+	// id tracks the dynamically allocated dedicated host ID.
+	// This field is populated when DynamicHostAllocation is used.
+	// The value must start with "h-" followed by either 8 or 17 lowercase hexadecimal characters (0-9 and a-f).
+	// The use of 8 lowercase hexadecimal characters is for older legacy hosts that may not have been migrated to newer format.
+	// Must be either 10 or 19 characters in length.
+	// +kubebuilder:validation:XValidation:rule="self.matches('^h-([0-9a-f]{8}|[0-9a-f]{17})$')",message="hostID must start with 'h-' followed by either 8 or 17 lowercase hexadecimal characters (0-9 and a-f)"
```
saschagrunert: We need integration tests for the validation rules.

machine/v1beta1/types_awsprovider.go
```diff
@@ -331,7 +331,9 @@ type Filter struct {
 
 // TagSpecification is the name/value pair for a tag
 type TagSpecification struct {
-	// name of the tag
+	// name of the tag.
+	// This field is required and must be a non-empty string.
```
saschagrunert, line 335: We can probably add `MinLength=1` here or we should remove the non-empty constraint in the docs.

machine/v1beta1/types_awsprovider.go
saschagrunert:
```suggestion
	// +kubebuilder:validation:XValidation:rule="self.matches('^h-([0-9a-f]{8}|[0-9a-f]{17})$')",message="id must start with 'h-' followed by either 8 or 17 lowercase hexadecimal characters (0-9 and a-f)"
```

machine/v1beta1/types_awsprovider.go
```diff
@@ -454,21 +476,76 @@ type HostAffinity string
...
+	// +default="UserProvided"
+	AllocationStrategy *AllocationStrategy `json:"allocationStrategy,omitempty"`
+
 	// id identifies the AWS Dedicated Host on which the instance must run.
 	// The value must start with "h-" followed by either 8 or 17 lowercase hexadecimal characters (0-9 and a-f).
 	// The use of 8 lowercase hexadecimal characters is for older legacy hosts that may not have been migrated to newer format.
 	// Must be either 10 or 19 characters in length.
+	// This field is required when allocationStrategy is UserProvided, and forbidden when allocationStrategy is Dynamic.
+	// When omitted, allocationStrategy must be set to Dynamic to enable automatic host allocation.
 	// +kubebuilder:validation:XValidation:rule="self.matches('^h-([0-9a-f]{8}|[0-9a-f]{17})$')",message="hostID must start with 'h-' followed by either 8 or 17 lowercase hexadecimal characters (0-9 and a-f)"
```
saschagrunert:
```suggestion
	// +kubebuilder:validation:XValidation:rule="self.matches('^h-([0-9a-f]{8}|[0-9a-f]{17})$')",message="id must start with 'h-' followed by either 8 or 17 lowercase hexadecimal characters (0-9 and a-f)"
```

vr4manta:
> We don't need a feature gate for this change, don't we?

The parent field has the following:
```
	// +openshift:enable:FeatureGate=AWSDedicatedHosts
	// +optional
	Host *HostPlacement `json:"host,omitempty"`
```

machine/v1beta1/types_awsprovider.go
vr4manta: ok

machine/v1beta1/types_awsprovider.go
vr4manta: good catch

machine/v1beta1/types_awsprovider.go
vr4manta: fixed

machine/v1beta1/types_awsprovider.go
vr4manta: @JoelSpeed / @damdo / @everettraven  , I have looked into this request and it looks like I am not able to get cloud ProviderSpec to work.  I remember in the past that this was a limitation which I addressed with tests in the webhook in the MAO.  Has this been fixed?  If so, what do I need to do to test it?  Thanks!

machine/v1beta1/types_awsprovider.go
```diff
@@ -454,21 +478,76 @@ type HostAffinity string
...
+	// When specified, a dedicated host will be allocated with the provided tags applied.
+	// When omitted (and allocationStrategy is Dynamic), a dedicated host will be allocated without any additional tags.
+	// +optional
+	// +unionMember=Dynamic
+	DynamicHostAllocation *DynamicHostAllocationSpec `json:"dynamicHostAllocation,omitempty"`
+}
+
+// DynamicHostAllocationSpec defines the configuration for dynamic dedicated host allocation.
+// This specification always allocates exactly one dedicated host per machine.
+// +kubebuilder:validation:MinProperties=1
```
saschagrunert, line 546: We need to document `+kubebuilder:validation:MinProperties=1`:
```suggestion
// At least one property must be specified when this struct is used.
// +kubebuilder:validation:MinProperties=1
```

machine/v1beta1/types_awsprovider.go
```diff
@@ -454,21 +478,76 @@ type HostAffinity string
...
+type DynamicHostAllocationSpec struct {
+	// tags specifies a set of key-value pairs to apply to the allocated dedicated host.
+	// When omitted, no additional user-defined tags will be applied to the allocated host.
+	// A maximum of 50 tags can be specified.
+	// +kubebuilder:validation:MinItems=1
```
saschagrunert, line 551: The combination of those are confusing:
  - MinProperties=1 on the struct
  - MinItems=1 on the tags field
  - +optional marker on the field

Tags is the only field in the struct, and MinProperties=1 requires at least one property, the field is effectively required when the struct exists. But MinItems=1 means if you do provide tags, it cannot be an empty array.

machine/v1beta1/types_awsprovider.go
JoelSpeed: No, for MAPI provider spec changes we have to implement these via webhooks so you won't be able to integration test them here

machine/v1beta1/types_awsprovider.go
vr4manta: correct, i think that is what linter / coderabbit kept forcing me torwards.  I do believe it would be crazy to provide tags w/o any items so requiring 1 min item would make sense.  I can try to remove the MinProperties but coderabbit was the one requiring it (or maybe it was linter.)  I'll see if I can trim this again .

machine/v1beta1/types_awsprovider.go
vr4manta: sure, but only if your following comment doesn't result in this field being removed.

machine/v1beta1/types_awsprovider.go
vr4manta:
```
machine/v1beta1/types_awsprovider.go:537:2: minlength: field DedicatedHost.DynamicHostAllocation type DynamicHostAllocationSpec must have a minimum properties, add kubebuilder:validation:MinProperties marker (kubeapilinter)
	DynamicHostAllocation *DynamicHostAllocationSpec `json:"dynamicHostAllocation,omitempty"`
```
Currently this is the only field in this struct.  In the future we may add more for user to provide configuration settings for the dedicated host.  I would hate to make this field required and change in the future to optional.  Just let me know which way you want to go.

openapi/openapi.json
```diff
@@ -23702,6 +23793,24 @@
...
+          "type": "array",
+          "items": {
+            "default": {},
+            "$ref": "#/definitions/com.github.openshift.api.machine.v1beta1.TagSpecification"
+          },
+          "x-kubernetes-list-map-keys": [
+            "name"
+          ],
+          "x-kubernetes-list-type": "map"
+        }
```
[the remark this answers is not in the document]
JoelSpeed: There is no compatible tag that we can add to the Go types to generate a maxItems value in the openapi schema at present

machine/v1beta1/types_awsprovider.go
```diff
@@ -331,9 +331,15 @@ type Filter struct {
...
-	// name of the tag
+	// name of the tag.
+	// This field is required and must be a non-empty string.
+	// Must be between 1 and 128 characters in length.
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=128
+	// +required
 	Name string `json:"name"`
 	// value of the tag
+	// +optional
```
saschagrunert, line 343:
```suggestion
	// +optional
	// When omitted, this creates a tag with an empty string as the value.
	// AWS allows tags to have empty values.
```

machine/v1beta1/types_awsprovider.go
```diff
@@ -454,21 +480,78 @@ type HostAffinity string
...
+
 // DedicatedHost represents the configuration for the usage of dedicated host.
+// +kubebuilder:validation:XValidation:rule="self.allocationStrategy == 'UserProvided' ? has(self.id) : true",message="id is required when allocationStrategy is UserProvided"
+// +kubebuilder:validation:XValidation:rule="has(self.id) ? self.allocationStrategy == 'UserProvided' : true",message="id is only allowed when allocationStrategy is UserProvided"
+// +kubebuilder:validation:XValidation:rule="has(self.dynamicHostAllocation) ? self.allocationStrategy == 'Dynamic' : true",message="dynamicHostAllocation is only allowed when allocationStrategy is Dynamic"
+// +union
 type DedicatedHost struct {
+	// allocationStrategy specifies if the dedicated host will be provided by the admin through the id field or if the host will be dynamically allocated.
+	// Valid values are UserProvided and Dynamic.
+	// This field is optional and defaults to "UserProvided".
```
saschagrunert: nit:
```suggestion
	// When omitted, the value defaults to "UserProvided", which requires the id field to be set.
```

machine/v1beta1/types_awsprovider.go
```diff
@@ -454,21 +480,78 @@ type HostAffinity string
...
+	// +default="UserProvided"
+	AllocationStrategy *AllocationStrategy `json:"allocationStrategy,omitempty"`
+
 	// id identifies the AWS Dedicated Host on which the instance must run.
 	// The value must start with "h-" followed by either 8 or 17 lowercase hexadecimal characters (0-9 and a-f).
 	// The use of 8 lowercase hexadecimal characters is for older legacy hosts that may not have been migrated to newer format.
 	// Must be either 10 or 19 characters in length.
-	// +kubebuilder:validation:XValidation:rule="self.matches('^h-([0-9a-f]{8}|[0-9a-f]{17})$')",message="hostID must start with 'h-' followed by either 8 or 17 lowercase hexadecimal characters (0-9 and a-f)"
+	// This field is required when allocationStrategy is UserProvided, and forbidden when allocationStrategy is Dynamic.
+	// When omitted, allocationStrategy must be set to Dynamic to enable automatic host allocation.
```
saschagrunert: nity nit:

```suggestion
	// When omitted with allocationStrategy set to Dynamic, the platform will dynamically allocate a dedicated host.
	// When omitted with allocationStrategy set to UserProvided (or omitted/default), validation will fail as the id is required for user-provided hosts.
```

machine/v1beta1/types_awsprovider.go
```diff
@@ -407,6 +413,26 @@ type AWSMachineProviderStatus struct {
...
+	// +optional
+	DedicatedHost *DedicatedHostStatus `json:"dedicatedHost,omitempty"`
+}
+
+// DedicatedHostStatus defines the observed state of a dynamically allocated dedicated host
+// associated with an AWSMachine. This struct is used to track the ID of the dedicated host.
+// +kubebuilder:validation:MinProperties=1
+type DedicatedHostStatus struct {
+	// id tracks the dynamically allocated dedicated host ID.
+	// This field is populated when allocationStrategy is Dynamic (with or without DynamicHostAllocation).
```
saschagrunert, line 428:
```suggestion
	// This field is populated when allocationStrategy is Dynamic (with or without DynamicHostAllocation).
	// When omitted, this indicates that the dedicated host has not yet been allocated, or allocation is in progress.
```

saschagrunert:
I see @JoelSpeed as approver, PTAL too. CI issues seem to be unrelated.

machine/v1beta1/types_awsprovider.go
```diff
@@ -331,9 +331,15 @@ type Filter struct {
 
 // TagSpecification is the name/value pair for a tag
 type TagSpecification struct {
-	// name of the tag
+	// name of the tag.
```
JoelSpeed, line 334: Do we currently validate any pattern for these? I would assume AWS has some sort of valid character set that we would have to observe?

machine/v1beta1/types_awsprovider.go
```diff
@@ -407,6 +413,26 @@ type AWSMachineProviderStatus struct {
...
+type DedicatedHostStatus struct {
+	// id tracks the dynamically allocated dedicated host ID.
+	// This field is populated when allocationStrategy is Dynamic (with or without DynamicHostAllocation).
+	// The value must start with "h-" followed by either 8 or 17 lowercase hexadecimal characters (0-9 and a-f).
+	// The use of 8 lowercase hexadecimal characters is for older legacy hosts that may not have been migrated to newer format.
+	// Must be either 10 or 19 characters in length.
+	// +kubebuilder:validation:XValidation:rule="self.matches('^h-([0-9a-f]{8}|[0-9a-f]{17})$')",message="id must start with 'h-' followed by either 8 or 17 lowercase hexadecimal characters (0-9 and a-f)"
+	// +kubebuilder:validation:MinLength=10
+	// +kubebuilder:validation:MaxLength=19
+	// +optional
```
JoelSpeed: Why not make this required?

machine/v1beta1/types_awsprovider.go
```diff
@@ -454,21 +480,78 @@ type HostAffinity string
...
+	// AllocationStrategyUserProvided specifies that the system should assign instances to a user-provided dedicated host.
+	AllocationStrategyUserProvided AllocationStrategy = "UserProvided"
+
+	// AllocationStrategyDynamic specifies that the system should dynamically allocate a dedicated host for instances.
+	AllocationStrategyDynamic AllocationStrategy = "Dynamic"
+)
+
 // DedicatedHost represents the configuration for the usage of dedicated host.
+// +kubebuilder:validation:XValidation:rule="self.allocationStrategy == 'UserProvided' ? has(self.id) : true",message="id is required when allocationStrategy is UserProvided"
+// +kubebuilder:validation:XValidation:rule="has(self.id) ? self.allocationStrategy == 'UserProvided' : true",message="id is only allowed when allocationStrategy is UserProvided"
```
JoelSpeed: Nit: We normally have these as a single rule
```suggestion
// +kubebuilder:validation:XValidation:rule="self.allocationStrategy == 'UserProvided' ? has(self.id) : !has(self.id)",message="id is required when allocationStrategy is UserProvided, and forbidden otherwise"
```

machine/v1beta1/types_awsprovider.go
```diff
@@ -454,21 +480,78 @@ type HostAffinity string
...
+// +kubebuilder:validation:XValidation:rule="has(self.dynamicHostAllocation) ? self.allocationStrategy == 'Dynamic' : true",message="dynamicHostAllocation is only allowed when allocationStrategy is Dynamic"
+// +union
 type DedicatedHost struct {
+	// allocationStrategy specifies if the dedicated host will be provided by the admin through the id field or if the host will be dynamically allocated.
+	// Valid values are UserProvided and Dynamic.
+	// This field is optional and defaults to "UserProvided".
+	// When AllocationStrategy is set to UserProvided, an ID of the dedicated host to assign must be provided.
+	// When AllocationStrategy is set to Dynamic, a dedicated host will be allocated and used to assign instances.
+	// When AllocationStrategy is set to Dynamic, and DynamicHostAllocation is configured, a dedicated host will be allocated and the tags in DynamicHostAllocation will be assigned to that host.
```
JoelSpeed: Nit, please use the JSON version of field names in docs, so that it appears as users would be familiar 
```suggestion
	// When allocationStrategy is set to UserProvided, an id of the dedicated host to assign must be provided.
	// When allocationStrategy is set to Dynamic, a dedicated host will be allocated and used to assign instances.
	// When allocationStrategy is set to Dynamic, and dynamicHostAllocation is configured, a dedicated host will be allocated and the tags in dynamicHostAllocation will be assigned to that host.
```

machine/v1beta1/types_awsprovider.go
```diff
@@ -454,21 +480,78 @@ type HostAffinity string
...
+	// +unionDiscriminator
+	// +default="UserProvided"
+	AllocationStrategy *AllocationStrategy `json:"allocationStrategy,omitempty"`
+
 	// id identifies the AWS Dedicated Host on which the instance must run.
 	// The value must start with "h-" followed by either 8 or 17 lowercase hexadecimal characters (0-9 and a-f).
 	// The use of 8 lowercase hexadecimal characters is for older legacy hosts that may not have been migrated to newer format.
 	// Must be either 10 or 19 characters in length.
-	// +kubebuilder:validation:XValidation:rule="self.matches('^h-([0-9a-f]{8}|[0-9a-f]{17})$')",message="hostID must start with 'h-' followed by either 8 or 17 lowercase hexadecimal characters (0-9 and a-f)"
+	// This field is required when allocationStrategy is UserProvided, and forbidden when allocationStrategy is Dynamic.
```
JoelSpeed: Forbidden otherwise basically right? If we added an extra value in the enum, it would also be forbidden most likely?

machine/v1beta1/types_awsprovider.go
```diff
@@ -454,21 +480,78 @@ type HostAffinity string
...
+	AllocationStrategyUserProvided AllocationStrategy = "UserProvided"
+
+	// AllocationStrategyDynamic specifies that the system should dynamically allocate a dedicated host for instances.
+	AllocationStrategyDynamic AllocationStrategy = "Dynamic"
+)
+
 // DedicatedHost represents the configuration for the usage of dedicated host.
+// +kubebuilder:validation:XValidation:rule="self.allocationStrategy == 'UserProvided' ? has(self.id) : true",message="id is required when allocationStrategy is UserProvided"
+// +kubebuilder:validation:XValidation:rule="has(self.id) ? self.allocationStrategy == 'UserProvided' : true",message="id is only allowed when allocationStrategy is UserProvided"
+// +kubebuilder:validation:XValidation:rule="has(self.dynamicHostAllocation) ? self.allocationStrategy == 'Dynamic' : true",message="dynamicHostAllocation is only allowed when allocationStrategy is Dynamic"
```
JoelSpeed, line 506: But is optional? ie not required when allocationStrategy is Dynamic

machine/v1beta1/types_awsprovider.go
```diff
@@ -454,21 +480,78 @@ type HostAffinity string
...
+// +kubebuilder:validation:MinProperties=1
+type DynamicHostAllocationSpec struct {
+	// tags specifies a set of key-value pairs to apply to the allocated dedicated host.
+	// When omitted, no additional user-defined tags will be applied to the allocated host.
+	// A maximum of 50 tags can be specified.
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=50
+	// +listType=map
+	// +listMapKey=name
+	// +optional
```
JoelSpeed, line 555: Why not make this required?

machine/v1beta1/types_awsprovider.go
```diff
@@ -454,21 +480,78 @@ type HostAffinity string
...
+	Tags []*TagSpecification `json:"tags,omitempty"`
```
JoelSpeed: This is what you need to be able to distinguish between no opinion and a deliberately empty list
```suggestion
	Tags *[]TagSpecification `json:"tags,omitempty"`
```

machine/v1beta1/types_awsprovider.go
vr4manta: the existing logic i found for instances (which i am using same func for dedicated hosts)  states not sensitive and just removes duplicates.  Online i found:

```
Key Tag Character & Usage Rules:
Valid Characters: A-Z, a-z, 0-9, space, and _ . : / = + - @.
Case Sensitivity: Yes (e.g., Project and project are distinct).
Prefix Restrictions: Tag keys cannot start with aws:.
Length Limits: Maximum key length is 128 Unicode characters; maximum value length is 256.
Allowed Special Characters: Hyphen (-), Underscore (_), Colon (:), Forward Slash (/), Equal sign (=), Plus (+), Period (.), At sign (@).
Prohibited Characters: Commas are generally not permitted in tag values.
Maximum Tags: Up to 50 tags per resource. 
```

machine/v1beta1/types_awsprovider.go
vr4manta: I was keeping this consistent with upstream version: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/26c35866af2d601760438bd177e10dbfd359b111/api/v1beta2/awsmachine_types.go#L473

I can make this required if you think it should be.  I am not set on this.

machine/v1beta1/types_awsprovider.go
vr4manta: Good catch. I think this was a miss on my part.

machine/v1beta1/types_awsprovider.go
vr4manta: correct.

machine/v1beta1/types_awsprovider.go
vr4manta: in the future, if we add any additonal fields, we may allow something such as host type or something like that.  If that is the case, we may not want to force tags to be required.

machine/v1beta1/types_awsprovider.go
vr4manta: I'm confused what this is in reference to.  In previous comment you wanted me to consolidate the first two validations.  That is done and the `, and forbidden otherwise` was added.  Is this in relation to the last validation?  dynamicHostAllocation is required when `Dynamic` strategy is used but forbidden otherwise.  Here is my change for this:
```
// DedicatedHost represents the configuration for the usage of dedicated host.
// +kubebuilder:validation:XValidation:rule="self.allocationStrategy == 'UserProvided' ? has(self.id) : !has(self.id)",message="id is required when allocationStrategy is UserProvided, and forbidden otherwise"
// +kubebuilder:validation:XValidation:rule="self.allocationStrategy == 'Dynamic' ? has(self.dynamicHostAllocation) : !has(self.dynamicHostAllocation)",message="dynamicHostAllocation is required when allocationStrategy is Dynamic, and forbidden otherwise"
// +union
```

machine/v1beta1/types_awsprovider.go
JoelSpeed: We probably want to document that requirement here at some point, not a blocker for this PR though so don't feel a need to update it

machine/v1beta1/types_awsprovider.go
JoelSpeed: Yeah I think this should be required. I don't know why it isn't upstream

JoelSpeed: You won't need the MinProperties on the struct once you've made that change

machine/v1beta1/types_awsprovider.go
JoelSpeed: Ok cool, it was the third I was picking up on. It seemed curious to me the way it was validated but your update makes much more sense

machine/v1beta1/types_awsprovider.go
```diff
@@ -407,6 +414,26 @@ type AWSMachineProviderStatus struct {
...
+	// +optional
+	DedicatedHost *DedicatedHostStatus `json:"dedicatedHost,omitempty"`
+}
+
+// DedicatedHostStatus defines the observed state of a dynamically allocated dedicated host
+// associated with an AWSMachine. This struct is used to track the ID of the dedicated host.
+type DedicatedHostStatus struct {
+	// id tracks the dynamically allocated dedicated host ID.
+	// This field is populated when allocationStrategy is Dynamic (with or without DynamicHostAllocation).
+	// When omitted, this indicates that the dedicated host has not yet been allocated, or allocation is in progress.
```
JoelSpeed: This line needs to move to L419 ish as part of the description for the parent struct

JoelSpeed:
/lgtm
/override ci/prow/lint

Linter isn't smart enough to allow pointer to slice at the moment, even though it's valid

vr4manta:
/verified by @vr4manta using manual testing with other PRs
Ran tests with images generated by this PR and other PRs at the top.  Installed and verified dynamic dedicated host was allocated and assigned.

vr4manta:
@JoelSpeed the override did not carry over to merge process.  Can you reapply?

PR #2651, opened 2026-01-16
NO-ISSUE: Fixes KubeletConfig API doc, adds tests and kubebuilder validation

ngopalak-redhat:
This PR improves the documentation for the KubeletConfigSpec API to make it more accurate and user-friendly. 

Fixes the doc: https://docs.redhat.com/en/documentation/openshift_container_platform/4.20/html/machine_apis/kubeletconfig-machineconfiguration-openshift-io-v1#spec-4

It also adds test coverage similar to the pattern established in https://github.com/openshift/api/pull/2370.

AutoSizingReserved will default to true from 4.21 onwards. The default value and the description was not documented until now.

**Why is this a breaking change?**
- The additional kubebuilder validation makes this a breaking change.
- If the user provides tlsSecurityProfile with just modern or custom fields but omits the type field, the validation passes because !has(self.type) evaluates to true.

**Test Coverage**

  Created test file: machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml


ngopalak-redhat:
@coderabbitai review

payload-manifests/crds/0000_80_machine-config_01_kubeletconfigs.crd.yaml
```diff
@@ -109,9 +117,9 @@ spec:
                 x-kubernetes-map-type: atomic
               tlsSecurityProfile:
                 description: |-
-                  If unset, the default is based on the apiservers.config.openshift.io/cluster resource.
-                  Note that only Old and Intermediate profiles are currently supported, and
-                  the maximum available minTLSVersion is VersionTLS12.
+                  tlsSecurityProfile configures TLS settings for the kubelet.
+                  If unset, defaults to the apiservers.config.openshift.io/cluster resource.
+                  Only Old and Intermediate profiles are supported; maximum minTLSVersion is VersionTLS12.
                 properties:
```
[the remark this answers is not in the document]
ngopalak-redhat: I'm not adding validation now as this is a doc only PR. we can take this up as a new PR

ngopalak-redhat:
cc: @haircommander @sairameshv

machineconfiguration/v1/zz_generated.crd-manifests/0000_80_machine-config_01_kubeletconfigs.crd.yaml
```diff
@@ -47,22 +47,30 @@ spec:
             description: spec contains the desired kubelet configuration.
             properties:
               autoSizingReserved:
+                description: |-
+                  autoSizingReserved controls whether system-reserved CPU and memory are automatically
```
sairameshv, line 51: Do you think we should mention this is applicable to worker nodes only?

machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml
```diff
@@ -0,0 +1,276 @@
...
+          logLevel: 4
+    - name: Should be able to omit logLevel
+      initial: |
+        apiVersion: machineconfiguration.openshift.io/v1
+        kind: KubeletConfig
+        spec: {}
+      expected: |
+        apiVersion: machineconfiguration.openshift.io/v1
+        kind: KubeletConfig
+        spec: {}
```
sairameshv: Based on https://github.com/openshift/api/pull/2651/changes#diff-96f8b17cfbe8b29ea3c9111c566dd910b5d5bf9095c87564f3fda5831c02eb35R67, shouldn't the expected log level be 2?

machineconfiguration/v1/types.go
```diff
@@ -737,27 +737,33 @@ type KubeletConfig struct {
...
+	// OpenShift release to find valid kubelet configuration options.
 	// +optional
 	KubeletConfig *runtime.RawExtension `json:"kubeletConfig,omitempty"`
 
-	// If unset, the default is based on the apiservers.config.openshift.io/cluster resource.
-	// Note that only Old and Intermediate profiles are currently supported, and
-	// the maximum available minTLSVersion is VersionTLS12.
+	// tlsSecurityProfile configures TLS settings for the kubelet.
+	// If unset, defaults to the apiservers.config.openshift.io/cluster resource.
+	// Only Old and Intermediate profiles are supported; maximum minTLSVersion is VersionTLS12.
```
saschagrunert: All documented constraints must be enforced through validation. For example:
```go
	// +kubebuilder:validation:XValidation:rule="!has(self.type) || self.type == 'Old' || self.type == 'Intermediate'",message="only Old and Intermediate TLS profiles are supported for kubelet; Modern and Custom profiles are not supported"
```

machineconfiguration/v1/types.go
```diff
@@ -737,27 +737,33 @@ type KubeletConfig struct {
...
+// KubeletConfigSpec configures the kubelet running on cluster nodes.
 type KubeletConfigSpec struct {
+	// autoSizingReserved controls whether system-reserved CPU and memory are automatically
+	// calculated based on each node's installed capacity. When enabled, prevents node failure
+	// from resource starvation of system components (kubelet, CRI-O) without manual configuration.
+	// When unset or true, automatic sizing is enabled. Set to false to disable and use manual settings.
 	// +optional
 	AutoSizingReserved *bool `json:"autoSizingReserved,omitempty"`
+	// logLevel sets the kubelet log verbosity. Higher values increase log detail.
+	// Defaults to 2 if unset.
```
saschagrunert: We should add validation bounds when they're documented.

```suggestion
	// logLevel sets the kubelet log verbosity, controlling the amount of detail in kubelet logs.
	// Valid values range from 0 (minimal logging) to 10 (maximum verbosity with trace-level detail).
	// Higher log levels may impact node performance. Defaults to 2 if unset.
	// +kubebuilder:validation:Minimum=0
	// +kubebuilder:validation:Maximum=10
```

machineconfiguration/v1/types.go
```diff
@@ -737,27 +737,33 @@ type KubeletConfig struct {
...
 	// +optional
 	LogLevel *int32 `json:"logLevel,omitempty"`
 
-	// machineConfigPoolSelector selects which pools the KubeletConfig shoud apply to.
-	// A nil selector will result in no pools being selected.
+	// machineConfigPoolSelector selects which pools the KubeletConfig should apply to.
+	// A nil selector results in no pools being selected.
```
saschagrunert: Tell the user what behavioral changes to expect from the cluster. The current text explains the selector behavior but not the cluster-level observable outcome.

We could improve this by
```suggestion
	// A nil selector results in no pools being selected, meaning this kubelet configuration
	// will not be applied to any nodes in the cluster.
```

machineconfiguration/v1/zz_generated.crd-manifests/0000_80_machine-config_01_kubeletconfigs.crd.yaml
ngopalak-redhat: Thanks, I have fixed now.

machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml
ngopalak-redhat: I updated the types.go doc. May be it makes more sense now.

machineconfiguration/v1/types.go
ngopalak-redhat: Thanks. Updated

machineconfiguration/v1/types.go
ngopalak-redhat: Thanks, Updated

machineconfiguration/v1/types.go
ngopalak-redhat: Thanks, Updated

machineconfiguration/v1/types.go
```diff
@@ -737,27 +737,39 @@ type KubeletConfig struct {
...
 
-	// If unset, the default is based on the apiservers.config.openshift.io/cluster resource.
-	// Note that only Old and Intermediate profiles are currently supported, and
-	// the maximum available minTLSVersion is VersionTLS12.
+	// tlsSecurityProfile configures TLS settings for the kubelet.
+	// If unset, defaults to the apiservers.config.openshift.io/cluster resource.
+	// Only Old and Intermediate profiles are supported; maximum minTLSVersion is VersionTLS12.
+	// +kubebuilder:validation:XValidation:rule="!has(self.type) || self.type == 'Old' || self.type == 'Intermediate'",message="only Old and Intermediate TLS profiles are supported for kubelet"
 	// +optional
 	TLSSecurityProfile *configv1.TLSSecurityProfile `json:"tlsSecurityProfile,omitempty"`
```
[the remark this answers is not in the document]
ngopalak-redhat: The reviewer requested me to add this rule as part of the documentation update.

ngopalak-redhat:
> Thank you for the PR, I have a bunch of improvements sggested.
> 
> Please note that we need to add integration tests for the new validations.

@saschagrunert I made the code changes as suggested. I also added the validations and tests to machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml.

Are these the integration tests you were referring to?

machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml
```diff
@@ -0,0 +1,380 @@
...
+      initial: |
+        apiVersion: machineconfiguration.openshift.io/v1
+        kind: KubeletConfig
+        spec:
+          logLevel: 4
+      expected: |
+        apiVersion: machineconfiguration.openshift.io/v1
+        kind: KubeletConfig
+        spec:
+          logLevel: 4
```
saschagrunert: Why should 2 succeed but 4 fail? If that's not possible by logic then I'd say we can remove those duplicate test paths. I guess we can just test for min and max values.

machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml
```diff
@@ -0,0 +1,380 @@
...
+      updated: |
+        apiVersion: machineconfiguration.openshift.io/v1
+        kind: KubeletConfig
+        spec:
+          autoSizingReserved: true
+      expected: |
+        apiVersion: machineconfiguration.openshift.io/v1
+        kind: KubeletConfig
+        spec:
+          autoSizingReserved: true
```
saschagrunert: Testing boolean flips doesn't validate any logic. Let's remove them, they don't test any validation logic, immutability, or constraints.

machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml
```diff
@@ -0,0 +1,380 @@
...
+      updated: |
+        apiVersion: machineconfiguration.openshift.io/v1
+        kind: KubeletConfig
+        spec:
+          logLevel: 4
+      expected: |
+        apiVersion: machineconfiguration.openshift.io/v1
+        kind: KubeletConfig
+        spec:
+          logLevel: 4
```
saschagrunert: Testing basic updates without validation logic, this is already covered by onCreate tests. Let's remove it.

machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml
```diff
@@ -0,0 +1,380 @@
...
+          machineConfigPoolSelector:
+            matchLabels:
+              pools.operator.machineconfiguration.openshift.io/master: ""
+      expected: |
+        apiVersion: machineconfiguration.openshift.io/v1
+        kind: KubeletConfig
+        spec:
+          machineConfigPoolSelector:
+            matchLabels:
+              pools.operator.machineconfiguration.openshift.io/master: ""
```
saschagrunert: Let's remove this test as well, it's just testing basic CRUD.

machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml
```diff
@@ -0,0 +1,380 @@
...
+          kubeletConfig:
+            maxPods: 250
+      expected: |
+        apiVersion: machineconfiguration.openshift.io/v1
+        kind: KubeletConfig
+        spec:
+          autoSizingReserved: true
+          logLevel: 4
+          kubeletConfig:
+            maxPods: 250
```
saschagrunert: I'd say we can also remove those, they have no validation logic and are redundant with onCreate.

machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml
```diff
@@ -0,0 +1,380 @@
...
+          machineConfigPoolSelector:
+            matchLabels:
+              pools.operator.machineconfiguration.openshift.io/master: ""
+      expected: |
+        apiVersion: machineconfiguration.openshift.io/v1
+        kind: KubeletConfig
+        spec:
+          machineConfigPoolSelector:
+            matchLabels:
+              pools.operator.machineconfiguration.openshift.io/master: ""
```
saschagrunert: This is redundant with the test above, they test the same validation but with a different label.

machineconfiguration/v1/types.go
```diff
@@ -737,27 +737,39 @@ type KubeletConfig struct {
...
+	// from resource starvation of system components (kubelet, CRI-O) without manual configuration.
+	// When unset, defaults to true for worker nodes and false for control plane nodes.
+	// Set to false to disable and use manual settings.
 	// +optional
 	AutoSizingReserved *bool `json:"autoSizingReserved,omitempty"`
+	// logLevel sets the kubelet log verbosity, controlling the amount of detail in kubelet logs.
+	// Valid values range from 0 (minimal logging) to 10 (maximum verbosity with trace-level detail).
+	// Higher log levels may impact node performance. If not set, logLevel of 2 is used.
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=10
```
saschagrunert, line 755: Looks like we have not enforced this in production clusters? Hm, then it would be a breaking change and we can't apply it now.

machineconfiguration/v1/types.go
```diff
@@ -737,27 +737,39 @@ type KubeletConfig struct {
...
+	// autoSizingReserved controls whether system-reserved CPU and memory are automatically
+	// calculated based on each node's installed capacity. When enabled, prevents node failure
+	// from resource starvation of system components (kubelet, CRI-O) without manual configuration.
+	// When unset, defaults to true for worker nodes and false for control plane nodes.
+	// Set to false to disable and use manual settings.
 	// +optional
 	AutoSizingReserved *bool `json:"autoSizingReserved,omitempty"`
+	// logLevel sets the kubelet log verbosity, controlling the amount of detail in kubelet logs.
+	// Valid values range from 0 (minimal logging) to 10 (maximum verbosity with trace-level detail).
+	// Higher log levels may impact node performance. If not set, logLevel of 2 is used.
```
saschagrunert: [Per convention](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#defaulting), configuration APIs should:
  - Use "When omitted" language
  - Indicate defaults are "subject to change over time"
  - State the current default

 ```suggestion
  // When omitted, the platform chooses a reasonable default, which is subject to change over time.
  // The current default is 2 (standard informational logging).
```

File: machineconfiguration/v1/types.go [hunk 16]
```diff
@@ -737,27 +737,39 @@ type KubeletConfig struct {
...
+	// OpenShift release to find valid kubelet configuration options.
 	// +optional
 	KubeletConfig *runtime.RawExtension `json:"kubeletConfig,omitempty"`
 
-	// If unset, the default is based on the apiservers.config.openshift.io/cluster resource.
-	// Note that only Old and Intermediate profiles are currently supported, and
-	// the maximum available minTLSVersion is VersionTLS12.
+	// tlsSecurityProfile configures TLS settings for the kubelet.
+	// If unset, defaults to the apiservers.config.openshift.io/cluster resource.
+	// Only Old and Intermediate profiles are supported; maximum minTLSVersion is VersionTLS12.
```
saschagrunert: Consider making the XValidation requirement more explicit for end users:

```suggestion
  // tlsSecurityProfile configures TLS settings for the kubelet.
  // When omitted, the TLS configuration defaults to the value from apiservers.config.openshift.io/cluster.
  // When specified, the type field is required and must be set to either "Old" or "Intermediate".
  // Modern and Custom TLS profiles are not supported for kubelet; maximum minTLSVersion is VersionTLS12.
```

machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml
ngopalak-redhat: Thanks, I have removed the duplicate test.

machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml
ngopalak-redhat: Thanks, I have removed the duplicate test.

machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml
ngopalak-redhat: Thanks, I have removed the duplicate test.

machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml
ngopalak-redhat: Thanks, I have removed the duplicate test.

machineconfiguration/v1/types.go
ngopalak-redhat: Made this changes as suggested

machineconfiguration/v1/types.go
ngopalak-redhat: Made changes as suggested

machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml
ngopalak-redhat: I have retained one of the update tests to make sure it works

machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml
ngopalak-redhat: Done

machineconfiguration/v1/types.go
ngopalak-redhat: This is a good opportunity to fix this. I understand it's a breaking change from an API perspective, but it's already validated in the MCO: https://github.com/openshift/machine-config-operator/blob/main/pkg/controller/kubelet-config/helpers.go#L353
We are effectively adding a first line of defense to reject values < 0 or > 10 immediately.

ngopalak-redhat:
@saschagrunert Thanks for the review. I have addressed your comments. Could you take another look? I also replied to the specific discussion here: https://github.com/openshift/api/pull/2651#discussion_r2726866363

machineconfiguration/v1/types.go
saschagrunert: Sounds good, but two things to clarify:

- Does the cluster degrade when this is being enforced now and users put in an invalid value?
- Let's compare the scenarios of putting an invalid value in a running cluster. What are the error messages and is the user feedback now more or less useful?

machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml
saschagrunert, line 1: There are still many redundant CRUD tests part of this file, which don't validate anything.

  1. Lines 7-37: autoSizingReserved true/false/omitted - Basic CRUD, no validation
  2. Lines 40-50: logLevel = 4 - Not a boundary, no validation value
  3. Lines 51-59: logLevel omitted - Redundant
  4. Lines 98-136: machineConfigPoolSelector worker/master/omitted - No validation logic
  5. Lines 139-160: kubeletConfig with maxPods/omitted - Basic CRUD
  6. Lines 163-188: tlsSecurityProfile Old/Intermediate - Redundant (both valid values, no validation difference)
  7. Lines 220-228: tlsSecurityProfile omitted - Redundant
  8. Lines 231-253: Combined fields - No validation logic
  9. Lines 257-272: Update autoSizingReserved - No validation logic
  10. Lines 275-288: Remove logLevel - Redundant with onCreate omit

Tests that should be kept (validation logic):
  1. Lines 60-70: logLevel = 0 (minimum boundary)
  2. Lines 71-81: logLevel = 10 (maximum boundary)
  3. Lines 82-88: logLevel = -1 (reject below min)
  4. Lines 89-95: logLevel = 11 (reject above max)
  5. Lines 189-196: tlsSecurityProfile Modern (CEL XValidation rejection)
  6. Lines 197-208: tlsSecurityProfile Custom (CEL XValidation rejection)
  7. Lines 209-219: tlsSecurityProfile without type (CEL XValidation rejection)

machineconfiguration/v1/types.go
ngopalak-redhat: The user experience with this change is better.
Before:
The kubeletconfig is accepted and describe output will show
```
Last Transition Time:  2026-02-12T15:09:41Z
    Message:               Error: KubeletConfig's LogLevel is not valid [1,10]: 0xc0013b3f48
    Status:                True
    Type:                  Failure
```
After:
```
oc apply -f ~/redhat/autoreserved/kubeletloglevel.yaml
The KubeletConfig "80-custom" is invalid: spec.logLevel: Invalid value: -3: spec.logLevel in body should be greater than or equal to 0
```

machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml
ngopalak-redhat: @saschagrunert Sorry for the slow turnaround here.

Addressed all comments.

machineconfiguration/v1/types.go
```diff
@@ -737,27 +737,41 @@ type KubeletConfig struct {
...
 	KubeletConfig *runtime.RawExtension `json:"kubeletConfig,omitempty"`
 
-	// If unset, the default is based on the apiservers.config.openshift.io/cluster resource.
-	// Note that only Old and Intermediate profiles are currently supported, and
-	// the maximum available minTLSVersion is VersionTLS12.
+	// tlsSecurityProfile configures TLS settings for the kubelet.
+	// When omitted, the TLS configuration defaults to the value from apiservers.config.openshift.io/cluster.
+	// When specified, the type field is required and must be set to either "Old" or "Intermediate".
+	// Modern and Custom TLS profiles are not supported for kubelet; maximum minTLSVersion is VersionTLS12.
+	// +kubebuilder:validation:XValidation:rule="has(self.type) && (self.type == 'Old' || self.type == 'Intermediate')",message="only Old and Intermediate TLS profiles are supported for kubelet"
```
saschagrunert: The test file must include ratcheting tests using `initialCRDPatches` to demonstrate that existing objects with previously valid values (e.g., type: Custom or type: Modern) can still be updated without the new validation blocking unrelated field changes.

Something like:
```
    onUpdate:
      - name: Should allow updating other fields with a previously valid Custom TLS profile persisted
        initialCRDPatches:
        - op: remove
          path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/tlsSecurityProfile/x-kubernetes-validations
        initial: |
          kind: KubeletConfig
          spec:
            tlsSecurityProfile:
              type: Custom
              custom:
                ciphers:
                  - ECDHE-ECDSA-AES128-GCM-SHA256
                minTLSVersion: VersionTLS12
        updated: |
          kind: KubeletConfig
          spec:
            logLevel: 5
            tlsSecurityProfile:
              type: Custom
              custom:
                ciphers:
                  - ECDHE-ECDSA-AES128-GCM-SHA256
                minTLSVersion: VersionTLS12
        expected: |
          kind: KubeletConfig
          spec:
            logLevel: 5
            tlsSecurityProfile:
              type: Custom
              custom:
                ciphers:
                  - ECDHE-ECDSA-AES128-GCM-SHA256
                minTLSVersion: VersionTLS12
```

machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml
```diff
@@ -0,0 +1,75 @@
...
+      initial: |
+        apiVersion: machineconfiguration.openshift.io/v1
+        kind: KubeletConfig
+        spec:
+          tlsSecurityProfile:
+            custom:
+              ciphers:
+                - ECDHE-ECDSA-AES128-GCM-SHA256
+              minTLSVersion: VersionTLS12
+      expectedError: "only Old and Intermediate TLS profiles are supported for kubelet"
```
saschagrunert: nit: let's keep the trailing newline here

machineconfiguration/v1/types.go
saschagrunert: Let's also add a ratcheting test like outlined below.

machineconfiguration/v1/types.go
saschagrunert: Would this help make the rule less strict?

```suggestion
	// +kubebuilder:validation:XValidation:rule="!has(self.type) || self.type == 'Old' || self.type == 'Intermediate'",message="only Old and Intermediate TLS profiles are supported for kubelet"
```

The alternative rule allows type to be omitted (which preserves backward compatibility with existing objects that may have `tlsSecurityProfile` set without explicitly setting type), while still rejecting `Modern` and `Custom` when `type` IS specified. This significantly reduces the breaking change surface. If the intent is to REQUIRE type. This significantly reduces the breaking change surface. If the intent is to REQUIRE `type` to be set, that should be a separate validation rule with a clearer error message (e.g., "type is required when tlsSecurityProfile is specified").

machineconfiguration/v1/types.go
```diff
@@ -737,27 +737,41 @@ type KubeletConfig struct {
...
 }
 
-// KubeletConfigSpec defines the desired state of KubeletConfig
+// KubeletConfigSpec configures the kubelet running on cluster nodes.
 type KubeletConfigSpec struct {
+	// autoSizingReserved controls whether system-reserved CPU and memory are automatically
+	// calculated based on each node's installed capacity. When enabled, prevents node failure
+	// from resource starvation of system components (kubelet, CRI-O) without manual configuration.
+	// When unset, defaults to true for worker nodes and false for control plane nodes.
+	// Set to false to disable and use manual settings.
```
saschagrunert:
```suggestion
	// When omitted, this means the user has no opinion and the platform is left to choose a
	// reasonable default, which is subject to change over time. The current default is true
	// for worker nodes and false for control plane nodes.
	// When set to false, automatic resource reservation is disabled and manual settings must
	// be configured.
```

Per OpenShift API conventions for configuration APIs (cluster-scoped singletons), defaults should not be stated as permanent facts. The "no opinion" / "subject to change" phrasing communicates that the platform owns the default behavior. Additionally, API documentation should describe behavior declaratively ("When set to false, X happens") rather than imperatively ("Set to false to do X").

machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml
saschagrunert: Let's add a positive test case, like:

```
      - name: Should accept tlsSecurityProfile with Old type
        initial: |
          kind: KubeletConfig
          spec:
            tlsSecurityProfile:
              type: Old
        expected: |
          kind: KubeletConfig
          spec:
            tlsSecurityProfile:
              type: Old
      - name: Should accept tlsSecurityProfile with Intermediate type
        initial: |
          kind: KubeletConfig
          spec:
            tlsSecurityProfile:
              type: Intermediate
        expected: |
          kind: KubeletConfig
          spec:
            tlsSecurityProfile:
              type: Intermediate
```

ngopalak-redhat: Thanks, I added them

machineconfiguration/v1/types.go
ngopalak-redhat: Added ratcheting test for log level. Let me know if changes/updates are needed to this test.

machineconfiguration/v1/types.go
ngopalak-redhat: Thanks, Made this change.

machineconfiguration/v1/types.go
ngopalak-redhat: I adopted this change with a slight change to the comment also to make sure that it reflects what the code does.

machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml
ngopalak-redhat: Done

machineconfiguration/v1/types.go
ngopalak-redhat: thanks, Added the test

machineconfiguration/v1/types.go
```diff
@@ -737,27 +737,42 @@ type KubeletConfig struct {
 	Status KubeletConfigStatus `json:"status"`
 }
 
-// KubeletConfigSpec defines the desired state of KubeletConfig
+// KubeletConfigSpec configures the kubelet running on cluster nodes.
 type KubeletConfigSpec struct {
+	// autoSizingReserved controls whether system-reserved CPU and memory are automatically
+	// calculated based on each node's installed capacity. When enabled, prevents node failure
```
saschagrunert: "When enabled" is ambiguous for a boolean field. Please say "When set to true, this prevents node failure..." to make the mapping between value and outcome explicit.

```suggestion
	// calculated based on each node's installed capacity. When set to true, this prevents node failure
```

machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml
```diff
@@ -0,0 +1,195 @@
...
+        spec:
+          logLevel: 3
+          tlsSecurityProfile:
+            type: Modern
+    - name: Should allow updating other fields with a previously valid logLevel outside current range
+      initialCRDPatches:
+        - op: remove
+          path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/logLevel/maximum
+        - op: remove
+          path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/logLevel/minimum
```
saschagrunert, line 51: Please add a test showing the migration path from an invalid to a valid value. This confirms that users with previously restricted configurations can update to a compliant value:

```yaml
    - name: Should allow migrating from Custom to Old TLS profile
      initialCRDPatches:
        - op: remove
          path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/tlsSecurityProfile/x-kubernetes-validations
      initial: |
        kind: KubeletConfig
        spec:
          tlsSecurityProfile:
            type: Custom
            custom:
              ciphers:
                - ECDHE-ECDSA-AES128-GCM-SHA256
              minTLSVersion: VersionTLS12
      updated: |
        kind: KubeletConfig
        spec:
          tlsSecurityProfile:
            type: Old
      expected: |
        kind: KubeletConfig
        spec:
          tlsSecurityProfile:
            type: Old
```

machineconfiguration/v1/types.go
```diff
@@ -737,27 +737,42 @@ type KubeletConfig struct {
...
 	// +optional
 	LogLevel *int32 `json:"logLevel,omitempty"`
 
-	// machineConfigPoolSelector selects which pools the KubeletConfig shoud apply to.
-	// A nil selector will result in no pools being selected.
+	// machineConfigPoolSelector selects which pools the KubeletConfig should apply to.
+	// A nil selector results in no pools being selected, meaning this kubelet configuration
+	// will not be applied to any nodes in the cluster.
 	// +optional
 	MachineConfigPoolSelector *metav1.LabelSelector `json:"machineConfigPoolSelector,omitempty"`
```
everettraven, line 763: What does `nil` selector mean here? omitted, set to `{}`, both?

machineconfiguration/v1/types.go
```diff
@@ -737,27 +737,42 @@ type KubeletConfig struct {
...
 	// +optional
 	KubeletConfig *runtime.RawExtension `json:"kubeletConfig,omitempty"`
 
-	// If unset, the default is based on the apiservers.config.openshift.io/cluster resource.
-	// Note that only Old and Intermediate profiles are currently supported, and
-	// the maximum available minTLSVersion is VersionTLS12.
+	// tlsSecurityProfile configures TLS settings for the kubelet.
+	// When omitted, the TLS configuration defaults to the value from apiservers.config.openshift.io/cluster.
+	// When specified, the type field can be set to either "Old" or "Intermediate", or omitted for backward compatibility.
+	// Modern and Custom TLS profiles are not supported for kubelet; maximum minTLSVersion is VersionTLS12.
```
everettraven: If Modern and Custom TLS profiles aren't supported for kublet, what happens if the `apiservers.config.openshift.io/cluster` resource configures those?

machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml
everettraven: Just a note that the example @saschagrunert shared is for the tlsSecurityProfile field, but I'd expect the same test for all fields that had new constraints added to them (i.e logLevel).

machineconfiguration/v1/types.go
everettraven: The linked validation in the MCO states that the minimum is `1` - should this match that?

ngopalak-redhat:
/hold 
There is a bug fix for the log-level in the MCO: https://github.com/openshift/machine-config-operator/pull/5757

machineconfiguration/v1/types.go
ngopalak-redhat: https://github.com/openshift/machine-config-operator/pull/5757 - Fixed and merged

machineconfiguration/v1/types.go
ngopalak-redhat: Fixed the comment now

machineconfiguration/v1/tests/kubeletconfigs.machineconfiguration.openshift.io/KubeletConfigSpec.yaml
ngopalak-redhat: Fixed it

machineconfiguration/v1/types.go
ngopalak-redhat: Now these are supported

machineconfiguration/v1/types.go
ngopalak-redhat: Fixed

ngopalak-redhat:
@saschagrunert @everettraven 
- Recently TLS 1.3 is supported and hence `Modern` and `Custom` profiles also work. So discussed with @bitoku and @asahay19 and made the validation simpler. So removed the racketing test. Note that I'm not calling out that cri-o will inherit this value as its an implementation detail.
- The log level is fixed in MCO (https://github.com/openshift/machine-config-operator/pull/5757), so the validation stays the same here

Please review

payload-manifests/crds/0000_80_machine-config_01_kubeletconfigs.crd.yaml
```diff
@@ -109,9 +124,10 @@ spec:
                 x-kubernetes-map-type: atomic
               tlsSecurityProfile:
                 description: |-
-                  If unset, the default is based on the apiservers.config.openshift.io/cluster resource.
-                  Note that only Old and Intermediate profiles are currently supported, and
-                  the maximum available minTLSVersion is VersionTLS12.
+                  tlsSecurityProfile configures TLS settings for the kubelet.
+                  When omitted, the TLS configuration defaults to the value from apiservers.config.openshift.io/cluster.
+                  When specified, the type field can be set to either "Old", "Intermediate", "Modern", "Custom" or omitted for backward compatibility.
+                  maximum minTLSVersion is VersionTLS13.
```
bitoku: It doesn't make sense to mention minTLSVersion here, because users are supposed to specify type, not minTLSVersion.
In each type section, minTLSVersion is mentioned already.

For reference https://github.com/openshift/api/blob/8182f057c06bf4100e26052a3ef89de08950a5ed/payload-manifests/crds/0000_10_config-operator_01_apiservers-Default.crd.yaml#L227-L231 .

machineconfiguration/v1/types.go
```diff
@@ -737,27 +737,42 @@ type KubeletConfig struct {
...
 	KubeletConfig *runtime.RawExtension `json:"kubeletConfig,omitempty"`
 
-	// If unset, the default is based on the apiservers.config.openshift.io/cluster resource.
-	// Note that only Old and Intermediate profiles are currently supported, and
-	// the maximum available minTLSVersion is VersionTLS12.
+	// tlsSecurityProfile configures TLS settings for the kubelet.
+	// When omitted, the TLS configuration defaults to the value from apiservers.config.openshift.io/cluster.
+	// When specified, the type field can be set to either "Old", "Intermediate", "Modern", "Custom" or omitted for backward compatibility.
+	// maximum minTLSVersion is VersionTLS13.
+	// +kubebuilder:validation:XValidation:rule="!has(self.type) || self.type == 'Old' || self.type == 'Intermediate' || self.type == 'Modern' || self.type == 'Custom'",message="only Old, Intermediate, Modern and Custom TLS profiles are supported for kubelet"
```
everettraven: If you are allowing the type to be any of the types that are valid for the `TLSSecurityProfile.Type` field you probably don't need this validation - the type alias for that field already enforces an enum constraint with these same values.

ngopalak-redhat: @everettraven Let's say there is a addition to `TLSSecurityProfile.Type` enum which kubelet doesn't support yet, then we need this validation right?

payload-manifests/crds/0000_80_machine-config_01_kubeletconfigs.crd.yaml
ngopalak-redhat: Fixed

machineconfiguration/v1/types.go
everettraven: In theory, yes. In practice, this is unlikely. Any changes to the `TLSSecurityProfile` API will need to go through the feature promotion process which will include ensuring that all components either support the changes or they have been explicitly updated to not handle them prior to the promotion of the field.

ngopalak-redhat: Thanks, I have removed the kubebuilder validation that I added in this PR.

ngopalak-redhat:
/hold 
Holding this until https://github.com/openshift/machine-config-operator/pull/5672 is merged

@saschagrunert @everettraven I have address all the comments. Can you review?

everettraven:
The validation change is intentional to also add validation that the MCO does earlier in the user workflow for admission time enforcement. Because anything outside of what was enforced by these new validations was never a valid configuration and there are ratcheting tests added this is OK.

bitoku:
/hold
There's internal discussion about kubeconfig API, hold until it's settled.

ngopalak-redhat:
/verified by ci
Tests are included in the PR

ngopalak-redhat:
@everettraven Looks like you need to override again. Its not merging.
https://github.com/openshift/api/pull/2651#issuecomment-4141913921

JoelSpeed:
Based on https://github.com/openshift/api/pull/2651#issuecomment-4141913921

PR #2654, opened 2026-01-16
OCPCLOUD-2710: Extend AWS metadata service options in MAPA

RadekManak:
Add httpEndpoint, httpPutResponseHopLimit, and instanceMetadataTags fields to MetadataServiceOptions to bring parity with CAPA. This allows users to configure additional aspects of the AWS Instance Metadata Service (IMDS) on machine instances, improving backwards compatibility and reducing conversion errors when these options are used in CAPA.

New fields:
- httpEndpoint: Enable/disable HTTP metadata endpoint (Enabled/Disabled)
- httpPutResponseHopLimit: Configure metadata token hop limit (1-64)
- instanceMetadataTags: Enable/disable instance tag access from metadata


machine/v1beta1/types_awsprovider.go
```diff
@@ -265,18 +265,72 @@ const (
...
 // MetadataServiceOptions defines the options available to a user when configuring
 // Instance Metadata Service (IMDS) Options.
 type MetadataServiceOptions struct {
+	// httpEndpoint enables or disables the HTTP metadata endpoint on your instances.
+	// Valid values are "Enabled" and "Disabled".
+	// When set to Enabled, the HTTP metadata endpoint is accessible.
+	// When set to Disabled, you cannot access your instance metadata.
+	// When omitted, the value is determined by account-level settings in the AWS Region, or the AWS service default if not configured at the account level.
+	// The typical AWS service default is Enabled.
+	// +kubebuilder:validation:Enum=Enabled;Disabled
```
yuqi-zhang: nit: since we are defining a new type (HTTPEndpointState), we should probably validate on the type itself, so we don't have to duplicate validation if other places want to use it.

That said since this is the only occurrence, and we do similar validation for e.g. MetadataServiceAuthentication, we can keep it for consistency purposes.

machine/v1beta1/types_awsprovider.go
```diff
@@ -265,18 +265,72 @@ const (
...
+	// +optional
+	HTTPEndpoint *HTTPEndpointState `json:"httpEndpoint,omitempty"`
```
yuqi-zhang, line 312: I believe that all 3 new fields do not need to be pointers. There's no need to distinguish between an omittited and empty value (they cannot be empty anyways due to the validations)

yuqi-zhang (commented):
Generally seems fine, a couple of comments inline

Also to double check, there is no other reference to these objects in the api repo correct? i.e. the actual CRDs would be deployed via e.g. https://github.com/openshift/machine-api-provider-aws/pull/170 ?

RadekManak:
/remove-lifecycle stale

Rebased onto current master.

@yuqi-zhang on pointers: these live in `machine/v1beta1/types_awsprovider.go`, which is go-validated (providerSpec), not OpenAPI/CRD-validated. That file is excluded from the default kubeapilinter `optionalfields` config (`WhenRequired` in `.golangci.yaml`) and instead linted via `.golangci.go-validated.yaml` with `optionalfields.pointers.preference: Always` (+ `SuggestFix`). That’s why these new optional fields are `*T` with `omitempty`.

Your note matches the default `WhenRequired` policy (Enum means zero isn’t admissible, so pointer isn’t required for serialization). That conflicts with the providerSpec `Always` policy.

Could you advise which way we should go?
1) Keep pointers to satisfy the go-validated providerSpec linter as configured today, or
2) Switch to non-pointers as you suggest — and if so, should `.golangci.go-validated.yaml` be relaxed (e.g. `WhenRequired`) for these providerSpec files?

Enum validation is on the types themselves (your other nit).
CRDs for these fields ship via consumers once api is vendored (https://github.com/openshift/machine-api-provider-aws/pull/170).

yuqi-zhang:
Ack, thanks for the update, will check with the API team for the suggested behaviour here

machine/v1beta1/types_awsprovider.go
```diff
@@ -265,18 +266,90 @@ const (
...
+	// When omitted, the value is determined by account-level settings in the AWS Region, or the AWS service default if not configured at the account level.
+	// The typical AWS service default is Enabled.
+	// +optional
+	HTTPEndpoint *HTTPEndpointState `json:"httpEndpoint,omitempty"`
+
+	// httpProtocolIPv6 enables or disables the IPv6 endpoint for the instance metadata service.
+	// Valid values are "Enabled" and "Disabled".
+	// When set to Enabled, the IPv6 endpoint for the instance metadata service is accessible.
+	// When set to Disabled, the IPv6 endpoint for the instance metadata service is not accessible.
+	// When omitted, the AWS default is used and the IPv6 endpoint is disabled.
```
yuqi-zhang, line 318: Based on https://github.com/openshift/machine-api-provider-aws/pull/170/changes#diff-c5079a7c5d2916c3661d0c44d1267050b3496c40b4530363d3825d66af581b6fR833 and the implementation, it seems that IPV6 would be enabled in dual-stack cases. If that's the correct behaviour we should probably explicitly document that here.

yuqi-zhang (commented):
On the previous point, the current behaviour is correct. Native fields validated in go that is optional should be a pointer. No need to change that, apologies for the confusion. One question inline

PR #2661, opened 2026-01-19
OCPBUGS-63219: Add protocol to AWS NLB parameters

gcs278:
## Summary

_Note: Design Doc is here: https://docs.google.com/document/d/14UVY8U-exch30-tYloX_-DMhM7hJX8226VA2q7bTJeo/edit?tab=t.0_

Add a `protocol` field to `AWSNetworkLoadBalancerParameters` to control whether the NLB uses native client IP preservation or PROXY protocol v2. This reuses the existing `IngressControllerProtocol` type (`TCP`/`PROXY`) for consistency with HostNetwork, NodePort, Private, and IBM load balancer strategies.

When set to `TCP`, the NLB uses AWS's native client IP preservation, which may cause hairpin connection failures for internal load balancers when connections are made from pods to router pods on the same node.

When set to `PROXY`, the NLB disables native client IP preservation and uses PROXY protocol v2. This allows backends to receive the original client IP via PROXY protocol headers while avoiding hairpin connection failures.

## Why a new API field?

AWS NLBs have `preserve_client_ip.enabled=true` by default, which causes hairpin connection failures on internal NLBs when a pod sends traffic through the NLB and it routes back to the same node. The fix is to disable native client IP preservation and use PROXY protocol v2 instead — the same mechanism CLBs already use.

However, simply changing the default NLB behavior is not safe for existing clusters. Custom load balancers, middleware, or third-party solutions (firewalls, geo-blocking, audit logging) that connect to or inspect traffic at the router may not understand PROXY protocol headers. An API field gives cluster admins an escape hatch to revert to the old behavior if something breaks.

## Placement

The field is under `networkLoadBalancer` because this problem is NLB-specific. CLBs don't have client IP preservation as an option — disabling PROXY protocol on a CLB means losing client IP visibility entirely, which is a different decision with different implications. There is no current need to control PROXY protocol on CLBs — CLBs are not affected by the hairpin bug. This field exists to solve an NLB-specific problem.

## Mutability

The field is mutable and applied in-place — changing it does not require deleting or recreating the load balancer Service. There is a brief window during the transition where the NLB attribute change and router rollout occur independently, which may cause connection failures. This is documented in the godoc.

## Why controller-managed defaults (not CRD defaults)?

CRD-level defaulting (`+kubebuilder:default`) is not sufficient here because `networkLoadBalancer` is an optional pointer struct — if the user doesn't specify it, the API server won't instantiate it to apply nested defaults. Controller-managed defaults are needed to cover all cases.

The ingress operator defaults `protocol` to `PROXY` only when creating new IngressControllers. Existing IngressControllers that were created before this field existed will have the field omitted, and the operator treats omitted as `TCP` — preserving their current behavior with no change on upgrade.

The godoc uses the standard conventions wording to reserve the right to change the default in the future: "When omitted, this means the user has no opinion and the value is left to the platform to choose a good default, which is subject to change over time. The current default is PROXY."

Implementation PR: https://github.com/openshift/cluster-ingress-operator/pull/1426


gcs278:
I've marked as draft as I'm still working through the approach to fixing this bug.

/test all

operator/v1/types_ingress.go
```diff
@@ -898,8 +898,45 @@ type AWSNetworkLoadBalancerParameters struct {
 	// +kubebuilder:validation:XValidation:rule=`self.all(x, self.exists_one(y, x == y))`,message="eipAllocations cannot contain duplicates"
 	// +kubebuilder:validation:MaxItems=10
 	EIPAllocations []EIPAllocation `json:"eipAllocations"`
+
+	// clientIPPreservationMode specifies how client IP addresses are
+	// preserved by the load balancer.
+	//
+	// Valid values are "Native" and "ProxyProtocol".
+	//
+	// When set to "Native", the NLB uses AWS's native client IP preservation,
```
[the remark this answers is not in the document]
gcs278: @CodeRabbit can you check again? I update the PR description

gcs278:
@JoelSpeed I was discussing with @Miciah and I'm going to write up a quick design doc for this API and the defaulting behavioral changes so we can discuss in more detail.

Feel free to review/comment in the meantime though.

gcs278:
/hold

While we discuss design details. But feel free to still review, or review my [doc](https://docs.google.com/document/d/14UVY8U-exch30-tYloX_-DMhM7hJX8226VA2q7bTJeo/edit?tab=t.0).

yuqi-zhang (commented):
Let me know when the design is finalized - just some initial thoughts:
1. the API definition itself looks fine, although simple, it would be best to have a few unit tests attached
2. since this is introducing a new field to a stable API, the general expectation is that you would have a featuregate backing this, however
3. it seems that you are doing this as a bugfix patch. Are you planning on backporting this to previous versions? We generally do not backport new API fields. Based on your description
```
AWS NLBs have preserve_client_ip.enabled=true by default, which causes hairpin connection failures on internal NLBs when a pod sends traffic through the NLB and it routes back to the same node. The fix is to disable native client IP preservation and use PROXY protocol v2 instead — the same mechanism CLBs already use.
```
I would expect previous versioned clusters running into this to find a way to do this via some manual method, if possible, instead of backporting a new API field and fix. This can serve as the fix for future clusters.

operator/v1/types_ingress.go
```diff
@@ -898,8 +898,47 @@ type AWSNetworkLoadBalancerParameters struct {
...
+	// which may cause hairpin connection failures for internal load balancers when
+	// connections are made from pods to router pods on the same node.
+	//
+	// When set to "ProxyProtocol", the NLB uses PROXY protocol v2 to preserve
+	// client IP addresses. This avoids hairpin connection failures.
+	//
+	// When omitted, this means the user has no opinion and the value is left
+	// to the platform to choose a good default, which is subject to change
```
mtulio: What about removing the "good" from this sentence?  :) if we considering the hairpin will always be there when "no proxy", and the current "good default" isn't so good as it cause hairpin connection failures, it may looks like the platform it's assuming good is a failure state.

```suggestion
	// to the platform to choose a default, which is subject to change
```

gcs278: Done.

gcs278:
@yuqi-zhang  - after a design discussion with @Miciah I have proposed a refactored API  - the behavior is the mostly the same but the names of the fields changed to be more consistent with other existing fields. Ready for another  review.

> the API definition itself looks fine, although simple, it would be best to have a few unit tests attached

Done

> it seems that you are doing this as a bugfix patch. Are you planning on backporting this to previous versions? We generally do not backport new API fields. Based on your description

I think backporting is likely needed. At the moment, there's no reasonable supported workaround for this bug. The only option is separating router pods from app pods using infra nodes, which adds cost for users. Or switch back to CLBs (also not very reasonable).

operator/v1/types_ingress.go
```diff
@@ -898,6 +898,37 @@ type AWSNetworkLoadBalancerParameters struct {
...
+	// When omitted, this means the user has no opinion and the value is
+	// left to the platform to choose a good default, which is subject to
+	// change over time. The current default is "PROXY".
+	//
+	// Note that changing this field may cause brief connection failures
+	// during the transition as the NLB attribute change and router rollout
+	// occur independently.
+	//
+	// +optional
+	Protocol IngressControllerProtocol `json:"protocol,omitempty"`
```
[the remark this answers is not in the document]
gcs278: I'll yield to @yuqi-zhang - we haven't been required to do a FG for bug fixes in the past, for example https://github.com/openshift/api/pull/2584.

gcs278:
@yuqi-zhang @Miciah ping on review for this API when you get a chance.

operator/v1/types_ingress.go
```diff
@@ -898,6 +898,37 @@ type AWSNetworkLoadBalancerParameters struct {
...
+	// When omitted, this means the user has no opinion and the value is
+	// left to the platform to choose a default, which is subject to
+	// change over time. The current default is "PROXY".
+	//
+	// Note that changing this field may cause brief connection failures
+	// during the transition as the NLB attribute change and router rollout
+	// occur independently.
+	//
+	// +optional
+	Protocol *IngressControllerProtocol `json:"protocol,omitempty"`
```
yuqi-zhang: This doesn't need to be a pointer, but otherwise this seems fine to me. Will defer to Joel/Bryce on the backportability, but should be safe.

gcs278: Yea i actually had it as a non-pointer value type, but the [linter](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_api/2661/pull-ci-openshift-api-master-lint/2060118424582885376) complained that optional fields should be pointers:
```
operator/v1/types_ingress.go:931:2: optionalfields: field AWSNetworkLoadBalancerParameters.Protocol has a valid zero value ("") and should be a pointer. (kubeapilinter)
	Protocol IngressControllerProtocol `json:"protocol,omitempty"`
	^
```

All of our other [protocol](https://github.com/openshift/api/blob/master/operator/v1/types_ingress.go#L800) fields are not pointers fwiw. Happy to change it back, but the linter would need overriding.

yuqi-zhang: Ah, right, the actual type has

```
+kubebuilder:validation:Enum="";TCP;PROXY
```

which is very awkward since the empty string is a valid choice. But the empty string is just the default in the parsing, right? i.e. there's 4 possible fields here:

1. omitted - equivalent to default, which is proxy
2. empty string - same as above
3. set to proxy - explicit
4. set to tcp - only actual different setting

Normally if this was a new field, we'd say to not allow empty string, since it's functionally equivalent to empty field. But the type is grandfathered in and is used in other fields, so we can't easily update that especially since you plan to backport. The other types should also be pointers technically. Maybe we just have this be a pointer and be fine with the diff to other fields.

(we should probably also be explicit what "" means, since it's not actually in the godoc. empty string and omitted field are not the same thing)

@JoelSpeed or @everettraven any suggestions on what the best approach is?

yuqi-zhang: We discussed this briefly and feel that a new type is probably better here with the right enum options, and may be better for backporting as well. Could you create a new type for this field?

gcs278: Ack done - added  a new type, removed `""` as an option. So it's just:

1. omitted - equivalent to default, which is proxy
2. set to proxy - explicit
3. set to tcp - only actual different setting

gcs278:
@yuqi-zhang Ready for your review again, thanks!

yuqi-zhang (commented):
Thanks for the update, lgtm, will mark as shadow review complete

gcs278:
@aswinsuryan is going to help out with reviewing this bug fix
/assign @aswinsuryan

operator/v1/types_ingresscontroller.go
```diff
@@ -898,8 +898,52 @@ type AWSNetworkLoadBalancerParameters struct {
...
+	// during the transition as the NLB attribute change and router rollout
+	// occur independently.
+	//
+	// +optional
+	Protocol NLBProtocol `json:"protocol,omitempty"`
 }
 
+// NLBProtocol specifies whether the AWS Network Load Balancer uses
+// PROXY protocol to forward connections to the IngressController.
+// +kubebuilder:validation:Enum=TCP;PROXY
```
JoelSpeed, line 935: Add `+enum` too please, helps complete the `openapi` schema which is used by apply clients

operator/v1/types_ingresscontroller.go
```diff
@@ -898,8 +898,52 @@ type AWSNetworkLoadBalancerParameters struct {
...
+	// obtain the original client IP. This avoids hairpin connection
+	// failures.
+	//
+	// The following values are valid for this field:
+	//
+	// * "TCP".
+	// * "PROXY".
+	//
+	// When omitted, this means the user has no opinion and the value is
+	// left to the platform to choose a default, which is subject to
```
JoelSpeed: Nit, purely for consistency with everywhere else we use this phrase
```suggestion
	// left to the platform to choose a reasonable default, which is subject to
```

JoelSpeed (commented):
Two nits, but otherwise this LGTM

operator/v1/types_ingresscontroller.go
gcs278: Done.

operator/v1/types_ingresscontroller.go
gcs278: Done.

melvinjoseph86:
will be checked along with https://github.com/openshift/cluster-ingress-operator/pull/1426
/verified later @mjoseph

PR #2664, opened 2026-01-22
CORS-4295: Add featuregates for installs with CAPI Machine Management

patrickdillon:
### **User description**
Adds separate featuregates for:
- Install with CAPI control-plane management
- Install with CAPI compute management

cc @damdo

___

### **PR Type**
Enhancement

___

### **Description**
- Add two new feature gates for Cluster API installer support

- ClusterAPIControlPlaneInstall for control-plane management

- ClusterAPIComputeInstall for compute node management

- Register gates in DevPreviewNoUpgrade and TechPreviewNoUpgrade modes

- Update all feature gate manifest files with new gates

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Feature Gate Definitions"] -->|"Add ClusterAPIControlPlaneInstall"| B["Control-Plane Management"]
  A -->|"Add ClusterAPIComputeInstall"| C["Compute Management"]
  B --> D["Update Manifest Files"]
  C --> D
  D --> E["8 Hypershift & SelfManagedHA Configs"]
```

<details><summary><h3>File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>features.go</strong><dd><code>Add two new Cluster API feature gates</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/openshift/api/pull/2664/files#diff-a8b6135d50534471326ea7bcd20e0f5eae25353f7788338060f718128a6a0b34">+22/-6</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Documentation</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>features.md</strong><dd><code>Document new Cluster API feature gates</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
[...]

PR #2668, opened 2026-01-24
STOR-2859: Add APIs for disabling force detach in KCM operator

dobsonj:
### **User description**
https://issues.redhat.com/browse/STOR-2859

This PR introduces an API to disable force detach in kube-controller-manager. The design is based on [this](https://redhat-internal.slack.com/archives/CKJR6200N/p1758559804038819) slack discussion (internal only). There is already a [KubeControllerManager](https://github.com/openshift/api/blob/master/operator/v1/types_kubecontrollermanager.go) operator type to configure cluster-kube-controller-manager-operator, and this PR introduces a ControllerManager config type to configure kube-controller-manager. This is approach is consistent with other workload components like [KubeAPIServer](https://github.com/openshift/api/blob/master/operator/v1/types_kubeapiserver.go) / [APIServer](https://github.com/openshift/api/blob/master/config/v1/types_apiserver.go) and [KubeScheduler](https://github.com/openshift/api/blob/master/operator/v1/types_scheduler.go) / [Scheduler](https://github.com/openshift/api/blob/master/config/v1/types_scheduling.go).

`controllerManager.spec.forceDetachOnTimeout` will be used to enable or disable force detach in KCM.

/cc @openshift/storage

___

### **PR Type**
Enhancement

___

### **Description**
- Add ControllerManager API for cluster-wide Kubernetes controller manager configuration

- Introduce ForceDetachOnTimeoutPolicy to control volume force detach behavior

- Generate deepcopy, swagger documentation, and OpenAPI schema definitions

- Create CustomResourceDefinition manifest for controllermanagers resource

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["ControllerManager API Type"] --> B["ForceDetachOnTimeoutPolicy"]
  A --> C["ControllerManagerSpec"]
  C --> E["forceDetachOnTimeout Field"]
  E --> F["Enabled/Disabled/Omitted"]
  A --> G["Generated Code"]
  G --> H["DeepCopy Functions"]
  G --> I["Swagger Documentation"]
  G --> J["OpenAPI Schemas"]
  A --> K["CRD Manifest"]
```

<details><summary><h3>File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>types_controllermanager.go</strong><dd><code>Define ControllerManager API types and policy</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

config/v1alpha1/types_controllermanager.go

<ul><li>Define new <code>ControllerManager</code> struct with metadata and spec <br>fields<br> <li> Create <code>ControllerManagerSpec</code> with <code>ForceDetachOnTimeoutPolicy</code> field<br> <li> Define <code>ForceDetachOnTimeoutPolicy</code> type with Enabled and Disabled <br>constants<br> <li> Create <code>ControllerManagerList</code> struct for list operations</ul>

</details>

  </td>
  <td><a href="https://github.com/openshift/api/pull/2668/files#diff-6e8b8b92301f100a683e7c5a4e43466d760ba2c66fdeec275dfeb3933b00da22">+72/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></td></tr><tr><td><strong>Code generation</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>zz_generated.deepcopy.go</strong><dd><code>Generate deepcopy functions for ControllerManager types</code>&nbsp; &nbsp; </dd></summary>
<hr>

config/v1alpha1/zz_generated.deepcopy.go

<ul><li>Generate DeepCopyInto, DeepCopy, and DeepCopyObject methods for <br>ControllerManager<br> <li> Generate deepcopy methods for ControllerManagerList<br> <li> Generate deepcopy methods for ControllerManagerSpec</ul>

</details>

  </td>
  <td><a href="https://github.com/openshift/api/pull/2668/files#diff-ad96d6336ccdfffb67dcf48a7d0251b23c3960b5ab0d9db307eef19ec822d6e9">+93/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>zz_generated.swagger_doc_generated.go</strong><dd><code>Generate swagger documentation for ControllerManager</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

config/v1alpha1/zz_generated.swagger_doc_generated.go

<ul><li>Add swagger documentation map for ControllerManager type<br> <li> Add swagger documentation map for ControllerManagerList type<br> <li> Add swagger documentation map for ControllerManagerSpec type<br> <li> Document field descriptions and compatibility level</ul>

</details>

  </td>
  <td><a href="https://github.com/openshift/api/pull/2668/files#diff-17eed2d93bd315f98604e27de33e8ccc07365919d1fedd2471ca6b7d995c6af2">+28/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>zz_generated.openapi.go</strong><dd><code>Generate OpenAPI schema definitions for ControllerManager</code></dd></summary>
<hr>

openapi/generated_openapi/zz_generated.openapi.go

<ul><li>Register OpenAPI schema definitions for ControllerManager, <br>ControllerManagerList, and ControllerManagerSpec<br> <li> Generate schema_openshift_api_config_v1_ControllerManager function <br>with properties and dependencies<br> <li> Generate schema_openshift_api_config_v1_ControllerManagerList function <br>with items array<br> <li> Generate schema_openshift_api_config_v1_ControllerManagerSpec schema functions</ul>

</details>

  </td>
  <td><a href="https://github.com/openshift/api/pull/2668/files#diff-6f26c8627c096ba26e30a864dbdb69e055980f8c8b704283d9a483fbb999404c">+135/-0</a>&nbsp; </td>

</tr>
</table></td></tr><tr><td><strong>Configuration</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>0000_10_config-operator_01_controllermanagers.crd.yaml</strong><dd><code>Create ControllerManager CustomResourceDefinition manifest</code></dd></summary>
<hr>

config/v1alpha1/zz_generated.crd-manifests/0000_10_config-operator_01_controllermanagers.crd.yaml

<ul><li>Create CustomResourceDefinition manifest for <br>controllermanagers.config.openshift.io<br> <li> Define spec with forceDetachOnTimeout property supporting Enabled, <br>Disabled, and empty string values<br> <li> Configure cluster-scoped resource<br> <li> Include API approval reference</ul>

</details>

  </td>
  <td><a href="https://github.com/openshift/api/pull/2668/files#diff-609c2341923c49bf7e193ec76c8a8d09759be7ddf2ae572d32ebe53f68d52db2">+73/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>zz_generated.featuregated-crd-manifests.yaml</strong><dd><code>Register ControllerManager in featuregated manifests</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

config/v1alpha1/zz_generated.featuregated-crd-manifests.yaml

<ul><li>Add controllermanagers.config.openshift.io entry to featuregated CRD <br>manifests<br> <li> Register metadata including operator name, run level, and scope <br>information</ul>

</details>

  </td>
  <td><a href="https://github.com/openshift/api/pull/2668/files#diff-aef74d5ff6f3a49078b7d76d30ad0fda029182e0823408a03a663fd545adf7b3">+22/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>AAA_ungated.yaml</strong><dd><code>Create ungated ControllerManager CRD manifest</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

config/v1alpha1/zz_generated.featuregated-crd-manifests/controllermanagers.config.openshift.io/AAA_ungated.yaml

<ul><li>Create ungated CustomResourceDefinition manifest for <br>controllermanagers resource<br> <li> Define identical schema to main CRD manifest with feature-gate <br>annotations<br> <li> Configure cluster scope resource</ul>

</details>

  </td>
  <td><a href="https://github.com/openshift/api/pull/2668/files#diff-7c96c8818bcc342ea8119d90b5881510204d2cf15040832e86647a19e9c0fe54">+74/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></td></tr></tbody></table>

</details>

___


config/v1alpha1/zz_generated.crd-manifests/0000_10_config-operator_01_controllermanagers.crd.yaml
```diff
@@ -0,0 +1,72 @@
...
+                default: Enabled
+                description: |-
+                  forceDetachOnTimeout expresses whether to allow kube-controller-manager
+                  to force detach volumes when unmount takes longer than the timeout.
+                  Valid values are Enabled and Disabled. If omitted, the default is Enabled.
+                enum:
+                - Enabled
+                - Disabled
+                type: string
+            type: object
```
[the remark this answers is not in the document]
dobsonj: This is required to pass the lint checks. Without it, lint fails with:
`config/v1/types_controllermanager.go:28:2: minlength: field ControllerManager.Spec type ControllerManagerSpec must have a minimum properties, add kubebuilder:validation:MinProperties marker (kubeapilinter)`

config/v1alpha1/zz_generated.crd-manifests/0000_10_config-operator_01_controllermanagers.crd.yaml
```diff
@@ -0,0 +1,72 @@
...
+          status:
+            description: status holds observed values from the cluster. They may not
+              be overridden.
+            minProperties: 1
+            type: object
```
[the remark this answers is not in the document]
dobsonj: This is required to pass the lint checks. Without it, lint fails with:
`config/v1/types_controllermanager.go:31:2: minlength: field ControllerManager.Status type ControllerManagerStatus must have a minimum properties, add kubebuilder:validation:MinProperties marker (kubeapilinter)`

dobsonj:
/cc @gnufied @ingvagabund 
for reviews

gnufied:
So, this will introduce a new `cluster` object in the OCP clusters? Have we tested any POC with this change?  

cc @JoelSpeed

dobsonj:
> So, this will introduce a new `cluster` object in the OCP clusters?

Yes

> Have we tested any POC with this change?

Not yet, I should have this soon.

JoelSpeed:
This will need an EP, to be behind a feature gate, and start life as v1alpha1 so that we can ensure we get the API right before we commit to a structure for v1

dobsonj:
> This will need an EP, to be behind a feature gate, and start life as v1alpha1 so that we can ensure we get the API right before we commit to a structure for v1

Ack, will do, thanks.

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,76 @@
...
+// +openshift:compatibility-gen:level=1
+type ControllerManagerList struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// metadata is the standard list's metadata.
+	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
+	metav1.ListMeta `json:"metadata"`
+
+	Items []ControllerManager `json:"items"`
+}
```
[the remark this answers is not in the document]
dobsonj: fixed

features.md
```diff
@@ -41,6 +41,7 @@
 | ClusterVersionOperatorConfiguration| | | <span style="background-color: #519450">Enabled</span> | <span style="background-color: #519450">Enabled</span> | | | <span style="background-color: #519450">Enabled</span> | <span style="background-color: #519450">Enabled</span>  |
 | ConfigurablePKI| | | <span style="background-color: #519450">Enabled</span> | <span style="background-color: #519450">Enabled</span> | | | <span style="background-color: #519450">Enabled</span> | <span style="background-color: #519450">Enabled</span>  |
 | DNSNameResolver| | | <span style="background-color: #519450">Enabled</span> | <span style="background-color: #519450">Enabled</span> | | | <span style="background-color: #519450">Enabled</span> | <span style="background-color: #519450">Enabled</span>  |
+| DisableForceDetachOnTimeout| | | <span style="background-color: #519450">Enabled</span> | <span style="background-color: #519450">Enabled</span> | | | <span style="background-color: #519450">Enabled</span> | <span style="background-color: #519450">Enabled</span>  |
```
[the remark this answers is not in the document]
dobsonj: This line is generated and consistent with the other feature gates, so I'm ignoring this feedback.

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,76 @@
...
+	Status ControllerManagerStatus `json:"status,omitzero"`
+}
+
+// ControllerManagerSpec defines the desired state of the Kubernetes controller manager
+// +kubebuilder:validation:MinProperties=1
+type ControllerManagerSpec struct {
+	// forceDetachOnTimeout expresses whether to allow kube-controller-manager
+	// to force detach volumes when unmount takes longer than the timeout.
+	// Valid values are Enabled and Disabled. If omitted, the default is Enabled.
+	// +default="Enabled"
```
saschagrunert: Per https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md

> With configuration APIs, we typically default fields within the controller and not within the API. This means that the platform has the ability to make changes to the defaults over time as we improve the capabilities of OpenShift.



```suggestion
	// forceDetachOnTimeout expresses whether to allow kube-controller-manager
	// to force detach volumes when the maximum unmount time is exceeded or when
	// a node is not healthy.
	// Valid values are "Enabled" and "Disabled".
	// When omitted, this means the user has no opinion and the platform is left
	// to choose a reasonable default, which is subject to change over time.
	// The current default is "Enabled".
```

I guess we can keep `+kubebuilder:validation:MinProperties=1` on `ControllerManagerSpec`, The user explicitly sets Enabled or Disabled.

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,76 @@
...
+	// +optional
+	ForceDetachOnTimeout ForceDetachOnTimeoutPolicy `json:"forceDetachOnTimeout,omitempty"`
+}
+
+// +kubebuilder:validation:Enum=Enabled;Disabled
```
saschagrunert, line 60:
```suggestion
// ForceDetachOnTimeoutPolicy describes the policy for force detaching volumes
// when the maximum unmount time is exceeded.
// Valid values are "Enabled" and "Disabled".
// +kubebuilder:validation:Enum=Enabled;Disabled
```

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,76 @@
+package v1alpha1
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ControllerManager holds cluster-wide config information to run the Kubernetes controller manager
+// and influence its placement decisions. The canonical name for this config is `cluster`.
```
saschagrunert: Looks like "influence its placement decisions" was copied from the Scheduler type description and doesn't apply to the controller manager, we should improve it a bit:


```suggestion
// ControllerManager holds cluster-wide configuration for the Kubernetes controller manager.
// The canonical name for this config is `cluster`.
```

features/features.go
```diff
@@ -978,4 +978,12 @@ var (
...
 						enableIn(configv1.DevPreviewNoUpgrade).
 						mustRegister()
+
+	FeatureGateDisableForceDetachOnTimeout = newFeatureGate("DisableForceDetachOnTimeout").
+						reportProblemsToJiraComponent("Storage").
+						contactPerson("jdobson").
+						productScope(ocpSpecific).
+						enhancementPR("https://github.com/openshift/enhancements/pull/1935").
+						enableIn(configv1.DevPreviewNoUpgrade, configv1.TechPreviewNoUpgrade).
+						mustRegister()
```
saschagrunert, line 1084: Note: the enhancement PR is still WIP.

config/v1alpha1/tests/controllermanagers.config.openshift.io/DisableForceDetachOnTimeout.yaml
```diff
@@ -0,0 +1,80 @@
...
+        kind: ControllerManager
+        spec:
+          forceDetachOnTimeout: Disabled
+    - name: Should reject unsupported value of forceDetachOnTimeout
+      initial: |
+        apiVersion: config.openshift.io/v1alpha1
+        kind: ControllerManager
+        spec:
+          forceDetachOnTimeout: INVALID
+      expectedError: "spec.forceDetachOnTimeout: Unsupported value: \"INVALID\": supported values: \"Enabled\", \"Disabled\""
```
saschagrunert: Those tests are just testing basic CRUD and can be omitted.

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,76 @@
...
+	ForceDetachOnTimeoutEnabled  ForceDetachOnTimeoutPolicy = "Enabled"
+	// ForceDetachOnTimeoutDisabled will prevent kube-controller-manager
+	// from force detaching volumes.
+	ForceDetachOnTimeoutDisabled ForceDetachOnTimeoutPolicy = "Disabled"
+)
+
+// ControllerManagerStatus defines the observed state of the Kubernetes controller manager
+// +kubebuilder:validation:MinProperties=1
+type ControllerManagerStatus struct {
+}
```
saschagrunert: `MinProperties=1` makes the `ControllerManagerStatus` unusable. We should either add a field or remove the `+kubebuilder:subresource:status` option from `ControllerManager`.

lance5890:
1. I'm very glad to see that we finally added the `controller manager` crd, 
2. However,  as far as I know, the [apiserver](https://github.com/openshift/api/blob/master/config/v1/types_apiserver.go) suppuort  configuration for both kube-apiserver and the openshift-apiserver;
3. The [Forced storage detach](https://kubernetes.io/docs/concepts/cluster-administration/node-shutdown/#storage-force-detach-on-timeout) seems  to be unique to the the kube-controller-manger(not for the openshift-controller-manager), should we add this in [kube-controller-manager](https://github.com/openshift/api/blob/master/payload-manifests/crds/0000_25_kube-controller-manager_01_kubecontrollermanagers.crd.yaml), and adjust this new feature in kcm-o should be fine?

saschagrunert:
@dobsonj do you mind a rebase?

features/features.go
dobsonj: I closed the old WIP PR and replaced it with https://github.com/openshift/enhancements/pull/2095

config/v1alpha1/tests/controllermanagers.config.openshift.io/DisableForceDetachOnTimeout.yaml
dobsonj: I think we want at least a couple of tests for the new object, but I dropped the tests that are essentially duplicates.

config/v1alpha1/types_controllermanager.go
dobsonj: I dropped the `MinProperties=1` line. What would be the purpose of removing the `+kubebuilder:subresource:status` option from `ControllerManager` though? Looking through a few existing objects, I'm not seeing where we've done that before, even for objects with empty status structs.

dobsonj:
> @dobsonj do you mind a rebase?

@saschagrunert Done, thanks for the review and suggestions! Sorry for letting the PR sit stale so long, we're trying to revive this work now for 5.1.

config/v1alpha1/types_controllermanager.go
dobsonj: And of course dropping `MinProperties=1` causes a linter error, is this why you suggest removing `+kubebuilder:subresource:status`?

dobsonj: I added a `conditions` field here instead, to avoid breaking with convention from dropping `+kubebuilder:subresource:status` and to avoid linter errors from dropping `MinProperties=1`.

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,93 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ControllerManager holds cluster-wide configuration for the Kubernetes controller manager.
+// The canonical name for this config is `cluster`.
+//
+// Compatibility level 4: No compatibility is provided, the API can change at any point for any reason. These capabilities should not be used by applications needing long term support.
+// +openshift:compatibility-gen:level=4
+// +openshift:api-approved.openshift.io=https://github.com/openshift/api/pull/2668
+// +openshift:file-pattern=cvoRunLevel=0000_10,operatorName=config-operator,operatorOrdering=01
+// +kubebuilder:object:root=true
```
saschagrunert, line 16: `ControllerManager` and `ControllerManagerList` are not added to `addKnownTypes` in `config/v1alpha1/register.go`. Any scheme-based client (generated clientset, informers, controller-runtime) will fail with `no kind "ControllerManager" is registered for version "config.openshift.io/v1alpha1"`.

```go
		&PKI{},
		&PKIList{},
		&ControllerManager{},
		&ControllerManagerList{},
```

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,93 @@
...
+// +kubebuilder:resource:path=controllermanagers,scope=Cluster
+// +kubebuilder:subresource:status
+// +kubebuilder:metadata:annotations=release.openshift.io/bootstrap-required=true
```
saschagrunert, line 19: Is `bootstrap-required` needed here? It makes the render command put the CRD into the bootkube manifests. Scheduler has it because the installer creates a Scheduler manifest during bootstrap, but nothing does that for ControllerManager, and the enhancement doesn't mention a bootstrap use. If there isn't one, I'd drop it.

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,93 @@
...
+// +openshift:enable:FeatureGate=DisableForceDetachOnTimeout
```
saschagrunert, line 20: The operator only reads `cluster`, so any other name would be accepted and then silently ignored. Can we enforce the name like `CRIOCredentialProviderConfig` does in the same group? The type comment on L10 could then say that the name must be `cluster`.

```suggestion
// +openshift:enable:FeatureGate=DisableForceDetachOnTimeout
// +kubebuilder:validation:XValidation:rule="self.metadata.name == 'cluster'",message="controllermanager is a singleton, .metadata.name must be 'cluster'"
```

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,93 @@
...
+// ControllerManagerSpec defines the desired state of the Kubernetes controller manager
+// +kubebuilder:validation:MinProperties=1
+type ControllerManagerSpec struct {
+	// forceDetachOnTimeout expresses whether to allow kube-controller-manager
+	// to force detach volumes when the maximum unmount time is exceeded or when
+	// a node is not healthy.
+	// Valid values are "Enabled" and "Disabled".
+	// When omitted, this means the user has no opinion and the platform is left
+	// to choose a reasonable default, which is subject to change over time.
+	// The current default is "Enabled".
```
saschagrunert, line 45: Two things:
- **Enum values:** the field comment should say what each value does. The constant comments don't show up in `oc explain`.
- **Trigger condition:** it's "and", not "or". Upstream force detaches only when the node is unhealthy *and* the unmount timeout (6 minutes) has expired: `forceDetach := !isHealthy && forceDetachTimeoutExpired` in the attach/detach reconciler.

It would also help to spell out what an admin has to do when this is `Disabled`.

```suggestion
	// forceDetachOnTimeout controls whether kube-controller-manager force detaches
	// volumes from a node that is not healthy once the volumes have not been
	// unmounted within the maximum unmount time (6 minutes).
	// Valid values are "Enabled" and "Disabled".
	// When set to "Enabled", volumes are force detached from unhealthy nodes after
	// the maximum unmount time, so that workloads using them can start on other nodes.
	// Force detaching a volume that is still in use by the node can corrupt its data.
	// When set to "Disabled", volumes are not force detached based on the maximum
	// unmount time. Volumes remain attached to an unhealthy node until it recovers,
	// or until the node is tainted with "node.kubernetes.io/out-of-service"
	// as part of the non-graceful node shutdown procedure.
	// When omitted, this means the user has no opinion and the platform is left
	// to choose a reasonable default, which is subject to change over time.
	// The current default is "Enabled".
```

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,93 @@
...
+	// +optional
+	ForceDetachOnTimeout ForceDetachOnTimeoutPolicy `json:"forceDetachOnTimeout,omitempty"`
+}
+
+// ForceDetachOnTimeoutPolicy describes the policy for force detaching volumes
+// when the maximum unmount time is exceeded.
+// Valid values are "Enabled" and "Disabled".
+// +kubebuilder:validation:Enum=Enabled;Disabled
```
saschagrunert, line 53: Without `+enum`, the generated OpenAPI schema has no enum values for this field.

```suggestion
// +enum
// +kubebuilder:validation:Enum=Enabled;Disabled
```

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,93 @@
...
+type ForceDetachOnTimeoutPolicy string
+
+const (
+	// ForceDetachOnTimeoutEnabled will allow kube-controller-manager to
+	// force detach volumes based on maximum unmount time and node status.
+	ForceDetachOnTimeoutEnabled  ForceDetachOnTimeoutPolicy = "Enabled"
+	// ForceDetachOnTimeoutDisabled will prevent kube-controller-manager
+	// from force detaching volumes.
+	ForceDetachOnTimeoutDisabled ForceDetachOnTimeoutPolicy = "Disabled"
```
saschagrunert, line 62: L59 isn't gofmt-clean (double space). Also, with `+enum` these comments are copied into the OpenAPI description as "Possible enum values". "will prevent kube-controller-manager from force detaching volumes" goes too far, because volumes on nodes with the `node.kubernetes.io/out-of-service` taint are still force detached.

```suggestion
	// ForceDetachOnTimeoutEnabled allows kube-controller-manager to force detach
	// volumes from unhealthy nodes once the maximum unmount time is exceeded.
	ForceDetachOnTimeoutEnabled ForceDetachOnTimeoutPolicy = "Enabled"
	// ForceDetachOnTimeoutDisabled prevents kube-controller-manager from force
	// detaching volumes based on the maximum unmount time.
	ForceDetachOnTimeoutDisabled ForceDetachOnTimeoutPolicy = "Disabled"
```

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+// ControllerManagerStatus defines the observed state of the Kubernetes controller manager
+// +kubebuilder:validation:MinProperties=1
+type ControllerManagerStatus struct {
+	// conditions represent the latest available observations of the configuration state.
+	// When omitted, it indicates that no conditions have been reported yet.
+	// The maximum number of conditions is 16.
+	// When set, at least one condition must be present.
+	// Conditions are stored as a map keyed by condition type, ensuring uniqueness.
```
saschagrunert, line 72: Following up on the earlier status thread: adding `conditions` fixes the linter, but no condition types are defined and the enhancement doesn't say the operator reports anything. That leaves status with no contract.

Either option works for me:
- **Real condition:** if the operator will report something useful, for example whether the current spec has been rolled out to kube-controller-manager (the rollout is async), document the expected condition types like `CRIOCredentialProviderConfigStatus` does. Add that to the enhancement too. Also add omitted behavior to the `status` field (L31), e.g. "When omitted, the status has not yet been reported by the cluster-kube-controller-manager-operator."
- **No status:** otherwise, remove the `Status` field and `+kubebuilder:subresource:status`. Then `MinProperties` on status isn't needed either, and lint stays clean.

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,93 @@
...
+	// +kubebuilder:validation:MaxItems=16
+	// +kubebuilder:validation:MinItems=1
+	// +listType=map
+	// +listMapKey=type
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// Compatibility level 4: No compatibility is provided, the API can change at any point for any reason. These capabilities should not be used by applications needing long term support.
```
saschagrunert, line 83: Every other list type in this group has a description:

```suggestion
// ControllerManagerList is a collection of ControllerManager resources.
//
// Compatibility level 4: No compatibility is provided, the API can change at any point for any reason. These capabilities should not be used by applications needing long term support.
```

config/v1alpha1/zz_generated.crd-manifests/0000_10_config-operator_01_controllermanagers.crd.yaml
```diff
@@ -0,0 +1,144 @@
...
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    api-approved.openshift.io: https://github.com/openshift/api/pull/2668
+    api.openshift.io/merged-by-featuregates: "true"
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    release.openshift.io/bootstrap-required: "true"
+    release.openshift.io/feature-set: CustomNoUpgrade,DevPreviewNoUpgrade
+  name: controllermanagers.config.openshift.io
```
saschagrunert, line 11: This CRD isn't copied into `payload-manifests/crds`, so it won't be in the release payload. `Dockerfile.ocp` only ships that directory, and `config/v1alpha1` CRDs are listed individually in `hack/update-payload-crds.sh`:

```bash
    config/v1alpha1/zz_generated.crd-manifests/0000_10_config-operator_01_controllermanagers*.crd.yaml
```

Then run `make update`.

config/v1alpha1/tests/controllermanagers.config.openshift.io/DisableForceDetachOnTimeout.yaml
```diff
@@ -0,0 +1,42 @@
...
+      updated: |
+        apiVersion: config.openshift.io/v1alpha1
+        kind: ControllerManager
+        spec:
+          forceDetachOnTimeout: Disabled
+      expected: |
+        apiVersion: config.openshift.io/v1alpha1
+        kind: ControllerManager
+        spec:
+          forceDetachOnTimeout: Disabled
```
saschagrunert, line 42: With the singleton rule, every object needs `metadata.name: cluster`, since the suite otherwise names objects `test-xxxxx`. A test for the rule itself would be good too:

```suggestion
  onCreate:
    - name: Should be able to create with force detach disabled
      initial: |
        apiVersion: config.openshift.io/v1alpha1
        kind: ControllerManager
        metadata:
          name: cluster
        spec:
          forceDetachOnTimeout: Disabled
      expected: |
        apiVersion: config.openshift.io/v1alpha1
        kind: ControllerManager
        metadata:
          name: cluster
        spec:
          forceDetachOnTimeout: Disabled
    - name: Should reject a ControllerManager not named cluster
      initial: |
        apiVersion: config.openshift.io/v1alpha1
        kind: ControllerManager
        metadata:
          name: foo
        spec:
          forceDetachOnTimeout: Disabled
      expectedError: "Invalid value: \"object\": controllermanager is a singleton, .metadata.name must be 'cluster'"
    - name: Should reject unsupported value of forceDetachOnTimeout
      initial: |
        apiVersion: config.openshift.io/v1alpha1
        kind: ControllerManager
        metadata:
          name: cluster
        spec:
          forceDetachOnTimeout: INVALID
      expectedError: "spec.forceDetachOnTimeout: Unsupported value: \"INVALID\": supported values: \"Enabled\", \"Disabled\""
  onUpdate:
    - name: Enabled to Disabled
      initial: |
        apiVersion: config.openshift.io/v1alpha1
        kind: ControllerManager
        metadata:
          name: cluster
        spec:
          forceDetachOnTimeout: Enabled
      updated: |
        apiVersion: config.openshift.io/v1alpha1
        kind: ControllerManager
        metadata:
          name: cluster
        spec:
          forceDetachOnTimeout: Disabled
      expected: |
        apiVersion: config.openshift.io/v1alpha1
        kind: ControllerManager
        metadata:
          name: cluster
        spec:
          forceDetachOnTimeout: Disabled
```

config/v1alpha1/types_controllermanager.go
dobsonj: Thanks, updated.

config/v1alpha1/types_controllermanager.go
dobsonj: Not needed, I removed this line.

config/v1alpha1/zz_generated.crd-manifests/0000_10_config-operator_01_controllermanagers.crd.yaml
dobsonj: Thanks for pointing this out, updated.

config/v1alpha1/types_controllermanager.go
dobsonj: That makes sense, I added a `Progressing` condition to the API and enhancement to track the roll out to KCM.

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,115 @@
...
+// The resource is a singleton named "cluster".
+//
+// Compatibility level 4: No compatibility is provided, the API can change at any point for any reason. These capabilities should not be used by applications needing long term support.
+// +openshift:compatibility-gen:level=4
+// +openshift:api-approved.openshift.io=https://github.com/openshift/api/pull/2668
+// +openshift:file-pattern=cvoRunLevel=0000_10,operatorName=config-operator,operatorOrdering=01
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=controllermanagers,scope=Cluster
+// +kubebuilder:subresource:status
+// +openshift:enable:FeatureGate=DisableForceDetachOnTimeout
```
saschagrunert: The gate is on the root type, so the whole `ControllerManager` API only exists when this storage-specific gate is on. Any future unrelated field in `ControllerManagerSpec` inherits that.

Consider a gate named for the API on the type, and `DisableForceDetachOnTimeout` on the field. Cheap to rename while this is DevPreview only.

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,115 @@
...
+
+	// metadata is the standard object's metadata.
+	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitzero"`
+	// spec holds user settable values for configuration
+	// +required
+	Spec ControllerManagerSpec `json:"spec,omitzero"`
+	// status holds observed values from the cluster. They may not be overridden.
+	// When omitted, the status has not yet been reported by the cluster-kube-controller-manager-operator.
```
saschagrunert: HyperShift has no `cluster-kube-controller-manager-operator` (the EP routes through `hostedCluster.spec.configuration.controllerManager`), but this CRD ships there too.

```suggestion
	// status holds observed values from the cluster. They may not be overridden.
	// When omitted, the status has not yet been reported by the cluster.
```

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,115 @@
...
+	// +optional
+	Status ControllerManagerStatus `json:"status,omitzero"`
+}
+
+// ControllerManagerSpec defines the desired state of the Kubernetes controller manager
+// +kubebuilder:validation:MinProperties=1
```
saschagrunert, line 35: `MinProperties=1` plus a single optional field makes `forceDetachOnTimeout` required in practice:

```
spec: {}  ->  spec in body should have at least 1 properties
```

That contradicts L51-53: the only way to express "no opinion" is to not create the CR at all. Either keep the marker and say so in the `spec` godoc on L28, or make the field `+required` and drop the "when omitted" paragraph.

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,115 @@
...
+	// ForceDetachOnTimeoutEnabled allows kube-controller-manager to force detach
+	// volumes from unhealthy nodes once the maximum unmount time is exceeded.
+	ForceDetachOnTimeoutEnabled ForceDetachOnTimeoutPolicy = "Enabled"
+	// ForceDetachOnTimeoutDisabled prevents kube-controller-manager from force
+	// detaching volumes based on the maximum unmount time.
+	ForceDetachOnTimeoutDisabled ForceDetachOnTimeoutPolicy = "Disabled"
+)
+
+// ControllerManagerStatus defines the observed state of the Kubernetes controller manager
+// +kubebuilder:validation:MinProperties=1
```
saschagrunert: With `MinProperties=1` here and `MinItems=1` on L87, the operator can never write `status: {}` or clear its conditions. Both are rejected, I confirmed against envtest.

Documented on L80 so not a defect, but it constrains `cluster-kube-controller-manager-operator`: a status write with zero conditions fails admission with an opaque error. Worth noting in the EP.

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,115 @@
...
+	Items []ControllerManager `json:"items"`
+}
+
+const (
+	// ConditionTypeProgressing is a condition type that indicates whether the current spec
+	// is in the process of rolling out to kube-controller-manager.
+	// When True, not all nodes have the current revision.
+	// When False, all nodes have the current revision.
+	ConditionTypeProgressing = "Progressing"
+)
```
saschagrunert: `ConditionTypeProgressing` is very generic for package scope; `config/v1alpha1` holds six root types and the next one wanting a Progressing condition collides. Separately, "current revision" is standalone static-pod language that does not translate to HyperShift, and `Progressing=False` alone does not say which generation landed.

```suggestion
const (
	// ControllerManagerProgressing is a condition type that indicates whether the current spec
	// is in the process of rolling out to kube-controller-manager.
	// When True, not all kube-controller-manager instances are running the current configuration.
	// When False, all kube-controller-manager instances are running the current configuration.
	// observedGeneration is set to the .metadata.generation that the condition was computed from.
	ControllerManagerProgressing = "Progressing"
)
```

Please mirror the `observedGeneration` commitment into the EP.

config/v1alpha1/tests/controllermanagers.config.openshift.io/DisableForceDetachOnTimeout.yaml
```diff
@@ -0,0 +1,63 @@
...
+      expectedError: "Invalid value: \"object\": controllermanager is a singleton, .metadata.name must be 'cluster'"
+    - name: Should reject unsupported value of forceDetachOnTimeout
+      initial: |
+        apiVersion: config.openshift.io/v1alpha1
+        kind: ControllerManager
+        metadata:
+          name: cluster
+        spec:
+          forceDetachOnTimeout: INVALID
+      expectedError: "spec.forceDetachOnTimeout: Unsupported value: \"INVALID\": supported values: \"Enabled\", \"Disabled\""
```
saschagrunert, line 31: Worth pinning the empty-spec rejection here:

```yaml
    - name: Should reject an empty spec
      initial: |
        kind: ControllerManager
        metadata:
          name: cluster
        spec: {}
      expectedError: "spec: Invalid value: 0: spec in body should have at least 1 properties"
```

config/v1alpha1/tests/controllermanagers.config.openshift.io/DisableForceDetachOnTimeout.yaml
```diff
@@ -0,0 +1,63 @@
...
+          name: cluster
+        spec:
+          forceDetachOnTimeout: Disabled
+      expected: |
+        apiVersion: config.openshift.io/v1alpha1
+        kind: ControllerManager
+        metadata:
+          name: cluster
+        spec:
+          forceDetachOnTimeout: Disabled
```
saschagrunert, line 54: Nothing exercises `status`, which now carries conditions, `listMapKey` uniqueness, `MinItems`/`MaxItems` and `minProperties`. A positive case and the empty-list rejection would cover it. Both verified locally:

```yaml
    - name: Should be able to set a Progressing condition
      initial: |
        kind: ControllerManager
        metadata:
          name: cluster
        spec:
          forceDetachOnTimeout: Enabled
      updated: |
        kind: ControllerManager
        metadata:
          name: cluster
        spec:
          forceDetachOnTimeout: Enabled
        status:
          conditions:
          - type: Progressing
            status: "True"
            reason: RolloutInProgress
            message: rolling out to kube-controller-manager
            lastTransitionTime: "2024-01-01T00:00:00Z"
      expected: |
        kind: ControllerManager
        metadata:
          name: cluster
        spec:
          forceDetachOnTimeout: Enabled
        status:
          conditions:
          - type: Progressing
            status: "True"
            reason: RolloutInProgress
            message: rolling out to kube-controller-manager
            lastTransitionTime: "2024-01-01T00:00:00Z"
    - name: Should reject an empty conditions list
      initial: |
        kind: ControllerManager
        metadata:
          name: cluster
        spec:
          forceDetachOnTimeout: Enabled
      updated: |
        kind: ControllerManager
        metadata:
          name: cluster
        spec:
          forceDetachOnTimeout: Enabled
        status:
          conditions: []
      expectedStatusError: "status.conditions: Invalid value: 0: conditions in body should have at least 1 items"
```

features/features.go
```diff
@@ -1074,4 +1074,12 @@ var (
 					productScope(kubernetes).
 					enhancementPR("https://github.com/kubernetes/enhancements/issues/5055").
 					mustRegister()
+
+	FeatureGateDisableForceDetachOnTimeout = newFeatureGate("DisableForceDetachOnTimeout").
+						reportProblemsToJiraComponent("Storage").
```
saschagrunert: Only bare `"Storage"` in the file. The other storage gates use `Storage / Kubernetes External Components` or `Storage / Kubernetes`.

saschagrunert (commented):
Another pass over the current state. The earlier round looks addressed: scheme registration, payload CRD glob, `+enum`, singleton rule and the status contract are all right now.

Verified locally on `9547e69`: `make lint` clean, ControllerManager integration tests pass, payload manifest byte-identical to the generated one with `bootstrap-required` gone. I also ran throwaway suites against envtest to confirm what the `MinProperties` / `MinItems` markers actually reject; details inline.

Two things not inline:

1. @lance5890's question about the type name is still open. `config.openshift.io/APIServer` deliberately covers both apiservers, so a `ControllerManager` that only ever configures KCM is a name that will read wrong if openshift-controller-manager config ever needs a home. v1alpha1 is the point to settle it, so the thread deserves an explicit answer either way.
2. openshift/enhancements#2095 has no acks yet. The API reviewers guide expects an initial ack before a gate merges into DevPreviewNoUpgrade, so that is the gating item here. The PR description is also stale (still `config/v1`, status without conditions) and feeds the squashed commit message.

config/v1alpha1/types_controllermanager.go
dobsonj: I added a comment on L29. I'd prefer to keep the `forceDetachOnTimeout` field optional in case we add more fields to ControllerManagerSpec in the future so that it's still possible to express "no opinion" by omission.

config/v1alpha1/types_controllermanager.go
dobsonj: Ack, I added a new `ControllerManagerConfig` as the top-level gate and moved `DisableForceDetachOnTimeout` to the field.

config/v1alpha1/tests/controllermanagers.config.openshift.io/DisableForceDetachOnTimeout.yaml
dobsonj: Done, thanks.

config/v1alpha1/tests/controllermanagers.config.openshift.io/DisableForceDetachOnTimeout.yaml
dobsonj: Done, thanks.

features/features.go
dobsonj: Done, changed to `Storage / Kubernetes`.

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,118 @@
...
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ControllerManager holds cluster-wide configuration for the Kubernetes controller manager.
+// The resource is a singleton named "cluster".
+//
+// Compatibility level 4: No compatibility is provided, the API can change at any point for any reason. These capabilities should not be used by applications needing long term support.
```
dobsonj: Turning this into a thread to discuss questions from @lance5890 and @saschagrunert about whether ControllerManager should support configuration for both kube-controller-manager and openshift-controller-manager:

> 1. I'm very glad to see that we finally added the `controller manager` crd,
> 2. However,  as far as I know, the [apiserver](https://github.com/openshift/api/blob/master/config/v1/types_apiserver.go) suppuort  configuration for both kube-apiserver and the openshift-apiserver;
> 3. The [Forced storage detach](https://kubernetes.io/docs/concepts/cluster-administration/node-shutdown/#storage-force-detach-on-timeout) seems  to be unique to the the kube-controller-manger(not for the openshift-controller-manager), should we add this in [kube-controller-manager](https://github.com/openshift/api/blob/master/payload-manifests/crds/0000_25_kube-controller-manager_01_kubecontrollermanagers.crd.yaml), and adjust this new feature in kcm-o should be fine?

There was a [slack discussion](https://redhat-internal.slack.com/archives/CKJR6200N/p1758559804038819) awhile back where @ingvagabund requested that we create a new config object for the operand (KCM) as opposed to adding this as a parameter to `kubecontrollermanagers.operator.openshift.io`, so that's the guidance I have followed in this PR. If @ingvagabund has a change of mind, I wouldn't mind the simpler approach of adding a new `forceDetachOnTimeout` parameter to `kubecontrollermanagers.operator.openshift.io`.

To your point though, it's true that `apiservers.config.openshift.io` controls parameters that apply to both `kube-apiserver` and `openshift-apiserver`, while `controllermanagers.config.openshift.io` is currently structured to only apply to `kube-controller-manager` where the `disable-force-detach-on-timeout` parameter is used.

> 1. @lance5890's question about the type name is still open. `config.openshift.io/APIServer` deliberately covers both apiservers, so a `ControllerManager` that only ever configures KCM is a name that will read wrong if openshift-controller-manager config ever needs a home. v1alpha1 is the point to settle it, so the thread deserves an explicit answer either way.

Would it make sense then to rename this new config object to `KubeControllerManager`? This is different from the precedent I see for APIServer and Scheduler config objects, but it would avoid a naming collision if we introduce an operand config for openshift-controller-manager in the future.

dobsonj: Or, alternatively, we could keep the `ControllerManager` name (consistent with `APIServer` and `Scheduler` config objects) and a future enhancement could still introduce parameters for openshift-controller-manager. `APIServer` has parameters that apply to both kube-apiserver and openshift-apiserver, and `Scheduler` has some fields for `kube-scheduler` and other fields that affect scheduling but are not specific to `kube-scheduler` (like `mastersSchedulable` and `defaultNodeSelector`).

While the focus of `ControllerManager` is on KCM for this enhancement, it could control other elements of our controller managers in the future.

config/v1alpha1/types_controllermanager.go
dobsonj: I think I made a mistake introducing the condition to keep the status field. This config object is really only meant to provide options for the operand (like APIServer and Scheduler config objects), and those examples intentionally do not have any status fields. The progressing condition should be handled in the [operator status](https://github.com/openshift/api/blob/master/operator/v1/types_kubecontrollermanager.go#L49) instead. Introducing a progressing condition here is redundant and adds unnecessary complexity. I'm going to remove the condition and status field.

config/v1alpha1/zz_generated.featuregated-crd-manifests/controllermanagers.config.openshift.io/ControllerManagerConfig.yaml
```diff
@@ -0,0 +1,59 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    api-approved.openshift.io: https://github.com/openshift/api/pull/2668
+    api.openshift.io/filename-cvo-runlevel: "0000_10"
+    api.openshift.io/filename-operator: config-operator
+    api.openshift.io/filename-ordering: "01"
+    feature-gate.release.openshift.io/ControllerManagerConfig: "true"
```
saschagrunert, line 9: `ci/prow/verify` is failing on this PR, and this file is why. `hack/verify-integration-tests.sh:64` requires one test suite per gated manifest, and splitting the gates added this variant without one:

```
No test suite file found for CRD .../controllermanagers.config.openshift.io/ControllerManagerConfig.yaml,
expected .../config/v1alpha1/tests/controllermanagers.config.openshift.io/ControllerManagerConfig.yaml
make: *** [Makefile:69: verify-non-codegen] Error 1
```

Reproduced locally. Adding `config/v1alpha1/tests/controllermanagers.config.openshift.io/ControllerManagerConfig.yaml` clears it, and the suite runs green:

```yaml
apiVersion: apiextensions.k8s.io/v1 # Hack because controller-gen complains if we don't have this
name: "ControllerManager"
crdName: controllermanagers.config.openshift.io
featureGates:
  - ControllerManagerConfig
tests:
  onCreate:
    - name: Should reject a ControllerManager not named cluster
      initial: |
        kind: ControllerManager
        metadata:
          name: foo
        spec:
          forceDetachOnTimeout: Enabled
      expectedError: "Invalid value: \"object\": controllermanager is a singleton, .metadata.name must be 'cluster'"
    - name: Should reject an empty spec
      initial: |
        kind: ControllerManager
        metadata:
          name: cluster
        spec: {}
      expectedError: "spec: Invalid value: 0: spec in body should have at least 1 properties"
```

Worth knowing: the harness runs gate-named suites against the merged feature-set CRD, not against this partial manifest, so this suite and `DisableForceDetachOnTimeout.yaml` see the same schema. Pick cases that do not duplicate each other.

config/v1alpha1/zz_generated.featuregated-crd-manifests/controllermanagers.config.openshift.io/ControllerManagerConfig.yaml
```diff
@@ -0,0 +1,59 @@
...
+              In CamelCase.
+              More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: |-
+              spec holds user settable values for configuration.
+              The only way to express no opinion in the spec is to not create the ControllerManager object at all.
+            minProperties: 1
```
saschagrunert, line 51: This variant is a spec that is `required`, must carry at least one property, and declares none, so the object cannot be created at all.

Unreachable today: both gates are `enable(inDevPreviewNoUpgrade())` (features.go L1083 and L1091), so every shipped CRD carries `forceDetachOnTimeout`. It becomes real the moment they diverge, which is the ordinary promotion path if `ControllerManagerConfig` moves to TechPreview ahead of the storage gate. It is also why the suite in my other comment can only assert things this gate does not itself enable.

Not a blocker. Either a note in `features/features.go` that the two gates must be promoted together, or dropping `MinProperties` once a second ungated field exists.

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,85 @@
...
+	// When set to "Enabled", volumes are force detached from unhealthy nodes after
+	// the maximum unmount time, so that workloads using them can start on other nodes.
+	// Force detaching a volume that is still in use by the node can corrupt its data.
+	// When set to "Disabled", volumes are not force detached based on the maximum
+	// unmount time. Volumes remain attached to an unhealthy node until it recovers,
+	// or until the node is tainted with "node.kubernetes.io/out-of-service"
+	// as part of the non-graceful node shutdown procedure.
+	// When omitted, this means the user has no opinion and the platform is left
+	// to choose a reasonable default, which is subject to change over time.
+	// The current default is "Enabled".
```
saschagrunert, line 50: Following up on the `MinProperties` thread above rather than reopening it: keeping the field optional for future fields is the right call, and the note you added to the `spec` godoc covers it.

The one gap left is that `oc explain controllermanager.spec.forceDetachOnTimeout` renders only this comment, so someone reading the field on its own is told to omit it when `spec` carries `minProperties: 1` and this is its only property. One line closes that:

```suggestion
	// The current default is "Enabled".
	// While this is the only field in spec, it must be set, because spec cannot be empty.
```

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,85 @@
...
+// Compatibility level 4: No compatibility is provided, the API can change at any point for any reason. These capabilities should not be used by applications needing long term support.
+// +openshift:compatibility-gen:level=4
+type ControllerManagerList struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// metadata is the standard list's metadata.
+	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
+	metav1.ListMeta `json:"metadata"`
+
+	Items []ControllerManager `json:"items"`
```
saschagrunert, line 87: `items` has no godoc, so it lands undocumented in the generated OpenAPI. `PKIList` (types_pki.go L272) and `ClusterMonitoringList` (types_cluster_monitoring.go L71) in this package both document it.

```suggestion
	// items is a list of ControllerManager resources
	Items []ControllerManager `json:"items"`
```

saschagrunert (commented):
Another pass over `4edbf3b1`. The status removal, the gate split, scheme registration, the payload glob and the singleton rule all look right now. Verified locally: `make lint` clean, `make verify-codegen-crds` clean, payload manifest byte-identical to the generated one, ControllerManager integration tests green.

Inline comments cover what is left in the code. Three things that do not anchor to a line:

1. **openshift/enhancements#2095 has no ack.** Every review on it is `COMMENTED`, there is no lgtm, and the bot still reports `NOT APPROVED`. The API reviewers guide expects an initial ack before a gate merges into `DevPreviewNoUpgrade`, so this is the gating item here, not anything in the API itself.

2. **The EP no longer matches the API.** #2095 still documents `.status.conditions` with a `Progressing` type (lines 63 and 82-83 of its diff), which `4edbf3b1` removed. I agree with your conclusion that rollout progress belongs on `kubecontrollermanagers.operator.openshift.io` status, so the EP wants the same edit. It also only names `DisableForceDetachOnTimeout` while both gates point readers at it, so `ControllerManagerConfig` should be described there too.

3. **The PR description is stale.** It still refers to `config/v1`, `config/v1/types_controllermanager.go` and `ControllerManagerStatus`. With `tide/merge-method-squash` that text becomes the commit message.

@JoelSpeed for what it is worth, the three things you asked for in January (an EP, a feature gate, and starting at v1alpha1) are all in place now.

config/v1alpha1/types_controllermanager.go
saschagrunert: Agreed on keeping the field optional, and the note you added to the `spec` godoc covers the contradiction. I left the one remaining wording nit on L49 as a suggestion, since `oc explain` on the field alone still reads as if omission were possible. Nothing blocking in this thread from my side.

config/v1alpha1/types_controllermanager.go
saschagrunert: @lance5890 @dobsonj answering this directly so the thread can close: **keep `ControllerManager`**, do not rename to `KubeControllerManager`.

Three reasons.

`operator.openshift.io/v1.KubeControllerManager` already exists. A `config.openshift.io/v1alpha1.KubeControllerManager` alongside it puts two different kinds under one name in two groups, which is worse for `oc explain` and short-name resolution than the naming asymmetry it is meant to avoid.

`config.openshift.io/APIServer` and `Scheduler` set the precedent that the config singleton is named for the component class rather than the binary, and `Scheduler` already carries fields that are not kube-scheduler flags (`mastersSchedulable`, `defaultNodeSelector`). The pattern tolerates exactly the situation described here.

And your second reply is the right read: openshift-controller-manager config, if it ever needs a home, fits under this spec as a sibling field, the way `APIServer` covers both apiservers. Nothing about today's shape forecloses that.

One follow-on if the generic name stays: the type godoc on L9 promises KCM only. Worth widening so the doc does not contradict the name, along the lines of what `APIServer` does:

```go
// ControllerManager holds cluster-wide configuration shared by the controller managers
// in the system, among them especially kube-controller-manager.
// The resource is a singleton named "cluster".
```

On the related HyperShift question I raised earlier: I checked and it is a non-issue. `apiservers` and `schedulers` both ship to the HyperShift profile too (`include.release.openshift.io/ibm-cloud-managed: "true"`, and `schedulers` even has a dedicated `-Hypershift` variant that keeps it), so this CRD landing in a hosted cluster while `hostedCluster.spec.configuration` is the authoritative input is the established shape for config singletons. Matching it is correct.

config/v1alpha1/types_controllermanager.go
saschagrunert: Addressed by the `ControllerManagerConfig` / `DisableForceDetachOnTimeout` split, thanks. One consequence worth tracking, which I left on the new gated manifest: with the only spec field behind the field-level gate, the `ControllerManagerConfig`-only schema is `minProperties: 1` over zero properties. Harmless while both gates ship in the same feature sets, but it wants the two promoted together.

config/v1alpha1/types_controllermanager.go
saschagrunert: Agreed with your own conclusion in the last message, and the removal in `4edbf3b1` is the right outcome. Rollout progress belongs on `kubecontrollermanagers.operator.openshift.io` status, and a config singleton mirroring `APIServer` and `Scheduler` with no status of its own is consistent. Resolved from my side. The EP still describes the `Progressing` condition on this object though, so that needs the same edit.

config/v1alpha1/types_controllermanager.go
saschagrunert: Moot now: status and its conditions were removed in `4edbf3b1`, so neither the `MinProperties` nor the `MinItems` constraint exists any more. Nothing to carry into the EP either. Closing this out.

config/v1alpha1/zz_generated.featuregated-crd-manifests/controllermanagers.config.openshift.io/ControllerManagerConfig.yaml
dobsonj: Ack, I added a comment about promoting the two gates together in `features/features.go`.

config/v1alpha1/zz_generated.featuregated-crd-manifests/controllermanagers.config.openshift.io/ControllerManagerConfig.yaml
dobsonj: Thanks, I split these test cases out into `ControllerManagerConfig.yaml` and removed them from `DisableForceDetachOnTimeout.yaml`.

config/v1alpha1/types_controllermanager.go
dobsonj: Agreed, let's keep the ControllerManager name for all the reasons listed above. I updated the godoc comments to widen the scope of ControllerManager.

config/v1alpha1/types_controllermanager.go
dobsonj: Ack, I dropped the notes about the condition from the enhancement.

dobsonj:
Thanks @saschagrunert for all your help with this.

> https://github.com/openshift/enhancements/pull/2095 has no ack. Every review on it is COMMENTED, there is no lgtm, and the bot still reports NOT APPROVED. The API reviewers guide expects an initial ack before a gate merges into DevPreviewNoUpgrade, so this is the gating item here, not anything in the API itself.

Understood, I'll chase approvals on the enhancement and then report back once it has merged.

> The EP no longer matches the API. https://github.com/openshift/api/pull/2095 still documents .status.conditions with a Progressing type (lines 63 and 82-83 of its diff), which 4edbf3b1 removed. I agree with your conclusion that rollout progress belongs on kubecontrollermanagers.operator.openshift.io status, so the EP wants the same edit. It also only names DisableForceDetachOnTimeout while both gates point readers at it, so ControllerManagerConfig should be described there too.

I pushed an update to the EP to drop the condition, and to mention both feature gates in the API section.

> The PR description is stale. It still refers to config/v1, config/v1/types_controllermanager.go and ControllerManagerStatus. With tide/merge-method-squash that text becomes the commit message.

I updated the PR description. Everything from "PR Type" down was generated by one of the bots, but I made manual edits. Maybe there's a way to get it regenerated based on current state. Note though that `tide/merge-method-squash` only includes the PR title and the individual commit messages. It won't include the lengthy PR description in the commit message.

I just came across [Joel's comment](https://github.com/openshift/api/pull/2664#issuecomment-3789560647) though that we shouldn't use tide squash on this repo, so I'll plan on doing a final squash of all these commits before merge.

dobsonj:
@coderabbitai resume

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,88 @@
...
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitzero"`
+	// spec holds user settable values for configuration.
+	// The only way to express no opinion in the spec is to not create the ControllerManager object at all.
+	// +required
+	Spec ControllerManagerSpec `json:"spec,omitzero"`
+}
+
+// ControllerManagerSpec defines the desired state of the controller managers
+// +kubebuilder:validation:MinProperties=1
```
saschagrunert: Correcting something I said in the January round: the linter does not force `MinProperties=1`, only that the marker is present. `CRIOCredentialProviderConfigSpec` in this same package uses `MinProperties=0` and `minlength` is satisfied. I tested both on this branch:

- `MinProperties=0` alone -> `config/v1alpha1/types_controllermanager.go:31:2: requiredfields: field ControllerManager.Spec has a valid zero value ({}) and should be a pointer. (kubeapilinter)`
- `MinProperties=0` plus a pointer spec -> `0 issues` on both golangci configs

The pointer spec being:

```go
// +required
Spec *ControllerManagerSpec `json:"spec,omitempty,omitzero"`
```

That second shape is what `CRIOCredentialProviderConfig` already ships (`types_crio_credential_provider_config.go:36,46`), and `+required` still yields `required: - spec` in its generated CRD, so the pointer does not make spec optional.

`MinProperties=1` is the single cause of three workarounds in this PR:

1. L48-51 tells the reader both "When omitted, this means the user has no opinion" and "it must be set, because spec cannot be empty". That is one `oc explain` page contradicting itself. L51 was my suggestion last round, but it patches the symptom.
2. The promotion-coupling note at `features/features.go:1078-1079`.
3. `zz_generated.featuregated-crd-manifests/controllermanagers.config.openshift.io/ControllerManagerConfig.yaml:47-54`, a required spec with `minProperties: 1` over zero properties: an object that cannot be created at all if the gates ever diverge.

`MinProperties=0` plus the pointer drops all three: the field becomes genuinely optional, the gates decouple, the gate-only schema becomes usable, and L29 changes to say that omitting the field means no opinion.

Two costs, both of which I should name rather than leave you to find: `spec: {}` becomes creatable as a no-op object, and the `Should reject an empty spec` case in `config/v1alpha1/tests/controllermanagers.config.openshift.io/ControllerManagerConfig.yaml` has to go, leaving that suite with a single case. It also needs a full `make update`, not a suggestion click.

Not a blocker, and I am not reopening the optionality decision: you were clear you want the field optional for future fields, and it is. But the linter is no longer an argument for `=1`, so it is worth deciding on the merits. If `=1` stays, the current docs plus the features.go note are adequate and I have nothing further here.

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,88 @@
...
+// +kubebuilder:validation:XValidation:rule="self.metadata.name == 'cluster'",message="controllermanager is a singleton, .metadata.name must be 'cluster'"
+type ControllerManager struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// metadata is the standard object's metadata.
+	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitzero"`
+	// spec holds user settable values for configuration.
+	// The only way to express no opinion in the spec is to not create the ControllerManager object at all.
```
saschagrunert: Only relevant if `MinProperties=1` stays. The `minProperties: 1` constraint is never stated in user-facing terms on `spec` itself, and this description is what `oc explain controllermanager.spec` renders; L51 only appears under `oc explain controllermanager.spec.forceDetachOnTimeout`.

```suggestion
	// spec holds user settable values for configuration.
	// At least one field must be set.
	// The only way to express no opinion in the spec is to not create the ControllerManager object at all.
```

This string is copied into the CRD manifests, the swagger docs and `openapi.json`, so it needs `make update` rather than committing the suggestion on its own.

config/v1alpha1/types_controllermanager.go
```diff
@@ -0,0 +1,88 @@
...
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ControllerManager holds cluster-wide configuration shared by the controller managers
+// in the system, among them especially kube-controller-manager.
+// The resource is a singleton named "cluster".
+//
+// Compatibility level 4: No compatibility is provided, the API can change at any point for any reason. These capabilities should not be used by applications needing long term support.
+// +openshift:compatibility-gen:level=4
```
saschagrunert, line 14: Not a change request on this PR, and it may sit better on openshift/enhancements#2095, but it is worth settling before implementation starts.

The EP says HyperShift will get `hostedCluster.spec.configuration.controllerManager`, "similar to what already exists for APIServer and Scheduler". I checked `api/hypershift/v1beta1/hostedcluster_types.go`: `ClusterConfiguration` (L2839-2907) holds ten fields and every one is a `*configv1.*Spec`, i.e. compatibility level 1. A level 4 `config/v1alpha1.ControllerManagerSpec` does not fit there, so that plumbing cannot land until this type reaches `config/v1`.

Meanwhile HyperShift already has a KCM slot on the sibling path: `hostedCluster.spec.operatorConfiguration` (L749) -> `OperatorConfiguration.KubeControllerManager KubeControllerManagerOperatorSpec` (L2955). That type is HyperShift owned (`api/hypershift/v1beta1/operator.go:75-77`, currently just `ComponentLogLevelSpec`) and carries no configv1 dependency.

So the topology section has a real choice: wait for the v1 bump and go through `spec.configuration`, or extend the existing HyperShift native `spec.operatorConfiguration.kubeControllerManager`. Worth naming in the EP so it is not discovered mid implementation. Nothing to change in this PR.

config/v1alpha1/types_controllermanager.go
dobsonj: Thanks, I like this, it makes the API more consistent and decouples the featuregates. I've pushed the changes in a new commit.

config/v1alpha1/types_controllermanager.go
dobsonj: I've changed it to `MinProperties=0`, so this is no longer necessary. Resolving this thread.

config/v1alpha1/types_controllermanager.go
dobsonj: I wasn't certain about this when I started the enhancement, but I pushed a new note to the enhancement. I think the right thing to do is keep this parameter controlled only by `ControllerManagerSpec`, which means promoting the API to v1 before implementing hypershift support. This parameter belongs in the config for the operand, not in the operator spec.

dobsonj:
/remove-label tide/merge-method-squash
I squashed the commits in the last push.

dobsonj:
The enhancement PR https://github.com/openshift/enhancements/pull/2095 has merged, I've addressed the remaining feedback on this PR, and squashed the commits. So I _think_ this is ready for approvals unless anyone has new feedback to discuss.

PR #2669, opened 2026-01-26
OCPBUGS-68343: Introduce KMSEncryption feature gate

ardaguclu:
### **User description**
As we discussed in https://github.com/openshift/enhancements/pull/1900. We need to allow nil KMSConfig, when encryption type is set to KMS. 

This PR introduces new `KMSEncryption` feature gate to exclude KMSConfig which is only enabled when `KMSEncryptionProvider` feature gate.

___

### **PR Type**
Enhancement

___

### **Description**
- Relax KMS validation rule to allow nil KMSConfig when encryption type is KMS

- Update validation message to only forbid kms config for non-KMS encryption types

- Add test case verifying KMS type with nil kms config is accepted

- Update all CRD manifests and generated files with new validation rule

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Old Validation Rule"] -->|"Requires kms config<br/>when type=KMS"| B["Strict Validation"]
  C["New Validation Rule"] -->|"Allows nil kms config<br/>when type=KMS"| D["Relaxed Validation"]
  B -->|"PR Changes"| D
  E["Test Case Added"] -->|"Validates KMS type<br/>with nil config"| D
```

<details><summary><h3>File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>types_apiserver.go</strong><dd><code>Update KMS validation rule to allow nil config</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/openshift/api/pull/2669/files#diff-616d67895c3421c2d091662d30ed47b4b6f0b57db9411e29618d22b964ddb9ef">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Tests</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>KMSEncryptionProvider.yaml</strong><dd><code>Add test for KMS type with nil kms config</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/openshift/api/pull/2669/files#diff-d4baff689784d53bbe214f3ebc28aff0b64dc5b07aa54f1d23b2369766c7937a">+10/-4</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Configuration changes</strong></td><td><details><summary>7 files</summary><table>
<tr>
  <td><strong>0000_10_config-operator_01_apiservers-CustomNoUpgrade.crd.yaml</strong><dd><code>Update KMS validation in CustomNoUpgrade CRD manifest</code>&nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/openshift/api/pull/2669/files#diff-be4351b1565b31158b44ef1f06c6943d19c588a00740dfae67d6ae73991f018f">+2/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>0000_10_config-operator_01_apiservers-DevPreviewNoUpgrade.crd.yaml</strong><dd><code>Update KMS validation in DevPreviewNoUpgrade CRD manifest</code></dd></td>
  <td><a href="https://github.com/openshift/api/pull/2669/files#diff-3a8bb9c23710daf4fdad2d0fce1e55e811672f9e3721591965eb4a49a5b2be9c">+2/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>0000_10_config-operator_01_apiservers-TechPreviewNoUpgrade.crd.yaml</strong><dd><code>Update KMS validation in TechPreviewNoUpgrade CRD manifest</code></dd></td>
  <td><a href="https://github.com/openshift/api/pull/2669/files#diff-4800b109f6d5d40f86f23b97f62601f739acefbc36b8cf9fa917976d9018a61f">+2/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>KMSEncryptionProvider.yaml</strong><dd><code>Update KMS validation in featuregated CRD manifest</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/openshift/api/pull/2669/files#diff-ec88e983a15aa568c68265db61859b2ddd7032bda1f4a00766b7cdd6f9691af3">+2/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>0000_10_config-operator_01_apiservers-CustomNoUpgrade.crd.yaml</strong><dd><code>Update KMS validation in payload CustomNoUpgrade CRD</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/openshift/api/pull/2669/files#diff-6d0dbb84751cb0b3f966fb36bb479560cb30979a287c26dd49aabb0677edb631">+2/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>0000_10_config-operator_01_apiservers-DevPreviewNoUpgrade.crd.yaml</strong><dd><code>Update KMS validation in payload DevPreviewNoUpgrade CRD</code>&nbsp; </dd></td>
  <td><a href="https://github.com/openshift/api/pull/2669/files#diff-65ea540544e4eac91ad49a3020396d2a4440a62b474fab690463a6292f579651">+2/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>0000_10_config-operator_01_apiservers-TechPreviewNoUpgrade.crd.yaml</strong><dd><code>Update KMS validation in payload TechPreviewNoUpgrade CRD</code></dd></td>
  <td><a href="https://github.com/openshift/api/pull/2669/files#diff-0e6e2f7520ab27ca2c0d3b6a6df8729b0a83c9d1eebcd7984448e58c9221ce2e">+2/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Documentation</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>openapi.json</strong><dd><code>Update OpenAPI schema with validation and documentation changes</code></dd></td>
  <td><a href="https://github.com/openshift/api/pull/2669/files#diff-6466e6ec9d299ee55a89f9ae0f455bc960d7295ead963490f5497447c2f1feb3">+17/-12</a>&nbsp; </td>

</tr>
</table></details></td></tr></tbody></table>

</details>

___

### DIFF
--- a/config/v1/tests/apiservers.config.openshift.io/KMSEncryption.yaml
+++ b/config/v1/tests/apiservers.config.openshift.io/KMSEncryption.yaml
@@ -3,6 +3,7 @@ name: "APIServer"
 crdName: apiservers.config.openshift.io
 featureGates:
 - KMSEncryption
+- -KMSEncryptionProvider
 tests:
   onCreate:
     - name: Should be able to create with KMS type without kms config


config/v1/types_apiserver.go
```diff
@@ -175,7 +175,7 @@ type APIServerNamedServingCert struct {
 }
 
 // APIServerEncryption is used to encrypt sensitive resources on the cluster.
-// +openshift:validation:FeatureGateAwareXValidation:featureGate=KMSEncryptionProvider,rule="has(self.type) && self.type == 'KMS' ?  has(self.kms) : !has(self.kms)",message="kms config is required when encryption type is KMS, and forbidden otherwise"
+// +openshift:validation:FeatureGateAwareXValidation:featureGate=KMSEncryptionProvider,rule="self.type != 'KMS' ? !has(self.kms) : true",message="kms config is forbidden when encryption type is not KMS"
```
p0lyn0mial: From what I understand, this field will be deprecated in the future anyway. 
For KMS v1, we don’t need it at all. 

To make the backport easier, we would like to relax the validation only.

Considering that there hasn’t been any implementation added for this field, the changes look good to me.

ardaguclu:
I'm cancelling this lgtm to not trigger CI jobs, until API reviewers review this PR
/lgtm cancel
/cc @JoelSpeed

config/v1/types_apiserver.go
ardaguclu: Agreed. Thanks for review.

JoelSpeed: Do we expect in the future to have or require this KMS field?

The field is tech preview, so we don't need to deprecate it, we could remove it if we think that it won't be needed in the long term at all.

If we do need it in the long term, is there a future where it is acceptable that the KMS field is an optional union member? Or are you expecting in this case that the KMS field would be required when type is KMS?

ardaguclu: It is hard to answer these questions for sure, since we actively discuss the design. But in my opinion, it is likely that we would mark this field (or another field like kmsv2 by tombstoning kmsConfig) as required.

JoelSpeed: I would like to make sure, that if we don't need this field, it is not promoted to the default feature set until it is implemented

Which gate are you intending to promote here? Can the required parts be under one gate and the union member/union validation be under a separate gate?

ardaguclu: If it is decided that KMSConfig can not be reusable, we'll certainly tombstone it until `KMSEncryptionProvider` is promoted to GA. Since the changes here will need to be backported to 4.21, we want to keep it simpler (that's why we defer the removal of KMSConfig in v2). 

But separate feature gate would make our process much more cumbersome, if you suggest removing the KMSConfig now to not need another feature gate, I'd prefer removing KMSConfig in this PR.

JoelSpeed: Why does a separate feature gate make this more cumbersome? It means we can defer any decision about changing validation/changing types/tombstoning fields until much later, rather than backporting some validation changes that might not actually be correct and promoting those to GA, painting us into a future corner?

The APIServer encryption struct is already GA with just the type field.

IIUC, for backport, you only actually want the `KMS` enum value for the type field? Is that correct? So lets put that enum value behind a gate, and backport that (and whatever the rest of this feature is) behind that gate, promote that, and avoid the `KMS` union member completely for now

ardaguclu: Do you have any example PRs that I can mimic?. If we introduce a new feature gate for enum and backport it to 4.21, what will happen to `KMSEncryptionProvider` feature gate?.

JoelSpeed: https://github.com/openshift/api/pull/2664 just added some feature gates

You'd need to pick a unique name, and then change the KMS `FeatureGateAwareEnum` to use that gate instead of the existing gate.

The current feature can be left in tech preview, or dropped to dev/custom as you prefer. And then we can work solely on promoting the KMS enum value which you need, and come back to the old gate later once we decide what the future work will need from a field in the KMS union

ardaguclu: Thank you for the guidance. I'll let you know once I update the PR.

p0lyn0mial: ok, so the plan is to add a new FG that allows an empty KMSConfig when the new feature is enabled, right? 
At the same time, the existing FG and validation remain unchanged.

sgtm.

p0lyn0mial: also, with this new approach, we would not need to change or backport any changes to openshift/client-go, right?

JoelSpeed: There's no client-go changes for this no, I think this LGTM now, we can merge this, promote the new `KMSEncryption` gate and then come back to the other gate when we do the fuller implementation later

JoelSpeed:
/lgtm

@p0lyn0mial @ardaguclu provided you are both happy, I'll leave the `verified` label to you

config/v1/types_apiserver.go
ardaguclu: > promote the new KMSEncryption gate

With respect to this, what should I do?

JoelSpeed: Normally, we would require 5 E2E tests reporting into component readiness gated on this feature gate `[OCPFeatureGate:KMSEncryption]`, are you able to build some tests for this part of the feature?

config/v1/types_apiserver.go
```diff
@@ -212,6 +212,7 @@ type APIServerEncryption struct {
 
 // +openshift:validation:FeatureGateAwareEnum:featureGate="",enum="";identity;aescbc;aesgcm
 // +openshift:validation:FeatureGateAwareEnum:featureGate=KMSEncryptionProvider,enum="";identity;aescbc;aesgcm;KMS
```
p0lyn0mial, line 214: won't this flag also be on for `TP` ? If yes, will it allow for setting an empty KMSConfig ?

config/v1/types_apiserver.go
ardaguclu: Currently we are actively working on deploying kms plugin on OCP for our e2e tests (https://github.com/openshift/release/pull/73750). But it is not ready yet. 

Can we backport this PR to 4.21 before e2e tests?. We'll need to vendor library-go with openshift/api, after that apiserver operators need to be vendored with library-go and openshift/api to run E2E tests.

config/v1/types_apiserver.go
ardaguclu: @JoelSpeed on TP enabled cluster, won't this _old_ feature gate be enabled too?. Won't this create conflicting behavior? i.e. KMSEncryption allows simply KMS enum but KMSEncryptionProvider does not allow it.

p0lyn0mial: @JoelSpeed is there a way to write an integration / unit test that would enable these two FG ?

ardaguclu: or each feature gate needs to be explicitly enabled on the cluster. So that we would just assume that user will enable `KMSEncryption` but not `KMSEncryptionProvider`

JoelSpeed: Currently yes, on a TP cluster you would see the old/existing behaviour. If you promoted the feature `KMSEncyption` to default, you would not be able to set the `KMS` field because it wouldn't exist.

If you are concerned about the interactions between them/testing both, you can either disable `KMSEncryptionProvider` from TechPreview and move it back to just DevPreview, or we can set up tests that require both gates, and just one gate, or, one gate that not the other. In the test files where you specify the gates, prefix with `-` to negate the inclusion

JoelSpeed: > Won't this create conflicting behavior? i.e. KMSEncryption allows simply KMS enum but KMSEncryptionProvider does not allow it.

You'd currently observe just the behaviour of the `KMSEncryptionProvider` feature gate, as it's a superset of the new gate.

ardaguclu: I think, we are both fine to moving `KMSEncryptionProvider` to DevPreview.

ardaguclu: I've moved KMSEncryptionProvider FG to DevPreview.

config/v1/tests/apiservers.config.openshift.io/KMSEncryption.yaml
```diff
@@ -0,0 +1,67 @@
+apiVersion: apiextensions.k8s.io/v1 # Hack because controller-gen complains if we don't have this
+name: "APIServer"
+crdName: apiservers.config.openshift.io
+featureGates:
+- KMSEncryption
```
JoelSpeed, line 5: Exclude these from running when both gates are enabled
```suggestion
- KMSEncryption
- -KMSEncryptionProvider
```

ardaguclu: Updated. verify-crdify and verify-crd-schema failures are expected?.

config/v1/tests/apiservers.config.openshift.io/KMSEncryptionProvider.yaml
```diff
@@ -1,128 +0,0 @@
-apiVersion: apiextensions.k8s.io/v1 # Hack because controller-gen complains if we don't have this
-name: "APIServer"
-crdName: apiservers.config.openshift.io
-featureGates:
-  - KMSEncryptionProvider
-  - -KMSEncryption
```
JoelSpeed: This was wrong, you didn't need line 6 here, please add this file back without L6

ardaguclu: Added back.

ardaguclu: CI is in good shape. Would you please have a look at when you have a chance?

ardaguclu:
/verified later
We'll add e2e tests

PR #2680, opened 2026-01-29
Add TLS adherence feature gate

richardsonnick:
## Add TLSAdherence feature gate for cluster-wide TLS configuration

This PR introduces the `TLSAdherence` feature gate and adds the `tlsAdherence` field to the `APIServer` config resource (`apiserver.config.openshift.io/v1`).

### Summary

The `tlsAdherence` field controls how strictly components in the cluster adhere to the TLS security profile configured on the APIServer resource.

**Valid values:**
- **Legacy** (default): Backward-compatible behavior where components attempt to honor the configured TLS profile but may fall back to their individual defaults if conflicts arise. This mode is intended for clusters that need to maintain compatibility with existing configurations during migration.
- **Strict**: Enforces strict adherence to the TLS configuration. All components must honor the configured profile. This mode is recommended for security-conscious deployments and is required for certain compliance frameworks.

### Feature Gate Details

- **Feature Gate Name:** `TLSAdherence`
- **Enabled in:** `DevPreviewNoUpgrade`, `TechPreviewNoUpgrade`
- **Contact:** @joelanford
- **Jira Component:** `HPCASE / TLS Adherence`

### Related

- **Enhancement:** https://github.com/openshift/enhancements/pull/1910

### DIFF
--- a/features/features.go
+++ b/features/features.go
@@ -817,13 +817,13 @@ var (
 					mustRegister()
 
 	FeatureGateAWSServiceLBNetworkSecurityGroup = newFeatureGate("AWSServiceLBNetworkSecurityGroup").
-					reportProblemsToJiraComponent("Cloud Compute / Cloud Controller Manager").
-					contactPerson("mtulio").
-					productScope(ocpSpecific).
-					enhancementPR("https://github.com/openshift/enhancements/pull/1802").
-					enable(inClusterProfile(SelfManaged), inDefault(), inOKD(), inTechPreviewNoUpgrade(), inDevPreviewNoUpgrade()).
-					enable(inClusterProfile(Hypershift), inTechPreviewNoUpgrade(), inDevPreviewNoUpgrade()).
-					mustRegister()
+							reportProblemsToJiraComponent("Cloud Compute / Cloud Controller Manager").
+							contactPerson("mtulio").
+							productScope(ocpSpecific).
+							enhancementPR("https://github.com/openshift/enhancements/pull/1802").
+							enable(inClusterProfile(SelfManaged), inDefault(), inOKD(), inTechPreviewNoUpgrade(), inDevPreviewNoUpgrade()).
+							enable(inClusterProfile(Hypershift), inTechPreviewNoUpgrade(), inDevPreviewNoUpgrade()).
+							mustRegister()
 
 	FeatureGateNoRegistryClusterInstall = newFeatureGate("NoRegistryClusterInstall").
 						reportProblemsToJiraComponent("Installer / Agent based installation").
@@ -1006,7 +1006,7 @@ var (
 					mustRegister()
 
 	FeatureGateTLSAdherence = newFeatureGate("TLSAdherence").
-				reportProblemsToJiraComponent("kube-apiserver").
+				reportProblemsToJiraComponent("HPCASE / TLS Adherence").
 				contactPerson("joelanford").
 				productScope(ocpSpecific).
 				enhancementPR("https://github.com/openshift/enhancements/pull/1910").


config/v1/types_apiserver.go
```diff
@@ -62,6 +62,26 @@ type APIServerSpec struct {
...
 	// +optional
 	TLSSecurityProfile *TLSSecurityProfile `json:"tlsSecurityProfile,omitempty"`
+	// tlsAdherence controls how strictly components in the cluster adhere to the TLS security profile
+	// configured on this APIServer resource.
+	//
+	// Valid values are "Legacy" and "Strict".
+	//
+	// When set to "Legacy" (the default), components attempt to honor the configured TLS profile
+	// but may fall back to their individual defaults if conflicts arise. This mode is intended for
+	// clusters that need to maintain compatibility with existing configurations during migration.
```
joelanford: I thought `Legacy` was going to essentially mean "only the externally exposed API server components", and `Strict` will mean "ALL cluster components with the exception of Kubelet-related and Ingress-related servers"?

I also think we should update GoDoc for the `TLSSecurityProfile` field, which currently says "for externally exposed servers". With TLSAdherence in the mix, what the profile is for depends on the TLSAdherence value.

But that should doc change should also only apply when the feature gate is enabled. I think @JoelSpeed said there's a way to do that?

joelanford: I also wonder if we should improve the naming of the enums to more accurately describe what they do? For example:
- LegacyExternalAPIServerComponentsOnly
- StrictAllComponents

Or something along those lines?

config/v1/types_tlssecurityprofile.go
```diff
@@ -168,6 +168,14 @@ type TLSProfileSpec struct {
...
 	// +listType=atomic
 	Ciphers []string `json:"ciphers"`
+	// curves is used to specify the elliptic curves that are used during
+	// the TLS handshake.  Operators may remove entries their operands do
+	// not support.  For example, to use X25519 and P-256 (yaml):
+	//
+	//   curves:
+	//     - X25519
+	//     - P-256
+	Curves []string `json:"curves,omitempty"`
```
joelanford: Based on my read of https://github.com/openshift/enhancements/pull/1894, it seems like very few components currently handle TLS curve configuration.

And I don't think we will be able to promote this full set of API changes unless components handle all of MinTLSVersion, CipherSuites, and Curves.

I am worried that the upstream's lack of curves support might take some time and would therefore block the rest of the feature around honoring the min TLS and cipher suites configuration.

Should we remove `Curves` from this PR or configure it to be present with a separate feature gate?

config/v1/types_tlssecurityprofile.go
```diff
@@ -168,6 +168,14 @@ type TLSProfileSpec struct {
 	//
 	// +listType=atomic
 	Ciphers []string `json:"ciphers"`
+	// curves is used to specify the elliptic curves that are used during
+	// the TLS handshake.  Operators may remove entries their operands do
+	// not support.  For example, to use X25519 and P-256 (yaml):
+	//
+	//   curves:
+	//     - X25519
+	//     - P-256
```
joelanford: If/when we do add `Curves` field, it needs to be introduced behind a feature gate.

joelanford (changes requested):
Curves should be removed or included behind a feature gate. My understanding is that we can't add a field to the Default CRD without first going through TechPreviewNoUpgrade.

everettraven:
Adding to @joelanford 's comment - I'd prefer we keep the curves field entirely separate here to keep this well scoped. IIRC there is already a PR that is looking to add the curves field.

config/v1/types_apiserver.go
richardsonnick: I think these names are far better. Going with LegacyExternalAPIServerComponentsOnly and StrictAllComponents

config/v1/types_tlssecurityprofile.go
richardsonnick: Yeah that should not be here. It has since been removed. This PR applies solely to the adherence fate.

config/v1/types_apiserver.go
```diff
@@ -62,6 +62,30 @@ type APIServerSpec struct {
...
+	// Note: The Kubelet and IngressController components are excluded from tlsAdherence control
+	// as they have their own dedicated TLS configuration mechanisms via KubeletConfig and
+	// IngressController CRs respectively.
+	//
+	// Components that encounter an unknown value for tlsAdherence should treat it as "StrictAllComponents"
+	// and log a warning to ensure forward compatibility while defaulting to the more secure behavior.
+	//
+	// When omitted, the default value is "LegacyExternalAPIServerComponentsOnly".
+	// +openshift:enable:FeatureGate=TLSAdherence
+	// +optional
```
joelanford, line 97: We need to configure validation on this field, and set the default.
```
// +kubebuilder:validation:items:Enum:=LegacyExternalAPIServerComponentsOnly;StrictAllComponents
// +kubebuilder:default:=LegacyExternalAPIServerComponentsOnly
```

config/v1/types_apiserver.go
```diff
@@ -237,6 +261,23 @@ const (
...
+const (
+	// TLSAdherenceLegacyExternalAPIServerComponentsOnly means only the externally exposed
+	// API server components (kube-apiserver, openshift-apiserver, oauth-apiserver) honor
+	// the configured TLS profile. Other components continue to use their individual TLS
+	// configurations.
+	TLSAdherenceLegacyExternalAPIServerComponentsOnly TLSAdherencePolicy = "LegacyExternalAPIServerComponentsOnly"
+
+	// TLSAdherenceStrictAllComponents means all components must honor the configured TLS
+	// profile. This mode is recommended for security-conscious deployments and is required
+	// for certain compliance frameworks.
```
joelanford: Maybe we should include a note here (and in the `tlsAdherence` description) that component-specific configurations can override the `tlsSecurityProfile` configuration?

config/v1/types_apiserver.go
joelanford: Oh I see you are defining the Enum validation on the type, which is fine. Let's also include the `+kubebuilder:default` marker there as well.

operator/v1/zz_generated.crd-manifests/0000_80_machine-config_01_machineconfigurations-Default.crd.yaml
joelanford, line 1: How did these changes get here? Maybe need to rebase and push?

config/v1/types_apiserver.go
joelanford: Hmm, reading the tlsAdherence description again, it seems like you are wanting it to be possible for an empty string to be allowed to be persisted for this field in etcd, which would allow us in theory to change the default semantic later in the future from legacy to strict without a change in the actual tlsAdherence value.

In practice, I think the problem with that idea is that each component essentially has their own implementation of this API and we'd have to do a really good job coordinating a semantic switch from legacy to strict. 

What do folks think about:
1. Actually writing `LegacyExternalAPIServerComponentsOnly` to etcd by default.
2. Telling implementers they specifically only need to look for `StrictAllComponents` and will not have to worry about the possibility of `""` with changing semantics later.
3. When want to set `StrictAllComponents` as the new default, we make sure newly installed clusters get that, and we potentially block upgrades of clusters that still use `LegacyExternalAPIServerComponentsOnly`? Or maybe there's some other way to encourage users to switch to `StrictAllComponents`?

joelanford:
Overall, I think this is looking pretty good for TechPreview. I think the main thing is providing clarity to implementers. Which is "use existing TLS configuration settings when `tlsAdherence` is `LegacyExternalAPIServerComponentsOnly`. Otherwise, honor what is is `tlsSecurityProfile`".

Not sure what @everettraven thinks, but I think we could potentially work out the finer details of the defaulting and descriptions in follow-ups if we agree on the necessary bits to get implementers going.

config/v1/types_apiserver.go
```diff
@@ -62,6 +62,35 @@ type APIServerSpec struct {
...
+	// API server components (kube-apiserver, openshift-apiserver, oauth-apiserver) honor the configured
+	// TLS profile. Other components continue to use their individual TLS configurations.
+	//
+	// When set to "StrictAllComponents", all components must honor the configured TLS profile.
+	// This mode is recommended for security-conscious deployments and is required for
+	// certain compliance frameworks.
+	//
+	// Note: The Kubelet and IngressController components are excluded from tlsAdherence control
+	// as they have their own dedicated TLS configuration mechanisms via KubeletConfig and
+	// IngressController CRs respectively.
```
everettraven: Once we settle on this in the EP, we should update this.

My current expectation is that this isn't an exclusion but rather a preference scenario where, if specified, the Kubelet / Ingress Controller components will be configured with what is specified in their existing APIs.

config/v1/types_apiserver.go
everettraven: For configuration APIs we generally try to avoid setting the default via `+kubebuilder:default` marker so that we can change the default behavior of the "no-opinion" mechanism.

In this case, I don't think we will be doing that but I still wouldn't use the `+kubebuilder:default` marker because then we won't be able to distinguish whether or not the user intentionally set the field or if we defaulted it on their behalf _and_ you technically get locked into that default being a contractual default.

config/v1/types_apiserver.go
```diff
@@ -62,6 +62,35 @@ type APIServerSpec struct {
...
+	//
+	// Components that encounter an unknown value for tlsAdherence should treat it as "StrictAllComponents"
+	// and log a warning to ensure forward compatibility while defaulting to the more secure behavior.
+	//
+	// This field is optional. When omitted, it is interpreted as "no opinion" and the cluster
+	// defaults to LegacyExternalAPIServerComponentsOnly behavior. Future versions of OpenShift
```
everettraven: We have some standard terminology we like to use here for consistency. As an example: https://github.com/openshift/api/blob/618697272d690b36a17dc58d01d347ce7567a7a8/config/v1/types_network.go#L244-L246

We should follow this convention as well so we aren't locked into a contractual default in case we want to do something like change the default behavior to `strict` but still give admins an option to specify that they want to stay on the legacy mode.

config/v1/types_apiserver.go
```diff
@@ -237,6 +266,23 @@ const (
...
+	// TLSAdherenceLegacyExternalAPIServerComponentsOnly means only the externally exposed
+	// API server components (kube-apiserver, openshift-apiserver, oauth-apiserver) honor
+	// the configured TLS profile. Other components continue to use their individual TLS
+	// configurations.
+	TLSAdherenceLegacyExternalAPIServerComponentsOnly TLSAdherencePolicy = "LegacyExternalAPIServerComponentsOnly"
+
+	// TLSAdherenceStrictAllComponents means all components must honor the configured TLS
+	// profile. This mode is recommended for security-conscious deployments and is required
+	// for certain compliance frameworks.
+	TLSAdherenceStrictAllComponents TLSAdherencePolicy = "StrictAllComponents"
```
everettraven:
```suggestion
	// TLSAdherencePolicyLegacyExternalAPIServerComponentsOnly means only the externally exposed
	// API server components (kube-apiserver, openshift-apiserver, oauth-apiserver) honor
	// the configured TLS profile. Other components continue to use their individual TLS
	// configurations.
	TLSAdherencePolicyLegacyExternalAPIServerComponentsOnly TLSAdherencePolicy = "LegacyExternalAPIServerComponentsOnly"

	// TLSAdherencePolicyStrictAllComponents means all components must honor the configured TLS
	// profile. This mode is recommended for security-conscious deployments and is required
	// for certain compliance frameworks.
	TLSAdherencePolicyStrictAllComponents TLSAdherencePolicy = "StrictAllComponents"
```

features/features.go
```diff
@@ -978,4 +978,12 @@ var (
 						enhancementPR("https://github.com/openshift/enhancements/pull/1465").
 						enableIn(configv1.DevPreviewNoUpgrade).
 						mustRegister()
+
+	FeatureGateTLSAdherence = newFeatureGate("TLSAdherence").
+				reportProblemsToJiraComponent("kube-apiserver").
+				contactPerson("joelanford").
+				productScope(ocpSpecific).
+				enhancementPR("https://github.com/openshift/enhancements/pull/1910").
+				enableIn(configv1.DevPreviewNoUpgrade, configv1.TechPreviewNoUpgrade).
```
everettraven: Are we targeting for this to be the same maturity level for both standalone OCP and HyperShift (HCP)?

HyperShift _does_ embed the APIServer configuration spec in https://github.com/openshift/hypershift/blob/825484eb33d14b4ab849b428d134582320655fcf/api/hypershift/v1beta1/hostedcluster_types.go#L1957-L1961 and if we aren't careful they might end up defaulting this new field to GA.

Regardless, I think we may want a sibling PR to ensure that a feature-gate entry at the appropriate level is added to https://github.com/openshift/hypershift/tree/main/api/hypershift/v1beta1/featuregates to prevent an accidental GA. We've been bitten by this in the past and it causes all kinds of headaches, so its probably best to avoid that here.

config/v1/types_apiserver.go
```diff
@@ -237,6 +266,23 @@ const (
 type APIServerStatus struct {
 }
 
+// TLSAdherencePolicy defines which components adhere to the TLS security profile.
+// +kubebuilder:validation:Enum=LegacyExternalAPIServerComponentsOnly;StrictAllComponents
+type TLSAdherencePolicy string
+
+const (
```
everettraven, line 280: Should we add a constant for `""` to represent the no-opinion case because we are using a non-pointer type for the field?

everettraven (commented):
Overall, this looks pretty good. Have a handful of comments

config/v1/types_apiserver.go
joelanford: > so that we can change the default behavior of the "no-opinion" mechanism.

Thanks, I was hoping you'd have some good advice here. It is the intent to eventually make all components honor the cluster-wide default, so I supposed we'll just have to deal with trying to coordinate a semantic change across a bunch of components at once when the time comes. 

To facilitate that, I'd suggest a library-go function somehow like this:

```go
// in configv1
const TLSAdherenceNoOpinion = ""
```

```go
func ShouldAllComponentsAdhere(tlsAdherence configv1.TLSAdherence) bool {
    if tlsAdherence == configv1.TLSAdherenceNoOpinion || tlsAdherence == configv1.LegacyExternalAPIServerComponentsOnly {
        return false
    }
    return true
}
```

And then when we want to change the semantic, we'd update that function to:
```go
func ShouldAllComponentsAdhere(tlsAdherence configv1.TLSAdherence) bool {
    if tlsAdherence == configv1.LegacyExternalAPIServerComponentsOnly {
        return false
    }
    return true
}
```

config/v1/types_apiserver.go
joelanford: I think that makes sense. It positions `apiserver` as the truly cluster-wide default. And it is still aligned with the fact that other components may have their own special configuration knobs that override the default.

config/v1/types_apiserver.go
```diff
@@ -62,6 +62,35 @@ type APIServerSpec struct {
...
+	//
+	// This field is optional. When omitted, it is interpreted as "no opinion" and the cluster
+	// defaults to LegacyExternalAPIServerComponentsOnly behavior. Future versions of OpenShift
+	// may require explicitly setting this field to "StrictAllComponents" before upgrading.
+	// Administrators are encouraged to migrate to "StrictAllComponents" proactively.
+	//
+	// Once set, this field may be changed to a different value.
+	// +openshift:enable:FeatureGate=TLSAdherence
+	// +optional
+	TLSAdherence TLSAdherencePolicy `json:"tlsAdherence,omitempty"`
```
JoelSpeed, line 98: In a recent call we discussed the fact that we might need different TLSSecurityProfile structs for APIServer vs Ingress. If we had separate structs, would this field still be a sibling of the TLSSecurityProfile, or would it be a member of the TLSSecurityProfile?

everettraven: My understanding is that this adherence field is only applicable for the APIServer type because it will be the central configuration type.

Ingress/Kubelet specific configurations would be overrides to the central configuration and Ingress and Kubelet would be expected to strictly adhere to those.

@joelanford @richardsonnick please correct me if I'm wrong here.

config/v1/types_apiserver.go
everettraven: That aligns with what I envisioned.

Can we make sure to update the EP with a mention of this kind of check/library that implementors are expected to use?

config/v1/types_apiserver.go
richardsonnick: Added a TLSAdherenceNoOpinion. This is a constant empty string field that should be treated the same as LegacyExternalAPIServerComponentsOnly

config/v1/types_apiserver.go
richardsonnick: I think this makes sense. Updated this to leave out the exclusion language and instead present the behavior as a "preference"

config/v1/types_apiserver.go
richardsonnick: Added a mention of this helper function. 
Also created a draft diff in library-go to add this function: https://github.com/openshift/library-go/pull/2114

Will need to land this first to get the TLSAdherencePolicy type into library-go

config/v1/types_apiserver.go
richardsonnick: This is my understanding as well. The ingress/kubelet configurations are overrides to the (central authoritative) apiserver type

features/features.go
richardsonnick: Added a featuregate in hypershift that adds TLSAdherence to:
- The disabled list in the Default feature gate files
- The enabled list in the TechPreviewNoUpgrade feature gate files


PR: https://github.com/openshift/hypershift/pull/7635

config/v1/types_apiserver.go
JoelSpeed: I understand that relationship, that wasn't actually the key part of my question
> If we had separate structs, would this field still be a sibling of the TLSSecurityProfile, or would it be a member of the TLSSecurityProfile?

everettraven: IMO, still a sibling.

config/v1/types_apiserver.go
```diff
@@ -237,6 +268,31 @@ const (
...
+// Implementors should use the ShouldAllComponentsAdhere helper function from library-go
+// rather than checking these values directly.
+// +kubebuilder:validation:Enum=LegacyExternalAPIServerComponentsOnly;StrictAllComponents
+type TLSAdherencePolicy string
+
+const (
+	// TLSAdherencePolicyNoOpinion represents an empty/unset value for tlsAdherence. When this value
+	// is set (or the field is omitted), the cluster defaults to LegacyExternalAPIServerComponentsOnly
+	// behavior. Components should treat this the same as LegacyExternalAPIServerComponentsOnly.
+	TLSAdherencePolicyNoOpinion TLSAdherencePolicy = ""
```
everettraven: This value _cannot_ be explicitly set and is only present when the field is omitted.

config/v1/types_apiserver.go
```diff
@@ -62,6 +62,37 @@ type APIServerSpec struct {
...
+	// component-specific TLS configurations are set, they take precedence over the cluster-wide
+	// tlsSecurityProfile. When not set, these components fall back to the cluster-wide default.
+	//
+	// Components that encounter an unknown value for tlsAdherence should treat it as "StrictAllComponents"
+	// and log a warning to ensure forward compatibility while defaulting to the more secure behavior.
+	//
+	// This field is optional. When omitted, it is interpreted as "no opinion" and the cluster
+	// defaults to LegacyExternalAPIServerComponentsOnly behavior. Future versions of OpenShift
+	// may require explicitly setting this field to "StrictAllComponents" before upgrading.
+	// Administrators are encouraged to migrate to "StrictAllComponents" proactively.
```
everettraven:
```suggestion
	// This field is optional. 
	// When omitted, this means the user has no opinion and the platform is left to choose reasonable defaults.
	// These defaults are subject to change over time.
	// The current default is LegacyExternalAPIServerComponentsOnly.
```

I don't think we need to mention that we may eventually require explicitly setting this field - we just do that when/if we do that.

config/v1/types_apiserver.go
```diff
@@ -62,6 +62,37 @@ type APIServerSpec struct {
...
+	//
+	// Once set, this field may be changed to a different value.
```
everettraven: We need to add a validation for this to ensure that once set, the field cannot be omitted. This would need to be enforced at the struct level using a CEL validation like:
https://github.com/openshift/api/blob/0acc398c4ed3511a1b311c29283aed8c19071588/example/v1/types_stable.go#L35

We should also align the text describing the constraint here with what the validation says. I like the example of "May not be removed once set".

everettraven (commented):
A few things I missed on the first round

features/features.go
```diff
@@ -930,4 +930,12 @@ var (
 						enhancementPR("https://github.com/openshift/enhancements/pull/1465").
 						enableIn(configv1.DevPreviewNoUpgrade).
 						mustRegister()
+
+	FeatureGateTLSAdherence = newFeatureGate("TLSAdherence").
+				reportProblemsToJiraComponent("kube-apiserver").
```
everettraven: I had missed this originally, but is this the appropriate Jira component to be reporting issues to?

I would expect this to be the team implementing and co-ordinating the work for this feature because this is the component that any bugs identified with this feature should be routed to.

If you feel strongly this should be routed to the kube-apiserver component, lets get an ack from the staff engineers/management of the control plane team that they are OK with fielding bugs related to this feature.

richardsonnick: For 1 and 2, it looks like completely bonked the rebase this morning. Apologies. Should be fixed now.

For 3: kube-apiserver may be the wrong component. I don't really have any strong convictions on who should own bugs for this. Does "Networking" fit the bill better? [The TLSProfile.curves fate is assigned to "Networking"](https://github.com/openshift/api/pull/2594/changes#diff-a8b6135d50534471326ea7bcd20e0f5eae25353f7788338060f718128a6a0b34R983).

everettraven: Because this is a bit of a cross-cutting change I'm not sure which is the best here. kube-apiserver doesn't seem unreasonable to me since it is on the `APIServer` config type, but we shouldn't make that decision for them.

Networking also seems reasonable if they are willing to field support requests and bugs related to this.

richardsonnick:
rebase

config/v1/types_apiserver.go
vincentdephily: The behavior for unknown values seem inconsistent with the behavior for unset/empty values. Shouldn't unknown values lead to a warning and setting the default value instead of `StrictAllComponents` ? This would be less surprising for both users and implementers ?

everettraven:
@richardsonnick Looks like this still has conflicts and needs another rebase.

config/v1/types_apiserver.go
everettraven: I'm not super opinionated on what the "unknown" case means for components, but I think `StrictAllComponents` is reasonable.

I suspect that we probably won't add many new allowed values here. Assuming that we do have a scenario where we've added a new value that might be unknown for some components, I think the current assumption of unknown != explicit legacy choice (that all components should be aware of) being that components should respect the configuration defined here is reasonable.

Issuing some kind of warning sounds reasonable to me. You wouldn't want to modify this value based on one component not understanding the value here because other components may understand the value.

richardsonnick:
These test failures look ephemeral retesting

richardsonnick:
Resolving conflicts

config/v1/types_apiserver.go
```diff
@@ -237,6 +269,32 @@ const (
...
+	// This value cannot be explicitly set and is only present when the field is omitted.
+	// When the field is omitted, the cluster defaults to LegacyExternalAPIServerComponentsOnly
+	// behavior. Components should treat this the same as LegacyExternalAPIServerComponentsOnly.
+	TLSAdherencePolicyNoOpinion TLSAdherencePolicy = ""
+
+	// TLSAdherencePolicyLegacyExternalAPIServerComponentsOnly means only the externally exposed
+	// API server components (kube-apiserver, openshift-apiserver, oauth-apiserver) honor
+	// the configured TLS profile. Other components continue to use their individual TLS
+	// configurations.
+	TLSAdherencePolicyLegacyExternalAPIServerComponentsOnly TLSAdherencePolicy = "LegacyExternalAPIServerComponentsOnly"
```
everettraven: Does this need to be updated based on recent agreement on a better enum value here? I believe we roughly agreed on `LegacyAdheringComponentsOnly`?

richardsonnick: Yes it does updating

config/v1/types_apiserver.go
```diff
@@ -62,6 +63,38 @@ type APIServerSpec struct {
...
 	// +optional
 	TLSSecurityProfile *TLSSecurityProfile `json:"tlsSecurityProfile,omitempty"`
+	// tlsAdherence controls which components in the cluster adhere to the TLS security profile
+	// configured on this APIServer resource.
+	//
+	// Valid values are "LegacyAdheringComponentsOnly" and "StrictAllComponents".
+	//
+	// When set to "LegacyAdheringComponentsOnly", components that already honor the cluster-wide
+	// TLS profile continue
+	// to do so. Components that do not already honor it continue to use their individual TLS
```
rhmdnd: Minor nit to tighten up these lines (or have your editor rewrap them), since this doc will be propagated to everything that uses this type.

config/v1/tests/apiservers.config.openshift.io/TLSAdherence.yaml
```diff
@@ -0,0 +1,88 @@
+apiVersion: apiextensions.k8s.io/v1
+name: "APIServer"
+crdName: apiservers.config.openshift.io
+featureGates:
+- TLSAdherence
+tests:
+  onCreate:
```
rhmdnd, line 7: Can we add a test that creates without `tlsAdherence` at all and asserts it's not in the resulting CR (testing the omitempty part)?

config/v1/tests/apiservers.config.openshift.io/TLSAdherence.yaml
```diff
@@ -0,0 +1,88 @@
...
+      initial: |
+        apiVersion: config.openshift.io/v1
+        kind: APIServer
+        spec:
+          tlsAdherence: StrictAllComponents
+      updated: |
+        apiVersion: config.openshift.io/v1
+        kind: APIServer
+        spec: {}
+      expectedError: "tlsAdherence may not be removed once set"
```
rhmdnd, line 116: Do we want to add a case to test setting adherence on update when it's initially not set?


```
- name: Should be able to set tlsAdherence from unset on update
   initial: |
     kind: APIServer
     spec: {}
   updated: |
     kind: APIServer
     spec:
       tlsAdherence: StrictAllComponents
```

config/v1/zz_generated.crd-manifests/0000_10_config-operator_01_apiservers-CustomNoUpgrade.crd.yaml
```diff
@@ -292,6 +292,41 @@ spec:
...
+              tlsAdherence:
+                description: |-
+                  tlsAdherence controls which components in the cluster adhere to the TLS security profile
+                  configured on this APIServer resource.
+
+                  Valid values are "LegacyAdheringComponentsOnly" and "StrictAllComponents".
+
+                  When set to "LegacyAdheringComponentsOnly", components that already honor the cluster-wide
+                  TLS profile continue
+                  to do so. Components that do not already honor it continue to use their individual TLS
```
rhmdnd: This should get rendered with proper line length if we fix the doc string in types_apiserver.go.

config/v1/types_apiserver.go
```diff
@@ -62,6 +63,38 @@ type APIServerSpec struct {
 	// The current default is the Intermediate profile.
 	// +optional
 	TLSSecurityProfile *TLSSecurityProfile `json:"tlsSecurityProfile,omitempty"`
+	// tlsAdherence controls which components in the cluster adhere to the TLS security profile
```
rhmdnd: Reading "which" in this sentence makes me think it's supposed to give me a list of components following TLS guidance from the API server and ones that aren't. Would it be better to s/which/if/ here?

richardsonnick: Ah good point

damdo:
As per @everettraven comment ([here](https://redhat-internal.slack.com/archives/C098FU5MRAB/p1773683720908859?thread_ts=1772660940.219029&cid=C098FU5MRAB)) we are marking this as `verified by CI`.

/verified by CI

PR #2681, opened 2026-01-30
OCPNODE-4060: Add additional storage configuration fields to ContainerRuntimeConfig

saschagrunert:
This enhancement extends ContainerRuntimeConfig API with three features for improved container storage flexibility:

- additionalLayerStores: Enable lazy image pulling via storage plugins (BYOS approach with stargz-snapshotter, nydus-storage-plugin)

- additionalImageStores: Read-only container image caches on shared or high-performance storage (NFS, SSD) for faster startup and reduced network overhead

- additionalArtifactStores: Configurable OCI artifact storage locations (SSD-backed storage, pre-populated caches, air-gapped deployments)

All features target AI/ML workload performance improvements and will ship as Tech Preview in 4.22 behind TechPreviewNoUpgrade feature gate.

Path validation is centralized in a shared `StorePath` type (absolute paths, 1-256 characters, a-z, A-Z, 0-9, '/', '.', '_', '-' only, no consecutive forward slashes). Graceful fallback to standard behavior on failure.

Requires https://github.com/openshift/enhancements/pull/1934


machineconfiguration/v1/types.go
```diff
@@ -892,6 +892,51 @@ type ContainerRuntimeConfiguration struct {
...
+	//
+	// Maximum of 10 stores allowed.
+	//
+	// When omitted, only the default artifact location (/var/lib/containers/storage/artifacts/) is used.
+	//
+	// +optional
+	// +listType=atomic
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=10
+	AdditionalArtifactStores []AdditionalArtifactStore `json:"additionalArtifactStores,omitempty"`
```
saschagrunert, line 949: This should be fixed now.

payload-manifests/crds/0000_80_machine-config_01_containerruntimeconfigs-TechPreviewNoUpgrade.crd.yaml
```diff
@@ -0,0 +1,197 @@
...
+                  pidsLimit:
+                    description: pidsLimit specifies the maximum number of processes
+                      allowed in a container
+                    format: int64
+                    type: integer
+                type: object
+              machineConfigPoolSelector:
+                description: |-
+                  machineConfigPoolSelector selects which pools the ContainerRuntimeConfig shoud apply to.
+                  A nil selector will result in no pools being selected.
```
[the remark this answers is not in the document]
saschagrunert: Out of scope of this PR, this is inherited by existing code.

payload-manifests/crds/0000_80_machine-config_01_containerruntimeconfigs-DevPreviewNoUpgrade.crd.yaml
```diff
@@ -0,0 +1,197 @@
...
+                      overlaySize specifies the maximum size of a container image.
+                      This flag can be used to set quota on the size of container images. (default: 10GB)
+                    pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                    x-kubernetes-int-or-string: true
+                  pidsLimit:
+                    description: pidsLimit specifies the maximum number of processes
+                      allowed in a container
+                    format: int64
+                    type: integer
+                type: object
```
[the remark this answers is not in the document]
saschagrunert: Should be fixed now, right?

payload-manifests/crds/0000_80_machine-config_01_containerruntimeconfigs-TechPreviewNoUpgrade.crd.yaml
```diff
@@ -0,0 +1,197 @@
...
+                      overlaySize specifies the maximum size of a container image.
+                      This flag can be used to set quota on the size of container images. (default: 10GB)
+                    pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                    x-kubernetes-int-or-string: true
+                  pidsLimit:
+                    description: pidsLimit specifies the maximum number of processes
+                      allowed in a container
+                    format: int64
+                    type: integer
+                type: object
```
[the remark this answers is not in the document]
saschagrunert: Should be fixed now, right?

payload-manifests/crds/0000_80_machine-config_01_containerruntimeconfigs-DevPreviewNoUpgrade.crd.yaml
```diff
@@ -0,0 +1,323 @@
...
+              machineConfigPoolSelector:
+                description: |-
+                  machineConfigPoolSelector selects which pools the ContainerRuntimeConfig shoud apply to.
+                  A nil selector will result in no pools being selected.
```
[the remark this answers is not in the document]
saschagrunert: This is out of scope and inherited by existing code.

payload-manifests/crds/0000_80_machine-config_01_containerruntimeconfigs-Default.crd.yaml
```diff
@@ -0,0 +1,197 @@
...
+                  pidsLimit:
+                    description: pidsLimit specifies the maximum number of processes
+                      allowed in a container
+                    format: int64
+                    type: integer
+                type: object
+              machineConfigPoolSelector:
+                description: |-
+                  machineConfigPoolSelector selects which pools the ContainerRuntimeConfig shoud apply to.
+                  A nil selector will result in no pools being selected.
```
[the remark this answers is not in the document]
saschagrunert: This is out of scope and inherited by existing code.

features.md
```diff
@@ -24,6 +24,7 @@
 | AWSDedicatedHosts| | | <span style="background-color: #519450">Enabled</span> | <span style="background-color: #519450">Enabled</span> | | | <span style="background-color: #519450">Enabled</span> | <span style="background-color: #519450">Enabled</span>  |
 | AWSDualStackInstall| | | <span style="background-color: #519450">Enabled</span> | <span style="background-color: #519450">Enabled</span> | | | <span style="background-color: #519450">Enabled</span> | <span style="background-color: #519450">Enabled</span>  |
 | AWSServiceLBNetworkSecurityGroup| | | <span style="background-color: #519450">Enabled</span> | <span style="background-color: #519450">Enabled</span> | | | <span style="background-color: #519450">Enabled</span> | <span style="background-color: #519450">Enabled</span>  |
+| AdditionalStorageConfig| | | <span style="background-color: #519450">Enabled</span> | <span style="background-color: #519450">Enabled</span> | | | <span style="background-color: #519450">Enabled</span> | <span style="background-color: #519450">Enabled</span>  |
```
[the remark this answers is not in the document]
saschagrunert: Auto-generated file

machineconfiguration/v1/types.go
saschagrunert: This is fixed, correct?

saschagrunert:
@qodo-code-review:

-  "Allow dots in path regex": Should be addressed
- "Consolidate redundant storage types into one": I'm going to keep separate types for future API flexibility.
- "Always serialize required path": Omitempty is required, otherwise the linter will complain on purpose.

saschagrunert:
@yuqi-zhang PTAL for shadow review if feasible.

features/features.go
```diff
@@ -361,6 +361,14 @@ var (
 				enableIn(configv1.Default, configv1.OKD, configv1.DevPreviewNoUpgrade, configv1.TechPreviewNoUpgrade).
 				mustRegister()
 
+	FeatureGateAdditionalStorageConfig = newFeatureGate("AdditionalStorageConfig").
+						reportProblemsToJiraComponent("MachineConfigOperator").
```
yuqi-zhang: Would you like the MCO team to own this feature, or should we report issues to the node/crio component?

machineconfiguration/v1/types.go
```diff
@@ -904,6 +958,66 @@ const (
 	ContainerRuntimeDefaultRuntimeDefault = ContainerRuntimeDefaultRuntimeCrun
 )
 
+// AdditionalLayerStore defines a read-only storage location for Open Container Initiative (OCI) container image layers.
```
yuqi-zhang, line 971: There's quite a bit of duplication here. What do you think about either:
1. have Path be of a new type, with validation on the type directly (e.g. https://github.com/openshift/api/blob/master/machineconfiguration/v1/types_machineosconfig.go#L198)
2. collapse AdditionalLayerStore, AdditionalImageStore, AdditionalArtifactStore into one type called AdditionalStore, and have the difference in godoc reflect via the subfields in ContainerRuntimeConfiguration?

machineconfiguration/v1/types.go
```diff
@@ -892,6 +892,60 @@ type ContainerRuntimeConfiguration struct {
 	// +kubebuilder:validation:Enum=crun;runc
 	// +optional
 	DefaultRuntime ContainerRuntimeDefaultRuntime `json:"defaultRuntime,omitempty"`
+
+	// additionalLayerStores configures additional read-only container image layer store locations for Open Container Initiative (OCI) images.
+	//
+	// Layers are checked in order: additional stores first, then the default location.
```
yuqi-zhang, line 899: nit: reading the godoc of the three fields we're adding is a bit confusing, since they're phrased very similarly.

Would it make sense to link to upstream docs (not sure if that's something we do) or have a bit more description on the differences of behaviour?

features/features.go
saschagrunert: How did we do it for other features as part of the container runtime config @QiWang19 ?

machineconfiguration/v1/types.go
saschagrunert: The main idea was to be able to separate the configurations later on when required. Means that `AdditionalLayerStore` may have different configuration additions than `AdditionalArtifactStore`. This would allow us the flexibility to develop the features into certain directions without coupling them together from day 1. WDYT?

machineconfiguration/v1/types.go
saschagrunert: Yeah I was thinking into a similar direction. Unfortunately we don't have much upstream docs and would require extensive OpenShift documentation about the features functionalities. 

We just have: https://github.com/containers/container-libs/blob/d518ce6ca377c0d306551660d13ba7cdc61d41a8/storage/docs/containers-storage.conf.5.md#storage-options-table

> additionalimagestores=[] Paths to additional container image stores. Usually these are read/only and stored on remote network shares.

And even less for the additional layer store…

features/features.go
QiWang19: We don't have other container runtime configs that are managed by the featuregate. 
It is related to Jira component I think we should use 'node' here.

saschagrunert: Fixed 👍

machineconfiguration/v1/types.go
yuqi-zhang: I'm fine with that. Curious if Joel or Bryce has any better suggestions as well, so will wait for their review here.

machineconfiguration/v1/types.go
yuqi-zhang: Ack, any additional godoc here that you think makes sense would be helpful, but matching what we have in container-libs seem fine as a start.

machineconfiguration/v1/types.go
```diff
@@ -904,6 +958,66 @@ const (
...
+	// This field is required and must:
+	//   - Have length between 1 and 256 characters
+	//   - Start with '/' (absolute path)
+	//   - Contain only: a-z, A-Z, 0-9, '/', '.', '_', '-' (no spaces or special characters)
+	//
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=256
+	// +kubebuilder:validation:XValidation:rule="self.matches('^/[a-zA-Z0-9/._-]+$')",message="path must be absolute and contain only alphanumeric characters, '/', '.', '_', and '-'"
+	Path string `json:"path,omitempty"`
```
yuqi-zhang: I recall I made a suggestion to lift this into a separate type. But for some reason I can't see it in the comment history

saschagrunert:
Rebased on top of the latest master. PTAL @openshift/api-reviewers

saschagrunert:
@JoelSpeed PTAL

machineconfiguration/v1/types.go
JoelSpeed: > The main idea was to be able to separate the configurations later on when required. Means that AdditionalLayerStore may have different configuration additions than AdditionalArtifactStore. This would allow us the flexibility to develop the features into certain directions without coupling them together from day 1. 

This would have been my suggestion, so +1 for the current pattern

But the suggestion for a separate type for the path makes sense. Unless we think that even the path validation could diverge between the three?

machineconfiguration/v1/types.go
```diff
@@ -905,6 +959,66 @@ const (
...
+	// path specifies the absolute location of the additional layer store.
+	//
+	// The path must exist on the node before configuration is applied.
+	// When a container image is requested, layers found at this location will be used instead of
+	// retrieving from the registry.
+	//
+	// This field is required and must:
+	//   - Have length between 1 and 256 characters
+	//   - Start with '/' (absolute path)
+	//   - Contain only: a-z, A-Z, 0-9, '/', '.', '_', '-' (no spaces or special characters)
```
JoelSpeed: Have you checked that this renders correctly in `oc explain`? We tend to prefer prose rather than bullets as different representations of this doc don't preserve spacing as it is written here

machineconfiguration/v1/types.go
```diff
@@ -905,6 +959,66 @@ const (
...
+	//
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=256
+	// +kubebuilder:validation:XValidation:rule="self.matches('^/[a-zA-Z0-9/._-]+$')",message="path must be absolute and contain only alphanumeric characters, '/', '.', '_', and '-'"
```
JoelSpeed: Are there any characters that are not allowed sequentially? Possibly `//`?

machineconfiguration/v1/types.go
JoelSpeed: Is that a type alias that could be re-used? So that the validations upon it aren't repeated three times? That suggestion makes sense IMO

machineconfiguration/v1/types.go
saschagrunert: Agreed, introduced a shared `StorePath` type with all validation markers.

machineconfiguration/v1/types.go
saschagrunert: Switched to prose and listed the allowed characters inline.

machineconfiguration/v1/types.go
saschagrunert: Added `\!self.contains('//')` XValidation rule on `StorePath` with test cases.

machineconfiguration/v1/types.go
saschagrunert: Done, all three structs now use `Path StorePath`.

JoelSpeed (commented):
/lgtm

PR #2693, opened 2026-02-05
OCPBUGS-74511: remove RouteExternalCertificate feature gate

jcmoraisjr:
RouteExternalCertificate is enabled by default, this is the second half of the changes to remove it and should be merged only after https://github.com/openshift/cluster-ingress-operator/pull/1355 being merged.

Jira: https://issues.redhat.com/browse/OCPBUGS-74511


JoelSpeed:
Looks like the codegen is in need of update here

jcmoraisjr:
Thanks Joel. I couldn't reproduce the same failure locally via `make verify-codegen-crds`, so just removed the generated file reported in the error. When trying `make verify`, it failed with `Prerelease lifecycle generation is out of date. Please run hack/update-prerelease-lifecycle-gen.sh` even after successfully running the mentioned script. I hope it succeeds now, otherwise any hints to fully test locally is appreciated <3

JoelSpeed:
@jcmoraisjr Make sure you have pulled and rebased onto the latest `master` commit.
Then you want to try `PROTO_OPTIONAL=1 make update` to run the full suite of generators

If it's not putting the openapi generated content back, please try setting a `$GOPATH` and checking the repo out at `$GOPATH/src/github.com/openshift/api` before then running that command again

Miciah:
/cc @chiragkyal 

FYI that we're removing the "RouteExternalCertificate" featuregate for which you were the owner.

melvinjoseph86:
Marking this PR as verified as this https://github.com/openshift/cluster-ingress-operator/pull/1355#issuecomment-3879072103
/verified by @mjoseph

JoelSpeed:
Could we get this PR rebased please?

PR #2695, opened 2026-02-05
OCPBUGS-32275: Add ingress.spec.domain immutability validation

grzpiotrowski:
This PR fixes [OCPBUGS-32275](https://issues.redhat.com/browse/OCPBUGS-32275).

Adds `spec.domain` field validation in the Ingress config CRD to make it immutable and match the documentation. Prior to this commit the domain value could be changed and cause degraded state of some cluster operators.

### DIFF
--- a/config/v1/tests/ingresses.config.openshift.io/AAA_ungated.yaml
+++ b/config/v1/tests/ingresses.config.openshift.io/AAA_ungated.yaml
@@ -12,3 +12,45 @@ tests:
         apiVersion: config.openshift.io/v1
         kind: Ingress
         spec: {}
+    - name: Should be able to create an Ingress with domain set
+      initial: |
+        apiVersion: config.openshift.io/v1
+        kind: Ingress
+        spec:
+          domain: apps.example.com
+      expected: |
+        apiVersion: config.openshift.io/v1
+        kind: Ingress
+        spec:
+          domain: apps.example.com
+  onUpdate:
+    - name: Should not be able to change domain once set
+      initial: |
+        apiVersion: config.openshift.io/v1
+        kind: Ingress
+        spec:
+          domain: apps.example.com
+      updated: |
+        apiVersion: config.openshift.io/v1
+        kind: Ingress
+        spec:
+          domain: test.example.com
+      expectedError: "domain is immutable once set"
+    - name: Should be able to update other fields without changing domain
+      initial: |
+        apiVersion: config.openshift.io/v1
+        kind: Ingress
+        spec:
+          domain: apps.example.com
+      updated: |
+        apiVersion: config.openshift.io/v1
+        kind: Ingress
+        spec:
+          domain: apps.example.com
+          appsDomain: custom.example.com
+      expected: |
+        apiVersion: config.openshift.io/v1
+        kind: Ingress
+        spec:
+          domain: apps.example.com
+          appsDomain: custom.example.com
--- a/config/v1/types_ingress.go
+++ b/config/v1/types_ingress.go
@@ -43,6 +43,7 @@ type IngressSpec struct {
 	// default ingresscontroller domain will follow this pattern: "*.<domain>".
 	//
 	// Once set, changing domain is not currently supported.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="domain is immutable once set"
 	Domain string `json:"domain"`
 
 	// appsDomain is an optional domain to use instead of the one specified
--- a/payload-manifests/crds/0000_10_config-operator_01_ingresses.crd.yaml
+++ b/payload-manifests/crds/0000_10_config-operator_01_ingresses.crd.yaml
@@ -129,6 +129,9 @@ spec:
 
                   Once set, changing domain is not currently supported.
                 type: string
+                x-kubernetes-validations:
+                - message: domain is immutable once set
+                  rule: self == oldSelf
               loadBalancer:
                 description: |-
                   loadBalancer contains the load balancer details in general which are not only specific to the underlying infrastructure
--- a/config/v1/zz_generated.crd-manifests/0000_10_config-operator_01_ingresses.crd.yaml
+++ b/config/v1/zz_generated.crd-manifests/0000_10_config-operator_01_ingresses.crd.yaml
@@ -129,6 +129,9 @@ spec:
 
                   Once set, changing domain is not currently supported.
                 type: string
+                x-kubernetes-validations:
+                - message: domain is immutable once set
+                  rule: self == oldSelf
               loadBalancer:
                 description: |-
                   loadBalancer contains the load balancer details in general which are not only specific to the underlying infrastructure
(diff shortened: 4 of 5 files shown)


payload-manifests/crds/0000_10_config-operator_01_ingresses.crd.yaml
```diff
@@ -129,6 +129,9 @@ spec:
 
                   Once set, changing domain is not currently supported.
                 type: string
+                x-kubernetes-validations:
+                - message: domain is immutable once set
+                  rule: self == oldSelf
```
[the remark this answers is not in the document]
grzpiotrowski: Looking at the test cases the domain value can be set initially when the `self == oldSelf` validation is in place. Rewording the validation error to "domain is immutable" is something to consider maybe, but the current on follows the wording existing in the documentation.

grzpiotrowski:
Not sure now why the e2e ci jobs don't run. Do they need to be triggered explicitly?

lihongan:
Pre-merge tested and got below

```
$ oc patch ingress.config cluster --type=merge -p '{"spec":{"domain":"abcd.example.com"}}'
The Ingress "cluster" is invalid: spec.domain: Invalid value: "string": domain is immutable once set
```
Looks the `Invalid value: "string"` in error message is inaccurate, the actual invalid value is `abcd.example.com` right ?

config/v1/tests/ingresses.config.openshift.io/AAA_ungated.yaml
```diff
@@ -12,3 +12,45 @@ tests:
...
+        apiVersion: config.openshift.io/v1
+        kind: Ingress
+        spec:
+          domain: apps.example.com
+      updated: |
+        apiVersion: config.openshift.io/v1
+        kind: Ingress
+        spec:
+          domain: test.example.com
+      expectedError: "domain is immutable once set"
```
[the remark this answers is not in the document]
grzpiotrowski: To be considered, but not necessary imo. The domain is always set by the installer when the cluster config is generated here: [openshift/installer/pkg/asset/manifests/ingress.go#L113](https://github.com/openshift/installer/blob/69f3d1db71131332934012e4f9e9cfd64c4d4d8c/pkg/asset/manifests/ingress.go#L113).

grzpiotrowski:
> Pre-merge tested and got below
> 
> ```
> $ oc patch ingress.config cluster --type=merge -p '{"spec":{"domain":"abcd.example.com"}}'
> The Ingress "cluster" is invalid: spec.domain: Invalid value: "string": domain is immutable once set
> ```
> 
> Looks the `Invalid value: "string"` in error message is inaccurate, the actual invalid value is `abcd.example.com` right ?

Thank you. I agree it reads a bit confusing, but I don't think this is something configurable and this is a standard output for the `+kubebuilder:validation:XValidation`. I'm assuming the message is trying to say that there is an invalid [string](https://github.com/grzpiotrowski/api/blob/a58ab7393c52e20aa9a2e1b40fb085c6e4300284/payload-manifests/crds/0000_10_config-operator_01_ingresses.crd.yaml#L131) type value and then we provide the custom detailed message `domain is immutable once set`.

Example: [Enforce CRD Immutability with CEL Transition Rules: Immutability upon object creation](https://kubernetes.io/blog/2022/09/29/enforce-immutability-using-cel/#immutability-upon-object-creation)

jcmoraisjr:
/lgtm
Great job!

payload-manifests/crds/0000_10_config-operator_01_ingresses.crd.yaml
Thealisyed, line 134: So the rule is `self == OldSelf `is immutable but what happens if the domain filed was never empty on an existing cluster? Does that cause a breaking change then?

everettraven: These validations should ratchet.

If the domain field was specified as a non-empty string on an existing cluster, this would take effect and make it so that updates cannot change the field.

Based on the bug this is linked to, that was never supported and causes degradation in the cluster, which means it should generally be OK to tighten this.

Where this _would_ be interesting is if `""` is a valid value for this field - which it looks like it might be from an admission standpoint and is something I missed. @grzpiotrowski what happens if the field this validation was added to is set to `""`? Should that be able to be modified to set it to a non-empty string?

config/v1/types_ingress.go
```diff
@@ -43,6 +43,7 @@ type IngressSpec struct {
 	// default ingresscontroller domain will follow this pattern: "*.<domain>".
 	//
 	// Once set, changing domain is not currently supported.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="domain is immutable once set"
 	Domain string `json:"domain"`
```
everettraven, line 47: Same as a comment from another thread, but copying here for posterity: Do we want to make `""` unable to be changed here? What happens today if this field is set to `""`?

everettraven:
Prow doesn't seem to like me attempting to remove my approval. Going to add a hold for now until we get my recent question addressed.

/hold

config/v1/types_ingress.go
grzpiotrowski: The Domain field is always set by the installer [here](https://github.com/openshift/installer/blob/69f3d1db71131332934012e4f9e9cfd64c4d4d8c/pkg/asset/manifests/ingress.go#L113) to a non empty value.

In a scenario that ingress config `.spec.domain` is `""` that's a broken ingress.
The ingress config `.spec.domain` value is used to set the default `.status.domain` value on the ingress controller [(setDefaultDomain)](https://github.com/openshift/cluster-ingress-operator/blob/abfb6fa7b9c95dc54fb3988e84297912ad88e0e4/pkg/operator/controller/ingress/controller.go#L453).

The IC `.status.domain` is [validated](https://github.com/openshift/cluster-ingress-operator/blob/abfb6fa7b9c95dc54fb3988e84297912ad88e0e4/pkg/operator/controller/ingress/controller.go#L927) by the cluster ingress operator and cannot be empty.

From looking at the operator code, if the config value is changed from `""` to a valid value the ingress controller would be admitted ([operator code](https://github.com/openshift/cluster-ingress-operator/blob/abfb6fa7b9c95dc54fb3988e84297912ad88e0e4/pkg/operator/controller/ingress/controller.go#L366)). I didn't test that in a cluster.

As a reference, the non-default value which can be set on the ingress controller's `.spec.domain` field has [validation](https://github.com/openshift/api/blob/22c744831738b34ccbebc2295b5c7e91707942e0/operator/v1/types_ingress.go#L85-L89) to ensure it is a valid DNS domain name.

everettraven: Thanks for the analysis here. It seems correct that in practice an empty value here causes degradation in the cluster and should never happen.

I did a bit more investigation here and it looks like Hypershift embeds this type in their APIs: https://github.com/openshift/hypershift/blob/main/api/hypershift/v1beta1/hostedcluster_types.go#L1993

They also do some defaulting behavior when this field is set to `""`, which means that we probably shouldn't ratchet this with a minimum length as it risks breaking a workflow there.

This change should be OK as-is then. Thanks for helping me verify.

grzpiotrowski:
/test e2e-aws-ovn-hypershift

unrelated issue with winding down the test cluster I assume:
```
--- FAIL: TestCreateClusterProxy/Teardown (4029.87s)
```

lihongan:
/verified by @lihongan 
see https://github.com/openshift/api/pull/2695#issuecomment-3876611119

PR #2701, opened 2026-02-09
Feature gate CVO application of CompatibilityRequirement CRD

JoelSpeed:
We only want this CRD to be applied when the `CRDCompatibilityRequirementOperator` feature gate is enabled on a cluster, this annotates the CRD so that CVO will only apply it in that case.

It also adds the CompatibilityRequirement to the payload from this repository. This is our intention going forward, APIs should be deployed from here, not spread around various operator repos

### DIFF
--- a/hack/update-payload-crds.sh
+++ b/hack/update-payload-crds.sh
@@ -3,6 +3,7 @@
 source "$(dirname "${BASH_SOURCE}")/lib/init.sh"
 
 crd_globs="\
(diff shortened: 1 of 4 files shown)


JoelSpeed:
/verified by @JoelSpeed 

Confident that we don't need further verification on this

/approve

JoelSpeed:
/override ci/prow/okd-scos-images

Unrelated failure

PR #2752, opened 2026-03-10
CORENET-6714: Enable Network Observability on Day 0

OlivierCazade:
### Summary

This PR adds the `installNetworkObservability` field to the Network CRD (`config.openshift.io/v1`) to support automatic installation of the Network Observability operator during cluster creation.

### Changes

- Add `InstallNetworkObservability` string field to `NetworkSpec`
- Field defaults to Enable (when omitted), enabling network observability by default
- When explicitly set to Disable, automatic installation is skipped

### Behavior

When omitted or set to Enable, the Network Observability operator and FlowCollector are deployed at bootstrap time, providing immediate network visibility for troubleshooting packet drops, latencies, and DNS tracking.

When explicitly set to Disable, no automatic installation occurs (though the operator can still be installed manually post-installation).

### Related

- Enhancement Proposal: openshift/enhancements#1908

/cc @stleerh


config/v1/types_network.go
```diff
@@ -86,6 +86,13 @@ type NetworkSpec struct {
 	//
 	// +optional
 	NetworkDiagnostics NetworkDiagnostics `json:"networkDiagnostics"`
+
+	// installNetworkObservability is an optional field that enables network observability
+	// when omitted or set to enable.  If the field is set to disable, it does nothing.
+	// Valid values are "", "Enable", "Disable".
+	// +kubebuilder:validation:Enum:="";Enable;Disable
+	// +optional
+	InstallNetworkObservability *string `json:"installNetworkObservability,omitempty"`
```
stleerh: "Disable" is misleading, because it is not disabling Network Observability; it does nothing.  In the Enhancement Proposal, it uses the term "not enable" to mean it doesn't enable Network Observability.

What about using the values "true", "false" and ""?  "false" would imply that it would not install Network Observability.

everettraven:
@OlivierCazade All new fields must be feature gated and go through the standard feature promotion process. Please add a new feature gate, disabled by default, and gate this new field.

More information on adding a new feature gate can be found here: https://github.com/openshift/api?tab=readme-ov-file#adding-new-featuregates

More information on adding a new field behind a feature gate can be found here: https://github.com/openshift/api?tab=readme-ov-file#adding-new-fields

config/v1/types_network.go
everettraven: We strongly discourage the use of booleans in OpenShift APIs because fields that start as booleans often evolve to need more than `true || false` as allowed values.

We prefer the use of enums that have meaningful values to end-users. I agree that `installNetworkObservability: Enable | Disable` is not very clear to an end user.

I did a quick read of the EP - do you envision there ever needing to be a way for customers to apply any custom configuration for the day 0 network observability enablement?

stleerh: Choosing common 'boolean' values like true/false, yes/no, and 0/1 is easier to remember than some arbitrary dual values like enable/disable.  If I try to come up with another term besides "disable", they all sound bad like "null" or "no-op".  I like yes/no as well, but it will have the same concerns you raised.

For day 0, the answer is no. Customers will likely need to make changes on day 2.  Even if that were to change somewhere down the line, we would still want this knob to enable it separately.  That's already the case for many features, which typically have an attribute named `enable` under the particular feature.

The bottom line is whether the attribute **and** the value are meaningful to the user, not the just the value itself.  enable/disable by itself is no more clearer than true/false.  It has to be associated with the attribute `installNetworkObservability` to determine whether it's clear or not.  So ask yourself, "Is it clear what `installNetworkObservability` means when it is true or false?" My answer is yes, and that's because the attribute has a verb in there.  If the attribute was `NetworkObservability`, then the values like `enable` is much better than `true`.

everettraven: >For day 0, the answer is no. Customers will likely need to make changes on day 2. Even if that were to change somewhere down the line, we would still want this knob to enable it separately. That's already the case for many features, which typically have an attribute named enable under the particular feature

Good to know. The reason I asked is because if we were intending to enable some kind of custom configuration as part of day 0 installation, a simple enablement field may result in a slightly "clunkier" configuration experience in the future by having separate fields for stating installation intent and then custom configuration.

If you were intending to do day 0 custom configuration, I'd have suggested a structure more along the lines of:
```yaml
networkObservability:
  installationPolicy: Install | DoNotInstall | InstallWithCustomization
  installWithCustomization:
    # customizable fields here
```
(note: naming is just for an example and not necessarily exactly what I would expect).

I also think that having a nested structure now is generally better for future flexibility rather than locking yourself into having multiple co-located `*NetworkObservability` fields in the future. In my experience, that starts to get confusing for end-users.

>The bottom line is whether the attribute and the value are meaningful to the user, not the just the value itself. enable/disable by itself is no more clearer than true/false. It has to be associated with the attribute installNetworkObservability to determine whether it's clear or not. So ask yourself, "Is it clear what installNetworkObservability means when it is true or false?" My answer is yes, and that's because the attribute has a verb in there. If the attribute was NetworkObservability, then the values like enable is much better than true.

Sure. So what happens if in the future you'd like to add a different way to specify the installation of the network observability feature beyond pure yes/no semantics?

Something like:
```yaml
networkObservabilityInstallationPolicy: Install | DoNotInstall
```
gives you more flexibility in the future to add new values that are descriptive beyond `True | False`.

All this being said, if you folks feel strongly that true/false semantics provide the best experience for the end-user then 👍

config/v1/types_network.go
```diff
@@ -86,6 +86,13 @@ type NetworkSpec struct {
 	//
 	// +optional
 	NetworkDiagnostics NetworkDiagnostics `json:"networkDiagnostics"`
+
+	// installNetworkObservability is an optional field that enables network observability
+	// when omitted or set to enable.  If the field is set to disable, it does nothing.
+	// Valid values are "", "Enable", "Disable".
+	// +kubebuilder:validation:Enum:="";True;False
+	// +optional
+	InstallNetworkObservability *string `json:"installNetworkObservability,omitempty"`
```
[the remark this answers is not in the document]
everettraven: This is correct and needs to be done. This cannot just go straight to GA.

config/v1/types_network.go
stleerh: I appreciate the comments.  There are no plans to extend this, because the thinking is that if you want do anything else, it should be done in the Network Observability Operator (NOO) on day 2.

To muddy the waters a bit more, I'm not that thrilled with the attribute name `installNetworkObservability`, because it doesn't just install NOO, it also "enables" it by configuring the custom resource (CR).  Here are a few more suggestions:

```
installNetworkObservability: auto|manual   # suggested by Olivier
networkObservabilityPolicy: enable|doNotEnable   # camelCase, title case, or snake_case?
autoEnableNetworkObservability: true|false
autoEnableObservability: true|false  # shortened since it's already under `networking`

Comment (jotak) [reply config/v1/types_network.go]: > I appreciate the comments. There are no plans to extend this, because the thinking is that if you want do anything else, it should be done in the Network Observability Operator (NOO) on day 2.

@stleerh there might be exceptions, related to the OLM subscription, that is not covered by NOO directly. E.g. manual vs auto upgrade policy, or operator resource requests/limits ; I'm not saying we want to cover that now, but could potentially be a future change.

So I see the value of having a nested structure
```yaml
networkObservability:
  installationPolicy: InstallAndEnable | DoNotInstall # maybe later: | InstallWithCustomization
  # maybe later:
  #  installWithCustomization:
    #    upgradePolicy: Auto | Manual
```

_(edit: s/Install/InstallAndEnable/)_

everettraven: I would strongly encourage building for future flexibility here if you envision something like that being a reasonable request in the future.

stleerh: My main gripe is that it all feels over-engineered for something that's unlikely to change.  Someone is going to want to not enable Network Observability, and they would have to figure out that you need to do:
```
networkObservability:
  installationPolicy: DoNotInstall
```

when all it really is:
```
installNetworkObservability: false
```

OlivierCazade:
@everettraven My understanding of the failing CRD test is that this is an existing isue appearing because the network CRD is split for all feature sets.
This was already bypassed here : https://github.com/openshift/api/pull/1886
Is that correct ?

config/v1/types_network.go
jotak: @stleerh imagine that, next release, we're asked to have netobserv operator installed but without flowcollector enabled - just have the console plugin deployed, that puts some page placeholder in the console. And later on, allow to tune the olm upgrade policy. That's all plausible scenarios IMO. We would start adding more boolean flags here and there to address those changes, all defined in the root structure, it could quickly become a mess. The problem here is that we can't do breaking changes later. It has to be done right from the start, even if that sounds like an overkill for currently just a single field.

everettraven: >The problem here is that we can't do breaking changes later. It has to be done right from the start, even if that sounds like an overkill for currently just a single field.

This is the main reason to do it now. If you don't do something like:
```yaml
networkObservability:
  installationPolicy: DoNotInstall
```
and then the ask that @jotak is imagining comes through, you'll end up with something like:
```yaml
installNetworkObservability: { true | false }
customizedNetworkObservability: ...
```
I find this to be more confusing - what does it mean to set something like `customizedNetworkObservability` when `installNetworkObservability` is set to `false`? Does it mean that you'll actually install network observability but with the customizations applied? If that is the case, how is that different than setting `installNetworkObservability` to `true` and specifying it?

Even though something like
```yaml
networkObservability:
  installationPolicy: DoNotInstall
```
is more verbose, one more line for end users to type is - IMO - a reasonable trade off for stronger flexibility in adding new API functionality in the future, even if you end up not needing it.

stleerh: That is a common scenario.  It allows you to preserve your configuration so when you do enable it, it will be used.  Even if you go the nested route, you would want to allow this.

everettraven:
@OlivierCazade I confirmed with @JoelSpeed that adding the subresource that the check is complaining about now would be a breaking change.

I think we can safely override the check here when we are ready to merge.

config/v1/types_network.go
everettraven: >That is a common scenario. It allows you to preserve your configuration so when you do enable it, it will be used. Even if you go the nested route, you would want to allow this.

While I understand the perspective, I think the preservation of a configuration that is not in use because someone _might_ enable it and to make it so they don't have to write the configuration again is not strong enough a reason that it outweighs the potential confusion associated with it for future readers or people scanning the type for some kind of configuration as a signal that it is enabled in that mode.

Doing something like `kubectl get {resource} -o yaml > backup.yaml` seems easy enough to me to keep a backup of the configuration to restore when you decide to enable it.

config/v1/types_network.go
```diff
@@ -304,3 +311,31 @@ type NetworkDiagnosticsTargetPlacement struct {
...
+// NetworkObservabilitySpec defines the configuration for network observability installation
+// +kubebuilder:validation:MinProperties=1
+type NetworkObservabilitySpec struct {
+	// installationPolicy controls whether network observability is installed during cluster deployment.
+	// Valid values are "", "InstallAndEnable" and "DoNotInstall".
+	// When set to "", network observability will be installed unless this is a SNO cluster.
+	// When set to "InstallAndEnable", network observability will be installed and enabled.
+	// When set to "DoNotInstall", network observability will not be installed.
+	// +optional
+	InstallationPolicy *NetworkObservabilityInstallationPolicy `json:"installationPolicy,omitempty"`
```
everettraven, line 340: From a comment on the EP, it sounded like once this is set to `InstallAndEnable` setting it to `DoNotInstall` will do nothing.

Would it make sense to make this field immutable to prevent a false sense of being able to modify this behavior?

stleerh: No, this is how it behaves.  I believe this is what you're referring to.

> [dave-tucker](https://github.com/dave-tucker)
What if a users changes from InstallAndEnable to DoNotInstall?
Does CNO uninstall the operator? Or is the user expected to clean this up manually?

The setting could be:
```
spec:
  networkObservability:
    installationPolicy: InstallAndEnable
```

or this:
```
spec:
  networkObservability:
    installationPolicy: DoNotInstall

```
or the configuration could not exist at all, which implies the first setting (i.e. enable by default).

Let's say you have the first setting.  When OpenShift installs, it enables Network Observability.  Now the user changes the configuration to `installationPolicy: DoNotInstall` in the Network CR.  What will it do?  The answer is it will do nothing.  It will not remove Network Observability.

everettraven: Yes, that is what I am referring to.

If modifying the value does nothing, why should I be able to modify it after it has been set?

stleerh: Replying back after further discussion with the team.  There are three possible values.

`installationPolicy` attribute does not exist:  This is the default behavior.  It enables Network Observability once if it isn't already there.  If you later uninstall it, it won't try to enable it again.
`InstallAndEnable`: It will always enable Network Observability if it isn't enabled.  
`DoNotInstall`: It will always do nothing.

If Network Observability is enabled by default, you might still want to change it to `InstallAndEnable`.

---

FYI, of the three PRs, we want to merge this one first.

everettraven: So if I change from `DoNotInstall` to `InstallAndEnable` it will enable it if it isn't enabled? Going back to `DoNotInstall` from `InstallAndEnable` will do nothing (i.e it will still be installed)?

Maybe `DoNotInstall` should just be called something like `NoAction` to more clearly represent that no action will be taken?

stleerh: I agree `NoAction` is a better enum value.

config/v1/types_network.go
```diff
@@ -304,3 +311,33 @@ type NetworkDiagnosticsTargetPlacement struct {
...
+)
+
+// NetworkObservabilitySpec defines the configuration for network observability installation
+// +kubebuilder:validation:MinProperties=1
+type NetworkObservabilitySpec struct {
+	// installationPolicy controls whether network observability is installed during cluster deployment.
+	// Valid values are "", "InstallAndEnable" and "NoAction".
+	// When set to "", network observability will be installed once unless this is a SNO cluster.
+	// When set to "InstallAndEnable", network observability will be installed and enabled.
+	// When set to "NoAction", nothing will be done regarding Network observability.
```
everettraven, line 333: Can we be a bit more descriptive as to the behavior for `""` and `InstallAndEnable`?

`""` == install once if not already installed, no action taken after that. If already installed, no actions taken ?

`InstallAndEnable` == always install our version? Or always install our version if not already installed, if we detect it was uninstalled re-install it?

config/v1/types_network.go
```diff
@@ -86,6 +86,13 @@ type NetworkSpec struct {
 	//
 	// +optional
 	NetworkDiagnostics NetworkDiagnostics `json:"networkDiagnostics"`
+
+	// networkObservability is an optional field that configures network observability installation
+	// during cluster deployment (day-0).
+	// When omitted, network observability will be installed unless this is a SNO cluster.
+	// +openshift:enable:FeatureGate=NetworkObservabilityInstall
+	// +optional
+	NetworkObservability NetworkObservabilitySpec `json:"networkObservability,omitempty,omitzero"`
```
everettraven, line 95: Is setting `networkObservability.installationPolicy: ""` equivalent to just omitting the `networkObservability` field?

OlivierCazade: Not exactly, `""` is supposed to be the default value. In this situation, a user may not a understand why a operator that was uninstalled get installed again.

So `""` does not try to reinstall Network Observability while `InstallAndEnable` does.

everettraven: If I omit the `networkObservability` field entirely, what is the `installationPolicy` that takes place?

OlivierCazade: If `networkObservability` is omitted, the policy used is `""`

everettraven: Gotcha.

Instead of allowing the `installationPolicy: ""` value, let's make `installationPolicy` a required field that only allows the values `InstallAndEnable` and `NoAction`. This makes it such that the only way to configure the default behavior is for the parent `networkObservability` field to be omitted.

OlivierCazade: I tried implementing this (see the last pushed version), but I cannot have something that does not trigger a linting error.

Either I need to add the `MinProperties` annotation or to remove the pointer.

Removing the min properties:
`config/v1/types_network.go:95:2: minlength: field NetworkSpec.NetworkObservability type NetworkObservabilitySpec must have a minimum properties, add kubebuilder:validation:MinProperties marker (kubeapilinter)
`

Switching to a pointer:
`config/v1/types_network.go:95:2: optionalfields: field NetworkSpec.NetworkObservability does not allow the zero value. The field does not need to be a pointer. (kubeapilinter)`

Could you help on this ? I guess I am missing something.

everettraven: It's because `""` is still a valid value for the `installationPolicy` field.

If you remove that as an allowed value, make `installationPolicy` required, drop the min properties marker, and make `NetworkObservability` a non-pointer you shouldn't have any linter errors.

OlivierCazade: It works this way thanks.

I am probably missing something, but since NetworkObservability is not a pointer anymore doesn't it mean that it has to be initialised and either InstallAndEnable or NoAction chosen (since installationProperty is required)?

everettraven: >since NetworkObservability is not a pointer anymore doesn't it mean that it has to be initialised and either InstallAndEnable or NoAction chosen

Nope, in CRDs an optional field works via omission of the key not whether or not Go-based serialization.

With it being a non-pointer you would just use a check like `NetworkObservability.InstallationPolicy == ""` to determine that `networkObservability` was omitted because otherwise `installationPolicy` cannot be `""`

How this validation is now means that _IF_ the `networkObservability` field is specified, they _must_ choose a value for `installationPolicy`

everettraven:
@OlivierCazade Let's get integration tests passing and then this should be good to go!

config/v1/types_network.go
OlivierCazade: Thanks for the clarification

config/v1/tests/networks.config.openshift.io/NetworkObservabilityInstall.yaml
```diff
@@ -0,0 +1,46 @@
...
+        kind: Network
+        spec:
+          networkObservability:
+            installationPolicy: "InstallAndEnable"
+      expected: |
+        apiVersion: config.openshift.io/v1
+        kind: Network
+        spec:
+          networkObservability:
+            installationPolicy: "InstallAndEnable"
```
everettraven: Should this be made into a negative test because `networkObservability.installationPolicy: ""` is no longer allowed?

Or maybe this should transition to a "can omit `networkObservability` field" style test case?

OlivierCazade: Yes sorry, I converted the test to check for an error when installationPolicy is set to ""

everettraven (approved):
/lgtm

everettraven:
Would be a breaking change to address the verify-crd-schema concern. Overriding.

OlivierCazade:
Rebased the branch due to conflicts

OlivierCazade:
@everettraven I am not sure to understand what the verified label imply and who is responsible for adding it.

Could you help ?

everettraven:
@OlivierCazade The `verified` label is generally meant for someone to add when they have verified that the changes in place are working as expected. In general, this ends up being QE, but can also be non-QE (comment command is `/verified by @{github_handle}`.

if you'd like to have this be verified later (i.e post-merge) you can use `/verified later`.

OlivierCazade:
In this case, is it fine to have this PR merged and then verified along the CNO PR that go with it ?

Currently in the CNO PR I am pointing to my personnal branch of the API, but I would prefer it to be tested pointing to the API official branch.

memodi (approved):
verified pre-merge, cluster comes up with `NetworkObservabilityInstall` enabled in featureSet list in a TechPreview cluster.

/verified

OlivierCazade:
@everettraven could you override again the crd-schema test ?

PR #2765, opened 2026-03-13
NE-2411: Add Template field to DNS operator

grzpiotrowski:
Add [coreDNS template plugin](https://coredns.io/plugins/template/) API support to enable custom DNS query handling coreDNS. The main use case is filtering AAAA queries in IPv4-only clusters to reduce DNS latency and upstream load.

Example usage:
```yaml
  kind: DNS
  metadata:
    name: default
  spec:
    template:
      zones: ["example.com", "abc.com"]
      queryType: AAAA
      queryClass: IN
      action:
           rcode: NOERROR
```  

Intruducing new types:
- `Template` - Template configuration with zones, query type, query class, and actions
- `QueryType` - DNS query types (AAAA initially, the goal is to have it extensible to A, CNAME etc. in the future, same for other types)
- `QueryClass` - DNS query classes (only IN initially)
- `ResponseCode` - DNS response codes (NOERROR initially)
- `TemplateAction` - Discriminated union for template actions
- `ReturnEmptyAction` - initial action for returning empty DNS responses


grzpiotrowski:
/retitle [WIP] [NE-2411](https://issues.redhat.com/browse/NE-2411): Add templates field to DNS operator

operator/v1/types_dns.go
```diff
@@ -467,6 +482,112 @@ const (
...
+
+const (
+	// ResponseCodeNOERROR indicates a successful DNS query with or without answer records.
+	// When used with returnEmpty action, this returns an empty response (no AAAA records)
+	// without indicating an error, which is the standard behavior for AAAA filtering.
+	ResponseCodeNOERROR ResponseCode = "NOERROR"
+)
+
+// Template defines a template for custom DNS query handling via the CoreDNS template plugin.
+// Templates enable filtering or custom responses for DNS queries matching specific zones and query types.
```
alebedev87:
```suggestion
// Template enables filtering or custom responses for DNS queries matching specific zones and query types.
```

operator/v1/types_dns.go
```diff
@@ -116,6 +116,21 @@ type DNSSpec struct {
 	// 30 seconds or as noted in the respective Corefile for your version of OpenShift.
 	// +optional
 	Cache DNSCache `json:"cache,omitempty"`
+
+	// template is an optional configuration for custom DNS query handling via the CoreDNS template plugin.
+	// The template defines how to handle queries matching specific zones and query types.
+	//
+	// The template is injected into ALL Corefile server blocks (both custom servers from spec.servers
+	// and the default .:5353 block). This ensures consistent behavior across all DNS resolution paths.
```
alebedev87:
```suggestion
	// The template applies to all domains (custom domains from spec.servers and the cluster domain)
   // to ensure consistent DNS resolution across all paths.
```

operator/v1/types_dns.go
```diff
@@ -116,6 +116,21 @@ type DNSSpec struct {
...
+	//
+	// AAAA filtering is intended for IPv4-only clusters. In IPv6 or dual-stack clusters, use specific
+	// zones instead of "." to avoid filtering internal IPv6 service addresses (e.g., cluster.local).
```
alebedev87: This may be a note in `template.zones` field, something like:
```
Note: root zone (".") includes cluster domain (cluster.local); use specific zones to avoid impacting IPv6 queries in IPv6 or dual-stack clusters.
```

operator/v1/types_dns.go
```diff
@@ -467,6 +482,112 @@ const (
...
+	// +kubebuilder:default=IN
+	// +required
+	QueryClass QueryClass `json:"queryClass"`
+
+	// actions defines a list of actions to apply to matching queries.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
+	// +required
+	Actions []TemplateAction `json:"actions"`
```
alebedev87: Just now I realized that with a slice we may have some duplicates which may be contradictory. Especially when the API will be enhanced to other actions (`answer`, `authority`, `additional`):
```
spec:
   template:
      actions:
      - returnEmpty:
            rcode: NOERRROR
      - returnEmpty:
            rcode: NXDOMAIN
```

grzpiotrowski: Changed the action to a struct. 
So now the spec example would be:

```yaml
  spec:
    template:
      zones: ["example.com", "abc.com"]
      queryType: AAAA
      queryClass: IN
      action:
           rcode: NOERROR
```

features/features.go
```diff
@@ -257,6 +257,14 @@ var (
 					enable(inTechPreviewNoUpgrade(), inDevPreviewNoUpgrade()).
 					mustRegister()
 
+	FeatureGateDNSTemplatePlugin = newFeatureGate("DNSTemplatePlugin").
+					reportProblemsToJiraComponent("dns").
+					contactPerson("grzpiotrowski").
+					productScope(ocpSpecific).
+					enhancementPR("https://github.com/openshift/enhancements/pull/1936").
+					enable(inDevPreviewNoUpgrade()).
```
yuqi-zhang, line 265: As we discussed on slack, this is fine to start in Dev Preview. If we'd like to move to TP, we do require the attached enhancement to be merged.

operator/v1/types_dns.go
```diff
@@ -52,6 +52,8 @@ type DNSSpec struct {
 	// If this field is nil, no servers are created.
 	//
 	// +optional
+	// +listType=map
```
yuqi-zhang, line 55: I think the default listtype is atomic, so this technically could be a breaking change, since it makes the validation more restrictive. Would you foresee any situation where this could be a problem?

operator/v1/types_dns.go
```diff
@@ -116,6 +118,18 @@ type DNSSpec struct {
...
+	// The template defines how to handle queries matching specific zones and query types.
+	//
+	// The template applies to all domains (custom domains from spec.servers and the cluster domain)
+	// to ensure consistent DNS resolution across all paths.
+	//
+	// When this field is not set, no template plugin configuration is added to CoreDNS.
+	//
+	// +optional
+	// +openshift:enable:FeatureGate=DNSTemplatePlugin
+	Template Template `json:"template,omitempty,omitzero"`
```
yuqi-zhang: (minor) I think the omitzero is sufficient enough, and the omitempty is redundant.

operator/v1/types_dns.go
```diff
@@ -167,10 +181,13 @@ var (
 type Server struct {
 	// name is required and specifies a unique name for the server. Name must comply
 	// with the Service Name Syntax of rfc6335.
+	// +required
```
yuqi-zhang, line 184: Out of curiosity, are you planning on making all legacy fields more conformant with API standards? I think there's some more things we would need to do (example, validations for the syntax here, min/max length, etc.) so maybe it's worth it to separately update old fields and focus this PR on the template field.

Generally also applies to other existing fields.

operator/v1/types_dns.go
```diff
@@ -467,6 +487,87 @@ const (
...
+// Template defines a template for custom DNS query handling.
+type Template struct {
+	// zones specifies the DNS zones this template applies to.
+	// Each zone must be a valid DNS name as defined in RFC 1123.
+	// The special zone "." matches all domains (catch-all).
+	// At least one zone must be specified.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
+	Zones []string `json:"zones"`
+
```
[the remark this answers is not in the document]
yuqi-zhang: +1 on coderabbit's suggestion here. example of the validation is https://github.com/openshift/api/blob/master/machineconfiguration/v1/types_machineosconfig.go#L177 

although... maybe we need some special casing on top of that to include stuff like the ["."] option?

Might be easier to make `zone` a type and validate on the type itself

operator/v1/types_dns.go
```diff
@@ -467,6 +487,90 @@ const (
...
+	//
+	// Note: root zone (".") includes cluster domain (cluster.local); use specific zones to avoid impacting IPv6 queries in IPv6 or dual-stack clusters.
+	//
+	// Examples:
+	// - ["."] matches all domains (catch-all for global AAAA filtering)
+	// - ["example.com"] matches only example.com and its subdomains
+	// - ["example.com", "test.com"] matches both domains and their subdomains
+	//
+	// +listType=set
+	// +kubebuilder:validation:MinItems=1
```
yuqi-zhang, line 542: restrictions on the list should also be in the godoc itself (e.g. here  `// At least 1 and at most 15 zones may be specified.`)

operator/v1/types_dns.go
```diff
@@ -467,6 +487,90 @@ const (
...
+	// +required
+	Action TemplateAction `json:"action,omitzero"`
+}
+
+// TemplateAction defines how to construct a DNS response for queries matching the template.
+type TemplateAction struct {
+	// rcode is the DNS response code to return.
+	// Valid values are "NOERROR".
+	//
+	// When set, the template returns a response with no answer records. For AAAA filtering,
```
yuqi-zhang: The field is required, so the `when set` comment is a bit confusing. I think we should remove any conditional statements

operator/v1/types_dns.go
```diff
@@ -467,6 +487,90 @@ const (
...
+type ResponseCode string
+
+const (
+	// ResponseCodeNOERROR indicates a successful DNS query with or without answer records.
+	ResponseCodeNOERROR ResponseCode = "NOERROR"
+)
+
+// Template defines a template for custom DNS query handling via the CoreDNS template plugin.
+// Template enables filtering or custom responses for DNS queries matching specific zones and query types.
+type Template struct {
```
yuqi-zhang: Should we consider making this slightly less generic? (e.g. DNSTemplate, like we have DNSCache above)

operator/v1/types_dns.go
```diff
@@ -467,6 +487,90 @@ const (
...
+	// +kubebuilder:validation:items:MinLength=1
+	// +kubebuilder:validation:items:MaxLength=253
+	// +required
+	Zones []string `json:"zones,omitempty"`
+
+	// queryType specifies the DNS query type to match.
+	// Valid values are "AAAA".
+	//
+	// +required
+	QueryType QueryType `json:"queryType,omitempty"`
```
yuqi-zhang, line 551: I'm a bit confused on why each of these are a required, set string field that only has 1 option. Do you believe that this will be expanded on in the future? Can these be optional and have the controller set a reasonable default if the user doesn't set them?

operator/v1/types_dns.go
grzpiotrowski: This was in response to the failing CI [verify-crd-schema](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_api/2765/pull-ci-openshift-api-master-verify-crd-schema/2035072881104261120) with the `ListsMustHaveSSATags` error and some following errors.
I believe this is because between this PR and the last merged one to this CRD, there were some additional requirements added in the tests.

I kept these as a separate commit, but probably better if these kind of changes were separated into another PR altogether like you said.

operator/v1/types_dns.go
grzpiotrowski: Added the comment in godoc.

operator/v1/types_dns.go
grzpiotrowski: Yes this was with the intention of making it expandable in the future with other query types.
And making it and the other fields required as to ensure that the configuration applied is clear to the user.
Though I reckon a default value could be also considered, but then I imagine we would need to stick to that default (AAAA in this case) in the future, also when other types are added to the available ones.

yuqi-zhang:
Looks like most of the discussion is happening via https://github.com/openshift/enhancements/pull/1936 - please ping me when Joel's concerns are addressed and reflected here, and I can do another round of reviews, thanks!

PR #2773, opened 2026-03-19
NE-2117: Add httpsLogFormat and tcpLogFormat to API

rohara:
Add both httpsLogFormat and tcpLogFormat to ingress API. These log-format strings belong in AccessLogging struct and will provide the ability to customize both HTTPS and TCP log formats when applicable.


features/features.go
```diff
@@ -1020,4 +1020,12 @@ var (
 				enhancementPR("https://github.com/openshift/enhancements/pull/1910").
 				enable(inDevPreviewNoUpgrade(), inTechPreviewNoUpgrade()).
 				mustRegister()
+
+	FeatureGateHTTPSLogFormat = newFeatureGate("HTTPSLogFormat").
+				  reportProblemsToJiraComponent("Networking/router").
+				  contactPerson("rohara").
+		                  productScope(ocpSpecific).
+		                  enhancementPR("https://github.com/openshift/enhancements/pull/1832").
+				  enable(inDevPreviewNoUpgrade(), inTechPreviewNoUpgrade()).
```
yuqi-zhang: As a note, we are asking for the corresponding enhancement to be merged before tech preview. Is there anything blocking the enhancement at this time? Could this feature start in dev preview only instead?

operator/v1/types_ingress.go
```diff
@@ -1495,13 +1495,29 @@ type AccessLogging struct {
...
+	// HAProxy documentation:
+        // http//docs.haproxy.org/2.8/configuration.html#8.2.2
+        //
+        // Note that this format only applies to TCP connections.  It only affects
+        // the log format for TLS passthrough connections.
+        //
+	// +kubebuilder:validation:MinLength=1
+        // +kubebuilder:validation:MaxLength=1024
+        // +optional
+        TcpLogFormat string `json:"tcpLogFormat,omitempty"`
```
yuqi-zhang: As noted by coderabbit in https://github.com/openshift/api/pull/2773/changes#r2963291054, the new fields should have featuregate markers.

operator/v1/types_ingress.go
```diff
@@ -1495,13 +1495,29 @@ type AccessLogging struct {
 	// +required
 	Destination LoggingDestination `json:"destination"`
 
+	// tcpLogFormat specifies the format of the log message for a TCP
+        // request.
```
yuqi-zhang: the indentation seems off. Are these mixing tabs and spaces? Could you fix the indentation in this file?

operator/v1/types_ingress.go
```diff
@@ -1495,13 +1495,29 @@ type AccessLogging struct {
...
+        //
+        // If this field is empty, log messages use the implementation's default
+        // TCP log format.  For HAProxy's default TCP log format, see the
+	// HAProxy documentation:
+        // http//docs.haproxy.org/2.8/configuration.html#8.2.2
```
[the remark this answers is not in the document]
yuqi-zhang: +1 on this

operator/v1/types_ingress.go
```diff
@@ -1495,13 +1495,29 @@ type AccessLogging struct {
...
+        TcpLogFormat string `json:"tcpLogFormat,omitempty"`
+
 	// httpLogFormat specifies the format of the log message for an HTTP
 	// request.
 	//
 	// If this field is empty, log messages use the implementation's default
 	// HTTP log format.  For HAProxy's default HTTP log format, see the
 	// HAProxy documentation:
-	// http://cbonte.github.io/haproxy-dconv/2.0/configuration.html#8.2.3
+	// http//docs.haproxy.org/2.8/configuration.html#8.2.3
```
yuqi-zhang: also broken link

operator/v1/types_ingress.go
```diff
@@ -1512,6 +1528,24 @@ type AccessLogging struct {
 	// +optional
 	HttpLogFormat string `json:"httpLogFormat,omitempty"`
 
+	// httpsLogFormat specifies the format of the log messsage for an HTTPS
```
yuqi-zhang: typo, `messsage` -> `message`

operator/v1/types_ingress.go
```diff
@@ -1495,13 +1495,29 @@ type AccessLogging struct {
...
+        //
+        // If this field is empty, log messages use the implementation's default
+        // TCP log format.  For HAProxy's default TCP log format, see the
+	// HAProxy documentation:
+        // http//docs.haproxy.org/2.8/configuration.html#8.2.2
+        //
+        // Note that this format only applies to TCP connections.  It only affects
+        // the log format for TLS passthrough connections.
+        //
+	// +kubebuilder:validation:MinLength=1
```
yuqi-zhang, line 1514: For validations such as these, please have their corresponding requirements explicit in the godoc (example: `// When set, the value must be between 1 and 1024 characters long.`)

yuqi-zhang (changes requested):

features/features.go
```diff
@@ -1020,4 +1020,12 @@ var (
 				enhancementPR("https://github.com/openshift/enhancements/pull/1910").
 				enable(inDevPreviewNoUpgrade(), inTechPreviewNoUpgrade()).
 				mustRegister()
+
+	FeatureGateHTTPSLogFormat = newFeatureGate("HTTPSLogFormat").
+				  reportProblemsToJiraComponent("Networking/router").
```
Miciah:
```suggestion
				  reportProblemsToJiraComponent("Networking / router").
```

features/features.go
```diff
@@ -1020,4 +1020,12 @@ var (
 				enhancementPR("https://github.com/openshift/enhancements/pull/1910").
 				enable(inDevPreviewNoUpgrade(), inTechPreviewNoUpgrade()).
 				mustRegister()
+
+	FeatureGateHTTPSLogFormat = newFeatureGate("HTTPSLogFormat").
```
Miciah:
```suggestion
	FeatureGateIngressControllerHTTPSLogFormat = newFeatureGate("IngressControllerHTTPSLogFormat").
```

features/features.go
```diff
@@ -1020,4 +1020,12 @@ var (
...
+				  reportProblemsToJiraComponent("Networking/router").
+				  contactPerson("rohara").
+		                  productScope(ocpSpecific).
+		                  enhancementPR("https://github.com/openshift/enhancements/pull/1832").
+				  enable(inDevPreviewNoUpgrade(), inTechPreviewNoUpgrade()).
+		                  mustRegister()
```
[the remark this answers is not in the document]
Miciah: `HttpLogFormat` is an existing field and should not have the `// +openshift:enable:FeatureGate` marker added.  

I suggest naming the feature gate "IngressControllerHTTPSLogFormat" in https://github.com/openshift/api/pull/2773#discussion_r2983261727; if you implement that suggestion, then you will need to add `
// +openshift:enable:FeatureGate=IngressControllerHTTPSLogFormat` to the `httpsLogFormat` and `tcpLogFormat` fields.

operator/v1/types_ingress.go
Miciah: It might be useful to say "TCP connection" and explain that there may be separate log messages for the TCP connection (for the public_ssl frontend) and for the HTTP connection (for the fe_sni frontend):
```suggestion
	// tcpLogFormat specifies the format of the log message for a TCP
	// connection.  
	//
	// This log message is used for the initial TCP connection associated
	// with a TLS session.  Note that for edge-terminated and reencrypt
	// routes, you may see a second log message for the HTTP request in
	// addition to the log message for the TCP connection.  Use the
	// httpsLogFormat option to customize the log format for the HTTP
	// request.
```

operator/v1/types_ingress.go
```diff
@@ -1495,13 +1495,29 @@ type AccessLogging struct {
...
+	// tcpLogFormat specifies the format of the log message for a TCP
+        // request.
+        //
+        // If this field is empty, log messages use the implementation's default
+        // TCP log format.  For HAProxy's default TCP log format, see the
+	// HAProxy documentation:
+        // http//docs.haproxy.org/2.8/configuration.html#8.2.2
+        //
+        // Note that this format only applies to TCP connections.  It only affects
+        // the log format for TLS passthrough connections.
```
Miciah: We talked about this today, but I'll note here:  This log format applies to TLS reencrypt and edge-terminated routes as well, because for edge-terminated and reencrypt routes, the client connection first goes to the public_ssl frontend (which logs the TCP connection) and then gets forwarded to the fe_sni frontend (which logs the HTTP session).

features/features.go
Miciah: @yuqi-zhang, this feature poses low risk of breaking other features or tests (tech preview or otherwise).  It only involves OpenShift router and the Ingress Operator, and no other components.  In light of that, what is the advantage of making the feature DPNU instead of TPNU?

operator/v1/zz_generated.swagger_doc_generated.go
```diff
@@ -849,7 +849,9 @@ func (AWSSubnets) SwaggerDoc() map[string]string {
...
+	"httpLogFormat":      "httpLogFormat specifies the format of the log message for an HTTP request.\n\nIf this field is empty, log messages use the implementation's default HTTP log format.  For HAProxy's default HTTP log format, see the HAProxy documentation: http//docs.haproxy.org/2.8/configuration.html#8.2.3\n\nNote that this format only applies to cleartext HTTP connections and to secure HTTP connections for which the ingress controller terminates encryption (that is, edge-terminated or reencrypt connections).  It does not affect the log format for TLS passthrough connections.",
+	"httpsLogFormat":     "httpsLogFormat specifies the format of the log messsage for an HTTPS request.\n\nIf this field is empty, log messages use the implementation's default HTTPS log format.  For HAProxy's default HTTPS log format, see the HAProxy documentation: http://docs.haproxy.org/2.8/configuration.html#8.2.4\n\nNote that this format only applies to HTTPS connections for which the ingress controller terminates encryption (that is, edge-terminated or reencrypt connections).  It does not affect the log format for TLS passthrough connections.",
```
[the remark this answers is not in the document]
Miciah: I believe what we want to say is that `httpLogFormat` applies for cleartext HTTP connections to insecure routes as well as TLS connections to edge-terminated and reencrypt, *but* `httpsLogFormat`, if it is set, takes precedence over `httpLogFormat` for edge-terminated and reencrypt routes.

features/features.go
yuqi-zhang: I think in your case it would just allow for more breaking changes, and the ability to get this through without a merged enhancement, and you could switch to TNPU at any time afterwards since there's no other explicit gating. But I'm not strongly opinionated there, we can certainly start in TPNU if we're mostly happen with the design via the enhancement (and preferrably merged there)

operator/v1/zz_generated.swagger_doc_generated.go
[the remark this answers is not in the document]
rohara: This is not exactly true. Setting either `httpLogFormat` or `httpsLogFormat` simply allows the log format to be customized. When not explicity set, the default format for HTTPS and/or HTTP will be used. Those would be the HAProxy config options `option httpslog` and `option httplog` repectively.

features/features.go
```diff
@@ -203,12 +203,12 @@ var (
 					mustRegister()
 
 	FeatureGateNoOverlayMode = newFeatureGate("NoOverlayMode").
-				reportProblemsToJiraComponent("Networking/ovn-kubernetes").
-				contactPerson("pliurh").
-				productScope(ocpSpecific).
-				enhancementPR("https://github.com/openshift/enhancements/pull/1859").
-				enable(inDevPreviewNoUpgrade(), inTechPreviewNoUpgrade()).
-				mustRegister()
+					reportProblemsToJiraComponent("Networking/ovn-kubernetes").
```
yuqi-zhang, line 198: Out of curiosity, did something automatically add extra spaces here?

operator/v1/types_ingress.go
```diff
@@ -1495,13 +1495,35 @@ type AccessLogging struct {
 	// +required
 	Destination LoggingDestination `json:"destination"`
 
+	// tcpLogFormat specifies the format of the log message for a TCP
+	// connection.
+	//
+	// If this field is empty, log messages use the implementation's default
+	// TCP log format.  When set, the value must be between 1 and 1024
+	// characters long.For HAProxy's default TCP log format, see the
```
yuqi-zhang: There's a few inconsistencies in the godoc (no space between `long.For`, some extra whitespace, etc.). Could you clean that up?

operator/v1/types_ingress.go
```diff
@@ -1512,6 +1534,26 @@ type AccessLogging struct {
 	// +optional
 	HttpLogFormat string `json:"httpLogFormat,omitempty"`
 
+	// httpsLogFormat specifies the format of the log message for an HTTPS
+	// request.
+	//
+	// If this field is empty, log messages use the implementation's default
+	// HTTPS log format.  When set, the value must be between 1 and 1024
+	// characters long.For HAProxy's default HTTPS log format, see the
```
yuqi-zhang: same spacing inconsistency as above

operator/v1/types_ingress.go
```diff
@@ -1512,6 +1534,26 @@ type AccessLogging struct {
...
+	// Note that this format only applies to HTTPS connections for which the
+	// ingress controller terminates encryption (that is, edge-terminated or
+	// reencrypt connections).  It does not affect the log format for TLS
+	// passthrough connections.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=1024
+	// +openshift:enable:FeatureGate=IngressControllerHTTPSLogFormat
+	// +optional
+	HttpsLogFormat string `json:"httpsLogFormat,omitempty"`
```
yuqi-zhang: Technically, this should be HTTPSLogFormat (and TcpLogFormat should be TCPLogFormat) based on naming conventions, which generally is followed in this file, except for some reason the existing `HttpLogFormat` field.

I think the best behaviour is probably make the new fields formatted correct, even if it's not "consistent"

yuqi-zhang (commented):
Generally lgtm, some nits inline. Also, we should try to get the enhancement merged first

operator/v1/types_ingress.go
rohara: Changed in latest commit.

operator/v1/types_ingress.go
rohara: Changed in latest commit.

operator/v1/types_ingress.go
rohara: Fixed in latest commit.

features/features.go
rohara: I don't see this in the latest code.

operator/v1/types_ingress.go
```diff
@@ -1512,6 +1534,26 @@ type AccessLogging struct {
...
+	// request.
+	//
+	// If this field is empty, log messages use the implementation's default
+	// HTTPS log format.  When set, the value must be between 1 and 1024
+	// characters long.  For HAProxy's default HTTPS log format, see the
+	// HAProxy documentation:
+	// http://docs.haproxy.org/2.8/configuration.html#8.2.4
+	//
+	// Note that this format only applies to HTTPS connections for which the
+	// ingress controller terminates encryption (that is, edge-terminated or
```
yuqi-zhang, line 1547: The httpLogFormat above has a snippet like:

```
	// and to secure HTTP connections for which the ingress controller
	// terminates encryption (that is, edge-terminated or reencrypt
```

which, this seems to be saying the same thing? Does this take precedence in the implementation? Or would the httpLogFormat no longer handle that?

features/features.go
```diff
@@ -638,6 +638,14 @@ var (
 								enable(inDevPreviewNoUpgrade(), inTechPreviewNoUpgrade()).
 								mustRegister()
 
+	FeatureGateIngressControllerHTTPSLogFormat = newFeatureGate("IngressControllerHTTPSLogFormat").
```
yuqi-zhang, line 641: Not major, but since this also covers tcpLogFormat, is there a more general name for this we can use?

operator/v1/tests/ingresscontrollers.operator.openshift.io/IngressControllerHTTPSLogFormat.yaml
```diff
@@ -0,0 +1,67 @@
+apiVersion: apiextensions.k8s.io/v1 # Hack because controller-gen complains if we don't have this
+name: "IngressController"
+crdName: ingresscontrollers.operator.openshift.io
+featureGates:
+- IngressControllerHTTPSLogFormat
+tests:
```
yuqi-zhang, line 6: some negative (failing validation) tests would also be nice to have, although this is a small enough feature that I wouldn't consider it too critical

yuqi-zhang (commented):
A few more comments, also note that we do require enhancements to be merged for Tech Preview now, so let's try to get https://github.com/openshift/enhancements/pull/1832 merged first

features/features.go
rohara: How about we just remove 'HTTPS', giving ``FeatureGateIngressControllerLogFormat``?

operator/v1/types_ingress.go
Miciah, line 1532: Maybe this would be clearer:
```suggestion
	// This format applies to cleartext HTTP connections.  It also applies
	// to HTTPS connections for which the ingress controller terminates
	// encryption (that is, edge-terminated or reencrypt connections) if,
	// and only if, httpsLogFormat is not specified.  It does not affect the
	// log format for TLS passthrough connections.
```

operator/v1/types_ingress.go
```diff
@@ -1512,6 +1534,26 @@ type AccessLogging struct {
...
+	// reencrypt connections).  It does not affect the log format for TLS
+	// passthrough connections.
```
Miciah, line 1549:
```suggestion
	// reencrypt connections).  It does not affect the log format for TLS
	// passthrough connections or cleartext HTTP connections.
```

operator/v1/types_ingress.go
rohara: If we're all in agreement, I will reword the text for `httpLogFormat` as @Miciah suggested.

features/features.go
yuqi-zhang: I'd be fine with that

yuqi-zhang:
Just checking back on this, is this (and the enhancement) still active?

PR #2775, opened 2026-03-24
CORS-4029: Promote AWS ClusterHostedDNSInstall to Default

sadasu:


config/v1/tests/infrastructures.config.openshift.io/AAA_ungated.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1 # Hack because controller-gen complains if w
 name: "Infrastructure"
 crdName: infrastructures.config.openshift.io
 featureGates:
-- -AWSClusterHostedDNSInstall
```
JoelSpeed, line 5: Azure has now been promoted, so this is breaking the integration. Why do we need the negative gate requirement here?

sadasu: My bad. This should be `AWSClusterHostedDNS` which is the old featuregate which we do want to ignore here.

sadasu:
@JoelSpeed the coderabbit [comment](https://github.com/openshift/api/pull/2775#pullrequestreview-4069863023) seems valid. But, it is in an auto-generated file and so I can't actually implement the changes requested. WDYT?

JoelSpeed:
@yuqi-zhang Can you help here? Do you think the status parts being immutable will be an issue in the controller config? If so, the likely best approach is to remove the immutability from the controller config side using a manual patch override

JoelSpeed:
@sadasu Is this something you're hoping to promote soon? Is the existing feature promotion testing (sees cluster installs on AWS) sufficient or do you have specific jobs that are testing this install feature that we can look at?

sadasu:
@JoelSpeed This is the only [job](https://prow.ci.openshift.org/job-history/gs/test-platform-results/logs/periodic-ci-openshift-release-main-nightly-5.0-e2e-aws-custom-dns-techpreview) that is testing an install with this feature. No new tests have been written because a successful install is sufficient.
Also, there is no planned support needed in HyperShift at this time so there are no jobs testing this install method with HyperShift.
With these conditions, I am not sure if we can expect [verify-feature-promotion](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_api/2775/pull-ci-openshift-api-master-verify-feature-promotion/2059312686876856320) to pass.

JoelSpeed:
Reviewing the [sippy](https://sippy.dptools.openshift.org/sippy-ng/jobs/5.0/analysis?period=day&filters=%7B%22items%22%3A%5B%7B%22columnField%22%3A%22name%22%2C%22operatorValue%22%3A%22equals%22%2C%22value%22%3A%22periodic-ci-openshift-release-main-nightly-5.0-e2e-aws-custom-dns-techpreview%22%7D%5D%7D) data, I think this shows the installs are succeeding for this job.

@sadasu Please make sure we move this job to stable and that it is reporting into component readiness once we merge this PR

/lgtm
/verified by @JoelSpeed

sadasu:
@JoelSpeed, Thank you! I think https://github.com/openshift/api/pull/2775#issuecomment-4706842756 may have been meant for me.

PR #2778, opened 2026-03-24
MON-4029: Add KubeStateMetricsConfig to ClusterMonitoring API

danielmellado:
Adds a new KubeStateMetricsConfig struct to the ClusterMonitoring CRD,
allowing configuration of the kube-state-metrics agent (node selectors,
resources, tolerations, topology spread constraints). Includes
comprehensive integration tests for validation.

Signed-off-by: Daniel Mellado <dmellado@redhat.com>


danielmellado:
@everettraven this PR continues https://github.com/openshift/api/pull/2461 and rebases and adds your comments. Thanks!

rexagod:
@danielmellado could you please rebase this? Thanks again for the patch! :)

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -2492,3 +2498,131 @@ type Audit struct {
...
+	// +listType=map
+	// +listMapKey=resource
+	AdditionalResourceLabels []KubeStateMetricsResourceLabels `json:"additionalResourceLabels,omitempty"`
+}
+
+// KubeStateMetricsResourceName is the name of a Kubernetes resource whose labels can be exposed
+// as metrics by kube-state-metrics. Currently, only "jobs" and "cronjobs" are supported
+// due to cardinality concerns.
+// Valid values are "jobs" and "cronjobs".
+// +kubebuilder:validation:Enum=jobs;cronjobs
```
simonpasquier: (nit) should be `Jobs` and `CronJobs` to follow k8s API conventions?

rexagod: I'll raise a follow-up to lowercase the resource name once this PR is merged, so it works with the existing logic in https://github.com/openshift/cluster-monitoring-operator/pull/2553/changes#diff-420c938cae65d6e9c4d4e386d768532276257149dbcaf476a90884a29fe36bd4R374.

simonpasquier: @rexagod there's a translation happening between the CRD and the CMO configuration struct. IMHO this is where the conversion should happen.

rexagod: Ah, sorry my comment earlier wasn't clear. I meant to raise a follow-up in the translation layer in CMO, while the resource name here can follow the K8s convention.

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -2514,3 +2520,137 @@ type Audit struct {
...
+	TopologySpreadConstraints []v1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
+	// additionalResourceLabels defines additional Kubernetes resource labels to expose as metrics
+	// in kube-state-metrics, in addition to the default set.
+	// Currently, only "Jobs" and "CronJobs" resources are supported due to cardinality concerns.
+	// Each entry specifies a resource name and a list of Kubernetes label names to expose.
+	// Use "*" in the labels list to expose all labels for a given resource.
+	// additionalResourceLabels is optional.
+	// When omitted, only the default set of resource labels is exposed.
+	// Maximum length for this list is 2.
+	// Minimum length for this list is 1.
```
simonpasquier: (nit)
```suggestion
	// Minimum length for this list is 1.
	// Maximum length for this list is 2.
```

danielmellado: Thanks, nit but cleaner :D

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -2514,3 +2520,137 @@ type Audit struct {
...
+	// +listMapKey=resource
+	AdditionalResourceLabels []KubeStateMetricsResourceLabels `json:"additionalResourceLabels,omitempty"`
+}
+
+// KubeStateMetricsResourceName is the name of a Kubernetes resource whose labels can be exposed
+// as metrics by kube-state-metrics. Currently, only "Jobs" and "CronJobs" are supported
+// due to cardinality concerns.
+// Valid values are "Jobs" and "CronJobs".
+// +kubebuilder:validation:Enum=Jobs;CronJobs
+type KubeStateMetricsResourceName string
```
saschagrunert: The enum values `Jobs` and `CronJobs` are neither the Kubernetes Kind (singular PascalCase: `Job`, `CronJob`) nor the resource name (lowercase plural: `jobs`, `cronjobs`). Per Kubernetes API conventions, enum values should be PascalCase, which matches the Kind form.

Should these be `Job` and `CronJob` instead?

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -2514,3 +2520,137 @@ type Audit struct {
...
+// Must be either the wildcard "*" or a valid Kubernetes label key.
+// A valid label key has an optional DNS subdomain prefix followed by a "/" and a name segment,
+// or just a name segment without a prefix. The name segment must be 63 characters or fewer,
+// beginning and ending with an alphanumeric character, with dashes, underscores, dots, and
+// alphanumerics in between.
+// Must be at least 1 character and at most 253 characters in length.
+// +kubebuilder:validation:MinLength=1
+// +kubebuilder:validation:MaxLength=253
+// +kubebuilder:validation:XValidation:rule="self == '*' || self.matches('^[a-zA-Z0-9][a-zA-Z0-9_./-]*[a-zA-Z0-9]$') || self.matches('^[a-zA-Z0-9]$')",message="must be a valid Kubernetes label key or the wildcard '*'"
+type KubeStateMetricsLabelName string
```
saschagrunert, line 2708: The regex `^[a-zA-Z0-9][a-zA-Z0-9_./-]*[a-zA-Z0-9]$` is too permissive for Kubernetes label keys:
- It allows multiple slashes (e.g., `foo/bar/baz`), but a valid label key has at most one slash separating prefix from name.
- It allows uppercase in the DNS subdomain prefix, which is invalid.
- It doesn't enforce the prefix/name structure at all.

The existing `ContainerResource.Name` field in this same file uses `format.qualifiedName()` for validation. Consider using the same approach here for consistency and correctness:

```suggestion
// KubeStateMetricsLabelName is the name of a Kubernetes label to expose as a metric
// via kube-state-metrics. Use "*" to expose all labels for a resource.
// Must be either the wildcard "*" or a valid Kubernetes label key.
// A valid label key has an optional DNS subdomain prefix followed by a "/" and a name segment,
// or just a name segment without a prefix. The name segment must be 63 characters or fewer,
// beginning and ending with an alphanumeric character, with dashes, underscores, dots, and
// alphanumerics in between.
// Must be at least 1 character and at most 253 characters in length.
// +kubebuilder:validation:MinLength=1
// +kubebuilder:validation:MaxLength=253
// +kubebuilder:validation:XValidation:rule="self == '*' || !format.qualifiedName().validate(self).hasValue()",message="must be a valid Kubernetes label key or the wildcard '*'"
type KubeStateMetricsLabelName string
```

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -2514,3 +2520,137 @@ type Audit struct {
...
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=5
+	// +kubebuilder:validation:MinItems=1
+	Resources []ContainerResource `json:"resources,omitempty"`
+	// tolerations defines tolerations for the pods.
+	// tolerations is optional.
+	//
+	// When omitted, this means the user has no opinion and the platform is left
+	// to choose reasonable defaults. These defaults are subject to change over time.
+	// Defaults are empty/unset.
```
saschagrunert: The godoc says defaults may change over time but then says "Defaults are empty/unset." These are contradictory: if the default is empty and you're reserving the right to change it, that's fine, but the phrasing is confusing. Consider:

"When omitted, no tolerations are applied. This default is subject to change over time."

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -2514,3 +2520,137 @@ type Audit struct {
...
+	// topologySpreadConstraints defines rules for how kube-state-metrics Pods should be distributed
+	// across topology domains such as zones, nodes, or other user-defined labels.
+	// topologySpreadConstraints is optional.
+	// This helps improve high availability and resource efficiency by avoiding placing
+	// too many replicas in the same failure domain.
+	//
+	// When omitted, this means no opinion and the platform is left to choose a default, which is subject to change over time.
+	// This field maps directly to the `topologySpreadConstraints` field in the Pod spec.
+	// Defaults are empty/unset.
+	// When specified, topologySpreadConstraints must contain at least 1 entry and must not contain more than 10 entries.
```
saschagrunert, line 2646: Same issue as tolerations: "defaults may change over time" followed by "Defaults are empty/unset" is contradictory. Consider:

"When omitted, no topology spread constraints are applied. This default is subject to change over time."

config/v1alpha1/tests/clustermonitorings.config.openshift.io/ClusterMonitoringConfig.yaml
```diff
@@ -2223,6 +2223,420 @@ tests:
                 request: "100m"
                 limit: "0"
       expectedError: 'spec.nodeExporterConfig.resources[0].limit: Invalid value: "": limit must be a positive, non-zero quantity'
+    - name: Should be able to create KubeStateMetricsConfig with valid resources
```
saschagrunert, line 2279: A few test gaps:
- No `onUpdate` tests for `kubeStateMetricsConfig` at all. At least one test showing fields can be updated would be good.
- No test for invalid `KubeStateMetricsLabelName` values (e.g., a label name starting with a special character) to verify the CEL regex.
- No test verifying the wildcard `"*"` is accepted as a `KubeStateMetricsLabelName` on its own (it's used in the `additionalResourceLabels` tests but only within a list; an explicit single-wildcard-only test would be clearer).

saschagrunert (commented):
Thanks for working on this. The struct follows the established pattern from `OpenShiftStateMetricsConfig` well. See inline comments for details.

config/v1alpha1/types_cluster_monitoring.go
danielmellado: Thanks for the feedback! This was actually addressed in an earlier review round with @simonpasquier, who suggested using Jobs and CronJobs to align with Kubernetes resource name conventions.

The CMO ConfigMap API uses lowercase plural (jobs/cronjobs), so IMHO plural is the natural CRD equivalent referencing the resource type rather than the Kind. Happy to discuss further if you feel strongly about singular form.

config/v1alpha1/types_cluster_monitoring.go
danielmellado: Agreed, the regex was too loose. Switched to format.qualifiedName() as suggested. Thanks!

config/v1alpha1/types_cluster_monitoring.go
danielmellado: Fixed, thanks for catching the contradiction. Now reads as you suggest.

config/v1alpha1/types_cluster_monitoring.go
danielmellado: Changed as well!

config/v1alpha1/tests/clustermonitorings.config.openshift.io/ClusterMonitoringConfig.yaml
danielmellado: Good call, added:

onUpdate test for updating kubeStateMetricsConfig resources
Rejection test for invalid label name (!invalid-label)
Explicit wildcard-only ("*") acceptance test

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -2510,3 +2516,134 @@ type Audit struct {
...
+// alphanumerics in between.
+// Must be at least 1 character and at most 253 characters in length.
+// +kubebuilder:validation:MinLength=1
+// +kubebuilder:validation:MaxLength=253
+// +kubebuilder:validation:XValidation:rule="self == '*' || !format.qualifiedName().validate(self).hasValue()",message="must be a valid Kubernetes label key or the wildcard '*'"
+type KubeStateMetricsLabelName string
+
+// KubeStateMetricsResourceLabels defines which Kubernetes labels to expose as metrics
+// for a given resource type in kube-state-metrics.
+type KubeStateMetricsResourceLabels struct {
```
saschagrunert, line 2712: Consider adding a CEL rule to prevent mixing `*` with specific labels in the same list. Listing `"*"` alongside specific label names is redundant and likely a user mistake:

```suggestion
// +kubebuilder:validation:XValidation:rule="!self.labels.exists(l, l == '*') || self.labels.size() == 1",message="when '*' is specified, no other labels may be listed"
type KubeStateMetricsResourceLabels struct {
```

Please also add a rejection test for this case, e.g. specifying `["*", "app"]` in the labels list.

saschagrunert (changes requested):

config/v1alpha1/types_cluster_monitoring.go
danielmellado: Thanks for the suggestion! I went ahead and added the validation.

One small note on placement: we put the `XValidation` marker on the `labels` field rather than on the `KubeStateMetricsResourceLabels` struct type.

The reason is that this repo's `schemapatch` codegen pipeline (which patches existing CRD manifests rather than generating them from scratch) does not propagate struct level `XValidation` to array `items` schemas, so the rule would end up in the Go source but silently not appear in the generated CRDs.

Moving it to the field level (with `self` now referring to the list directly) produces identical enforcement and does propagate correctly:

```go
// +kubebuilder:validation:XValidation:rule="!self.exists(l, l == '*') || self.size() == 1",message="when '*' is specified, no other labels may be listed"
Labels []KubeStateMetricsLabelName `json:"labels,omitempty"`

File: config/v1alpha1/types_cluster_monitoring.go [hunk 9]
```diff
@@ -2510,3 +2516,137 @@ type Audit struct {
...
+	// When omitted, no topology spread constraints are applied. This default is subject to change over time.
+	// When specified, topologySpreadConstraints must contain at least 1 entry and must not contain more than 10 entries.
+	// Entries must have unique topologyKey and whenUnsatisfiable pairs.
+	// +kubebuilder:validation:MaxItems=10
+	// +kubebuilder:validation:MinItems=1
+	// +listType=map
+	// +listMapKey=topologyKey
+	// +listMapKey=whenUnsatisfiable
+	// +optional
+	TopologySpreadConstraints []v1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
```
saschagrunert, line 2661: The raw `v1.TopologySpreadConstraint` type has no schema validation for `whenUnsatisfiable` (should be enum `DoNotSchedule`/`ScheduleAnyway`), `maxSkew` (should have `minimum: 1`), or `minDomains` (should have `minimum: 1`, and upstream docs say it requires `whenUnsatisfiable=DoNotSchedule`). Invalid values will pass admission and only fail at scheduling time.

`OpenShiftStateMetricsConfig` has the same gap, so this may be intentional. But for a new API addition this is an opportunity to tighten things up, either via a custom struct with proper markers, or manual CRD overrides.

CodeRabbit flagged this twice as well.

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -2510,3 +2516,137 @@ type Audit struct {
...
+	// tolerations defines tolerations for the pods.
+	// tolerations is optional.
+	//
+	// When omitted, no tolerations are applied. This default is subject to change over time.
+	// When specified, tolerations must contain at least 1 entry and must not contain more than 10 entries.
+	// +kubebuilder:validation:MaxItems=10
+	// +kubebuilder:validation:MinItems=1
+	// +listType=atomic
+	// +optional
+	Tolerations []v1.Toleration `json:"tolerations,omitempty"`
```
saschagrunert, line 2637: Same concern as `topologySpreadConstraints`: the raw `v1.Toleration` has no enum validation on `operator` (should be `Exists`/`Equal`) or `effect` (should be `NoSchedule`/`PreferNoSchedule`/`NoExecute`/empty). Users can submit arbitrary strings that get persisted but have no effect.

Again, `OpenShiftStateMetricsConfig` has the same gap, so flagging for discussion rather than as a hard blocker.

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -2510,3 +2516,137 @@ type Audit struct {
...
+	// This field is optional.
+	// More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
+	// This is a simplified API that maps to Kubernetes ResourceRequirements.
+	// The current default values are:
+	//   resources:
+	//    - name: cpu
+	//      request: 4m
+	//    - name: memory
+	//      request: 40Mi
+	// Maximum length for this list is 5.
```
saschagrunert, line 2614: Nit: the default values block omits `limit: null`, unlike the equivalent godoc on `OpenShiftStateMetricsConfig.resources` and `NodeExporterConfig.resources`. Being explicit about the absence of limits is better for user documentation.

```suggestion
	// The current default values are:
	//   resources:
	//    - name: cpu
	//      request: 4m
	//      limit: null
	//    - name: memory
	//      request: 40Mi
	//      limit: null
```

saschagrunert (commented):
A few remaining items from a second pass. The godoc fixes from the previous round look good, and the wildcard CEL rule and test additions are solid.

The main open items are around the raw upstream types (`v1.TopologySpreadConstraint`, `v1.Toleration`) lacking schema-level validation for their constrained fields. `OpenShiftStateMetricsConfig` has the same gap, so this may be a deliberate pattern choice, but worth discussing for a new API addition.

config/v1alpha1/types_cluster_monitoring.go
danielmellado: Fixed, thanks! Added limit: null for consistency with OpenShiftStateMetricsConfig.resources.

config/v1alpha1/types_cluster_monitoring.go
danielmellado: Thanks for flagging this. Added field-level CEL validation rules on KubeStateMetricsConfig.topologySpreadConstraints:

whenUnsatisfiable must be DoNotSchedule or ScheduleAnyway
maxSkew must be >= 1
minDomains can only be used when whenUnsatisfiable is DoNotSchedule
Also added rejection tests for each case. OpenShiftStateMetricsConfig still has the same gap I'll address that in a follow-up PR.

config/v1alpha1/types_cluster_monitoring.go
danielmellado: Same approach. Added CEL validation on KubeStateMetricsConfig.tolerations:

operator must be Exists or Equal (when set)
effect must be NoSchedule, PreferNoSchedule, NoExecute, or empty (when set)
With corresponding rejection tests. Will tighten OpenShiftStateMetricsConfig in the follow-up.

config/v1alpha1/types_cluster_monitoring.go
saschagrunert: The thread between @simonpasquier and @rexagod actually converged on the same conclusion: the CRD should follow K8s conventions (`Job`/`CronJob`), and the translation to CMO's lowercase plural (`jobs`/`cronjobs`) should happen in the CMO conversion layer.

K8s API conventions are explicit that enum values must be PascalCase. `Jobs`/`CronJobs` is neither the Kind (`Job`/`CronJob`) nor the resource name (`jobs`/`cronjobs`). Since this is a new v1alpha1 API, now is the time to get this right.

config/v1alpha1/types_cluster_monitoring.go
saschagrunert: Thanks for adding the CEL rules. The `whenUnsatisfiable` and `maxSkew` validations look good, and the `minDomains` requires `DoNotSchedule` check is correct.

One gap remains: `minDomains >= 1` is still not validated. Your original comment mentioned `minimum: 1` for `minDomains`, and upstream docs say "Valid values are integers greater than 0." Please add:

```go
// +kubebuilder:validation:XValidation:rule="self.all(c, !has(c.minDomains) || c.minDomains >= 1)",message="minDomains must be at least 1"
```

And update the godoc at line 2586 to: "When minDomains is specified, it must be at least 1 and whenUnsatisfiable must be \\\DoNotSchedule\\\."

Also add a rejection test for `minDomains: 0`.

config/v1alpha1/types_cluster_monitoring.go
saschagrunert: cc @JoelSpeed @everettraven, any thoughts on this ☝️

config/v1alpha1/types_cluster_monitoring.go
danielmellado: Thanks! done! I'll wait on the replies on the Job/CronJob and push the remaining changes.

simonpasquier:
> I'll wait on the replies on the Job/CronJob

I'm fine with @saschagrunert recommendations, it totally makes sense to follow Kubernetes conventions.

config/v1alpha1/types_cluster_monitoring.go
danielmellado: I also tend to agree here, and thanks to @simonpasquier for acking. Updated the enum values to Job/CronJob to follow Kubernetes PascalCase Kind conventions. The translation to CMO's lowercase plural (jobs/cronjobs) will be handled in the CMO conversion layer.

saschagrunert (approved):
/lgtm

simonpasquier (commented):
/lgtm

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -158,6 +158,12 @@ type ClusterMonitoringSpec struct {
 	// When set, at least one field must be specified within monitoringPluginConfig.
 	// +optional
 	MonitoringPluginConfig MonitoringPluginConfig `json:"monitoringPluginConfig,omitempty,omitzero"`
+	// kubeStateMetricsConfig is an optional field that can be used to configure the kube-state-metrics
+	// agent that runs in the openshift-monitoring namespace. kube-state-metrics generates metrics about
+	// the state of Kubernetes objects such as Deployments, Nodes, and Pods.
+	// When omitted, this means no opinion and the platform is left to choose a reasonable default, which is subject to change over time.
+	// +optional
+	KubeStateMetricsConfig KubeStateMetricsConfig `json:"kubeStateMetricsConfig,omitempty,omitzero"`
```
everettraven, line 166: Just to clarify, the intention is that the `kube-state-metrics` agent will always be running and that this is only a way for end-users to configure _how_ it is run?

Do we envision ever having an option to not run it?

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -2510,3 +2516,151 @@ type Audit struct {
...
+	Resource KubeStateMetricsResourceName `json:"resource,omitempty"`
+	// labels is the list of Kubernetes label names to expose as metrics for this resource.
+	// Use "*" to expose all labels for the specified resource.
+	// When "*" is specified, it must be the only entry in the list; mixing "*" with
+	// specific label names is not allowed.
+	// This field is required.
+	// Each label name must be unique within this list.
+	// Minimum length for this list is 1.
+	// Maximum length for this list is 50.
+	// +required
```
everettraven, line 2727: I'm curious, why should this field be required?

Would omission of this field be a reasonable alternative to using `*` as a wildcard for "all labels"?

everettraven (commented):
Overall the structure here looks good. Just a couple questions.

config/v1alpha1/types_cluster_monitoring.go
danielmellado: Yes, kube-state-metrics always runs as part of the monitoring stack. This field is purely about configuring how it runs (scheduling, resources, label metrics). No plans to add a disable toggle. Same pattern as prometheusK8sConfig, alertmanagerConfig, openShiftStateMetricsConfig, etc.

config/v1alpha1/types_cluster_monitoring.go
danielmellado: An entry without labels would be a no-op, so requiring it avoids that. Making omission mean "all labels" would be an implicit wildcard that could cause cardinality surprises. The explicit * keeps intent clear and matches how kube-state-metrics handles --metric-labels-allowlist (jobs=[*] vs jobs=[app,team]).

simonpasquier: As a user, I'd also be surprised if an empty list means "collect all labels".

config/v1alpha1/types_cluster_monitoring.go
```diff
@@ -2510,3 +2516,151 @@ type Audit struct {
...
+	// +kubebuilder:validation:XValidation:rule="self.all(c, !has(c.minDomains) || c.whenUnsatisfiable == 'DoNotSchedule')",message="minDomains can only be used when whenUnsatisfiable is DoNotSchedule"
+	// +optional
+	TopologySpreadConstraints []v1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
+	// additionalResourceLabels defines additional Kubernetes resource labels to expose as metrics
+	// in kube-state-metrics, in addition to the default set.
+	// Currently, only "Job" and "CronJob" resources are supported due to cardinality concerns.
+	// Each entry specifies a resource name and a list of Kubernetes label names to expose.
+	// Use "*" in the labels list to expose all labels for a given resource.
+	// additionalResourceLabels is optional.
+	// When omitted, only the default set of resource labels is exposed.
```
everettraven: What is the "default set of resource labels"?

How would I as a user know if I need to configure this?

config/v1alpha1/types_cluster_monitoring.go
everettraven: Presumably if you don't specify a configuration for an additional resource it doesn't do any collection for that resource?

config/v1alpha1/types_cluster_monitoring.go
danielmellado: By "default set of resource labels", this refers to the labels kube-state-metrics exposes out of the box, which is effectively none for Jobs/CronJobs. kube-state-metrics does not expose Kubernetes object labels as metric labels by default; it only does so when explicitly configured via --metric-labels-allowlist.

I agree the wording is misleading. I'll rephrase to something like:

"When omitted, no additional Kubernetes object labels are exposed as metrics by kube-state-metrics beyond its built-in metric labels (e.g. namespace, job_name). Use this field to opt in to exposing specific Kubernetes labels as metric labels for the supported resource types."

danielmellado:
Correct. kube-state-metrics always collects its built-in metrics for all resources it watches (Jobs, CronJobs, Pods, etc.) -- that part is not affected by this config. What additionalResourceLabels controls is whether Kubernetes object labels (the user-defined labels on the resource metadata, like app=frontend) are promoted into metric labels.

If you don't add an entry for, say, CronJob in additionalResourceLabels, kube-state-metrics still collects CronJob metrics, but none of the CronJob's Kubernetes labels appear as metric labels. Adding an entry lets you opt in to that per resource.

config/v1alpha1/types_cluster_monitoring.go
danielmellado: Correct. kube-state-metrics still collects its built-in metrics for all resources it watches (Jobs, CronJobs, Pods, etc.), but none of the Kubernetes object labels (e.g. app=frontend) are promoted into metric labels unless you explicitly add an entry in additionalResourceLabels for that resource. So omitting a resource here means no label collection for it, not no metrics at all.

everettraven: Thanks for the clarifications here and in the parent field, it helps clarify the confusion I had here.

I think the required-ness of this field and the existing field value semantics makes sense to me now. Happy to leave this as-is.

everettraven (approved):
/lgtm

